PR #1, opened 2016-04-26
First pass at a page to collect all GOV.UK repos

tijmenb:
This commit is a first attempt at a page that collects the all alphagov-repos that are used to run GOV.UK. It's a simple static Jekyll site, with content in YAML files, augmented with GitHub API data.

screen shot 2016-04-26 at 17 48 07

Caveats & notes:

- The list of repos is not complete, since there's no authoritative list of GOV.UK-owned repos.
- The descriptions are taken from GitHub and are sometimes non-descriptive or lacking.
- Jekyll was chosen because of ease of deployment and because the resulting pages will be really fast. Downside is that the pages will not be updated automatically (when a description changes for example). We could get around that by triggering builds from our Jenkins CI infrastructure.
- This uses a stripped down version of the `govuk_template` and some things from the govuk-frontend-toolkit, because it provides a nice font and grid. It doesn't have the black bar because it's not an official
service - we should probably discuss the branding.


alexmuller:
It's worth noting this kind of existed already on https://alphagov.github.io/ :)

tijmenb:
We've been chatting about this offline too.

alphagov.github.io is not really maintained anymore, and because it's alphagov-wide not really appropriate for GOV.UK-specific documentation. Perhaps it could be a place to link to the projects that use the alphagov org (PaaS, Pay, GOV.UK). 

It's also a 400MB repo that [contains all kinds of fun binaries](https://github.com/alphagov/alphagov.github.io/tree/master/maven) (lol).

PR #3, opened 2016-04-29
Fix port number for local running instructions

dsingleton:
Running `bundle exec jekyll serve --watch --baseurl ''` starts the app
on `4000`, not `4567` for me.

Also drop the period, so the address is easier to copy and paste.


tijmenb:
Cheers!

PR #4, opened 2016-04-29
Misc improvements (fix broken links, add missing app, updated GH descriptions)

dsingleton:
- Fix broken links for GHE repos
- Add https://github.com/alphagov/feedback to list of FE apps
- Re-generate data, which also picked up some descriptions I updated.


tijmenb:
👍

PR #5, opened 2016-05-04
Fix some templating issues & duplication

tijmenb:
See individual commits for more info.


css/main.scss
```diff
@@ -39,3 +39,19 @@ h2 {
...
+
+  li {
+    list-style: none;
+    padding: 0;
+    display: inline-block;
+    margin-right: 20px;
+  }
+
+  a.active {
+    color: #333;
```
Davidslv: If you added a new colour I would add it in the pallete file, see here: https://github.com/alphagov/govuk-developers/tree/gh-pages/css/toolkit/colours

tijmenb: Good call. Used something from the palette now. PR updated.

PR #6, opened 2016-09-08
Content format docs

tijmenb:
This  PR adds auto-generated documentation for the govuk-content-schemas. The idea is that we have a nice way to view what fields a schema has and what the attributes mean.

This has been languishing on a branch on my machine for a while now. The code is a bit odd and doesn't have any tests (maybe okay because the rake task is test enough).

I'm not 100% convinced this is useful (we'll see after merge) and if is the right place (could live in the govuk-content-schema repo itself).

screen shot 2016-09-08 at 15 25 18


timblair:
I think this would be better placed within the [alphagov/govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas) repo itself, especially now we can [publish the project page's `/docs` dir to GitHub Pages](https://github.com/blog/2233-publish-your-project-documentation-with-github-pages).

PR #12, opened 2016-11-28
Configure for Heroku hosting

tijmenb:
We can't use the crown & font on non-gov.uk sites. Hosting on Heroku allows us to use HTTP auth.

Also brings us awesomeness like auto-deploying review apps for each PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .sass-cache
+build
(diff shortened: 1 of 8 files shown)


rubenarakelyan:
Nice 🚢

fofr:
The basic auth is frustrating. We don't do it on Elements which has the same issue: http://govuk-elements.herokuapp.com/

PR #14, opened 2016-11-30
Pull in rummager's search documentation

tijmenb:
This PR adds pulls in the rummager documentation page for the search API.

https://github.com/alphagov/rummager/blob/master/docs/search-api.md

I think this is probably the most useful page for rummager. I've considered also pulling in pages like [indexing](https://github.com/alphagov/rummager/blob/master/docs/documents.md) docs, but they seem less complete at the moment. Perhaps we can start with this page and see how it's used?

Thoughts @MatMoore?

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -15,6 +15,7 @@ configure :build do
 end
 
 require_relative './lib/dashboard/dashboard'
+require_relative './lib/external_doc'
 
 helpers do
   def dashboard
--- a/lib/external_doc.rb
+++ b/lib/external_doc.rb
@@ -1,0 +1,10 @@
+class ExternalDoc
+  def self.fetch(url_to_markdown)
+    markdown = Faraday.get(url_to_markdown).body
+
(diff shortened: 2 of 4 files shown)


MatMoore:
Was there a problem linking out to more project focused documentation?

http://alphagov.github.io/rummager was serving documentation for a while, and I found it really useful! What I like about github pages is the docs live really close to the code, and updating them is really low maintenance.

I'm not a huge fan of the `search-api.md` page, which is why I've been looking at swagger-based tools to automatically generate it, but it's the most complete thing we have for now. Tara's blog post might be a better introduction though, it's got lots of examples: https://gdsdata.blog.gov.uk/2016/05/26/use-the-search-api-to-get-useful-information-about-gov-uk-content/

https://github.com/alphagov/rummager/blob/master/docs/adding-new-fields.md is another one I've referred back to recently (even though I think I wrote it)

PR #18, opened 2016-12-14
First iteration of the application pages

tijmenb:
This adds application pages to the docs. Each application page contains a description (fetched from GitHub), links to the repo, puppet config and deploy scripts. If provided, it shows which team owns the applications.

The applications are configured using a single YAML file. The puppet/deploy/prod links are mostly auto-generated, but we allow them to be overridden in case of snowflakiness.

## Screenshots!

screen shot 2016-12-14 at 11 07 42

screen shot 2016-12-14 at 11 08 12

Note that this is an MVP. The correct teams per repo can be added in subsequent PRs (I'm leaving them empty on purpose, so that more people can get involved).

Things we'd like to do in the future with these pages:

- Showing the dependencies of an application
- Showing examples of content published / rendered by the app
- Show screen shots

https://trello.com/c/dr5JzgVD/22-mvp-for-application-pages


rubenarakelyan:
Love. It.

PR #23, opened 2016-12-21
Run builds on Circle to test project

tijmenb:
This project doesn't have tests at the moment. This makes circle run a build. If it fails, the build will fail as well.


tijmenb:
This doesn't work, because GitHub doesn't appear to respect the caching headers of Faraday from Circle. I'll investigate later.

PR #33, opened 2017-02-01
Configure markdown renderer option to not use intra-word emphasis

klssmith:
- The no_intra_emphasis option for the markdown renderer is now set to
true, which stops the renderer from trying to parse the inside of words.
- Without this setting, words with underscores in the middle of them cause
 the formatting to be incorrect.

#### Before: Markdown not being rendered properly in the 'Link' paragraph
screen shot 2017-02-01 at 11 13 44

#### After: With markdown being rendered in the same way as its source repo in the 'Link' paragraph
screen shot 2017-02-01 at 11 12 47

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -7,7 +7,8 @@ set :markdown,
       with_toc_data: true
     ),
     fenced_code_blocks: true,
-    tables: true
+    tables: true,
+    no_intra_emphasis: true
 
 configure :development do
   activate :livereload


klssmith:
Internal links on the page don’t seem to be working anymore because the heading ids have now changed, for example the Link heading now has an id of `linking-3-link` instead of an id of `link`.

Need to check if this is a problem being caused by this commit, or if this is ok and just caused by a recent change to how the template works, for example [in this file](https://github.com/alphagov/govuk-developer-docs/blob/ef5b8e8d4b1c8c909a9c6c15d7fb4387111d40ec/lib/unique_identifier_generator.rb).

PR #34, opened 2017-02-01
Simplify document types

tijmenb:
We currently have to make one request per document_type to find the example pages for each type.

Since https://github.com/alphagov/rummager/pull/743 we can request examples for each facet. This saves us a lot of requests when building the pages.

cc @rubenarakelyan


rubenarakelyan:
🎉

PR #36, opened 2017-02-02
Add guide to A/B testing

tijmenb:
The new A/B testing method using Fastly is a fairly complicated system that touches a few parts of a stack. This is an attempt at a guide for developers that want to use the system.

This is the first "guide" on the docs site. The idea is that this is a more appropriate place than the opsmanual (which is closed and only accessible via the VPN) and the wiki (which is closed and not user friendly). Feedback welcome on that too.

I've copied most from https://gov-uk.atlassian.net/wiki/pages/viewpage.action?pageId=128385031 by @suzannehamilton. We should probably replace that page with this.

https://trello.com/c/bh7iFV7G

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: git://github.com/alphagov/table_of_contents.git
-  revision: ee6d26c6ead059e60b212f10f906766ec89b25df
+  revision: 2b3639c05eb68e5d00e63469f784b203eb6257a1
   specs:
     table_of_contents (0.1.0)
       nokogiri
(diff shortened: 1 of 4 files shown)


suzannehamilton:
This all looks correct. 👍 

Is it worth explicitly mentioning the pros and cons of the JS and Fastly options? It's useful to know that the JS is much simpler to configure but suffers from content flashing.

tijmenb:
@suzannehamilton yes! Do you want to add them in a separate PR?

suzannehamilton:
Sure. :)

PR #46, opened 2017-02-09
Add page about the taxonomy

tijmenb:
This is a first pass at a page to explain the technical stuff behind the taxonomy.

It's intended for GOV.UK developers that are interested in using the taxonomy, like @pmanrubia and @tomnatt currently.

I *think* that it makes sense for a page like this to exist on this site, rather than say content-tagger, because it's a concept that touches many apps on GOV.UK - content-store, rummager, content-tagger, publishing-api and soon publishing apps and collections.

Thoughts? @pmanrubia @tomnatt @MatMoore @carvil  

**Preview**: https://govuk-tech-docs-pr-46.herokuapp.com/guides.html#taxonomy

### DIFF
--- a/source/guides.html.md.erb
+++ b/source/guides.html.md.erb
@@ -3,6 +3,8 @@ layout: layout
 title: Guides
 ---
 
+<%= partial 'guides/taxonomy' %>
+
 # A/B testing
 
 ## 1. Overview
--- a/source/guides/_taxonomy.html.md.erb
+++ b/source/guides/_taxonomy.html.md.erb
@@ -1,0 +1,41 @@
+# Taxonomy
+
+GOV.UK's "single subject taxonomy" intends to describe all content on GOV.UK. It is being developed theme-by-theme, starting with education.
+
+## 1. Editing taxonomy
+
+The taxonomy is managed in [content-tagger][edit-taxonomy].
+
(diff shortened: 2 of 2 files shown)


MatMoore:
🚀 🏂 🎉 💯 💃  These guides are great!

We've been thinking about creating a shared library for accessing the taxonomy (as a tree structure), so I'll make sure to link to anything that comes out of that.

I think under editing taxonomy we could mention the `parent_taxon` link type stores the relationship between taxons, and there is currently no support for multiple parents per taxon.

I was thinking it might be good to create another guide that covers some the older tagging methods on GOV.UK, what they're used for, and maybe what the long term plan is for simplifying tagging. Like what's on the finding things wall at the moment. What do you think? Should I create a separate PR for that?

carvil:
This is great, thanks @tijmenb !

At some point when we have the new navigation live behind the A/B test, we should expand on what components make up for the navigation using the new taxonomy. But that's for later.

:shipit:

pmanrubia:
@tijmenb looks really good. Thank you 👍 .

tijmenb:
> I think under editing taxonomy we could mention the parent_taxon link type stores the relationship between taxons, and there is currently no support for multiple parents per taxon.

@MatMoore added a commit!

PR #48, opened 2017-02-14
Update the A/B testing guide

tijmenb:
Add another trade-off to the A/B testing guide.

@suzannehamilton

### DIFF
--- a/source/guides.html.md.erb
+++ b/source/guides.html.md.erb
@@ -26,6 +26,7 @@ Which implementation you choose depends on your A/B testing needs:
 
 * **Pro:** No content flashing
 * **Con:** Requires changes to the CDN configuration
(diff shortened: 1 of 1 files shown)


source/guides.html.md.erb
```diff
@@ -26,6 +26,7 @@ Which implementation you choose depends on your A/B testing needs:
 
 * **Pro:** No content flashing
 * **Con:** Requires changes to the CDN configuration
+* **Con:** Because it depends on Fastly, this can only be tested on staging and production
```
suzannehamilton: "fully tested', maybe? You can still test your A and B versions in other environments, it's just not a completely realistic test without the full stack.

tijmenb: 👍 Yes!

PR #51, opened 2017-02-16
Allow apps to be marked "retired"

tijmenb:
Setting `retired: true` for an application will:

- Hide it from the dashboard
- Hide it from the sidebar on the applications page
- Doesn't count it in lists of active apps
- Shows "(retired)" in the page title, shows a link to gds-attic repo

screen shot 2017-02-16 at 10 21 32

Also marks panopticon as retired - https://trello.com/c/dGPNmVub/482-retire-panopticon

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -34,8 +34,8 @@ helpers do
     PublishingApiDocs.pages
   end
 
-  def app_pages
-    AppDocs.pages.sort_by(&:title)
+  def active_app_pages
+    AppDocs.pages.reject(&:retired?).sort_by(&:app_name)
   end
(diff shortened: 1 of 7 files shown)


timblair:
Knowing that an app _is_ retired is one thing, but knowing _why_ it's been retired would be very useful from a historical point of view.

tijmenb:
Good idea, I'll add something to the description.

tijmenb:
@timblair:

screen shot 2017-02-16 at 15 04 22

timblair:
🎉 👍

PR #56, opened 2017-03-02
Update name of govuk_schemas gem

klssmith:
This repo still has the old name for the gem (govuk_schemas_gem instead
of govuk_schemas) which is stopping the homepage from displaying.


tijmenb:
Thank you!!

PR #57, opened 2017-03-02
Documentation for the content-store api.

danielroseman:
This is intended as the start of documenting the data we expose
externally via the public /api/content endpoint; putting it in the
developer docs for now for want of a better place.

Obviously there is some overlap with the schemas documentation which
already exists, but the audience is different.

### DIFF
--- a/source/apis/content-store.html.md.erb
+++ b/source/apis/content-store.html.md.erb
@@ -1,0 +1,142 @@
+---
+layout: api_layout
+title: Content Store
+parent: /apis.html
+---
+
+The [Content Store][content-store-repo] is the store for all published content on
+GOV.UK. When given a path, it responds with the content that should be displayed
+on that path.
+
+While it is primarily intended for the use of the frontend apps, it is also
+exposed externally at `https://www.gov.uk/api/content/<path>`, where `<path>` is
+the full path of the item on GOV.UK.
+
+The full list of fields in a document depends on the document type and schema
+name, which are themselves fields included at the top level; these schemas are
+defined in the [govuk-content-schemas repository][schemas-repo]. However, the
+bulk of the fields are common between all schemas, and these are primarily
+contained in three sections, as detailed below.
+
+## Top level fields
+
+These fields are constant across all document types.
+
+`base_path`: The main path that this content item is served on. Some items will
+contain data for multiple pages; these will all be sub-pages contained under the
+base_path.
+
+`content_id`: The unique identifier for this content item. This uuid remains
+constant throughout the lifetime of a piece of content, and is used to identify
+an item across our systems.
+
+`schema_name`: The individual schema in [govuk-content-schemas][schemas-repo]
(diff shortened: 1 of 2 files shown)


source/apis/content-store.html.md.erb
```diff
@@ -0,0 +1,143 @@
+---
+layout: api_layout
+title: Content Store
+navigation_weight: 15
```
klssmith: We're not using navigation weights, so I don't think we need this.

klssmith:
Looks good! The Heroku build might work if you rebase onto master - there was a small fix added a few days ago that might solve the problem, but it all seems to work locally anyway.

danielroseman:
Thanks.
Removed redundant line and rebased onto master, and Heroku build has worked.

source/apis/content-store.html.md.erb
```diff
@@ -0,0 +1,142 @@
...
+
+The full list of fields in a document depends on the document type and schema
+name, which are themselves fields included at the top level; these schemas are
+defined in the [govuk-content-schemas repository][schemas-repo]. However, the
+bulk of the fields are common between all schemas, and these are primarily
+contained in three sections, as detailed below.
+
+## Top level fields
+
+These fields are constant across all schemas.
```
cbaines: I think would be good to clarify here that you are not talking about the schemas, which themselves are JSON objects with fields, but about the different types of content. For example, I think `These fields are constant across all document types` is clearer.

source/apis/content-store.html.md.erb
```diff
@@ -0,0 +1,142 @@
+---
+layout: api_layout
+title: Content Store
+parent: /apis.html
+---
+
+The [Content Store][content-store-repo] is the central store for all published content on
```
cbaines: Its unclear to me what is meant by central here.

source/apis/content-store.html.md.erb
```diff
@@ -0,0 +1,142 @@
...
+`publishing_app`: The app that created this content and is responsible for
+editing it. Examples include "whitehall", "specialist-publisher".
+
+`rendering_app`: The frontend app that is responsible for displaying this
+content. Examples include "government-frontend", "specialist-frontend".
+
+`locale`: The language of the content, eg "en", "ar".
+
+`need_ids`: The IDs of the user needs that this content serves, as denoted by
+the Need API. (Deprecated?)
```
cbaines: I believe this is deprecated, so it would be good to say so.

source/apis/content-store.html.md.erb
danielroseman: Removed.

source/apis/content-store.html.md.erb
danielroseman: Used document types throughout.

source/apis/content-store.html.md.erb
danielroseman: Done.

PR #61, opened 2017-03-08
Add example pages for apps

tijmenb:
This adds a list of pages published or rendered to the application pages.

This data is fetched from the search API. Unfortunately most of the `rendering_app`/`publishing_app` is still missing, so not all pages will have the examples.

## Screenshots

screen shot 2017-03-08 at 09 12 52
screen shot 2017-03-08 at 09 13 06
screen shot 2017-03-08 at 09 13 19


lib/app_data.rb
```diff
@@ -0,0 +1,24 @@
+class Appdata
```
rubenarakelyan: Should that be `AppData` for consistency?

tijmenb: Good catch, amended.

PR #70, opened 2017-03-27
Improve visibility in search engines

tijmenb:
This adds a bunch of meta tags for search engines, Facebook & Twitter (and [by extension Slack][slack]). This should improve the looks of the docs in those services. Also makes sure that search engines can find our sitemap via robots.txt. The sitemap can also be submitted via Google Webmaster Tools, to use that we need to verify this domain with a verification HTML.

[slack]: https://medium.com/slack-developer-blog/everything-you-ever-wanted-to-know-about-unfurling-but-were-afraid-to-ask-or-how-to-make-your-e64b4bb9254#.7762cma17


rubenarakelyan:
Sweet 🎉

PR #71, opened 2017-03-27
Retire multipage-frontend

tijmenb:


rubenarakelyan:
👋  Bye bye multipage-frontend

PR #72, opened 2017-04-03
Add Emergency publishing docs

tijmenb:
This is the first page from the opsmanual that we're moving to this repo.

Best reviewed on preview app:

https://govuk-tech-docs-pr-72.herokuapp.com/opsmanual/emergency-publishing.html

PR on opsmanual: https://github.gds/gds/opsmanual/pull/915

PR #73, opened 2017-04-03
Add an "Opsmanual" section to the docs

tijmenb:
This adds an "Opsmanual" section to the docs. We intend to move the current opsmanual into here (first PR for that is https://github.com/alphagov/govuk-developer-docs/pull/72). I've also moved the two guides into the new section.

### DIFF
--- a/source/opsmanual.html.erb
+++ b/source/opsmanual.html.erb
@@ -1,0 +1,26 @@
+---
+layout: false
+title: Opsmanual
+---
+
+<% content_for :sidebar do %>
+  <ul>
+    <li>
+      <% sub_pages('opsmanual/').each do |page| %>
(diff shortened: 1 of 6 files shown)


source/opsmanual.html.erb
```diff
@@ -0,0 +1,26 @@
...
+<% end %>
+
+<% wrap_layout :layout_with_sidebar do %>
+  <%= partial 'partials/header' %>
+
+<blockquote>
+  <strong>Work in progress</strong><br/><br/>
+
+  We're currently moving <a href='https://github.gds/pages/gds/opsmanual'>
+  the opsmanual</a> to this site. Follow along option
```
whoojemaflip: `under_construction.gif`

PR #76, opened 2017-04-03
Add "Keeping software current"

tijmenb:
We currently have some documentation in the confluence wiki. We'd like to move most of this to this site, so that it's easier to find.

This page comes from:

https://gov-uk.atlassian.net/wiki/display/TECH/Keeping+GOV.UK%27s+software+current.

I've removed some outdated bits.

https://trello.com/c/RfRRgVqU

### DIFF
--- a/source/opsmanual/keeping-software-current.html.md
+++ b/source/opsmanual/keeping-software-current.html.md
@@ -1,0 +1,35 @@
+---
+title: Keeping software current
+parent: /opsmanual.html
+---
+
+# Keeping software current
+
+One of our core values is to use secure and up to date software. This document lays out the recommendations for keeping our Ruby on Rails software current.
+
+## Introduction
+
+We run a lot of Rails applications. This means that we have dependencies on both Rails and Ruby versions.
+
+### Upgrading Rails
+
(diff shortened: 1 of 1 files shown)


source/opsmanual/keeping-software-current.md
```diff
@@ -0,0 +1,35 @@
...
+## Introduction
+
+We run a lot of Rails applications. This means that we have dependencies on both Rails and Ruby versions.
+
+### Upgrading Rails
+
+It's very important that we're running a currently supported version of Rails for all applications, otherwise we aren't covered by security fixes. We should:
+
+- Be running on the current major version - this currently means 4.y.z
+- Maintain our applications at the latest current bugfix release for the minor version we're on (expressed in Gemfile syntax as: ~> X.Y.Z) - this currently means 4.1.8 and 4.2.3
```
tomnatt: These version numbers are going to go out of date very quickly. Can we link to a place which lists latest releases?

tijmenb: I've removed that specific reference. Maybe we should add a ticket to look at adding a page with the currently supported versions somewhere?

PR #79, opened 2017-04-04
Add "Content Preview" documentation

tijmenb:
This will replace:

https://github.gds/pages/gds/opsmanual/2nd-line/applications/content-preview.html

For the desks of @kevindew or @danielroseman.

https://trello.com/c/ZErfvzXs

### DIFF
--- a/source/opsmanual/content-preview.html.md
+++ b/source/opsmanual/content-preview.html.md
@@ -14,15 +14,11 @@ sent for publication. This allows content editors to preview their
 content before making it live, without having to manually copy it into a
 separate version of the publishing tool.
 
-Content Preview only contains applications that read from the
-content-store. Initially, these are:
-
--   government-frontend (Whitehall case studies only)
--   manuals-frontend (manuals published by specialist publisher)
--   contacts-frontend
--   static (supplies assets and templates for the other apps)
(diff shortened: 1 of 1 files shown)


source/opsmanual/content-preview.html.md
```diff
@@ -0,0 +1,80 @@
...
+app, which sits between Varnish and the router.
+
+Draft content items can contain an optional `access_limited` hash in
+their metadata, which contains a list of signon user UIDs corresponding
+to the users who are permitted to see that content. All other users will
+see a 403 Forbidden page.
+
+(This is achieved by the API adapters passing the UID in the headers of
+the request to content-store, in the same way as the govuk-request-id.)
+
```
danielroseman: This needs updating to take account of the auth bypass stuff. Something like:

Publishing apps can also add an `auth_bypass_ids` list to the access limited hash, to allow unauthenticated access for preview or fact checking. The ID is encoded in a JSON Web Token and appended to the URL provided to the users. Authenticating-proxy decodes the token and extracts the ID, and again passes it to the content-store in the request headers, where it is compared with the ID stored on the requested content item.

tijmenb: Updated

PR #82, opened 2017-04-05
Fix in-page navigation for emergency page

tijmenb:
There's a small bug in the template that means that the in-page nav will show the content of a `<ul>` if it directly follows the heading tag. Having a paragraph after the heading fixes this.

## Before

screen shot 2017-04-05 at 13 35 30

## After

screen shot 2017-04-05 at 13 37 48

cc @36degrees

### DIFF
--- a/source/opsmanual/emergency-publishing.html.md
+++ b/source/opsmanual/emergency-publishing.html.md
@@ -29,6 +29,8 @@ required will be supplied by the GOV.UK on-call escalations contact:
 
 ### Prerequisites
 
+You'll need this:
+
 - Text for the heading
 - Text for 'extra info', which is a sentence displayed under the heading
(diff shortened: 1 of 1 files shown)


36degrees:
That's a bit odd… 😕

36degrees:
```html
<h3 id="prerequisites" class="anchored-heading">
        <a href="#prerequisites" class="anchored-heading__icon" aria-hidden="true"></a>
        Prerequisites
      </h2>
```

🤔

[Yeah, that'd do it.](https://github.com/alphagov/govuk-developer-docs/blob/master/lib/tech_docs_html_renderer.rb#L22)

tijmenb:
Haha okay, our bad :)

PR #83, opened 2017-04-05
Add page with GitHub topics + show topics on app pages

tijmenb:
Topics are tags for repositories (https://github.com/blog/2309-introducing-topics). 

We've tagged most of the GOV.UK repos with `govuk` now. This page explains that, and links to other topics that have been used. Also shows the topics on the application pages.

https://trello.com/c/CdGSaozQ

### DIFF
--- a/lib/app_docs.rb
+++ b/lib/app_docs.rb
@@ -1,12 +1,16 @@
 class AppDocs
   def self.pages
-    YAML.load_file('data/applications.yml').map do |app_data|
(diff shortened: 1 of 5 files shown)


lib/app_docs.rb
```diff
@@ -1,6 +1,6 @@
 class AppDocs
   def self.pages
-    YAML.load_file('data/applications.yml').map do |app_data|
+    @pages ||= YAML.load_file('data/applications.yml').map do |app_data|
```
whoojemaflip: this will work but it's 😨

PR #85, opened 2017-04-05
Getting started guide

whoojemaflip:
Initial integration of the Getting Started documentation into this app.

screen shot 2017-04-05 at 17 12 35

### DIFF
--- a/config/tech-docs.yml
+++ b/config/tech-docs.yml
@@ -8,7 +8,7 @@ description: Documentation for developers working on GOV.UK
 
 header_links:
   Dashboard: /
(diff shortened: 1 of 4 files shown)


config.rb
```diff
@@ -51,6 +51,13 @@ def opsmanual_pages
 
 ignore 'templates/*'
 
+GettingStarted.pages.each do |page|
+  proxy "/getting_started/#{page.path}.html", "templates/getting_started.html", locals: {
```
tijmenb: I think we use dashes in other pages (`getting-started`).

PR #89, opened 2017-04-06
Use GOV.UK CI instead of CircleCI 

tijmenb:
https://trello.com/c/qqdCvswe


whoojemaflip:
Will there still be a branch deploy to Heroku after this change?

tijmenb:
Yup, that's a separate process.

PR #95, opened 2017-04-07
Add minimum viable search via Google

andrewgarner:
See also:

- https://trello.com/c/vSbk1HoF
- https://github.com/alphagov/tech-docs-template/issues/66


tijmenb:
👍 cool

For context, I wanted to use [Algolia docsearch](https://community.algolia.com/docsearch/), a system that is used by among others React, Middleman and Stripe. But there's some contention over this because the search doesn't work with a screen reader. We'll revisit this after the import of the opsmanual.

PR #101, opened 2017-04-10
Move dev-docs specific changes to the HTML renderer out of /lib

whoojemaflip:
To keep our application specific code seperate from the template app,
we are trying to keep our changes in the /app folder.


app/developer_docs_renderer.rb
```diff
@@ -0,0 +1,19 @@
+class DeveloperDocsRenderer < TechDocsHTMLRenderer
+  def header(text, header_level)
+    anchor = githubify_fragment_id(text)
+    tag = "h" + header_level.to_s
+    "<#{tag} id='#{anchor}' class='anchored-heading'>#{text}</#{tag}>"
```
tijmenb: I think we don't need `class='anchored-heading'` here

PR #110, opened 2017-04-11
Fix titles and sections

tijmenb:
This fixes a bunch of titles to make them descriptive and consistent. See commits for details.

https://trello.com/c/l33LPxcU

### DIFF
--- a/source/manual/assets.html.md
+++ b/source/manual/assets.html.md
@@ -1,5 +1,5 @@
 ---
-title: GOV.UK assets
+title: "Assets: how they work"
 section: Assets
 layout: manual_layout
 parent: "/manual.html"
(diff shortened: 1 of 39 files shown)


source/manual/assets.html.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: GOV.UK assets
+title: Assets: how they work
```
klssmith: This title needs to be in quotes because of the colon - it raises an error that the page doesn't have a title otherwise.

klssmith:
Looks good to me 👍

source/manual/assets.html.md
tijmenb: Thanks, fixed!

PR #114, opened 2017-04-11
Replace all mentions of opsmanual_layout

klssmith:
A few pages still had the layout as "opsmanual_layout" and the page
parent as the opsmanual instead of the new names for these. This was
causing the build to fail and these pages not to display.


tijmenb:
Thank you!

PR #124, opened 2017-04-12
Add an "API" for page freshness

tijmenb:
This JSON endpoint can be called by a hypothetical bot that announces soon to expire pages to Slack.

https://trello.com/c/ZUd0ITTZ


whoojemaflip:
this could probably use a test, if we're going to integrate our hypothetical bot

tijmenb:
@whoojemaflip added

PR #125, opened 2017-04-12
Fix doc urls

andrewgarner:
Adds a number of filters that ensure external markdown docs render correctly:

- Loads external sources
- Rewrites relative links to work in this context
- Rewrites links to other markdown links to their dev docs equivalent

We have had to duplicate some of the TOC logic as this is only applied to markdown templates. Ideally we could do this for all ERB templates.

https://trello.com/c/nGPt3dpG

### DIFF
--- a/app/external_doc.rb
+++ b/app/external_doc.rb
@@ -1,13 +1,138 @@
+require 'html/pipeline'
+require 'uri'
+
 class ExternalDoc
-  def self.fetch(url_to_markdown)
-    markdown = HTTP.get(url_to_markdown)
+  def self.fetch(repository:, path:)
+    contents = HTTP.get(
+      "https://raw.githubusercontent.com/#{repository}/master/#{path}",
+    )
 
-    # remove the own title of the page
-    markdown = markdown.lines[2..-1].join.strip
+    context = {
+      base_url: URI.join(
+        'https://github.com',
+        "#{repository}/blob/master/",
+      ),
 
-    # Make sure we link to the pages hosted here
-    markdown = markdown.gsub('.md', '.html')
+      image_base_url: URI.join(
+        'https://raw.githubusercontent.com',
+        "#{repository}/master/",
+      ),
+    }
 
-    markdown
+    context[:subpage_url] =
+      URI.join(context[:base_url], File.join('.', File.dirname(path), '/'))
+
+    context[:image_subpage_url] =
+      URI.join(context[:image_base_url], File.join('.', File.dirname(path), '/'))
+
+    filters = [
+      HTML::Pipeline::MarkdownFilter,
+      HTML::Pipeline::AbsoluteSourceFilter,
+      PrimaryHeadingFilter,
+      HeadingFilter,
+      AbsoluteLinkFilter,
+      MarkdownLinkFilter,
+    ]
+
+    HTML::Pipeline
+      .new(filters)
+      .to_html(contents, context)
+  end
+
+private
+
+  # When we import external documentation it can contain relative links to
+  # source files within the repository that the documentation resides. We need
+  # to filter out these types of links and make them absolute so that they
+  # continue to work when rendered as part of GOV.UK Developer Docs.
+  #
+  # For example a link to `lib/link_expansion.rb` would be rewritten to
+  # https://github.com/alphagov/publishing-api/blob/master/lib/link_expansion.rb
+  class AbsoluteLinkFilter < HTML::Pipeline::Filter
+    def call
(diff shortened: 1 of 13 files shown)


app/github_repository.rb
```diff
@@ -1,6 +1,6 @@
 require 'octokit'
 
-class GitHub
+class GitHubRepository
```
tijmenb: I think this is "repository" as in "repository pattern"? Maybe this could be confused with a GitHub repo? Perhaps something like `GithubRepoFetcher` or something?

app/external_doc.rb
```diff
@@ -1,13 +1,121 @@
...
+  end
+
+  class PrimaryHeadingFilter < HTML::Pipeline::Filter
+    def call
+      doc.at('h1:first-of-type').unlink
+      doc
+    end
+  end
+
+  class TableOfContentsFilter < HTML::Pipeline::Filter
```
tijmenb: I'm not sure what this is doing, but is this re-implementing the anchored-heading stuff from the docs? I think that's done with JS so this might not be necessary.

app/external_doc.rb
```diff
@@ -1,13 +1,121 @@
...
+          element['href'] = uri.to_s
+        end
+      end
+
+      doc
+    end
+  end
+
+  class PrimaryHeadingFilter < HTML::Pipeline::Filter
+    def call
```
tijmenb: Perhaps add a comment:

> Remove the H1 from the page so that we can choose our own title

app/external_doc.rb
```diff
@@ -1,13 +1,121 @@
...
+      doc
+    end
+
+    def subpage_url
+      context[:subpage_url]
+    end
+  end
+
+  class MarkdownLinkFilter < HTML::Pipeline::Filter
+    def call
```
tijmenb: Can you add a short comment here explaining the intention?

app/external_doc.rb
```diff
@@ -1,13 +1,121 @@
...
+
+    HTML::Pipeline
+      .new(filters)
+      .to_html(contents, context)
+  end
+
+private
+
+  class AbsoluteLinkFilter < HTML::Pipeline::Filter
+    def call
```
tijmenb: Can you add a short comment here explaining the intention?

app/external_doc.rb
```diff
@@ -1,13 +1,121 @@
...
+      HTML::Pipeline::AbsoluteSourceFilter,
+      PrimaryHeadingFilter,
+      TableOfContentsFilter,
+      AbsoluteLinkFilter,
+      MarkdownLinkFilter,
+    ]
+
+    HTML::Pipeline
+      .new(filters)
+      .to_html(contents, context)
```
tijmenb: I'm probably wrong, but could this thing output markdown so that middleman will do the markdown -> HTML?

app/publishing_api_docs.rb
```diff
@@ -21,12 +21,24 @@ def initialize(filename, title)
       @title = title
     end
 
+    def branch
+      'master'
```
tijmenb: I think we can make `master` the only option, maybe we shouldn't allow other branches.

app/external_doc.rb
andrewgarner: Fixed with 75fa6ce52e4a994ca7f23175a703c3f67938800a

app/external_doc.rb
andrewgarner: Fixed with b2361e5

app/external_doc.rb
andrewgarner: Fixed with 6c5c60c

app/external_doc.rb
andrewgarner: We need to add a unique ID to all the header elements so that the table of contents behaviour upstream can work. The class name was a bit misleading. I've fixed with 1020a22.

app/publishing_api_docs.rb
andrewgarner: I've removed with 3ca61b0 since it's no longer needed here.

app/external_doc.rb
andrewgarner: We could maybe use something like https://github.com/xijo/reverse_markdown to return Markdown. That doesn't quite feel right though. I think we'd better off extracting some of the behaviour from `DeveloperDocsRenderer` and making it available to all rendering engines.

app/github_repository.rb
andrewgarner: Fixed with 483087e

PR #126, opened 2017-04-12
Make context of Icinga alert pages clearer

tijmenb:
The pages in the "Icinga alerts" section [like this one][1]  are linked from the Icinga UI:

screen shot 2017-04-12 at 17 15 40

However, when you visit the page the context isn't really made clear. 

screen shot 2017-04-12 at 17 18 39

This PR adds a block to 1) communicate that this is a page that deals with a specific alert, 2) links back to govuk-puppet, so that the user can find out what triggered the alert. See [these search results for example][2].

screen shot 2017-04-12 at 17 19 08

cc @nickcolley @fofr because I'm randomly adding a new UI element here, what do you think?

[1]: https://docs.publishing.service.gov.uk/manual/alerts/rabbitmq-high-watermark.html
[2]: https://github.com/alphagov/govuk-puppet/search?utf8=%E2%9C%93&q=rabbitmq-high-watermark

### DIFF
--- a/source/stylesheets/screen.css.scss
+++ b/source/stylesheets/screen.css.scss
@@ -27,12 +27,9 @@
   background-color: $govuk-blue;
   color: $page-colour;
 
-  a:link {
-    color: $page-colour;
-  }
-
+  a:link,
   a:visited {
-    color: $page-colour;
+    color: inherit;
   }
 
   a:hover {
--- a/source/stylesheets/screen.css.scss
+++ b/source/stylesheets/screen.css.scss
@@ -19,3 +19,27 @@
   border: 1px solid $error-colour;
   color: $error-colour;
 }
+
+.icinga-alert-backlink {
+  padding: $gutter;
+  margin-top: $gutter;
+  margin-bottom: $gutter;
+  background-color: $govuk-blue;
+  color: $page-colour;
+
+  a:link {
+    color: $page-colour;
+  }
+
+  a:visited {
+    color: $page-colour;
+  }
+
+  a:hover {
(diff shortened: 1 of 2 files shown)


fofr:
A list of our content call out patterns are towards the bottom here:
http://govuk-component-guide.herokuapp.com/components/govspeak

Though I quite like what you've done. We also know that those patterns aren't doing their job well.

nickcolley:
I'd prefer to have consistency in the callouts but we don't have consistency with the current callouts yet anyway so this seems fine.

source/stylesheets/screen.css.scss
```diff
@@ -19,3 +19,27 @@
   border: 1px solid $error-colour;
   color: $error-colour;
 }
+
+.icinga-alert-backlink {
```
nickcolley: Rather than giving this a name that is specific to it's use you could consider it's purpose, that way it can be reused across the manual if needed.

For example 'information-callout'

source/stylesheets/screen.css.scss
```diff
@@ -19,3 +19,27 @@
...
+  padding: $gutter;
+  margin-top: $gutter;
+  margin-bottom: $gutter;
+  background-color: $govuk-blue;
+  color: $page-colour;
+
+  a:link {
+    color: $page-colour;
```
nickcolley: You could simplify this by doing the following

```css
a:link,
a:visited {
  color: inherit;
}
```

tijmenb:
Thanks @nickcolley @fofr !

PR #142, opened 2017-04-13
Add a page for the link-checker-api

tijmenb:
I don't know which team owns this.

@kevindew

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -120,6 +120,8 @@
 
 - github_repo_name: link-checker-api
   type: APIs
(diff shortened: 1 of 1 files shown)


kevindew:
Thanks for adding this @tijmenb - it's owned by Content History team, Antonia is the product manager. Aside from this is there anything I should raise for these docs?

tijmenb:
Updated. @kevindew nope, this will add a link on the homepage and create the page. If you have API docs, they're welcome over here: https://docs.publishing.service.gov.uk/apis.html

PR #144, opened 2017-04-18
Change to Operations Manual

surminus:
Everything in this part of the manual used to be in the opsmanual, so
change the title to reflect it's mostly about operations.

### DIFF
--- a/source/manual.html.erb
+++ b/source/manual.html.erb
@@ -1,6 +1,6 @@
 ---
 layout: false
-title: Dev manual
+title: Operations Manual
 ---
 
 <% wrap_layout :header_footer_only do %>


tijmenb:
So it's _mostly_ about operations, but not entirely (https://docs.publishing.service.gov.uk/manual/ab-testing.html, https://docs.publishing.service.gov.uk/manual/keeping-software-current.html for example). Plus I'm hoping there's going to be more docs & guides in this section (the RFCs, how to upgrade to newer rails version, what libraries to use). 

The reason not to call it Opsmanual was that it might easily be confused with the old Opsmanual, which unfortunately needed to stay around. "Operations manual" might have the same problem?

surminus:
Hmm, I think you're probably right, "dev manual" just sounds like it's for developers rather than anyone else. I'm not sure now.

tijmenb:
"Tech manual", or maybe just "Manual"?

MatMoore:
Devops manual!

nickcolley:
It should definitely reflect Ops not be generic 👍

surminus:
Tech Manual would work for me.

tijmenb:
New attempt: https://github.com/alphagov/govuk-developer-docs/pull/170

PR #150, opened 2017-04-18
Remove "Create a new environment for GOV.UK"

tijmenb:
I have a hunch this page is no longer useful. What do you think @surminus?


surminus:
This is still relevant if we ever need to rebuild a VMWare vCloud environment somewhere. Until we've moved elsewhere I think we keep it, but it may need a check to see if it's up to date.

tijmenb:
👍

PR #156, opened 2017-04-19
A11Y, Spacing fixes

nickcolley:
- Aligns grid columns by adding `grid-row` class
- Adds label to search input for assistive tech
- Moves search input into the navigation so it uses the same padding and aligns, means it's hidden in the table contents dropdown smaller screens though 🤔 
- Update team tools sub heading so the IDs do not conflict

Before:

screen shot 2017-04-19 at 11 47 07

After:

screen shot 2017-04-19 at 11 46 54

---

Before:

screen shot 2017-04-19 at 11 50 23

After:

screen shot 2017-04-19 at 11 50 12

---

Before:

screen shot 2017-04-19 at 11 53 51

After:

screen shot 2017-04-19 at 11 53 57

### DIFF
--- a/source/stylesheets/screen.css.scss
+++ b/source/stylesheets/screen.css.scss
@@ -54,3 +54,37 @@
   background-color: $canvas-colour;
   padding: 0 20px 20px 20px;
 }
+
+.technical-documentation {
+  .entry-list {
+    list-style: none;
+    padding-left: 0;
+  }
+  .entry-list__item {
+    margin-top: 0;
+    margin-bottom: $gutter-half;
+  }
+  .entry-list__key {
+    display: block;
+  }
+  @include media(tablet) {
+    .entry-list {
+      display: table;
+    }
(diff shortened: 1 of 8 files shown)


source/stylesheets/screen.css.scss
```diff
@@ -54,3 +54,37 @@
...
+  .entry-list {
+    list-style: none;
+    padding-left: 0;
+  }
+  .entry-list__item {
+    margin-top: 0;
+    margin-bottom: $gutter-half;
+  }
+  .entry-list__key {
+    display: block;
```
tijmenb: Can we set a minimum width for this, so that most sections will align?

nickcolley: It looks nicer but I think it makes it harder for some of the shorter keys.

tijmenb: Ok 👍

PR #158, opened 2017-04-19
Document search field definitions

MatMoore:
This is a first stab at making the documentation better for rummager. I've added an autogenerated list of document fields.

This pulls descriptions out of the JSON file that configures the rummager schema.

I've also put a notice on the search api page itself as it's quite wordy and hasn't been updated for a long time. I'd like to split this information out into shorter pages, eg a quickstart for using the api, and a guide to using filters and facets.

I'm also wondering if the README itself would make a better landing page for the search API?
https://github.com/alphagov/rummager/blob/master/README.md

### DIFF
--- a/source/layouts/api_layout.html.erb
+++ b/source/layouts/api_layout.html.erb
@@ -10,7 +10,10 @@
         <% end %>
       </ul>
     </li>
-    <li><%= sidebar_link 'Search API', '/apis/search-api.html' %></li>
+    <li><%= sidebar_link 'Search API', '/apis/search/search-api.html' %></li>
+    <ul>
+      <li><%= sidebar_link "Field reference", "/apis/search/fields.html" %></li>
+    </ul>
   </ul>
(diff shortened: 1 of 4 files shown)


source/layouts/api_layout.html.erb
```diff
@@ -10,7 +10,10 @@
         <% end %>
       </ul>
     </li>
-    <li><%= sidebar_link 'Search API', '/apis/search-api.html' %></li>
+    <li><%= sidebar_link 'Search API', '/apis/search/search-api.html' %></li>
+    <ul>
```
dwhenry: I feel like this ul should be nested inside the li tag if it's a submenu item. :)

MatMoore: This one is supposed to be top level. I'm trying to follow the same pattern as publishing api - some kind of overview page with a bunch of child pages.

PR #171, opened 2017-04-24
Update create-a-gpg-key.html.md

davidbasalla:
I followed the docs but got the following error when sending to the default keyserver (from gpgtools):

```
gpg: sending key XXXX to hkps server hkps.pool.sks-keyservers.net
gpgkeys: HTTP post error 60: SSL certificate problem: Invalid certificate chain
```

Updates the docs with a hint for anyone else running into this...


tijmenb:
🔑

PR #175, opened 2017-04-26
Add Icinga alert pages for search healthchecks

suzannehamilton:
https://trello.com/c/TXeHcOnY/53-revive-the-search-healthcheck

I've set the review date to be in two months' time because we may make some changes to the healthcheck this quarter.

### DIFF
--- a/source/manual/alerts/search-spelling-suggestions.html.md
+++ b/source/manual/alerts/search-spelling-suggestions.html.md
@@ -1,0 +1,23 @@
+---
+owner_slack: '#search-team'
+review_by: 2017-06-26
+title: 'Check for spelling suggestions failed'
+parent: /manual.html
+layout: manual_layout
+section: Icinga alerts
+---
+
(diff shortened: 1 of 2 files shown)


source/manual/alerts/search-spelling-suggestions.html.md
```diff
@@ -0,0 +1,23 @@
+---
+owner_slack: '#search-team'
+review_by: 2017-06-26
+title: 'Test spelling suggestions work for example queries failed'
```
tijmenb: I had to read this a couple of times before I understood what was going on :)

Would something like "Check for spelling suggestions failed" work?

suzannehamilton: Yeah, this error message is a bit clunky because Jenkins adds the word "failed" to the service name. I've used your suggestion and updated alphagov/govuk-puppet#5836.

PR #176, opened 2017-04-26
Update deploying.html.md - guest not room

paulheron:
Updated to reflect that 'GOV.UK Release Calendar' needs to be added as a guest (rather than as a room).


issyl0:
I'll do the Jenkins dance to make GitHub enable the merge button...

PR #177, opened 2017-04-26
Add a redirect for the virus scanning page that Google has indexed

issyl0:
- Google has indexed the old URL (`manual/virus-scanning.html`) and so
  it 404s. Redirect to what appears to be the actual page.


sihugh:
Naughty Google.  They need some sort of link checker 🤔

PR #178, opened 2017-04-27
Change GPG Key server address

brenetic:
Port number is required to send the user to the search page for that particular key server.

Also added a link to an overview about GPG key server pools.

### DIFF
--- a/source/manual/create-a-gpg-key.html.md
+++ b/source/manual/create-a-gpg-key.html.md
@@ -52,9 +52,10 @@ Send your key to a keyserver by doing:
 ```
 gpg --send-keys $KEYID
 ```
-If you are having problems uploading your key, it's worth trying another keyserver. Ultimately the uploaded key will be synced to the other keyservers so it doesn't matter which one you chose.
+If you are having problems uploading your key, it's worth trying another keyserver. Those trying to receive your key may be connecting to a different keyserver than the one you sent your key to. This is fine, as the keyservers synchronise, but this may take some time to happen.
 
-You now should be able to find your key on <http://keys.gnupg.net/>
+You now should be able to find your key on <http://keys.gnupg.net:11371/>
 
-It occasionally takes a while for the keyserver to display pushed keys
-due to caching.
+It occasionally takes a while for the keyserver to display pushed keys due to caching.
+
+You can find an overview of the GPG keyserver pools [here](https://sks-keyservers.net/overview-of-pools.php)


cbaines:
> The previous address was redirecting to a random site. The MIT server is currently correct.

This is intentional, as `keys.gnupg.net` is a CNAME to `pool.sks-keyservers.net`, which is a set of keyservers that meet certain criteria [1]. I think its better to use a pool than a single keyserver, as you get more redundancy, and balance the load around.

1: https://sks-keyservers.net/overview-of-pools.php

source/manual/create-a-gpg-key.html.md
```diff
@@ -54,7 +54,7 @@ gpg --send-keys $KEYID
 ```
 If you are having problems uploading your key, it's worth trying another keyserver. Ultimately the uploaded key will be synced to the other keyservers so it doesn't matter which one you chose.
 
-You now should be able to find your key on <http://keys.gnupg.net/>
```
cbaines: You might want to consider adding the port 11371 to the docs, so http://keys.gnupg.net:11371/ . I forget the details of how SKS (the bit of software most if not all of the keyservers will be running) works, but this seems to help to access the web interface.

brenetic:
Not sure I know enough about the pools, but if the docs are meant to send someone to a place where they can search for their uploaded (or any) key then whatever page we send them to should have that ability.

Should we add extra docs on the pooling?

brenetic:
@cbaines Updated that 👍

cbaines:
> Not sure I know enough about the pools, but if the docs are meant to send someone to a place where they can search for their uploaded (or any) key then whatever page we send them to should have that ability.
> 
> Should we add extra docs on the pooling?

I think one potential issue with this documentation at the moment is that when you send a key to a pool, it will take a little while for every server in the pool to receive it (as I don't think the server <-> server synchronisation is necessary low latency with respect to distributing keys).

A paragraph below says:
> It occasionally takes a while for the keyserver to display pushed keys due to caching.

I don't know what caching its referring to, and most people by default won't be using "the" keyserver, as they will be using a pool of keyservers, but I think the point about it not being instant is good. 

It might be simplest to change this to be:

> Send your key to a keyserver by doing:
> 
> `gpg --send-keys $KEYID`
> 
> Those trying to receive your key may be connecting to a different keyserver than the one you sent your key to, this is fine, as the keyservers synchronise, but this may take some time to happen.

What do you think?

brenetic:
Yup that sounds pretty reasonable. I have added it to the PR and updated

PR #182, opened 2017-04-28
Update Transition docs

Rosa-Fox:
[Trello](https://trello.com/c/Rzil6Irn/74-consolidate-transition-and-bouncer-documentation)

### DIFF
--- a/source/manual/bouncer.html.md
+++ b/source/manual/bouncer.html.md
@@ -1,0 +1,94 @@
+---
+owner_slack: '#navigation'
+title: Bouncer
+section: Transition
+layout: manual_layout
+parent: "/manual.html"
+old_path_in_opsmanual: "../opsmanual/2nd-line/applications/bouncer-and-transition.html"
+last_reviewed_on: 2017-05-03
+review_in: 6 months
+---
+
+# Bouncer
+
+[Bouncer](https://github.com/alphagov/bouncer) is a Ruby/Rack web app that receives requests for the URLs of government sites that have either been transitioned to GOV.UK, archived or removed. It queries the database it shares with Transition and replies with a
+redirect, an archive page or a 404 page. It also handles `/robots.txt`
+and `/sitemap.xml` requests.
+
+Transition is a Rails app that allows users in transitioning
+organisations and at GDS to view, add and edit the mappings used by
+Bouncer. It also presents traffic data sourced from CDN logs and logs
+provided by transitioning organisations (though this latter activity has
+now ended).
+
+## Bouncer's stack
+
+### DNS
+
+When sites transition they are generally CNAMEd to a domain we control that
+points to our CDN (an A record is used for root domains which can't be CNAMEd).
+
+Some sites partially transition, which means that they redirect some paths to
+their AKA domain, which is CNAMEd to us.
+
+GDS doesn't control the DNS for most transitioned domains, except for some domains such as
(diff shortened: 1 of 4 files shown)


source/manual/bouncer.html.md
```diff
@@ -0,0 +1,89 @@
...
+which adds and removes domains to and from the service.
+That job fetches the list of domains which should be configured at the CDN from
+Transition's [hosts API](https://transition.publishing.service.gov.uk/hosts), so
+will fail if that is unavailable.
+
+[More information about Bouncer's Fastly service](https://docs.publishing.service.gov.uk/manual/cdn.html#bouncer39s-fastly-service)
+
+### Machines
+
+Bouncer runs on 3 machines in the `redirector` vDC (`bouncer-[1-3].redirector`),
```
jennyd: It's probably worth mentioning the DR machines and automatic fallback here too.

source/manual/bouncer.html.md
```diff
@@ -0,0 +1,89 @@
...
+title: Bouncer
+section: Transition
+layout: manual_layout
+parent: "/manual.html"
+old_path_in_opsmanual: "../opsmanual/2nd-line/applications/bouncer-and-transition.html"
+---
+
+# Bouncer
+
+[Bouncer](https://github.com/alphagov/bouncer) is a Ruby/Rack web app that receives requests for the old URLs,
```
36degrees: I'm not sure it's clear what 'old URLs' refers to here – can this be a little more specific?

source/manual/bouncer.html.md
```diff
@@ -0,0 +1,89 @@
...
+### DNS
+
+When sites transition they are generally CNAMEd to a domain we control that
+points to our CDN (an A record is used for root domains which can't be CNAMEd).
+
+Some sites partially transition, which means that they redirect some paths to
+their AKA domain, which is CNAMEd to us.
+
+GDS doesn't control the DNS for most transitioned domains, except for
+`*.direct.gov.uk`, `*.businesslink.co.uk`, `*.alphagov.co.uk` etc. If the DNS
```
36degrees: The use of etc is confusing here. Is this is an exhaustive list or not?

source/manual/data-sources-for-transition.html.md
```diff
@@ -0,0 +1,26 @@
...
+
+# Data sources for Transition
+
+Site configuration is automatically imported on an hourly basis via
+[deploy.publishing](https://deploy.publishing.service.gov.uk/job/Transition_load_site_config/) from
+[transition-config](https://github.com/alphagov/transition-config).
+
+Pre-transition traffic data is imported from
+[pre-transition-stats](https://github.com/alphagov/pre-transition-stats),
+based on logs provided by transitioning organisations. This was updated
```
36degrees: Has the whole thing (importing traffic data) come to an end, or just the updating by hand? If this is historical, suggest rewriting the whole paragraph so it's in the past tense.

jennyd: Importing data for sites before they transition happens on a one-off basis - if we decide we need to do this for a particular site, we'd get logfiles from the org, process them manually, commit them to this repo and they would get imported - that only needs to happen once.

We did this for local.direct.gov.uk last year, and discussed doing it for at least one of the upcoming PHE transitions, and it should be considered as an option for other future ones, even if we only do it rarely.

Rosa-Fox:
Thanks @36degrees and @jennyd. I have made the changes, let me know if you have any other suggestions.

PR #183, opened 2017-04-28
Review pact-broker and blocking-apps-from-release pages for 2ndline

issyl0:
- Seems to all make sense and the links work.


36degrees:
Is there a reason for the two different dates? Is it just 'pick a random date in the not too distant future' or is there meant to be a system?

issyl0:
In this case it's because I can't do basic arithmetic. I meant to pick three months in the future. :-P

issyl0:
I just rebased this on top of Tijmen's changes and can confirm that the last reviewed at approach is easier!

PR #184, opened 2017-04-28
Add some information about Fastly logging

boffbowsh:
Specifies that the logging isn’t configured using VCL or other script.

### DIFF
--- a/source/manual/cdn.html.md
+++ b/source/manual/cdn.html.md
@@ -32,6 +32,12 @@ Most of the CDN config is versioned and scripted:
 These are deployed to [staging][staging_cdn] and [production][production_cdn].
 There is no CDN in integration.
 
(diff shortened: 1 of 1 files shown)


source/manual/cdn.html.md
```diff
@@ -32,6 +32,11 @@ Most of the CDN config is versioned and scripted:
 These are deployed to [staging][staging_cdn] and [production][production_cdn].
 There is no CDN in integration.
 
+Some configuration isn't scripted, such as logging. The www, bouncer and assets
+services stream logs to the `logs-cdn-1` box in each environment. We also log
```
jennyd: Bouncer logs also go to transition-logs-1.

boffbowsh: Added.

PR #185, opened 2017-04-28
Make review_by easier to work with

tijmenb:
Since the `review_by` date was introduced I've seen a few questions come up around what a good review date is (https://github.com/alphagov/govuk-developer-docs/pull/183 for example). I've also noticed that I find it hard to pick a good date (since it involves some arithmetic).

This PR tries make it easier by replacing `review_by` with `last_reviewed_at` and `review_in`. The idea is that you don't have to think about the dates anymore, just insert todays date.

Thoughts?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem 'webmock', '~> 2.1'
 gem 'activesupport', '~> 5.0'
 gem 'govuk-lint', '~> 2.1'
 gem 'simplecov'
+gem 'chronic'
 
 # Middleman Gems
 gem 'middleman', '>= 4.0.0'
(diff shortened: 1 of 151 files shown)


36degrees:
Seems like a sensible approach.

Looks like some of the content is being nuked (by the script?) as part of the PR though.

tomnatt:
Seeing the hard coded Slack channels raises some questions for me about the ownership. The channels are tied to the teams, and the teams change so we'll need to review all of the ownerships in detail every quarter.

That doesn't affect this PR though - it's a good change.

tijmenb:
Updated PR.

timblair:
In Ruby-land, due to ActiveRecord, an `_at` suffix implies a datetime. For just a date, I'd prefer `last_reviewed_on`.

PR #192, opened 2017-05-02
Correct information about clearing whitehall cache

bevanloon:
Whitehall uses memcached. The document originally suggested reloading
whitehall to clear the cache. This does not work since reloading Whitehall is done gracefully. Restarting memcached on whitehall-frontend machines clears whitehall-frontend caches properly since at the time of writing, memcached does not persist the cache between restarts (https://github.com/memcached/memcached/wiki/ProgrammingFAQ#can-you-persist-cache-between-restarts).

Also add some information about testing with cache-bust strings.


thomasleese:
Approved as it reads fine and makes sense, but I have no knowledge about whether running `app.restart:memcached` will do the right thing.

bevanloon:
👍 thanks. Added a quick link to memcached docs - basically cache is not persisted between restarts (https://github.com/memcached/memcached/wiki/ProgrammingFAQ#can-you-persist-cache-between-restarts)

PR #198, opened 2017-05-03
Generate nicer snippets for sharing

tijmenb:
The manual pages currently fall back to the default description "Documentation for developers working on GOV.UK". That's not super informative.

This generates a 300 character snippet from the page HTML. The goal is to generate a representative piece of text from the body, so:

- Only text after the h1 (to skip the "This page was imported" text)
- No HTML
- No headings

### DIFF
--- a/app/snippet.rb
+++ b/app/snippet.rb
@@ -1,0 +1,10 @@
+class Snippet
+  def self.generate(html)
+    html
+      .split('</h1>')[1..-1].join # make sure we skip anything before <h1>
+      .gsub(/<h.+?>.+?<\/h.>/, "") # remove headings
+      .gsub(/<\/?[^>]*>/, "") # remove other HTML but keep the contents
+      .squish # remove whitespace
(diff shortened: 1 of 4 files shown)


app/snippet.rb
```diff
@@ -0,0 +1,10 @@
+class Snippet
+  def self.generate(html)
+    html
+      .split('</h1>')[1..-1].join # make sure we skip anything before <h1>
+      .gsub(/(\<h.+\>)/, "") # remove headings
```
whoojemaflip: Is this regex aiming to remove `<h*>` tags or the entire header? I think it's excessively gready either way.

tijmenb: The entire header. Sometimes pages begin with a header (https://docs.publishing.service.gov.uk/manual/deploying.html). I think it's clearer if we skip that. Also avoids a weird sentence with missing punctuation. What would you do for that?

whoojemaflip: It will also remove any other content in an HTML tag on the same line as the header.

Maybe it won't be a problem, but you can be sure by using a less greedy regex.

eg `/<h.>.+?<\/h.>/`

PR #200, opened 2017-05-03
Add jenkins upgrade steps

surminus:


source/manual/jenkins-ci.html.md
```diff
@@ -133,6 +131,82 @@ With this configuration, in our jobs we can choose between 'GitHub' and 'GDS Git
 access the source code repository. The API endpoint name is referenced in Puppet to configure jobs automatically so
 we shouldn't update it.
 
+## Patching
```
tijmenb: Would it make sense to create a separate page for this so it can be found more easily?

surminus: Yeah, makes sense.

surminus:
I added it Patching Jenkins as it's own page.

PR #203, opened 2017-05-04
Remove old emergency banner docs

bevanloon:
Old emergency publishing process has been replaced with a new improved version. 
We kept both versions of the docs for a while so we could easily compare the new and old process and also so we could carry out product review with 2nd line.

We can now confidently deploy and remove banners with the new process so the old one can be deprecated.

Trello card: https://trello.com/c/GZh4Rvpz/923-remove-old-emergency-banner-code-partials-and-tests

[Related PR 1018 in static](https://github.com/alphagov/static/pull/1018)
[Related PR 1217 in frontend](https://github.com/alphagov/frontend/pull/1217)
[Related PR 244 in fabric-scritps](https://github.com/alphagov/fabric-scripts/pull/244)


source/manual/emergency-publishing.html.md
```diff
@@ -1,162 +0,0 @@
----
```
tijmenb: You could set up a redirect for this file here: https://github.com/alphagov/govuk-developer-docs/blob/master/data/redirects.yml

bevanloon: Instead of the renaming of the new version you mean?

tijmenb: Sorry, meant setting up a redirect from `/emergency-publishing-redis.html` -> `/emergency-publishing.html`

PR #207, opened 2017-05-08
Suggest using our airbrake fork when upgrading to rails 5

h-lame:
We have forked airbrake and made a branch from the last release of 4.x that includes a fix to silence the deprecation warnings you get when using rails 5 and airbrake 4.x.  This way we can keep the deployment recording functionality in errbit.

### DIFF
--- a/source/manual/upgrade-rails.html.md
+++ b/source/manual/upgrade-rails.html.md
@@ -29,8 +29,15 @@ will ensure the classes are loaded in both production and development. See
 
 ### Airbrake
 
-Upgrading to Rails 5 requires upgrading the Airbrake gem. This breaks some compatibility with
+Upgrading to Rails 5 requires upgrading the Airbrake gem to avoid deprecation 
+warnings about middleware. However, Airbrake 5 breaks some compatibility with
 our Errbit installation, namely that deployments are no longer recorded.
+
+Instead of upgrading to a released version of the Airbrake 5 gem you can use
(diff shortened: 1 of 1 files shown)


source/manual/upgrade-rails.html.md
```diff
@@ -29,9 +29,16 @@ will ensure the classes are loaded in both production and development. See
 
 ### Airbrake
 
-Upgrading to Rails 5 requires upgrading the Airbrake gem. This breaks some compatibility with
+Upgrading to Rails 5 requires upgrading the Airbrake gem to avoid deprecation 
+warnings about middleware. However, Airbrake 5 breaks some compatibility with
 our Errbit installation, namely that deployments are no longer recorded.
 
+Instead of upgrading to a Airbrake 5 release gem you can use [a branch in our
```
binaryberry: Would "upgrading to **the** Airbrake 5 release gem" be better?

h-lame: `upgrading to a released version of the Airbrake 5 gem` maybe?  (There's more than one version so *the* doesn't seem right, but what I had before is equally clumsy).

binaryberry: yes that sounds good! It's the double 'a' (a Airbrake) that made me tick.

h-lame: Of course, changed in a rebase.

PR #213, opened 2017-05-10
Mark specialist-frontend as retired

fofr:
specialist-frontend was used to render specialist documents. Those were moved to government-frontend in April 2017.

Part of https://trello.com/c/UM7aecOQ/194-final-stage-decomission-specialist-frontend-2


tijmenb:
🔥

PR #217, opened 2017-05-12
Move DB sync info to a new guide

timblair:
Info about syncing data locally was spread about a bit. There's now a new guide in the developer docs about it (see: alphagov/govuk-puppet#5909), so this is just to minimise the info in the Getting Started guide, and link off to the new one.

### DIFF
--- a/source/manual/copying-data-locally.html.md
+++ b/source/manual/copying-data-locally.html.md
@@ -1,0 +1,63 @@
+---
+owner_slack: "#2ndline"
+title: Copying application data locally for development
+section: Manual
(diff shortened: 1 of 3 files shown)


source/manual/copying-data-locally.html.md
```diff
@@ -0,0 +1,63 @@
+---
+owner_slack: "#2ndline"
+title: Copying application data locally for development
```
tijmenb: Since we often refer to this as "replicating" maybe "Replicate application data locally for development" will make it easier to find?

PR #218, opened 2017-05-12
Update the deployment process

tomnatt:
We are experimenting with slotless deployment and this is an updated process.
The old process should be removed if / when we decide to adopt the new one
permanently.

### DIFF
--- a/source/manual/deploying.html.md
+++ b/source/manual/deploying.html.md
@@ -6,7 +6,7 @@ layout: manual_layout
 section: Deployment
 important: true
 last_reviewed_on: 2017-05-12
-review_in: 6 months
+review_in: 1 months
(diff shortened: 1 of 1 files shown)


source/manual/deploying.html.md
```diff
@@ -5,7 +5,7 @@ parent: "/manual.html"
 layout: manual_layout
 section: Deployment
 important: true
-last_reviewed_on: 2017-02-02
+last_reviewed_on: 2017-05-12
 review_in: 6 months
```
tijmenb: Maybe set this lower so that we don't forget to update after the experiment.

tomnatt: Good idea.

PR #224, opened 2017-05-16
Review some expired pages

tijmenb:


source/manual/alerts/defined-cpu-type-does-not-match.html.md
```diff
@@ -1,10 +1,10 @@
 ---
 owner_slack: "#2ndline"
-title: defined cpu type does not match
+title: Defined cpu type does not match
```
andrewgarner: Should we say "CPU"?

source/manual/clone-mysql-slave.html.md
```diff
@@ -1,12 +1,12 @@
 ---
-owner_slack: "#2ndline"
+owner_slack: "#infrastructure"
```
andrewgarner: Should this be `#govuk-infrastructure`?

PR #230, opened 2017-05-18
Update emergency publishing to reflect redesign

bevanloon:
The emergency banner has been redesigned. As a result new options are now being used for the campaign class or type of the emergency banner.

Update the documentation to reflect this. Also neaten up a few phrases.

Related [Trello](https://trello.com/c/eSfTdFEx)
Related PR: https://github.com/alphagov/static/pull/1049


andrewgarner:
Given "a notable death" isn't an emergency is "emergency banner" still a good name for this thing?

bevanloon:
@andrewgarner - possibly not. What do you think would be better? "emergency publishing" is also non descriptive...."exceptional circumstances publishing/banner"?

leenagupte:
@bevanloon When did we removed examples of what the banner looks like?

bevanloon:
@leenagupte - not sure? did we ever have them in the new docs?

leenagupte:
@bevanloon I thought we had, but considering we're drastically changing what they look like on 99% of the pages, it might be good to add examples of what they look like somewhere.

bevanloon:
@leenagupte 👍 indeed.....the example images that are there are currently the old design. I'll grab some new design screenshots to include.

PR #240, opened 2017-05-24
Remove references to the Content API

alecgibson:
The [`content_api`](https://github.com/alphagov/govuk_content_api) has
been retired, and all content has been migrated to the
[`content_store`](https://github.com/alphagov/content-store).

Therefore, we can remove all references to the Content API, and rely
solely on the Content Store.

### Trello

https://trello.com/c/TsUsUwD7/153-remove-documentation-on-content-api

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -107,10 +107,13 @@
   product_manager: Rob Rankin
 
 - github_repo_name: govuk_content_api
+  retired: true
   type: APIs
   puppet_name: contentapi
-  product_manager: Holly Garrett
-  team: Navigation
+  description: |
+    API that used to store and serve published content to the rendering applications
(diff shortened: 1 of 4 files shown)


data/applications.yml
```diff
@@ -106,12 +106,6 @@
   team: Worldwide Publishing
   product_manager: Rob Rankin
 
-- github_repo_name: govuk_content_api
```
tijmenb: This needs to stay, but you can add a `retired: true` flag, add a description what it used to be and remove team/PM:

https://github.com/alphagov/govuk-developer-docs/blob/master/data/applications.yml#L213-L223

That way the page will not 404 and we keep a bit of a record about apps, should that be needed.

alecgibson:
@tijmenb done

PR #243, opened 2017-05-26
Add notes about scripts for refreshing nfs cache

h-lame:
In https://github.com/alphagov/govuk-puppet/pull/5981 we added script versions of the `ls` and `find` commands given in these docs.  Now that is merged we can describe the scripts in the docs and complete the circle.


kevindew:
Just merged this into: https://github.com/alphagov/govuk-developer-docs/pull/250

PR #249, opened 2017-05-30
Review root filesystem is readonly

kevindew:
Update this documentation to have more of an emphasis on who has access to vCloud director instance

### DIFF
--- a/source/manual/alerts/root-filesystem-is-readonly.html.md
+++ b/source/manual/alerts/root-filesystem-is-readonly.html.md
@@ -4,7 +4,7 @@ title: root filesystem is readonly
 parent: "/manual.html"
 layout: manual_layout
 section: Icinga alerts
-last_reviewed_on: 2016-11-27
+last_reviewed_on: 2017-05-30
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manual/alerts/root-filesystem-is-readonly.html.md
```diff
@@ -13,9 +13,15 @@ read only. This could be caused by a cloud hosting provider having
...
 
 The machine will need to have a filesystem check forced on reboot. To do
-this log into the vCloud Director instance for the environment, find the
-affected machine and click on the 'console' image to bring up a web
-based console. Click the 'reset' icon in the top bar and then click to
-focus on the console. Press the 'f' key when prompted (or repeatedly if
-there's no output) to force a filesystem check.
+this you will need access to the vCloud Director instance for the 
+environment, this access is typically only available to members
+of the infrastructure team.
```
surminus: "...of the GOVUK Infrastructure team. However, if you do on-call, you should have access to the vCloud Production environment."

kevindew: Thanks @surminus - I should probably get a login to that before I'm next on call.

PR #258, opened 2017-06-02
Update ab-testing.html.md

tijmenb:
This looks nice.


spec/pages/manual_page_spec.rb
```diff
@@ -3,6 +3,10 @@
     raw = File.read(filename)
     frontmatter = YAML.load(raw.split('---')[1])
 
+    it "uses the correct spelling of GOV.UK" do
+      expect(raw).not_to match "Gov.uk"
```
chao-xian: Would it be good if it was to match `GOV.UK` so that it would also catch `gov.uk`?

PR #261, opened 2017-06-07
Fix 'reboot required by apt' instruction

klssmith:


klssmith:
@tomnatt Thanks for reviewing - I've updated the second example now.

tomnatt:
We probably don't want to run unlock across every machine in an environment. Could you use an example which targets a specific host? Something like:

`fab <environment> -H <machine-name> locksmith.unlock:"<machine-name>"`

PR #263, opened 2017-06-07
Review elasticsearch cluster health page

MatMoore:
- Add some links to ES docs
- Put more useful stuff higher up
- Discourage restarting nodes to fix cluster health issues. This seems to be a solution for a specific situation, but I think there are situations where it will make things worse too.
- Make search team do the next review (we should know about this stuff)

### DIFF
--- a/source/manual/alerts/elasticsearch-cluster-health.html.md
+++ b/source/manual/alerts/elasticsearch-cluster-health.html.md
@@ -1,15 +1,15 @@
 ---
-owner_slack: "#2ndline"
+owner_slack: "#search-team"
 title: Elasticsearch cluster health
 parent: "/manual.html"
 layout: manual_layout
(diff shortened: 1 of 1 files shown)


source/manual/alerts/elasticsearch-cluster-health.html.md
```diff
@@ -28,13 +28,42 @@ solution.
...
+To use this, forward port 9200 from the Elasticsearch box to your localhost:
+
+```
+ssh -L9200:localhost:9200 logs-elasticsearch-1.management.staging
+```
+
+Access the UI at <http://localhost:9200/_plugin/head/>
+
+The tabs on top right corner for Cluster Status & Cluster Health
+come handy
```
suzannehamilton: Should be "come in handy"?

PR #264, opened 2017-06-07
Update elasticsearch backup guidance

MatMoore:
I want to make sure we can follow this guidance for managing GOV.UK
search (the api-elasticsearch cluster) so I've assigned our team as
owner.

I think this is still out of date for logs-elasticsearch because we now
store snapshots on s3 using
https://www.elastic.co/guide/en/elasticsearch/reference/1.7/modules-snapshots.html
but I've left it out of this PR.

Review this in 3 months because we're likely to setup s3 backups for
api-elasticsearch too at some point.

### DIFF
--- a/source/manual/elasticsearch-dumps.html.md
+++ b/source/manual/elasticsearch-dumps.html.md
@@ -1,18 +1,20 @@
 ---
-owner_slack: "#2ndline"
+owner_slack: "#search-team"
 title: 'Elasticsearch: dump and restore indices'
 parent: "/manual.html"
 layout: manual_layout
 section: Backups
-last_reviewed_on: 2017-04-06
-review_in: 6 months
+last_reviewed_on: 2017-06-07
+review_in: 3 months
 ---
 
 ## Creating a dump
 
 Sometimes you may need to take a backup of Elasticsearch before a
-critical operation and all Elasticsearch servers create their own
-nightly backups via the following cronjob:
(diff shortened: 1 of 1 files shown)


source/manual/elasticsearch-dumps.html.md
```diff
@@ -1,49 +1,73 @@
...
+iso_date="$(date --iso-8601=seconds|cut --byte=-19|tr [:upper:] [:lower:])z"
+real_name="$alias_name-$(date --iso-8601=seconds|cut --byte=-19|tr [:upper:] [:lower:])z-00000000-0000-0000-0000-000000000000"
+
+es_dump_restore restore_alias "http://localhost:9200/" "$alias_name" "$real_name" "$backupfile" '{}' $BATCH_SIZE
 ```
 
-If the import fails (as it sometimes does), you can simply re-run the command.
+This will restore each backup into a new index, and then move the alias to point to it.
+
+If you need to change the alias back for any reason, you can run the rummager rake task `rummager:switch_to_named_index[foo]`, where `foo` is the name of the index you want to point the alias to.
```
suzannehamilton: Might be worth mentioning that this automatically matches the alias to the new index name. Otherwise it's a bit surprising that you don't need to specify the alias.

suzannehamilton:
One comment, but otherwise looks OK to me. (With the caveat that I haven't tried restoring an index before.)

PR #272, opened 2017-06-14
Update docs on retiring an application

carvil:
More specifically:
- updated point on DNS changes;
- added an extra point on unpublishing relevant content.

Trello: https://trello.com/c/2MF97xce/154-retire-content-api

### DIFF
--- a/source/manual/retiring-an-application.html.md
+++ b/source/manual/retiring-an-application.html.md
@@ -4,7 +4,7 @@ title: Retire an application
 section: Packaging
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2017-05-03
+last_reviewed_on: 2017-06-14
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manual/retiring-an-application.html.md
```diff
@@ -119,8 +119,13 @@ application should be removed.
 ## 12. Update DNS
 
 Request any public DNS entries be removed. If the app had an admin UI, it will
-have had public DNS entries in the `publishing.service.gov.uk` domain. Request
-that the infrastructure team remove these, or, if appropriate, transition them.
+have had public DNS entries in the `publishing.service.gov.uk` domain.
+
+Follow the instructions [here][dns-changes] in order to remove these, and ask
```
cbaines: I think `Follow the [instructions for DNS changes][dns-changes] in order to remove these` would be better semantically and for accessibility.

carvil: That makes sense, I will change.

PR #278, opened 2017-06-15
Add info on how to fix full MongoDB

chao-xian:
A number of devs have encountered this problem, so it's common enough to warrant adding some documentation around.


chao-xian:
thanks @cbaines

PR #280, opened 2017-06-19
Update pages about redirects

rubenarakelyan:
This commit merges two manual pages related to redirects into one and redirects the deleted page to the current one. It also adds information about using short-url-manager in preference to router-data.


rubenarakelyan:
Good point, I've added some information about that as the first point.

cbaines:
> Good point, I've added some information about that as the first point.

Great, thanks :)

PR #285, opened 2017-06-23
Update encrypted-hiera-data.html.md

Davidslv:
Update links to point to https://github.digital.cabinet-office.gov.uk/gds/deployment/tree/master/pass/2ndline as it seems the previous path doesn't exist.

Let me know if I should point the links somewhere else.


cbaines:
The Rakefile seems to talk about mounting the 2ndline credentials store, so I'm unsure this change is correct.

PR #290, opened 2017-06-26
Add a page for content indexed in search

MatMoore:
See https://github.com/alphagov/rummager/pull/820


tijmenb:
Do we need to add a page like https://github.com/alphagov/govuk-developer-docs/blob/master/source/apis/search/fields.html.md.erb first?

MatMoore:
Yes we do. That was just me failing to use git properly. :man_facepalming:

PR #299, opened 2017-06-27
Replace slave to backup mysql command

SamLR:
This changes the advice for copying backups between servers from copying
private keys to one of the servers to instead use agent forwarding.

Whilst agent forwarding is not ideal ([see this blog for
details](https://lyte.id.au/2012/03/19/ssh-agent-forwarding-is-a-bug/))
it is a strict improvement on copying private keys to a remote server
and having to remember to remove them.

Please review this carefully and confirm the command (I've personally tested it but I'd like to be sure).

### DIFF
--- a/source/manual/clone-mysql-slave.html.md
+++ b/source/manual/clone-mysql-slave.html.md
@@ -5,7 +5,7 @@ section: Databases
 layout: manual_layout
 parent: "/manual.html"
 old_path_in_opsmanual: "../opsmanual/infrastructure/howto/clone-mysql-slave.md"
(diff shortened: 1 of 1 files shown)


stephenharker:
This change is an improvement over copying your private key up to the slave for sure. Better would be to deploy a dedicated and restricted user and keypair to the slaves so they can copy to each other and nowhere else for this purpose.

It's also better I think to use Percona Xtrabackup with "--slave-info" or mysqldump with "--dump-slave" rather than just copying over /var/lib/mysql as those will be more reliable. Perhaps for another pull request...

PR #301, opened 2017-06-27
Fix common 404

tijmenb:
This fixes a number of pages that are currently 404'ing. Their sometimes caused by old links and incorrect relative links.

Found using kibana:

```
@fields.http_host:"docs.publishing.service.gov.uk" AND
@fields.status:404
```


MatMoore:
I also noticed if you chop of the end of a path you can sometimes get a 404. Like [/apis/search/](https://docs.publishing.service.gov.uk/apis/search/)

tijmenb:
@MatMoore I'm not sure how to solve those yet. These redirects are created by creating a HTML file that redirects, so they're not actual 301s.

PR #311, opened 2017-07-05
Add instructions on processing virus scanning queue

suzannehamilton:
Two of the links in this guide were broken because they linked to the old Ops Manual. To fix this, copy two sections from https://github.digital.cabinet-office.gov.uk/attic/opsmanual/blob/master/2nd-line/applications/whitehall.rst:

- Background on how attachments are uploaded and processed by the ClamAV virus scanner
- Instructions on how to bypass the queue and publish all clean files immediately

---

I've tested that the links work and that the directories still exist. I haven't actually tested the command to process the backlog.

I'd appreciate a review from someone who knows about asset uploading - the content I've copied from the ops manual is very whitehall-specific but it sounds like it might apply to other publishing apps too. Is that right?

### DIFF
--- a/source/manual/alerts/virus-scanning.html.md
+++ b/source/manual/alerts/virus-scanning.html.md
@@ -5,22 +5,84 @@ section: Icinga alerts
 layout: manual_layout
 parent: "/manual.html"
 old_path_in_opsmanual: "../opsmanual/2nd-line/alerts/virus-scanning.md"
-last_reviewed_on: 2016-12-27
+last_reviewed_on: 2017-07-05
 review_in: 6 months
 ---
 
-> **This page was imported from [the opsmanual on GitHub Enterprise](https://github.com/alphagov/govuk-legacy-opsmanual)**.
-It hasn't been reviewed for accuracy yet.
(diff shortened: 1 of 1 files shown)


suzannehamilton:
Oh, I missed that this content still exists at https://github.com/alphagov/govuk-legacy-opsmanual/blob/master/2nd-line/applications/whitehall.rst. 🤦‍♀️ Should we just link to that?

sihugh:
If this content can go in developer docs, then it should so that we can find it :-) .  Once it's moved, we should remove it from the legacy ops manual

danielroseman:
For the sake of completeness, we should also document the fact that a separate script - https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/files/node/s_asset_base/virus_scan.sh - regularly rescans all the previously uploaded files in both clean and draft-clean, to catch any newly-released virus signatures. The script is configured in cron to run every hour, but actually takes over 2 days to complete (and starts under a lock so that only one runs at a time).

suzannehamilton:
Thanks Daniel! I've added that to a "Detecting new viruses" subsection.

PR #312, opened 2017-07-06
Add Fastly redirect

sihugh:
As part of adding this, I've sorted the redirects alphabetically and as a result discovered a duplicate.  (`manual/mirror-fallback.html: manual/fall-back-to-mirror.html` was in twice)


data/redirects.yml
```diff
@@ -1,5 +1,22 @@
...
 dictionary.html: manual/dictionary.html
 guides.html: manual.html
+manual/alerts/applications/sidekiq-monitoring.html: manual/setting-up-new-sidekiq-monitoring-app.html
+manual/alerts/publisher-app-health-check-not-ok.html: manual/alerts/publisher-app-healthcheck-not-ok.html
+manual/alerts/whitehall-app-health-check-not-ok.html: manual/alerts/whitehall-app-healthcheck-not-ok.html
+manual/create-page.html: manual/docs-style-guide.html
+manual/creating-pages.html: manual/docs-style-guide.html
+manual/elasticsearch.html: manual/elasticsearch-dumps.html
+manual/emergency-publishing-redis.html: manual/emergency-publishing.html
+manual/fastly-error-rate.html: /manual/alerts/fastly-error-rate.html
```
tijmenb: @sihugh this might not work because of the leading slash.

sihugh: https://github.com/alphagov/govuk-developer-docs/pull/317

PR #323, opened 2017-07-11
Add links from app pages to API docs

suzannehamilton:
Several applications have APIs which are documented separately. Make it easier for users to find docs for the relevant API from the application page.

Joe L was looking for the content-store API docs yesterday and couldn't find them from the app page. It would also have helped me when I was first using the docs - I didn't realise there were separate API docs.


suzannehamilton:
Good call. 👍 I've changed it to use full URLs and added a Mapit link.

PR #332, opened 2017-07-13
Update redirect routes

cbaines:


source/manual/archiving-and-redirecting-content.html.md
```diff
@@ -1,77 +0,0 @@
----
```
tijmenb: Can you set up a redirect for this page? 😄

cbaines: Done in 858d492

PR #336, opened 2017-07-17
Add smart-answers to list of apps

chao-xian:
Smart Answers is also using Heroku Review apps so should be listed.

The GOV.UK Developer Docs repo itself also uses it so should be added
to the list.

### DIFF
--- a/source/manual/review-apps.html.md
+++ b/source/manual/review-apps.html.md
@@ -13,7 +13,7 @@ allow us to automatically deploy application changes into a brand
 new Heroku app so that each Pull Request reviewer has a chance to see how it
 looks with live data.
 
-These are 2 examples of GOV.UK apps that have review apps enabled:
+These are examples of GOV.UK apps that have review apps enabled:
 
 - [collections](https://github.com/alphagov/collections)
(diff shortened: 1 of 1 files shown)


source/manual/review-apps.html.md
```diff
@@ -17,6 +17,8 @@ These are 2 examples of GOV.UK apps that have review apps enabled:
 
 - [collections](https://github.com/alphagov/collections)
 - [government-frontend](https://github.com/alphagov/government-frontend)
+- [govuk-developer-docs](https://github.com/alphagov/) - this project
```
cbaines: I think the link should be `https://github.com/alphagov/govuk-developer-docs`?

source/manual/review-apps.html.md
```diff
@@ -17,6 +17,8 @@ These are 2 examples of GOV.UK apps that have review apps enabled:
```
cbaines: On line 16, it says there are "2 examples". Now that there are more, it would be good to just say something more general.

chao-xian: Good spot @cbaines - fixed

PR #345, opened 2017-07-20
Set parent page for apps by team page

tijmenb:
This will highlight the "Apps" link in the main navigation.


tijmenb:
:man_facepalming:

PR #347, opened 2017-07-25
Add information about training environment

rubenarakelyan:
This commit adds some information about the upcoming training environment.


rubenarakelyan:
👍  I've updated the date and shortened the review time to 3 months so that it gets reviewed roughly after the next firebreak.

PR #349, opened 2017-07-27
Update the Publishing to Puppet Forge docs

issyl0:
- Use `pass` instead of the credstore.
- The password for the Puppet Forge account has moved to the 2ndline
  password store.
- Render numbered lists properly, delimited with `.` instead of `)`.
- Split the "Releasing" section into two - via the website, or via the
  command line, to make it clearer that they're not linked.


issyl0:
Oh, I entirely missed that this had been approved!

PR #351, opened 2017-07-28
Add manual page to show how to search for hardcoded markup

nickcolley:
cc @fofr @kevindew

### DIFF
--- a/source/manual/howto-find-hardcoded-markup-govspeak.html.md
+++ b/source/manual/howto-find-hardcoded-markup-govspeak.html.md
@@ -1,0 +1,53 @@
+---
+owner_slack: '#publishing-frontend'
+title: Find hardcoded markup in GovSpeak
+section: Frontend
+layout: manual_layout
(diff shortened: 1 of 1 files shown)


kevindew:
Looks good. I notice you're looking up HTML strings. Will this be a problem for the (planned to be) growing number of Editions which just have govspeak stored and not the resulting HTML? As in that case would be more sorted to iterate through the content stores.

source/manual/howto-find-hardcoded-markup-govspeak.html.md
```diff
@@ -0,0 +1,53 @@
+---
+owner_slack: '#publishing-frontend'
+title: Find hardcoded markup in GovSpeak
```
fofr: `govspeak` I think

nickcolley:
@kevindew I believe that for the buttons, this would still work since the markup is hardcoded by content designers and is not a govspeak extension.

So this should still be somewhat useful?

PR #354, opened 2017-07-31
Add implied step to PAYE tools release checklist

MatMoore:
The guide said to prep the content, but didn't explicitly say when to publish it.

Also, I almost used the wrong manifest by copy pasting commands. Using an invalid version number should make sure future me doesn't do anything destructive.

### DIFF
--- a/source/manual/hmrc-paye-files.html.md
+++ b/source/manual/hmrc-paye-files.html.md
@@ -30,7 +30,7 @@ The way the updates work:
 -   Binaries are provided for Windows, OS X and Linux
 -   Deltas of the binaries are also provided for partial updates
 -   The software hits a manifest file
-    (`realtimepayetools-update-v14.xml`) to determine which files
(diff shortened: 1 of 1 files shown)


source/manual/hmrc-paye-files.html.md
```diff
@@ -30,7 +30,7 @@ The way the updates work:
 -   Binaries are provided for Windows, OS X and Linux
 -   Deltas of the binaries are also provided for partial updates
 -   The software hits a manifest file
-    (`realtimepayetools-update-v14.xml`) to determine which files
+    (`realtimepayetools-update-vXX.xml`) to determine which files
     to download. This manifest contains the version number and partial
     paths to each of the available binaries. This manifest URL is
     hard-coded in the desktop software and must remain stable
```
dwhenry: As discussed can we clarify that this should remain stable based on the major release version

PR #358, opened 2017-08-03
Add a note about size of files in support-api

SamLR:
The disk comparison check between assets slave & master can fail if
large reports have been compiled against zendesk. This adds a note to
this effect.

### DIFF
--- a/source/manual/alerts/asset-master-slave-disk-space-comparison.html.md
+++ b/source/manual/alerts/asset-master-slave-disk-space-comparison.html.md
@@ -38,3 +38,5 @@ If the disk usage is slowly increasing, the most likely cause of the alert is
 that a large volume of files has been uploaded to Whitehall. In this case the
 alert will only resolve itself when the daily sync job runs so the alert is a
 false alarm and can be acknowledged.
+
+It is worth checking the `support-api/csvs` directory as any reports generated against zendesk get created here, if the report is poorly constrained this can be large (>1GB). These files will be synced daily as part of the `rsync-uploads` task. Use `ls -hltr` to check if any of the recent files are particularly large.


deanwilson:
Can files found in this way be deleted or do they back in use reports?

SamLR:
@deanwilson I have no idea. I think that they're only temporary but would need confirmation from someone who understands zendesk, and I don't know who that might be

deanwilson:
@paulheron - any thoughts?

boffbowsh:
Those reports aren't for ZenDesk, they're reports of anonymous feedback in Feedback Explorer (the What were you doing?, What went wrong? boxes on all GOV.UK pages). We could put them in S3 under a short lifecycle policy as an alternative in the future.

boffbowsh:
The way it works is someone requests an export which is generated in the background. The job emails them when the report is ready and links to the generated CSV. We'd need to make a call on how long to retain them. Maybe a month? Once we've decided on that we could create a cron job to automate it, rather than documenting.

SamLR:
@boffbowsh Unless we set the cronjob to run very regularly and all files are stored on S3 it wont fix the underlying problem which is that a large generated csv (>1GB) will exceed the threshold for difference between master & slave disk usage.

boffbowsh:
Also a good point. In that case is it worth adding a note to just run an rsync manually to fix the problem?

tuzz:
Bump @SamLR

SamLR:
@tuzz  @boffbowsh I'm not sure re-running the sync job is really necessary, it's tricky and the data is (as far as I'm aware) all derived so it's not vital that it's kept. I guess it depends on how bad we think it is to recommend that people ack an alert?

deanwilson:
Does anyone object to what's already been written being merged in? it can then be expanded upon as people get more time.

deanwilson:
Bump @boffbowsh / @SamLR

SamLR:
@deanwilson fine by me

PR #359, opened 2017-08-03
Document how to set up db GUI tools with the dev vm

tuzz:
https://govuk-tech-docs-pr-359.herokuapp.com/manual/using-database-gui-tools.html

Setting up these tools can be a bit fiddly so I thought it might help others to share this information. I'm not sure if there's a better way to deal with the Postgres `pg_hba` trust issue so suggestions there would be appreciated. As it stands, this only requires running a command every ~3 days or so depending on how frequently you run `govuk_puppet`.


davidbasalla:
Great, seems really useful! We could add a mongo section too (not sure if there's a dominant tool, I use Robomongo and Mongohub), but I'm happy to add that a later point.

Strangely, when previewing this file within Github's UI, some of the sub headers don't render correctly. I can't see any problems with the markup code so I assume it'll render fine...

tuzz:
Good idea about Mongo. I don't know the tools personally but if anyone has a recommended tool we could add it to this document.

I'm not sure why that's happening in GutHub's preview. It's rendering ok in the Manual so I'm going to merge.

PR #362, opened 2017-08-07
Update the location of router-data

cbaines:
The repository has moved from GitHub Enterprise, to a private
repository on GitHub.

### DIFF
--- a/data/dashboard.yml
+++ b/data/dashboard.yml
@@ -56,10 +56,7 @@
         - fastly-configure
         - govuk-cdn-config
         - router
-      sites:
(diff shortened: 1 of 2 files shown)


data/dashboard.yml
```diff
@@ -57,8 +57,8 @@
         - govuk-cdn-config
         - router
       sites:
```
tijmenb: You can also remove this section and add `router-data` to the `repos:` array above, so that the repo description is auto-updated.

cbaines: Ok, I've updated the commit.

PR #377, opened 2017-08-17
Add instructions to view Sidekiq queues locally

binaryberry:
cc @brenetic @dwhenry

### DIFF
--- a/source/manual/monitor-sidekiq-workers.html.md
+++ b/source/manual/monitor-sidekiq-workers.html.md
@@ -22,6 +22,14 @@ Then visit [http://127.0.0.1:9000](http://127.0.0.1:9000) to see a list of [Side
 
 You can also monitor Sidekiq queue lengths using [this Grafana dashboard](https://grafana.publishing.service.gov.uk/dashboard/file/sidekiq.json).
 
+To view your local Sidekiq queue, go to the [sidekiq-monitoring app](https://github.com/alphagov/sidekiq-monitoring) in the vm and run `./bin/foreman start` for all applications, or `./bin/foreman run <app_name>` for a specific app. In the output of sidekiq-monitoring, you'll find the port number of the app whose queue you want to view. Make a note of it, then in your local machine, go to govuk-puppet/development-vm and run:
+
+```bash
+$ vagrant ssh  -- -CNL <port_number>:localhost:<port_number>
+```
+
+Then visit `http://localhost:<port_number>/<app_name>` to see a list of Sidekiq configurations you can monitor.
+
 See also: [Add sidekiq-monitoring to your application](setting-up-new-sidekiq-monitoring-app.html).
 
 [gov.uk]: https://www.gov.uk/


source/manual/monitor-sidekiq-workers.html.md
```diff
@@ -22,6 +22,20 @@ Then visit [http://127.0.0.1:9000](http://127.0.0.1:9000) to see a list of [Side
 
 You can also monitor Sidekiq queue lengths using [this Grafana dashboard](https://grafana.publishing.service.gov.uk/dashboard/file/sidekiq.json).
 
+Or to view your local Sidekiq queue, within your vm go to govuk-puppet/development-vm and run:
```
cbaines: The below command doesn't look like it will work within the govuk-puppet devleopment vm.

dwhenry: It should be within your local environment. :)

source/manual/monitor-sidekiq-workers.html.md
```diff
@@ -22,6 +22,20 @@ Then visit [http://127.0.0.1:9000](http://127.0.0.1:9000) to see a list of [Side
...
+
+```bash
+$ vagrant ssh  -- -CNL 3091:localhost:3091
+```
+
+Then in the [sidekiq-monitoring app](https://github.com/alphagov/sidekiq-monitoring), run:
+
+```bash
+./bin/foreman run rummager
```
dwhenry:
```
./bin/foreman start
```
to get all applications or you can run it for an individual application using:
```
./bin/foreman run <app_name>
```

source/manual/monitor-sidekiq-workers.html.md
binaryberry: Thanks both, updated now!

source/manual/monitor-sidekiq-workers.html.md
binaryberry: Thanks, updated now! I've made a couple of additional changes to make the instructions non-app specific.

source/manual/monitor-sidekiq-workers.html.md
```diff
@@ -22,6 +22,16 @@ Then visit [http://127.0.0.1:9000](http://127.0.0.1:9000) to see a list of [Side
 
 You can also monitor Sidekiq queue lengths using [this Grafana dashboard](https://grafana.publishing.service.gov.uk/dashboard/file/sidekiq.json).
 
+To view your local Sidekiq queue, go to the [sidekiq-monitoring app](https://github.com/alphagov/sidekiq-monitoring) in the vm and run `./bin/foreman start` for all applications, or `./bin/foreman run <app_name>` for a specific app.
+
+Make a note of the port number of the app whose queue you want to view. Then in your local machine, go to govuk-puppet/development-vm and run:
```
suzannehamilton: Is there somewhere you can look this up? Puppet?

Also, is it definitely the port number of the app? I would have guessed that it would be the port number for something to do with Sidekiq, but I don't really understand how it all hooks together. :)

binaryberry: Following our chat in meatspace, I've changed the wording to make it clear where you can find the port number: "in the output of sidekiq-monitoring, you'll find the port number of the app whose queue you want to view. ". Does that work better?

PR #385, opened 2017-08-25
Review and update "Data sync" manual page

leenagupte:
Update the "Data sync" manual page to use the word "Copy"

The jobs that perform the data sync are called "Copy ...." if you
search the manual with the name of the job you'd never find this page
because the word "copy" doesn't appear anywhere.

### DIFF
--- a/source/manual/alerts/data-sync.html.md
+++ b/source/manual/alerts/data-sync.html.md
@@ -5,7 +5,7 @@ section: Icinga alerts
 layout: manual_layout
 parent: "/manual.html"
 old_path_in_opsmanual: "../opsmanual/2nd-line/alerts/data-sync.md"
-last_reviewed_on: 2017-02-17
+last_reviewed_on: 2017-07-25
 review_in: 6 months
 ---
 
@@ -18,3 +18,11 @@ Data is synced from production to staging and integration every night.
 
 Check the output of the production Jenkins job to see which part of
 the data sync failed. It may be safe to re-run part of the sync.
+
+The Jenkins jobs included in the sync are:
+
+* Copy Data to Staging
+* Copy Data to integration
+* Copy Licensify Data to staging
+
+See the [source code](https://github.com/alphagov/env-sync-and-backup/tree/master/jobs) of the jobs for more information about how they work.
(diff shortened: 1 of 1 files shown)


source/manual/alerts/data-sync.html.md
```diff
@@ -18,3 +18,13 @@ Data is synced from production to staging and integration every night.
 
 Check the output of the production Jenkins job to see which part of
 the data sync failed. It may be safe to re-run part of the sync.
+
+Warning: the mysql backup will cause signon in staging to point to production until the `Data Sync Complete` job runs and renames the hostnames copied from production to back to their staging equivalents.  This means that any deploys to staging that rely on GDS API Adapters are likely to fail.
```
davidbasalla: How does this relate to GDS API Adapters?

leenagupte: When the Data Sync job was running, a deploy of static threw a `GdsApi::TimedOutException: GdsApi::TimedOutException`.  See [console message](https://deploy.staging.publishing.service.gov.uk/job/Deploy_App/4426/console). We assumed it was because static was trying to read from publishing api in production, rather than staging.

davidbasalla: Ah ok... from that console log it looks like it timed out when trying to publish to the PublishingAPI. I'm quite hazy on how this stuff hangs together (signon, gds api adapters, deploying apps and the renaming of hostnames) and I'm not sure if the text addition would improve my understanding of it. I think @h-lame and @fofr ran into this problem recently? Might be useful to get a 2nd opinion from them on this...

leenagupte: I'm not either. I just want to warn people that copying Signon data to Staging in the middle of the day is not the best idea.

davidbasalla: Had a quick slack chat with Murray about this. It makes more sense in my head now. Still not 100% clear on the gds-api-adapters but definitely valuable to warn about this so I'm happy to merge this. You could maybe add something like ('due to authentication failures') at the end, but up to you!

source/manual/alerts/data-sync.html.md
```diff
@@ -5,7 +5,7 @@ section: Icinga alerts
 layout: manual_layout
 parent: "/manual.html"
 old_path_in_opsmanual: "../opsmanual/2nd-line/alerts/data-sync.md"
-last_reviewed_on: 2017-02-17
+last_reviewed_on: 2017-07-25
```
tijmenb: This should probably be `2017-08-25` (or `2017-08-29` by now)

leenagupte: Ooops! Will fix. Thanks

PR #387, opened 2017-08-29
Add some docs on uptime metrics

thomasleese:
This adds some documentation on how the uptime metrics are collected.

[Trello Card](https://trello.com/c/zveUxj8x/943-spike-uptime-and-availability-metrics-timebox-3-days)


thomasleese:
Depends on https://github.com/alphagov/govuk-puppet/pull/6353 being merged.

source/manual/uptime-metrics.html.md
```diff
@@ -0,0 +1,36 @@
...
+parent: "/manual.html"
+last_reviewed_on: 2017-08-25
+review_in: 6 months
+related_applications: [publishing-api, content-store]
+---
+
+Uptime metrics are collected for the `content-store` and the `publishing-api`,
+they are available as [a Grafana dashboard][grafana-dashboard].
+
+![Uptime Metrics](images/uptime-metrics.png)
```
kevindew: Can we do any power moves here we write html that embeds the graph? Might be pure wishful thinking on my part.

source/manual/uptime-metrics.html.md
```diff
@@ -0,0 +1,36 @@
...
+
+They are available broken down into a day by day view, highlighted in different
+colours representing the level of uptime. Green means 100%, orange means above
+90% and red for everything else.
```
kevindew: I think we'd probably want to adjust these guages - I'd imagine orange is less than 5 minutes, whereas 90% equates to 144 minutes. So for it to be 5 minutes I think it'd be 99.7% and above (if my Maths is right - which it probably isn't)

thomasleese: Yes, I agree that 90% is too low. I was wondering if 99.7% was too high though, I would tempted to go to 10 minutes.

kevindew: Yeah that's cool - we'll get a better feel for it anyway as the data builds up. At the mo I'm not really imagining we'll see it slip below 100% much. Is that 99.3%?

Wonder if it's hard to represent it as minutes. We could do `86400 - (value * 86400)` to get seconds but we'd need grafana to have a method to show the seconds as minutes.

thomasleese: Yes, that would be 99.3%.

Looks like maths operators are not yet supported in Graphite directly: https://github.com/graphite-project/graphite-web/issues/295 I'll see if we can do it in Grafana, but I suspect not.

source/manual/uptime-metrics.html.md
thomasleese: Yes, some experimentation looks like we can do this. But I think we have to wait until it's deployed to get the URL as I'm not sure which panel and dashboard ID it will be.

PR #393, opened 2017-08-31
Link to better VPN instructions

h-lame:
The google doc linked to as the "normal GDS vpn" didn't include details for using openconnect insead of cisco anyconnect.  The page about fixing problems on the dev VM suggests we use openconnect (see: https://docs.publishing.service.gov.uk/manual/fix-problems-with-vagrant.html#are-you-on-a-vpn) instead to fix some VM issues with anyconnect so it feels like it'd be useful if this page did the same.


h-lame:
Should I be bumping the `last_reviewed_on` frontmatter of this page in this PR too?

davidbasalla:
Sure!

h-lame:
Thanks @davidbasalla, I have done so.

PR #400, opened 2017-09-01
Add frontend components page to manual

fofr:
Detail what components are, where principles and conventions live and where to find component guides.

### DIFF
--- a/source/manual/components.html.md
+++ b/source/manual/components.html.md
@@ -1,0 +1,39 @@
+---
+owner_slack: '#publishing-frontend'
+title: Components
+section: Frontend
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2017-09-01
+review_in: 6 months
+---
+
+Components are packages of template, style, behaviour and documentation that live in your application or in static.
+
+## Component guides
+
+Components in applications are documented in component guides using the [govuk_publishing_components gem](https://github.com/alphagov/govuk_publishing_components). This mounts a component guide at the path `/component-guide` in development, test and on Heroku review apps.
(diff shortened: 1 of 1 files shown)


source/manual/components.html.md
```diff
@@ -0,0 +1,38 @@
...
+
+Find components in these guides:
+
+* [static component guide](https://govuk-static.herokuapp.com/component-guide/)
+* [government-frontend component guide](https://government-frontend.herokuapp.com/component-guide/)
+
+## Building components
+
+A component must:
+* [meet the component principles](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_principles.md)
```
andysellick: I think there needs to be an empty line here - it's not rendering the list properly.

source/manual/components.html.md
```diff
@@ -0,0 +1,38 @@
...
+
+Components are packages of template, style, behaviour and documentation that live in your application or in static.
+
+## Component guides
+
+Components in applications are documented in component guides using the [govuk_publishing_components gem](https://github.com/alphagov/govuk_publishing_components). This mounts a component guide at the path `/component-guide` in development, test and on Heroku review apps.
+
+Find components in these guides:
+
+* [static component guide](https://govuk-static.herokuapp.com/component-guide/)
```
andysellick: This link doesn't work yet - I assume that's okay?

fofr: It'll work now.

source/manual/components.html.md
fofr: Fixed :+1:

PR #410, opened 2017-09-05
Add documentation for bulk-resetting signon password

boffbowsh:
This is a leftover action from an incident in February 2016. This
documentation was in the old opsmanual but was lost in the move to this
manual.


tijmenb:
At the time I put this into https://github.com/alphagov/signon/issues/543 because we felt it was too app-specific for the manual. We could add this page to the `/docs` folder in signon?

boffbowsh:
Ah ok, good call. I'll sort that out.

PR #412, opened 2017-09-05
Add puppet_name for asset-manager

chrisroos:
This fixes the broken link to the Deployment dashboard on the
apps/asset-manager.html page.

The link changes from:

  https://grafana.publishing.service.gov.uk/dashboard/file/deployment_asset_manager.json

to:

  https://grafana.publishing.service.gov.uk/dashboard/file/deployment_asset-manager.json

I've tried this locally and have confirmed that the new link works as
expected.

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -148,6 +148,7 @@
 - github_repo_name: asset-manager
   type: APIs
   team: "#2ndline"
+  puppet_name: asset-manager
 
 - github_repo_name: router-api
   type: APIs


data/applications.yml
```diff
@@ -148,6 +148,7 @@
 - github_repo_name: asset-manager
   type: APIs
   team: "#2ndline"
+  puppet_name: asset-manager
```
tijmenb: This would break the link to the puppet config. I think the correct solution would be to use the `app_name` in the deploy dashboard URL:

https://github.com/alphagov/govuk-developer-docs/blob/master/app/app_docs.rb#L82-L84

Looks like that matches a lot better to the list used here:

https://github.com/alphagov/govuk-puppet/blob/ed07581c5054c8ef9183e4101571f6cd2fe5f60e/hieradata_aws/common.yaml#L873-L985

chrisroos: Aha. I did wonder whether adding the puppet_name might have unintended side effects. I'll take another look at this.

tijmenb:
@chrisroos feel free to reopen this if necessary.

PR #413, opened 2017-09-05
Add documentation around GOR traffic replay for rummager

dwhenry:
https://trello.com/c/CdZiSBIb/280-deploy-es-to-production

### DIFF
--- a/source/manual/rummager-traffic-replay.html.md
+++ b/source/manual/rummager-traffic-replay.html.md
@@ -1,0 +1,33 @@
+---
+owner_slack: "#search-team"
+title: 'Replay search indexing traffic after restoring a backup'
+parent: "/manual.html"
+layout: manual_layout
+section: Backups
+last_reviewed_on: 2017-09-05
+review_in: 3 months
+---
+
+If the index should fail and we need to [restore from backup](https://docs.publishing.service.gov.uk/manual/elasticsearch-dumps.html) we can then replay this traffic rather than trying to resend from the source applications.
+
+We are using [GOR](https://github.com/buger/goreplay) to store POST and DELETED requests made to the rummager servers.
+
+Data for each of the rummager machine is stored locally on the machine at:
+
+```
+/var/log/gor_dumps
+```
+
+Traffic will needs to be replayed from each server using the command:
+
+```
+sudo GODEBUG="netdns=go" /usr/local/bin/gor --input-file "/var/log/gor_dump/*.log|1000%" --output-http http://localhost:3009 --output-http-timeout 30s
+```
+
+You can run the command with debug logging to get a better idea of progress using the below command:
+
+```
+sudo GODEBUG="netdns=go" /usr/local/bin/gor --input-file "/var/log/gor_dump/*.log|1000%" --output-http http://localhost:3009 --output-http-timeout 30s --verbose --debug
+```
+
+The above command replays the traffic at 10 times (1000%) the speed is was captured, this can be adjusted by changing the percentage value in the command.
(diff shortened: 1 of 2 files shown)


source/manual/rummager-traffic-replay.html.md
```diff
@@ -0,0 +1,35 @@
...
+owner_slack: "#search-team"
+title: 'Replay HTTP traffic on rummager machines'
+parent: "/manual.html"
+layout: manual_layout
+section: Backups
+last_reviewed_on: 2017-09-05
+review_in: 3 months
+---
+
+Have have started using [GOR](https://github.com/buger/goreplay) to store POST and DELETED requests
```
MatMoore: Have Have

source/manual/rummager-traffic-replay.html.md
```diff
@@ -0,0 +1,35 @@
...
+to the rummager servers.
+
+At is stored on the individual machine at:
```
MatMoore: At is

source/manual/rummager-traffic-replay.html.md
```diff
@@ -0,0 +1,35 @@
...
+title: 'Replay HTTP traffic on rummager machines'
+parent: "/manual.html"
+layout: manual_layout
+section: Backups
+last_reviewed_on: 2017-09-05
+review_in: 3 months
+---
+
+Have have started using [GOR](https://github.com/buger/goreplay) to store POST and DELETED requests
+to the rummager servers.
```
MatMoore: This should probably say "requests made to the rummager servers" as it sounds a bit like the rummager servers are just the place where you're storing stuff.

source/manual/rummager-traffic-replay.html.md
```diff
@@ -0,0 +1,35 @@
...
+
+At is stored on the individual machine at:
+
+```
+/var/log/gor_dumps
+```
+
+If the index should fail and we need to restore from backup we can then replay this traffic rather than
```
MatMoore: Since this is the only use case we're considering, maybe we should title the page something like "Replay search indexing traffic after restoring a backup".

Then we can move this right to the top of the page.

We should also link to this page from https://docs.publishing.service.gov.uk/manual/elasticsearch-dumps.html

source/manual/rummager-traffic-replay.html.md
MatMoore: Actually this could just say "We are using", instead of started.

source/manual/rummager-traffic-replay.html.md
```diff
@@ -0,0 +1,35 @@
...
+trying to resend from the source applications.
+
+Traffic needs to be replayed from each server using the command:
+
+```
+sudo GODEBUG="netdns=go" /usr/local/bin/gor --input-file "/var/log/gor_dump/*.log|1000%" --output-http http://localhost:3009 --output-http-timeout 30s
+```
+
+You can run the command with debug logging to get a better idea of process using the below command:
```
MatMoore: s/process/progress ?

source/manual/rummager-traffic-replay.html.md
dwhenry: Done

source/manual/rummager-traffic-replay.html.md
dwhenry: Done

source/manual/rummager-traffic-replay.html.md
dwhenry: reworded

source/manual/rummager-traffic-replay.html.md
dwhenry: done

source/manual/rummager-traffic-replay.html.md
dwhenry: Done

PR #420, opened 2017-09-06
Retire metadata-api

elenatanasoiu:
Mark the `metadata-api` as retired and add a description of what it used to do.

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -168,8 +168,16 @@
   api_docs_url: https://mapit.mysociety.org/docs/
 
 - github_repo_name: metadata-api
+  retired: true
   type: APIs
   team: "#survey-and-support"
+  description: |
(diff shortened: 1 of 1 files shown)


data/applications.yml
```diff
@@ -168,8 +168,16 @@
   api_docs_url: https://mapit.mysociety.org/docs/
 
 - github_repo_name: metadata-api
+  retired: true
   type: APIs
   team: "#survey-and-support"
+  description: |
+    API written in Go that was used to get metadata about given URLs on GOV.UK.
+    The api would communicate with the `backdrop read API` on behalf of `info-frontend`
+    and retrieve information about user needs. The `backdrop read API` would talk
```
h-lame: I am fairly sure that metadata API spoke to need API directly, not via backdrop.

elenatanasoiu: Corrected

PR #425, opened 2017-09-07
Add tips on dealing with a broken bowl

### LINKED ISSUE #424: Problem with 'Get started on GOV.UK'
Problem with 'Get started on GOV.UK' (https://docs.publishing.service.gov.uk/manual/get-started.html)

Section 8: Run your apps
This seems to be out of date. I can't run "bowl x" but have to run "bundle exec bowl x" to get any output. What then happens is the app runs and quits with a message that seems to suggest bundle exec wasn't previously run, even when it was:
/usr/lib/rbenv/versions/1.9.3-p550/lib/ruby/gems/1.9.1/gems/bundler-1.7.4/lib/bundler/source/git.rb:188:in `rescue in load_spec_files': git://github.com/alphagov/airbrake.git (at silence-dep-warnings-for-rails-5) is not yet checked out. Run `bundle install` first. (Bundler::GitError)

I've been pointed at an alternate way of running the apps which is to go into their directory and run the startup file from in there - does the documentation need to switch to instructing users to do this instead?

h-lame:
Sometimes the VM ends up created without a fully working `bowl` command
so the advice in the getting started guide about running apps seems like
it is out of date.  The resolution is to manually install the bowler gem
rather than trying to use `bundle exec bowl` or ignoring `bowl` in favour
of using `startup.sh` in each app folder directly.

It's not clear to me why some VMs get into this state, and others don't,
but it's happened enough that some docs on solving it wouldn't go amiss.

[Fixes #424]

### DIFF
--- a/source/manual/fix-problems-with-vagrant.html.md
+++ b/source/manual/fix-problems-with-vagrant.html.md
@@ -36,6 +36,51 @@ $ cd /var/govuk/govuk-puppet/development-vm
 $ foreman start rummager
 ```
 
+Most apps also have a `startup.sh` script in their root folder, and you can
+run that directly too:
+
+```shell
+$ cd /var/govuk/rummager
+$ ./startup.sh
+````
+
+## Using `bowl` fails with bundler error
+
+Sometimes the `bowl` command is not correctly installed on the VM.  If you
+get an error like:
+
+```shell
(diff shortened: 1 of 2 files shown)


fofr:
16261022-broken-bowl-on-white-background-stock-photo

PR #430, opened 2017-09-11
Update Heroku review apps to include ruby-version

nickcolley:
We have had issues with static not running correctly based on the ruby version being different on Heroku and production, this change ensures the version is the same


nickcolley:
@tijmenb I'm not sure why https://github.com/alphagov/static/pull/1140 is failing but we should resolve that before merging this.

PR #431, opened 2017-09-11
Retire Need API

elenatanasoiu:
Mark `govuk_need_api` as retired and add a description.

For: https://trello.com/c/ClCFMBlG/217-retire-need-api

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -118,9 +118,14 @@
     [content-store](/apps/content-store.html).
 
 - github_repo_name: govuk_need_api
+  retired: true
   type: APIs
   puppet_name: need_api
   team: "#survey-and-support"
+  description: |
+    A JSON read and write API for information about user needs on GOV.UK. It was a Rails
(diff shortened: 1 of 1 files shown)


data/applications.yml
```diff
@@ -118,9 +118,14 @@
...
 
 - github_repo_name: govuk_need_api
+  retired: true
   type: APIs
   puppet_name: need_api
   team: "#survey-and-support"
+  description: |
+    A JSON read and write API for information about user needs on GOV.UK. It was a Rails
+    app which was part of the GOV.UK Publishing Platform. The need information is now
+    being fetched from the content-store.
```
cbaines: I'd say something like "stored in the Publishing API, and available in the Content Store", rather than just "being fetched from the content-store".

I realise that this is the architectural change from using the Metadata API, to using the Content Store for the Info Frontend, but I think its better to put something more general here.

elenatanasoiu: I've changed the description. Thanks :+1:

cbaines: Great 👍

PR #441, opened 2017-09-14
Add step for removing from Sentry

elenatanasoiu:
Applications should be removed from error tracking after they've been retired. This commit adds that step in the list.

Following changes for: https://trello.com/c/ClCFMBlG/217-retire-need-api

### DIFF
--- a/source/manual/retiring-an-application.html.md
+++ b/source/manual/retiring-an-application.html.md
@@ -134,10 +134,14 @@ Some applications are responsible for publishing certain routes. If you're
 retiring a publishing application, make sure you check if any of its content
 items need to be unpublished and do it via the Publishing API.
 
-## 17. Move repo
+## 17. Remove from Sentry
+
+Since the application has been retired, it shouldn't be tracked in Sentry. 
+
+## 18. Move repo to attic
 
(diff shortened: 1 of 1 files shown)


source/manual/retiring-an-application.html.md
```diff
@@ -134,10 +134,14 @@ Some applications are responsible for publishing certain routes. If you're
 retiring a publishing application, make sure you check if any of its content
 items need to be unpublished and do it via the Publishing API.
 
-## 17. Move repo
+## 17. Remove from Errbit / Sentry
+
+Since the application has been retired, it shouldn't be tracked in Errbit or Sentry.
```
h-lame: Are there any docs about how we do this?  I was able to do sentry, but I think only because I have some elevated permissions, no idea how to do it on errbit.

elenatanasoiu: I couldn't find any documentation on how to remove an app from Sentry, except for the API version. 

I've removed the errbit part since it's being retired.

h-lame: Cool.

The commit message still references errbit though.  Change that and you're good to merge I reckon.

PR #449, opened 2017-09-21
govuk-puppet development vm fix additions

cbaines:
See the individual commits for the details.


tijmenb:
I think the title is a bit verbose. And I don't think anybody would refer to it as the "govuk-puppet development VM"?

I'd rather see this page split up into smaller pages, so that the content is more easily findable. Perhaps move https://github.com/alphagov/govuk-developer-docs/pull/449/commits/64d8b5090c4a1ec8e768e3d4f6d01775cefadb66 to a new page?

cbaines:
> I think the title is a bit verbose. And I don't think anybody would refer to it as the "govuk-puppet development VM"?
> 
> I'd rather see this page split up into smaller pages, so that the content is more easily findable. Perhaps move 64d8b50 to a new page?

Ok, I'll close this in favour of refactoring the documentation first.

PR #463, opened 2017-09-27
Update docs on Airbrake/Sentry

tijmenb:
We've migrated to Sentry, this updates the docs.


source/manual/upgrade-to-sentry.html.md
```diff
@@ -1,45 +0,0 @@
...
-layout: manual_layout
-section: Monitoring
-owner_slack: "@tijmen"
-last_reviewed_on: 2017-09-04
-review_in: 1 month
----
-
-We are [migrating all apps][trello] to use Sentry.
-
-## 1. Make the application changes
```
fofr: Is any of this still useful for adding Sentry to a new application?

tijmenb: I think there's enough here now: https://github.com/alphagov/govuk_app_config#usage

PR #470, opened 2017-09-28
Update Search Benchmarking failure page

tomnatt:


source/manual/alerts/search-benchmarking.html.md
```diff
@@ -4,18 +4,16 @@ title: Benchmark search queries failed
...
 (https://deploy.publishing.service.gov.uk/job/search_benchmark/) has failed.
 
 The healthcheck sends queries to the search API and compares the results to the
-expected pages.
+expected pages. It will only fail if something actually goes wrong - for example, if a request to the search API fails. Check the output of the job on the relevant environment for more information.
 
-The Jenkins job will still succeed even if the search results are poor and will
-only fail if something actually goes wrong. For example, if a request to the
-search API fails.
+The search benchmark checks if various terms appear at expected positions in search lists, however the pass mark is less than 100% as data moves around over time and some tests fail to pass as a result. The data is useful to analyse over time - it is not a point in time indication something is wrong.
```
cbaines: I don't know if there is a convention for wrapping text in this repository, but normally I'd try to be consistent within the same file.

source/manual/alerts/search-benchmarking.html.md
```diff
@@ -4,18 +4,16 @@ title: Benchmark search queries failed
...
 
-Check the output of the job on the relevant environment for more information.
+The Jenkins job was disabled 28th September 2017.
```
cbaines: If this is the case, why not remove the page?

tomnatt: The job still exists - the team is deciding what to do with it long term. If / when it's gone then agreed the page should go away too.

MatMoore: We've decided to keep it now. The job can be reenabled - I've made it a bit less noisy with https://github.com/alphagov/rummager/pull/941

tomnatt:
@MatMoore can you have another quick check and give me a thumbs up / down please?

source/manual/alerts/search-benchmarking.html.md
```diff
@@ -8,14 +8,9 @@ last_reviewed_on: 2017-10-02
 review_in: 3 months
 ---
 
-Indicates that the Jenkins [search_benchmark healthcheck job]
-(https://deploy.publishing.service.gov.uk/job/search_benchmark/) has failed.
+Indicates that the Jenkins [search_benchmark healthcheck job] (https://deploy.publishing.service.gov.uk/job/search_benchmark/) has failed.
 
 The healthcheck sends queries to the search API and compares the results to the
-expected pages.
+expected pages. It will only fail if something actually goes wrong - for example, if a request to the search API fails. Check the output of the job on the relevant environment for more information.
```
MatMoore: This is essentially correct although I've made it swallow the exception if it's a server error.

tomnatt: So will this error ever appear?

MatMoore: @tomnatt as discussed it'll just be a lot less likely to appear.

tomnatt: So page is still relevant (as per Chris's original question). I will merge.

PR #471, opened 2017-09-28
Add two new component guides

fofr:
Collections and finder-frontend now use the govuk_publishing_components gem and have Heroku review apps and pipelines set up.

### DIFF
--- a/app/app_docs.rb
+++ b/app/app_docs.rb
@@ -87,6 +87,10 @@ class AppDocs
       app_data["api_docs_url"]
     end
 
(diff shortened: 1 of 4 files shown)


tijmenb:
How about doing something like this https://github.com/alphagov/govuk-developer-docs/pull/323 for the component guides too?

fofr:
@tijmenb Added links to guides on app pages  in ad49fd8

tijmenb:
Very cool!

PR #475, opened 2017-09-29
Add section on transition to dashboard

h-lame:
`bouncer` and `transition` are apps that we use to support sites
transitioning to gov.uk, but there are a handful of other tools we use as
aprt of this process.  We collect them all here in a new "Transition"
section on the dashboard.  The repos were found via [the
`govuk-transition` topic on github][govuk-transition].  Some of the repos
we link to are private, so they have to be added as sites.

[govuk-transition]: https://github.com/search?q=topic%3Agovuk-transition+org%3Aalphagov&type=Repositories

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -214,16 +214,6 @@
   product_manager: "@martinlugton"
   team: "#publishing-platform"
 
-- github_repo_name: bouncer
-  type: Supporting apps
-  product_manager: "@lukemalcher"
-  team: "#taxonomy"
-
-- github_repo_name: transition
-  type: Supporting apps
-  product_manager: "@lukemalcher"
-  team: "#taxonomy"
-
 - github_repo_name: release
   type: Supporting apps
   product_manager: "@robrankin"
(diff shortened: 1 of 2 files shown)


data/dashboard.yml
```diff
@@ -4,6 +4,14 @@
     - name: Publishing apps
     - name: APIs
     - name: Supporting apps
+      sites:
```
tijmenb: I'm not sure this is an application like the others (it isn't a web thing at least). Perhaps a new section on this dashboard for transition-y stuff like https://github.com/search?q=topic%3Agovuk-transition+org%3Aalphagov?

h-lame: Good point.  I've retooled the commit.

data/applications.yml
```diff
@@ -347,3 +337,18 @@
...
+  type: Transition apps
+  product_manager: "@lukemalcher"
+  team: "#taxonomy"
+
+- github_repo_name: transition
+  type: Transition apps
+  product_manager: "@lukemalcher"
+  team: "#taxonomy"
+
+- github_repo_name: side-by-side-browser
```
tijmenb: Most of the links on the application page (Sentry / Puppet configuration / Deploy scripts / Deployment dashboard) won't work for this app because it's not hosted on our own infrastructure (it's on Heroku). Perhaps move it to the `Supporting transition` in dashboard.yml?

h-lame: Makes sense.  Rebased.

PR #488, opened 2017-10-04
Details on previewing the message sent via rabbit MQ

dwhenry:
https://trello.com/c/KDfyOeS1/331-specialist-publisher-implementation

### DIFF
--- a/source/manual/rabbitmq.html.md
+++ b/source/manual/rabbitmq.html.md
@@ -78,3 +78,19 @@ Note: there is a risk that you might delete the wrong message(s). This
 is because the contents of the queue may have changed.
 
 Repeat, but change the "Requeue" option to "No".
+
+## Previewing a message for a document_type
+
+The publishing API generates messages when content is updated and posts them
+to the rabbitMQ exchange. Each message has a shared format, however the contents
+of the message is affected by the publishing app and what data it sends to the
+publishing API.
+
+As messages for different formats can vary, we have created a rake task in the
+publishing API app to allow us to easily generate example messages. The example
+message is generated from the most recently published message (based off of
+last public_updated_at) for the entered document type:
+
+```
+bundle exec rake queue:preview_recent_message[<document_type>]
+```


source/manual/rabbitmq.html.md
```diff
@@ -78,3 +78,14 @@ Note: there is a risk that you might delete the wrong message(s). This
...
+
+## Previewing a message for a document_type
+
+As the message which is published for each document type can vary based on
+what infomation is passed in from the publishing app, this allows us to easily
+generate an example message based off the last published message for a given
+document type:
+
+```
+bundle exec rake queue:preview_last_message[<document_type>]
```
cbaines: Sounds good, but it's unclear where this rake task sits?

MatMoore:
I think this needs a bit more context since the page is about rabbitmq in general. That command is something you run against the publishing api, if you're interested in whats published to the `published_documents` exchange. Don't think that's obvious because we use rabbitmq for some other things as well.

source/manual/rabbitmq.html.md
```diff
@@ -78,3 +78,14 @@ Note: there is a risk that you might delete the wrong message(s). This
 is because the contents of the queue may have changed.
 
 Repeat, but change the "Requeue" option to "No".
+
+## Previewing a message for a document_type
+
+As the message which is published for each document type can vary based on
+what infomation is passed in from the publishing app, this allows us to easily
+generate an example message using a recently published message (based off
+last public_updated_At) for a given document type:
```
MatMoore: `public_updated_at` should be lowercase here.

dwhenry:
@MatMoore updated and added content.

source/manual/rabbitmq.html.md
```diff
@@ -78,3 +78,19 @@ Note: there is a risk that you might delete the wrong message(s). This
 is because the contents of the queue may have changed.
 
 Repeat, but change the "Requeue" option to "No".
+
+## Previewing a message for a document_type
+
+The publishing API generates messages when content is updated and posts them
+to the rabbit MQ exchange. Each message has a shared format, however teh contents
+of the message is affected by the publoishing app and what data it sends to the
+publioshing API.
```
MatMoore: There's a few typos here, can you do another edit?

- publoishing/publishging
- teh
- RabbitMQ
- to allow easily generate
- generetd
- As message for different formats can vary

I'd also explicitly name the exchange (published_documents) so people can easily track down the messages in the management UI.

dwhenry: @brenetic can you pair with me on this so we can get it merged. :)

brenetic:
@MatMoore Updated for your viewing pleasure.

Rosa-Fox:
This is really useful! We have been using it today in integration. Will wait for @MatMoore  to give the final approval though.

PR #489, opened 2017-10-04
Document how to whitelist site in Transition

tomsabin:
Without the admin permission for Transition, it is not obvious that
there is whitelisting functionality in the app. By documenting the fact
that such a feature exists, this should help others find some
information when a whitelist request comes in.

### DIFF
--- a/source/manual/whitelist-site-in-transition.html.md
+++ b/source/manual/whitelist-site-in-transition.html.md
@@ -1,0 +1,18 @@
+---
+owner_slack: '#taxonomy'
+title: Whitelist site to allow redirection away from GOV.UK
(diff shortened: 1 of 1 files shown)


tomsabin:
@tijmenb I've added it to these docs because other similar manuals show within the "[relevant manual pages](https://docs.publishing.service.gov.uk/apps/transition.html)", so if we want to move app-specific docs to within the app repository then we should probably pull along these two manual pages too.

However, as a new joiner, my initial thoughts were actually to look in the developer docs before heading to GitHub. It's much easier to search for specific terms (e.g. "whitelist, transition") on this app.

PR #516, opened 2017-10-09
Update docs for github.com being down

surminus:
Disabling security is trivial in newer versions of Jenkins, so this should be our emergency process if github.com is down. The Jenkins UI should still be protected by IP whitelisting.

/cc @boffbowsh 

https://trello.com/c/5TGUzAGn/743-find-alternative-way-to-login-to-jenkins-if-github-is-down


issyl0:
@boffbowsh I updated the docs with this note.

tomnatt:
I think @boffbowsh's points have been addressed, but he's not here to ok them so I'm going to dismiss his review as it'll block pushing these changes live.

PR #518, opened 2017-10-10
Add page for the architecture diagram

tijmenb:
This diagram contains almost all applications and the relationships between them. It's permanent work in progress.


h-lame:
👏 Typo on the image: "Blue line means B writes from A" probably should be "Purple line means B writes to A"

h-lame:
Oh, also, don't suppose the image is available as SVG / graphviz / some other textual format so it's easier to update?

davidbasalla:
Is Licensify the missing app you mentioned at TechSAD? Do I win? 😛

tijmenb:
@davidbasalla It isn't, but I did forget it! There's at least one more :)

surminus:
Just spotted this PR - this is super useful. I wonder if @stephenharker has any further diagrams we could add here?

PR #519, opened 2017-10-10
Add an on-call checklist

boffbowsh:
Things you should do before going on call

### DIFF
--- a/source/manual/on-call.html.md
+++ b/source/manual/on-call.html.md
@@ -23,6 +23,26 @@ to keep GOV.UK running at night, on the weekends and on public holidays.
 - Logs are not as important as being available - if you need to lose some logs
   in order to bring the site back up, that's probably a good trade-off to make
 
+## On call checklist
+
+You should do these things before going on call so you're prepared.
+
+1. Have the numbers of other people on your shift saved in your phone. This
+   includes whoever is on Escalations. Get these numbers from PagerDuty.
+2. Ensure you have an up to date local copy of the [Developer Docs][docs] repository
+   and that you can build it.
+3. Make sure your [`fabric-scripts`][fabric] are up to date.
+4. Make sure you can VPN to the office or disaster recovery location.
+5. Ensure your PagerDuty alert settings will wake you if you're called. You might want
+   to install the [PagerDuty App](https://www.pagerduty.com/features/mobile-incident-management/)
+   on your phone.
+6. Ensure you can [decrypt secrets][govuk-secrets] with your GPG setup.
+7. Ensure you can access [vCloud Director][vcloud] in production.
+8. Read these documents:
+    - [So, you're having an incident](https://gov-uk.atlassian.net/wiki/spaces/PLOPS/pages/46301531/So+you+re+having+an+incident)
+    - [Falling back to the static mirror](/manual/fall-back-to-mirror.html)
+    - [Emergency publishing](/manual/emergency-publishing.html)
+
 ## Things that may result in you being contacted
 
 ### Automated monitoring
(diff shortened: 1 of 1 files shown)


source/manual/on-call.html.md
```diff
@@ -23,6 +23,22 @@ to keep GOV.UK running at night, on the weekends and on public holidays.
...
+includes whoever is on Escalations. Get these numbers from PagerDuty.
+2. Ensure you have an up to date local copy of the [Developer Docs][docs] repository
+and that you can build it.
+3. Make sure your [`fabric-scripts`][fabric] are up to date.
+4. Make sure you can VPN to the office or disaster recovery location
+5. Read these documents:
+    - [Stopping Nginx on the cache machines to fail to our static mirror](https://github.com/alphagov/fabric-scripts/blob/master/incident.py)
+    - [Emergency publishing](/manual/emergency-publishing.html)
+    - [Emergency publishing while origin is unavailable](/manual/fall-back-to-mirror.html)
+    - [Responding to an outage](https://gov-uk.atlassian.net/wiki/display/PLOPS/GOV.UK+Incidents)
```
danielroseman: This page is a list of incidents, not instructions on responding to an outage. Is there a more appropriate page we could link to?

boffbowsh: Good call, fixed.

danielroseman: :+1: (of course, the next question is, why is that page in Confluence rather than in the docs here?)

source/manual/on-call.html.md
```diff
@@ -23,6 +23,22 @@ to keep GOV.UK running at night, on the weekends and on public holidays.
...
+You should do these things before going on call so you're prepared.
+
+1. Have the numbers of other people on your shift saved in your phone. This
+includes whoever is on Escalations. Get these numbers from PagerDuty.
+2. Ensure you have an up to date local copy of the [Developer Docs][docs] repository
+and that you can build it.
+3. Make sure your [`fabric-scripts`][fabric] are up to date.
+4. Make sure you can VPN to the office or disaster recovery location
+5. Read these documents:
+    - [Stopping Nginx on the cache machines to fail to our static mirror](https://github.com/alphagov/fabric-scripts/blob/master/incident.py)
```
rjw1: Is this still a thing people should know about?

source/manual/on-call.html.md
```diff
@@ -23,6 +23,22 @@ to keep GOV.UK running at night, on the weekends and on public holidays.
 - Logs are not as important as being available - if you need to lose some logs
   in order to bring the site back up, that's probably a good trade-off to make
 
+## On call checklist
+
+You should do these things before going on call so you're prepared.
+
+1. Have the numbers of other people on your shift saved in your phone. This
```
rjw1: Possible talk about looking at your alert settings in pager duty and indeed maybe installing the pager duty app.

source/manual/on-call.html.md
boffbowsh: Changed for the generic "fallback to static mirror" advice page

source/manual/on-call.html.md
boffbowsh: Good shout, done.

source/manual/on-call.html.md
```diff
@@ -23,6 +23,24 @@ to keep GOV.UK running at night, on the weekends and on public holidays.
...
+   includes whoever is on Escalations. Get these numbers from PagerDuty.
+2. Ensure you have an up to date local copy of the [Developer Docs][docs] repository
+   and that you can build it.
+3. Make sure your [`fabric-scripts`][fabric] are up to date.
+4. Make sure you can VPN to the office or disaster recovery location
```
jennyd: I have a few more things in my list to check I can access:

- decrypting 2nd line pass and encrypted hieradata (upgrades have been known to break my gpg setup...)
- ssh to mirror boxes (this may not be that important any more? depends on whether we now have other ways of handling the various times when we might need to urgently update content on the mirrors)
- production vCloud Director

This covers the things I would usually use less frequently, on the assumption that I would have noticed already if I couldn't get to production machines or log into jenkins.

boffbowsh: Thanks, added.

PR #520, opened 2017-10-10
Update documentation on changing an organisation's slug

tuzz:
The Need API has been retired to gds-attic and the
cache of Organisation in Maslow is only an hour so
we can probably let this expire organically.


tuzz:
Yeah :) much simpler now.

PR #128, opened 2017-04-12
[WIP] Algolia DocSearch

andrewgarner:
Work in progress. The search results are not read out by a screen reader when tabbed through.

See https://trello.com/c/vSbk1HoF/51-add-search-to-docs & https://github.com/alphagov/tech-docs-template/issues/66

PR #522, opened 2017-10-11
Add Algolia DocSearch

tijmenb:
This adds [Algolia DocSearch][al]. It's a free service used by prominent projects such as React, Stripe and Scala. We have an account set up, which is configured in https://github.com/algolia/docsearch-configs/blob/master/configs/publishing_service_gov_uk.json.

We previously discussed this in https://github.com/alphagov/govuk-developer-docs/pull/128. We didn't go for this then because there were questions about accessibility. I think it would make sense to implement it now after all because the Google is very slow, causing new pages to be indexed weeks after first being published. 

oct-11-2017 13-57-42

[al]: https://community.algolia.com/docsearch

### DIFF
--- a/source/layouts/header_footer_only.erb
+++ b/source/layouts/header_footer_only.erb
@@ -15,6 +15,7 @@
     <!--[if lte IE 8]><%= stylesheet_link_tag 'screen-old-ie', media: 'screen' %><![endif]-->
 
     <link rel="canonical" href="<%= canonical_url %>">
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/docsearch.js/2/docsearch.min.css"/>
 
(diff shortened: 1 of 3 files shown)


fofr:
There are still some serious accessibility problems with this aside from the voiceover problems:

1. Search results are not keyboard navigable, you cannot tab to the first item in the list
2. The highlighted item in the search results isn't necessarily the focussed one (VoiceOver reads out something different to what's highlighted)
3. The focus state has a contrast that's too low
4. There is no fallback to the Google Search

tijmenb:
Updated contrasts:

screen shot 2017-10-11 at 14 59 53

tijmenb:
screen shot 2017-10-11 at 15 37 36

PR #525, opened 2017-10-12
Add new document about how to add a new document type

deborahchua:
This document covers how to add a new document type to govuk-content-schemas and making it available to search.

### DIFF
--- a/source/manual/add-a-new-document-type.html.md
+++ b/source/manual/add-a-new-document-type.html.md
@@ -1,0 +1,79 @@
+---
+title: Add a new document type
+parent: "/manual.html"
+layout: manual_layout
+section: Tools
+owner_slack: "@davidslv and @matmoore"
+last_reviewed_on: 2017-08-08
+review_in: 3 months
+---
+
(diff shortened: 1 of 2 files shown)


Davidslv:
@deborahchua given that the process is changing to add a new document type we should wait for that change to happen so it can be documented after.
Although we can cut the first part of this documentation and release only the search section (meaning: remove anything before "Make the new document type available to search") and maybe create a new section for this document - I'm just not sure about the naming, let me know what you think of it.

deborahchua:
@Davidslv @tijmenb I have renamed the doc and kept the second half about making the document type searchable. 
I have kept it as a manual as it lists the steps but I have linked it to Rummager at the bottom of the page.
@MatMoore these are the steps you wrote and passed onto David, are there any changes you want me to make to it?

MatMoore:
@deborahchua no, I think that's everything I'm aware of for search. Did it work ok for task lists?

The current structure looks ok to me. We could make "Make the new document type available to search" a separate page, but if we do this I think I'd prefer it to stay in the dev manual, as it's easier to review that way and keep it focused on one use case. There is already some documentation in rummager about working the elasticsearch schemas, but I don't think you should have to know all the details to introduce a new document type.

PR #543, opened 2017-10-20
Add guide to uploading an asset to whitehall

suzannehamilton:
Document how we manually upload large attachments to whitehall, since we cannot use the asset-manager process for this.

@fofr and @tuzz - would you mind taking a look?

Note that I've skipped the republishing step that Paul ran. I think this was only necessary because we fixed the file size metadata after the document was published. Does that make sense?

### DIFF
--- a/source/manual/upload-asset-to-whitehall.html.md
+++ b/source/manual/upload-asset-to-whitehall.html.md
@@ -1,0 +1,50 @@
+---
+owner_slack: "#2ndline"
+title: Upload an asset to whitehall
+section: Assets
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2017-10-20
+review_in: 6 months
+---
+
+Sometimes publishers ask us to help them upload very large attachments to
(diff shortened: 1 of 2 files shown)


source/manual/upload-asset-to-whitehall.html.md
```diff
@@ -0,0 +1,50 @@
...
+title: Upload an asset to whitehall
+section: Assets
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2017-10-20
+review_in: 6 months
+---
+
+Sometimes publishers ask us to help them upload very large attachments to
+documents in whitehall because they see file size errors when they try to upload
```
fofr: In our case the user was getting a whitehall timeout (after about 10 seconds), possibly because the 80mb zip filed unzipped to a 2gb CSV.

suzannehamilton: 👍 I've changed the wording to "timeout".

PR #544, opened 2017-10-25
Improve Logit sign-in instructions

floehopper:
I ran into some problems following the instructions, so I've attempted to improve them in light of my experience.


floehopper:
@surminus I hope you don't mind that I've requested a review from you, but you were suggested by GitHub as a reviewer.

PR #546, opened 2017-10-25
Document redirecting Whitehall assets

gpeng:


fofr:
Is this completely true? An attachment that's replaced should redirect?

https://www.gov.uk/guidance/how-to-publish-on-gov-uk/creating-and-updating-pages#replace-edit-and-delete-attachments

> You can replace and edit published attachments.
>
>In an open document, select the ‘Attachments’ tab.
>Click ‘Edit’ next to the attachment you want to change.
>Upload a new file and it will overwrite the old file at the same URL location. The Markdown for this file will stay the same so doesn’t need to be changed.
>Edit the title or other metadata, if necessary.

There's a larger issue here around making _replacement_ the default choice of users, and delete being a more extreme, less used option.

PR #480, opened 2017-10-02
[DO NOT MERGE] Update docs on how logging works

surminus:
This should not be merged until we're using Logit everywhere.

This PR reflects any changes to our documentation.

https://trello.com/c/YrTOpoq7/746-logit-migration

Ideally users should check this branch when we're testing on Integration.

PR #551, opened 2017-10-27
Fix Kibana queries and links

suzannehamilton:
This just fixes Kibana queries on three pages:
- Tools
- Kibana
- Fastly error rate

There are probably other pages which still use the old fields and syntax, but this is all I had time to review while on 2ndline.

### DIFF
--- a/source/manual/alerts/fastly-error-rate.html.md
+++ b/source/manual/alerts/fastly-error-rate.html.md
@@ -5,14 +5,9 @@ section: Icinga alerts
 layout: manual_layout
 parent: "/manual.html"
 old_path_in_opsmanual: "../opsmanual/2nd-line/alerts/fastly-error-rate.md"
-last_reviewed_on: 2017-07-13
+last_reviewed_on: 2017-10-20
 review_in: 6 months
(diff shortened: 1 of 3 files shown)


source/manual/alerts/fastly-error-rate.html.md
```diff
@@ -37,7 +33,7 @@ The alert appears on `monitoring-1.management`. Collectd uses the Fastly API to
 
 To prove collectd is the problem, use this query in Kibana:
 
-`@source_host:monitoring-1.management AND @fields.syslog_program:collectd`
+`syslog_hostname:monitoring-1.management AND syslog_program:collectd`
 
 You will see many reports simlilar to:
```
rubenarakelyan: Would be nice to fix this spelling too while we're in this file :)

suzannehamilton: Fixed. 👍

suzannehamilton:
@MatMoore No, just some of them, unfortunately. #480 fixed a lot of the queries, but there are probably a few more lying around.

PR #554, opened 2017-11-01
Instructions on how to send an email using the dev VM

1pretz1:
Guide on how to send a test email to a specific email
address using Notify

### DIFF
--- a/source/manual/send-test-email.html.md
+++ b/source/manual/send-test-email.html.md
@@ -1,0 +1,25 @@
+---
+owner_slack: "#email"
+title: Send a test email via Notify
+section: Development VM
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2017-11-01
+review_in: 6 months
+---
+
+To send a test email using the dev VM you will need to follow these steps;
+
+1. Sign in/create a Notify account via [Notify](https://www.notifications.service.gov.uk)
+2. Add a new service
+3. Click on 'Templates' and create a new template with subject `((subject))` and message `((body))`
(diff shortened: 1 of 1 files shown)


source/manual/send-test-email.html.md
```diff
@@ -0,0 +1,25 @@
+---
+owner_slack: "#email"
+title: Send a test email via Notify
+section: Development VM
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2017-11-1
```
tuzz: Should this be `-01` instead of `-1` at the end? It might complain if it's not two-digits.

source/manual/send-test-email.html.md
```diff
@@ -0,0 +1,25 @@
...
+
+To send a test email using the dev VM you will need to follow these steps;
+
+1. Sign in/create a Notify account via [Notify](https://www.notifications.service.gov.uk)
+2. Add a new service
+3. Create a new template with subject `((subject))` and message `((body))`
+4. Add the test email address to the whitelist(under API integration) if it's different from your Notify email
+5. Create an API key
+6. Select the Key 'Team and whitelist'
+7. Run the rake task in email-alert-api using the generated key, template ID and test email
```
tuzz: Looks good. I was wondering if it might help to say exactly where in the Notify interface these things are.

For example, you put `under API integration` in parentheses, but it might be clearer to write:
> Click on 'API Integration' and add the test email ... 

And:
> Click on 'Templates' and ....

Maybe?

1pretz1: Okay, yes good idea! I'll make the changes.

PR #559, opened 2017-11-03
Add section for Incidence Management Guidance

pmanrubia:
We are not moving the content wiki into the Developers docs,
as there are plans to standardize this information before
making it public.

We have decided to add a link to the docs, so developers,
which are usually incident leads, can find easily this
information.


tijmenb:
@pmanrubia sorry just saw the URL - should it read `incident-management-guidance`?

pmanrubia:
Yes @tijmenb: I have created this: https://github.com/alphagov/govuk-developer-docs/pull/560

PR #566, opened 2017-11-07
Remove the environments page

MatMoore:
I don't think there is a real need for this page. It's trying
to describe what all the environments for while at the same time
giving a bit of technical detail about where they run.

If any of this information is actually valuable I think it should
be moved to other pages or separated out.

There is already a page about what the CI environment does here:
https://docs.publishing.service.gov.uk/manual/jenkins-ci.html

The training environment is not being used at the moment, and there are no
plans to do so until we've moved to AWS.

It would be useful to have something that describes the deployment process
in more detail, but the basic information about what environments exist is something
devs pick up when they join a team, and the getting started guide introduces them.


tijmenb:
Not sure about removing it entirely, but I agree with removing the technical details. I think it'd still be useful to point people to somewhere that explains what "integration" is.

MatMoore:
That was the approach I started with, but I realised I was just rewriting the entire page and wasn't sure how useful that was. I think it would be better to create a diagram of our deploy pipeline instead. Something where you see how code gets from one place to another. This could also inform discussions about moving towards continuous deployment.

PR #567, opened 2017-11-07
Update rails upgrade guide to include config guide

danielburnley:
As part of the dependency upgrade project we are working on standardising rails configs. This outlines the initial guidelines that were created

### DIFF
--- a/source/manual/upgrade-rails.html.md
+++ b/source/manual/upgrade-rails.html.md
@@ -10,6 +10,17 @@ parent: "/manual.html"
 
 When upgrading our apps between Rails major and minor versions, follow the [official Rails guides][guide]
 
+## Rails Configs
+
+Going forward we are trying to standardise our Rails config files. The guidelines
+for these are as follows:
+
+- If `load_defaults` is used, it should be top of the config
+- Config options should be alphabetised
+- Group together options that relate to the same option
+  - E.g. `config.assets.enabled` and `config.assets.prefix` should be grouped together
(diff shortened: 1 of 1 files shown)


source/manual/upgrade-rails.html.md
```diff
@@ -10,6 +10,17 @@ parent: "/manual.html"
 
 When upgrading our apps between Rails major and minor versions, follow the [official Rails guides][guide]
 
+## Rails Configs
+
+Going forward we are trying to standardise our Rails config files. The guidelines
+for these are as follows:
+
+- If `load_defaults` is used, it should be top of the config
+- Config options should be alphabetised
```
h-lame: I'd argue that the options in `config/application.rb`, and `config/environments/*.rb` should preserve the order of options generated by `rails new` (or `rails app:update`) to make it easier to compare your app's version to the new defaults when doing an upgrade.

Perhaps we should say (some more brief version of) the above and introduce the alphabetising and grouping rules for _new_ options, which should be added at the bottom of the file (as again this helps with diffs when upgrading).

andrewgarner: 👍 

I've found you get nice clean diffs each time you upgrade if you keep the comments and example config as they are generated by Rails and tailor any custom configuration to follow the conventions introduced in the upgraded. For example moving custom asset config to`/config/assets.rb` when Rails introduced that convention.

With that approach your upgraded app not only runs the new version of Rails, it also looks as if it could have been started at that version of Rails.

PR #571, opened 2017-11-10
Add info about integration shutdown/startup

chrisroos:
I had to find this information by asking in Slack. This should hopefully help people with the same question in future.


chrisroos:
Thanks @rubenarakelyan!

PR #578, opened 2017-11-16
Archive repos, don't send them to the attic

tijmenb:
This is the GDS Way now: https://github.com/alphagov/gds-way/pull/93

### DIFF
--- a/source/manual/retiring-an-application.html.md
+++ b/source/manual/retiring-an-application.html.md
@@ -4,7 +4,7 @@ title: Retire an application
 section: Packaging
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2017-10-17
+last_reviewed_on: 2017-11-16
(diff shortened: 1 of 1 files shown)


source/manual/retiring-an-application.html.md
```diff
@@ -139,10 +139,8 @@ items need to be unpublished and do it via the Publishing API.
...
 
-## 18. Move repo to attic
+## 18. Archive the repo
 
-Move the repository to the attic. Go into the repository settings in Github, and
-change the owner to [gds-attic][gds-attic]. Once moved, delete any webhooks and
-services that exist for it.
+Go into the repository settings in Github, and [archive the repo](https://github.com/blog/2460-archiving-repositories).
 
 [gds-attic]: https://github.com/gds-attic
```
dwhenry: Can we remove this line as well as it is no longer required

PR #582, opened 2017-11-17
Add pre-commit hook with vale linter

surminus:
This adds:

Vale lint: https://github.com/ValeLint/vale
pre-commit: https://github.com/pre-commit/pre-commit

Pre-commit hooks are a useful alternative to CI tests, and when someone adds or edits markdown documents we can run tests against the changes before they are commit.

Vale lint has some styleguides which it checks against, and throws errors for violations, and warnings for suggestions.

This will help us ensure our documentation is clear, concise and not full of filler language which makes it harder to understand and less accessible to people who do not have English as their first language.


nickcolley:
This is awesome, I like the `vale.GenderBias = warning` addition too.

Is this a pre-commit thing because we don't have CI for this repo?

surminus:
I added it as a pre-commit hook as it's good practice to catch problems before sending them to CI (if the tests are light and quick enough to run). 

If we wanted I don't think there would be a problem adding it to CI, but would require a little more work, so I wanted to see if people were happy getting the changes using pre-commit hooks.

I would be more than happy to set this up in the CI environment, because we could then add it a method in the Jenkins library, and if we wanted to, set it as a default to check markdown files against all our repositories.  

I'll wait for general consensus if we want it added to CI.

README.md
```diff
@@ -68,8 +68,24 @@ This will create a bunch of static files in `/build`.
 This project is re-deployed by a Jenkins task every hour (to pick up external
 changes). It is [hosted on S3][terraform].
 
+## Pre-commit hooks
```
MatMoore: Is it possible to make all of this available on the dev VM, or do you have to run `pre-commit install` per project?

surminus: They're linked to git config, so I think you would have to install for each project where the `.pre-commit-config.yaml` file exists.

.vale
```diff
@@ -0,0 +1,6 @@
+# Global settings (applied to every syntax)
+[*]
+# List of styles to load
+BasedOnStyles = vale
```
MatMoore: The PlainLanguage style also looks really useful - looks like they moved it out of the core one. https://github.com/ValeLint/vale/tree/master/styles/PlainLanguage

README.md
MatMoore: Ah ok. I think the git hook is still more useful than a CI job would be. Should we add some linux installation instructions here as well?

surminus: Will do.

.vale
surminus: That does look good. I'll try enabling it to see what output we return.

surminus: I enabled PlainLanguage, and most results were about slashes, and so there were quite a few false positives:

```
41:37  error    Use either 'or' or 'and' in     PlainLanguage.Slash
                        'alphagov/govuk'
```

I'm going to set that to `warning` only, but in the future I may look at increasing "wordiness" and "sentencelength" to `error` as these seem really useful from this style.

surminus:
I'm going to merge this because it only affects a user if they install the pre-commit hooks. I've just put myself on the agenda for the Tech Fortnightly so will talk about it then.

PR #583, opened 2017-11-20
Add docs for stopping integration shutdown and data sync.

dwhenry:
This is the first draft and still requires review and spell checking before being merged.


dwhenry:
This is not the process for this

PR #596, opened 2017-11-27
Update the taxonomy page

cbaines:
This is from the vision work done by the Taxonomy team in Q3. Drop the
review_in to 3 months, as the taxonomy is regularly changing at the
moment.


cbaines:
> Looks great

Thanks for reviewing Nick :)

PR #603, opened 2017-12-04
Fix page name and add tests

surminus:


tools/page-name-check.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
tijmenb: Do you think we can reuse https://github.com/alphagov/govuk-developer-docs/blob/master/spec/pages/manual_page_spec.rb for this?

surminus: Yup! I'll do that.

surminus:
@tijmenb updated.

PR #613, opened 2017-12-07
Add docs about user management in AWS

surminus:
Add initial docs about user management in AWS.

This has two links missing:

 - official guidance on the central GDS users account, which should be provided by Reliability Engineering
 - details on GOV.UK accounts and IDs. These could potentially be stored in the password store.

https://trello.com/c/hH4jI14U/873-add-docs-on-user-management

### DIFF
--- a/source/manual/user-management-in-aws.html.md
+++ b/source/manual/user-management-in-aws.html.md
@@ -1,0 +1,113 @@
+---
+owner_slack: "#govuk-infrastructure"
+title: User Management in AWS
+section: AWS
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2017-12-07
+review_in: 1 month
+---
+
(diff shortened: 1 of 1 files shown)


source/manual/user-management-in-aws.html.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+If successful, this will output some credentials. Store them in your environment using
+these environment variables:
+
+```
+AWS_ACCESS_KEY_ID
+AWS_SECRET_ACCESS_KEY
+AWS_SESSION_TOKEN
+```
```
afda16: Probably we could add how to switch role in the web console too

afda16:
Should the document be in a new AWS section of the manual?

surminus:
I don't think so. The header itself sets the section it's part of.

source/manual/user-management-in-aws.html.md
surminus: I will expand this along with the linked documentation.

PR #626, opened 2017-12-19
Update ssh tunnel and local dev vm docs for sidekiq-monitoring

h-lame:
The sidekiq-monitoring app now listens on port 3211 instead of port 80
so we need to change the port we tunnel to and from.  The dev VM now can
host the sidekiq-monitoring app directly without needing a tunnel so we
can remove that section entirely.

See: https://github.com/alphagov/govuk-puppet/pull/6981 for the changes
that support documentation.

### DIFF
--- a/source/manual/monitor-sidekiq-workers.html.md
+++ b/source/manual/monitor-sidekiq-workers.html.md
@@ -4,7 +4,7 @@ title: Monitor sidekiq queues for your application
 section: Monitoring
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2017-06-28
+last_reviewed_on: 2017-12-19
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manual/monitor-sidekiq-workers.html.md
```diff
@@ -15,20 +15,19 @@ We have restricted public access as the Web UI allows modifying the state of [Si
...
-To view your local Sidekiq queue, go to the [sidekiq-monitoring app](https://github.com/alphagov/sidekiq-monitoring) in the vm and run `./bin/foreman start` for all applications, or `./bin/foreman run <app_name>` for a specific app. In the output of sidekiq-monitoring, you'll find the port number of the app whose queue you want to view. Make a note of it, then in your local machine, go to govuk-puppet/development-vm and run:
+To view your local Sidekiq queue, go to the [sidekiq-monitoring app](https://github.com/alphagov/sidekiq-monitoring) in the vm and run `./bin/foreman start` for all applications, or `./bin/foreman run <app_name>` for a specific app.
 
-```bash
-$ vagrant ssh  -- -CNL <port_number>:localhost:<port_number>
-```
+Then visit:
 
-Then visit `http://localhost:<port_number>/<app_name>` to see a list of Sidekiq configurations you can monitor.
+* [`http://sidekiq-monitoring.dev.gov.uk:3211/`](http://sidekiq-monitoring.dev.gov.uk:3211/) to see a list of all the sidekiq monitoring apps that are available
```
andrewgarner: ~~sidekiq~~ Sidekiq

source/manual/monitor-sidekiq-workers.html.md
```diff
@@ -15,20 +15,19 @@ We have restricted public access as the Web UI allows modifying the state of [Si
...
+* `http://sidekiq-monitoring.dev.gov.uk/<app_name>` to directly a specific app
```
andrewgarner: Maybe "to directly monitor a specific app"?

source/manual/monitor-sidekiq-workers.html.md
andrewgarner: "sidekiq monitoring apps" is maybe a bit confusing. There is one Sidekiq monitoring app that lets you monitor multiple queues or GOV.UK applications?

source/manual/monitor-sidekiq-workers.html.md
h-lame: Gosh, you'd think I would have read this before creating the PR.  Thanks!

source/manual/monitor-sidekiq-workers.html.md
h-lame: I take your point.  So the reality is that there's one app called sidekiq-monitoring that acts as a proxy to all the individual sidekiq web ui sinatra apps that let you monitor the queues for a given GOV.UK application.  But that's a mouthful, so I've gone with `to see a list of all the GOV.UK applications whose Sidekiq status you can monitor` ?

h-lame:
Thanks for the review @andrewgarner, addressed your comments via a rebase.

PR #641, opened 2018-01-02
Update assets documentation

floehopper:
I've reviewed the the following pages and added notes about the ongoing asset management development work. I've also reduced the `review_in` period from 6 months to 1 month in both cases to reflect the ongoing work.

* "Assets: how do they work"
* "Attachment backups"


floehopper:
@danielroseman Thanks!

PR #648, opened 2018-01-03
Change root_taxons to level_one_taxons

cbaines:
The name of the link type is changing.


cbaines:
Thanks for reviewing Mat :)

PR #653, opened 2018-01-04
Removed the section about restarting frontend.

koetsier:
During the emergency banner drill the step to reload government
frontend was not needed and so the corresponding section was
removed from the manual.


Davidslv:
I think this is still a necessary step in a real case scenario.

PR #660, opened 2018-01-05
Taxonomy docs changes

cbaines:


cbaines:
Thanks for reviewing Mat :)

PR #661, opened 2018-01-05
Add some documentation on Metrics

cbaines:
High level stuff, just explaining the tooling.


cbaines:
> Might be good to link to https://github.com/etsy/statsd/blob/master/docs/metric_types.md for statsd. I found it useful for understanding gauges.

Thanks for reviewing Mat :) I've added a link in to the StatsD docs.

PR #671, opened 2018-01-09
Update SSH docs

surminus:
Now DNS is migrated we can make the docs more consistent.

I have also added a (simplified) SSH configuration for people to use in the future.


tuzz:
Makes sense to me. I'll let a second pair of eyes approve. 👀  👍

surminus:
I repushed to pass the tests by adding additional `#`.

PR #675, opened 2018-01-11
Review infrastructure AWS docs

issyl0:


issyl0:
Should we set the channel to not be govuk-infrastructure given that's going away? Who looks after these docs once we're... _sniff_... gone?

issyl0:
On second thoughts, I'll do that in a separate PR as there are probably loads more.

PR #676, opened 2018-01-11
Change #govuk-infrastructure docs to be owned by #2ndline

issyl0:


tijmenb:
Should these be assigned to `reliability-eng` instead?

surminus:
I would argue no as they discuss things owned by GOVUK rather than Reliability Engineering.

PR #678, opened 2018-01-15
Update Kibana query for puppet last run errors

suzannehamilton:
Fix quote style in Kibana query and explain the `hostname` format.


suzannehamilton:
That's a good point - I hadn't thought about the AWS hostnames. I can't find a better field to query instead.

If it makes it harder to debug things, maybe we should work out how to label these errors with more human-readable machine names.

PR #685, opened 2018-01-24
Remove the friendlygiraffe username

chao-xian:
The example is not helpful as 2 developers took it at face value
and tried to ssh with that username within hours of each other.

Also added a missing `.integration` so that people can actually
ssh to the right box.

### DIFF
--- a/source/manual/howto-ssh-to-machines-in-aws.html.md
+++ b/source/manual/howto-ssh-to-machines-in-aws.html.md
@@ -46,7 +46,7 @@ ip-10-1-6-88.eu-west-1.compute.internal
 
 SSH with agent forwarding to the jumpbox:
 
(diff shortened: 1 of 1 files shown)


source/manual/howto-ssh-to-machines-in-aws.html.md
```diff
@@ -46,7 +46,7 @@ ip-10-1-6-88.eu-west-1.compute.internal
 
 SSH with agent forwarding to the jumpbox:
 
-        ssh -A friendlygiraffe@integration
+        ssh -A <EXAMPLEUSER>@integration
```
bevanloon: perhaps `$USERNAME` instead of `<EXAMPLEUSER>` since that should just work even if copy-and-pasted.

PR #687, opened 2018-01-25
Add content schema keywords

deborahchua:
The Modelling Services team have been working on the content schema for step by step navigtion (previously called `task lists`).

We had to do a bit of research to find out how to represent certain properties of our task list, and thought it would be good to add to the Developer docs.

Please feel free to add to this.

screen shot 2018-01-25 at 16 39 13

### DIFF
--- a/source/content-schemas.html.md.erb
+++ b/source/content-schemas.html.md.erb
@@ -5,4 +5,14 @@ title: Content schemas
 
 The content schemas are defined in the [govuk-content-schemas repo][repo].
 
+| Keyword       | Explanation                                                                            |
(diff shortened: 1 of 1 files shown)


tijmenb:
Very nice! I wonder if this is better in the govuk-content-schemas directory. There we already link to the json-schema site, which has a bunch of examples (http://json-schema.org/examples.html). Those might be good to link to as well. https://spacetelescope.github.io/understanding-json-schema is also a good resource.

deborahchua:
Thanks @tijmenb, I have moved this to govuk-content-schemas now: https://github.com/alphagov/govuk-content-schemas/pull/708

PR #693, opened 2018-01-29
Update application ownership for teams

tijmenb:
See individual commits for details.


tijmenb:
Not for all the teams!

PR #705, opened 2018-02-07
add instruction to the getting started page

hannako:
This PR adds an instruction on how to import a new ssh key into keychain.

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -75,6 +75,10 @@ To create a new account, start by creating an SSH key at least 4096 bits long. F
 
     mac$ ssh-keygen -t rsa -b 4096 -C "your_email@example.com" -f ~/.ssh/alphagov
 
+Import the SSH key into your keychain.
+
+    mac$ /usr/bin/ssh-add -K ~/.ssh/alphagov
+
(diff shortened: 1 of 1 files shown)


source/manual/get-started.html.md
```diff
@@ -75,6 +75,10 @@ To create a new account, start by creating an SSH key at least 4096 bits long. F
 
     mac$ ssh-keygen -t rsa -b 4096 -C "your_email@example.com" -f ~/.ssh/alphagov
 
+Import the SSH key into your keychain.
+
+    mac$ /usr/bin/ssh-add -K ~/.ssh/alphagov
```
andrewgarner: To be consistent we shouldn't include the absolute path to `ssh-add`. This follows the precedence on line 78 with `ssh-keygen`.

hannako: Thanks Andrew. Do you want to apply the same logic to the previous section as well in that case? Ie line 60

andrewgarner: Just looked into this some more. I think maybe ignore my comment. The explicit path is because the `-K` argument that adds the key to your keychain only ships with the default macOS `ssh-add` command. Without the path you may end up invoking a version installed by homebrew.

PR #711, opened 2018-02-09
Update assets.html.md

chrisroos:
Based on the work we've been doing recently.

I've set the review date to be a week from now in the hope that we'll
have made some more changes that need to be documented.


chrisroos:
Thanks @floehopper!

PR #712, opened 2018-02-09
Adds docs about how to remove a change note

elenatanasoiu:
For: https://trello.com/c/sOPRBrrT/63-2-remove-change-note-and-redirect-html-attachment-url

This documents the process to go through when having to remove a change note from Whitehall.

### DIFF
--- a/source/manual/howto-remove-change-note-from-whitehall.html.md
+++ b/source/manual/howto-remove-change-note-from-whitehall.html.md
@@ -1,0 +1,39 @@
+---
+owner_slack: "#2ndline"
+title: Remove a change note in Whitehall
+section: Whitehall
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2018-02-09
+review_in: 6 months
+---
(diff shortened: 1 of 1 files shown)


source/manual/howto-remove-change-note-from-whitehall.html.md
```diff
@@ -0,0 +1,39 @@
...
+and presents them to the Content Store. You can read more about this in the 
+Publishing API [docs](https://docs.publishing.service.gov.uk/apis/publishing-api/model.html#changenote).
+
+### Remove a Change Note
+
+Steps to undertake when someone has requested that you remove a change note 
+from Whitehall: 
+
+1. Obtain the ID of the edition on which the change note was created. 
+2. Create a migration in Whitehall where you search for the edition by ID
```
dwhenry: it is best practice to number these all `1.` so that adding/removing a step is easier to do in the future.

Markdown will automatically generate the correct numbering sequence.

PR #718, opened 2018-02-14
Update docs for SSHing to environment

surminus:
It's always a good idea to abstract the technical detail into a wrapper coded in bash, so add documentation about how to use this particular wrapper.

This documents how to use the `govukcli` tool which will hopefully make SSHing into different environments easier. This also allows the use of one method to use in different environments while we migrate.

### DIFF
--- a/source/manual/howto-ssh-to-machines-in-aws.html.md
+++ b/source/manual/howto-ssh-to-machines-in-aws.html.md
@@ -4,8 +4,8 @@ title: SSH into AWS machines
 section: AWS
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2018-01-11
-review_in: 3 months
+last_reviewed_on: 2018-02-14
+review_in: 2 months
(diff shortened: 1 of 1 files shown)


MatMoore:
This looks great, just have a couple of questions:

- Is there an easy way to list node classes using this tool? (I usually forget what things are called)
- Should installing it be a step in the getting started guide?

surminus:
@MatMoore two good comments! I'll think about how we can list classes using it, and yes, getting started seems like a good place :)

surminus:
@MatMoore I have added this functionality in this PR: https://github.com/alphagov/govuk-aws/pull/486

Try switching to that branch and giving it a shot. It works something like:

```
🐆  ~ govukcli ssh node-types
apt
asset_master
backend
bouncer
cache
calculators_frontend
content_store
...
```

Only available in AWS as I'm using AWS tags to find them, but hopefully this is useful!

PR #720, opened 2018-02-16
Add ValeLint to CI

surminus:
I added Vale as a pre-commit hook in b3dcb10783b238de3fbc0acc808da2647ee851f7.

The idea is that we can "lint" our documentation for grammar and readability. Documentation is often as important as the code we have to support, so we should ensure it is concise and accessible, particularly to people where English is not their first language.

This adds the tool to the CI chain. It can be further configured using `.vale`, and further styles may be added to `styles/`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,6 +38,10 @@ node {
       govuk.rubyLinter("app lib helpers spec bin")
     }
 
+    stage("Lint documentation") {
+      sh "vale --glob='*.md' ."
+    }
+
     stage("Tests") {
       govuk.runTests()
     }
--- a/.vale
+++ b/.vale
@@ -1,12 +1,10 @@
 StylesPath = styles/
-MinAlertLevel = warning
+MinAlertLevel = error
 
 # Global settings (applied to every syntax)
 [*]
 # List of styles to load
 BasedOnStyles = vale, PlainLanguage
-# Set level for styles
-vale.GenderBias = warning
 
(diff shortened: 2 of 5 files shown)


surminus:
Oof, Jenkins didn't like it.

Jenkinsfile
```diff
@@ -38,6 +38,10 @@ node {
       govuk.rubyLinter("app lib helpers spec bin")
     }
 
+    stage("Lint documentation") {
+      sh "find . -name '*.md' |xargs ./vale-lint.sh"
```
timblair: Vale has a `--glob` option. Could we use that instead?

```sh
vale --glob='*.md' .
```

If we do, it avoids using the separate shell script, and avoids the cycle of "fix file → check → fail on new file → fix new file..." that happens at the moment, because the script exits on the first failure.

timblair: Ah, unless you did it like that on purpose to only highlight errors, rather than outputting a deluge of warnings on every run? Might want to tweak the shell script to output all failures then, rather than exiting on the first one it finds.

surminus: I did it like that I think originally for use with pre-commit hooks, but CI it's different so glob may be more appropriate as like you said it does fail -> fix -> fail -> fix. 

I'll see if I can suppress all the warnings as well so we just throw errors.

surminus:
I updated this again to only show errors, and do a general run across all files.

PR #736, opened 2018-02-27
Point to saved searches

MatMoore:
Some of the kibana examples here are still useful because 
they show how queries can be made more specific, and saved searches
aren't parameterisable.


h-lame:
@MatMoore - do you want to reword that commit message - it looks to be missing some words.

MatMoore:
Whoops. Have amended.

PR #737, opened 2018-02-27
Review

maxgds:
Several approvals after review and a couple of minor tweaks.

### DIFF
--- a/source/manual/mysql.html.md
+++ b/source/manual/mysql.html.md
@@ -4,7 +4,7 @@ title: MySQL backups
 section: Backups
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2017-08-09
+last_reviewed_on: 2018-02-27
 review_in: 6 months
 ---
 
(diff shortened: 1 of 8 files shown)


source/manual/mysql.html.md
```diff
@@ -4,15 +4,15 @@ title: MySQL backups
...
+last_reviewed_on: 2018-02-27
 review_in: 6 months
 ---
 
 ## automysqlbackup
 
 This is how MySQL backups have traditionally been taken on the GOV.UK Infrastructure.
 
-A third-party script called [automysqlbackup](http://www.debianhelp.co.uk/mysqlscript.htm) takes a `mysqldump` every night and stores them on disk
+A third-party script called automysqlbackup takes a `mysqldump` every night and stores them on disk
```
dwhenry: I think this is the backup script we are now using: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_mysql/templates/automysqlbackup if we want to update it

maxgds:
@dwhenry Good to go, now?

PR #738, opened 2018-02-27
Renaming doc

elenatanasoiu:
This will make it clear it is an alert and not a manual page.
We've also reassigned the category to Icinga Alerts.


source/manual/alerts/recreate-a-mongo-instance-in-aws.html.md
```diff
@@ -1,7 +1,7 @@
 ---
 owner_slack: "#2ndline"
 title: Re-create an AWS mongo instance
-section: Development VM
+section: Icinga Alerts
```
suzannehamilton: I think the sections are case-sensitive, so this might need to be "Icinga alerts".

PR #741, opened 2018-03-02
UpdateJenkinsfile to use external library

surminus:


surminus:
I created this PR as a test for a script I'm going to run across everything. Worked as intended! Closing and another one will automagically be opened.

PR #746, opened 2018-03-05
Add an MFA serial to the gds AWS profile

boffbowsh:
This is needed by govukcli for `aws sts get-session-token` in https://github.com/alphagov/govuk-aws/pull/507


issyl0:
Can we recommend using `govukcli` as a wrapper? It would save a lot of the faff or manually exporting credentials, or people writing their own wrappers.

PR #754, opened 2018-03-06
Use an explicit fab class for emergency publishing

boffbowsh:
A recent fabric-scripts change broke fabric's ability to find
suitable servers from the role attached to a task definition.
Until this is fixed, we need to specify the class explicitly
when flushing caches.


kevindew:
There's also the clear_cached_templates job which you may need to break up as it runs different jobs on different classes.

Did you mean to include the stylesheet changes?

boffbowsh:
One day I'll learn to stop just doing `git add .`

PR #778, opened 2018-03-13
Update last reviewed on date

KyleMacPherson:
Reviewed the doc and updated last_reviewed_on


kevindew:
I've just added a commit to fix some of the headers. I'd also suggest that you use the titles of these PRs to state the files eg: "Bump review data for database guis". This would make the various PRs easier to distinguish between (and would be easier reading in the git log)

PR #780, opened 2018-03-13
Clarify how to stop hourly copy.

KyleMacPherson:
- Specificy `govuk_sync_mirror` as the cron job that needs to be stopped
in order to stop the hourly copy from the mirrorer machine overwriting the
copy.
- Update `last_reviewed_on`


tijmenb:
@KyleMacPherson can this be merged?

KyleMacPherson:
@tijmenb No probably not, opened this PR 3 months ago so probably worth someone having a look at it with a fresh pair of eyes as I do not know how much has changed since then. Will close this PR now.

PR #785, opened 2018-03-15
Update components.html

fofr:
* Add link to gem component guide
* Indicate where shared components now live and that static components are being deprecated


fofr:
I wanted to add links to some of the new component guides, now we’ve added components to a lot more apps, but I don’t think we’ve set up shareable links to those guides using Heroku yet

PR #786, opened 2018-03-15
Replicate app data locally fixes

andrewgarner:
Fixes the 'AWS access' section so that it displays correctly as an ordered list. We need to indent the content within each list item or the markdown renderer interprets each list item as a new list.

Also contains a number of spelling, grammar and punctuation fixes.

#### Before:

screen shot 2018-03-15 at 12 19 58

#### After:

screen shot 2018-03-15 at 12 19 31

### DIFF
--- a/source/manual/replicate-app-data-locally.html.md
+++ b/source/manual/replicate-app-data-locally.html.md
@@ -50,7 +50,7 @@ want to do developer replication you will need to:
 
 1. Install the AWS CLI application locally
 
-    On MacOS this can be done using `brew install awscli`. This is being installed locally with the assumption you will be doing a two step backup process, i.e. download the files on your local machine and then update you VM. This is the recommended process as the download is quicker to you local machine for most users.
+    On macOS this can be done using `brew install awscli`. This is being installed locally with the assumption you will be doing a two step backup process, i.e. download the files on your local machine and then update you VM. This is the recommended process as the download is quicker to you local machine for most users.
 
 1. Setup your AWS access config and credentials files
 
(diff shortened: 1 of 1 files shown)


source/manual/replicate-app-data-locally.html.md
```diff
@@ -36,54 +36,55 @@ want to do developer replication you will need to:
...
 
-Role ARN: should be one of the following depending on the level of access you have been configured with [here](https://github.com/alphagov/govuk-aws-data/blob/master/data/infra-security/integration/common.tfvars):
+    Role ARN: should be one of the following depending on the level of access you have been configured with [here](https://github.com/alphagov/govuk-aws-data/blob/master/data/infra-security/integration/common.tfvars):
 
-* govuk-admins: arn:aws:iam::210287912431:role/govuk-administrators
-* govuk-poweruser: arn:aws:iam::210287912431:role/govuk-poweruser
-* govuk-users: arn:aws:iam::210287912431:role/govuk-users
+        * govuk-admins: arn:aws:iam::210287912431:role/govuk-administrators
+        * govuk-poweruser: arn:aws:iam::210287912431:role/govuk-poweruser
+        * govuk-users: arn:aws:iam::210287912431:role/govuk-users
```
cbaines: I don't think this bit needs to be indented this much, as I think this will lead to it being rendered as code, rather than a list.

andrewgarner: I thought that too, but without it the subsequent paragraph rendered as a code block. It looked fine when GitHub rendered the markdown but not when rendered in the tech docs template.

PR #789, opened 2018-03-20
Update testing Email Alert API emails documentation

kevindew:
This:

- tries to make it clearer that emails refer to Email Alert API;
- rephrases the instructions for greater clarity;
- includes the instruction about adding yourself to Notify;
- includes information on the rake task to test process.


source/manual/receiving-emails-from-email-alert-api-in-integration-and-staging.html.md
```diff
@@ -0,0 +1,34 @@
+---
+owner_slack: "#email"
```
bevanloon: I wonder if we should push this to platform support given email doesn't look like it's carrying on next quarter?

kevindew: I think taxonomy would own it next quarter. Could be we set the review date lower because of that though?

I'd not want to unnecessarily put it into platform support as it may just get lost amongst all the docs they have.

bevanloon: That sounds fair enough. A review date towards the end of the next quarter would probably be good because then we can figure out which team should own it when we have an idea of what teams will exist.

PR #791, opened 2018-03-21
Amend docs on cookies and sessions

rubenarakelyan:
This commit amends the documentation on cookies and sessions now that email-alert-frontend uses both.


source/manual/cookies-and-sessions-in-frontend-apps.html.md
```diff
@@ -4,29 +4,28 @@ parent: "/manual.html"
 layout: manual_layout
 section: CDN & Caching
 owner_slack: "@bevan"
```
tijmenb: Should we change the owner to 2nd line or devs?

rubenarakelyan: Yup, changed it to #govuk-developers

PR #796, opened 2018-03-22
Review manual page: on call

laurentqro:


source/manual/on-call.html.md
```diff
@@ -4,7 +4,7 @@ title: Out of hours support (on-call)
 section: Support
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2017-09-21
+last_reviewed_on: 2018-09-22
```
thomasleese: Missed the month here :)

PR #828, opened 2018-03-27
Use correct case for 'Elasticsearch'

andrewgarner:


andrewgarner:
@tijmenb good idea. The regex might be a bit complex so we don't get false positives for references to the `elasticsearch` command though.

PR #831, opened 2018-03-28
Update HMRC PAYE tool instructions

suzannehamilton:
- Explain how to scp files to the upload directory as the assets user
- Fix the fab script for purging the cache

I'm not entirely sure why the old versions of these commands didn't work, so I'd welcome suggestions for alternative fixes.

I wonder whether the owner of `/mnt/uploads/whitehall/clean/uploaded/hmrc/` changed to `assets` as part of the recent asset migration, but I haven't been able to track down the change.

I don't know whether the fab scripts changed to require a host class or if the `cdn:purge_all` command never worked exactly as written.

### DIFF
--- a/source/manual/hmrc-paye-files.html.md
+++ b/source/manual/hmrc-paye-files.html.md
@@ -4,7 +4,7 @@ title: Upload HMRC PAYE files
 section: Publishing
 layout: manual_layout
 parent: "/manual.html"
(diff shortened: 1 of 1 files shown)


chrislo:
FYI: HMRC assets are still being served from Whitehall. There's some more information in https://github.com/alphagov/asset-manager/issues/217.

PR #860, opened 2018-04-04
Update Assets documentation

chrislo:
I've made some small changes to the existing docs to reflect the current reality but I think we could also add some more detail. What do you think @floehopper @chrisroos?

### DIFF
--- a/source/manual/assets.html.md
+++ b/source/manual/assets.html.md
@@ -44,18 +44,15 @@ There are currently two systems for uploading, managing and serving
 user-supplied assets on GOV.UK.
 
 Asset Manager is an API that is called internally by Publisher, Specialist
-Publisher, Manuals Publisher and Travel Advice Publisher to manage their
+Publisher, Manuals Publisher, Travel Advice Publisher and Whitehall to manage their
 uploads. It serves the uploaded assets on assets.publishing.service.gov.uk/media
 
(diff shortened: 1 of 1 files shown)


source/manual/assets.html.md
```diff
@@ -44,27 +44,24 @@ There are currently two systems for uploading, managing and serving
...
+Publisher, Manuals Publisher, Travel Advice Publisher and Whitehall to manage their
 uploads. It serves the uploaded assets on assets.publishing.service.gov.uk/media
 
-Whitehall is a standalone publishing app that manages its own uploaded assets.
-It also serves those assets, via both
-assets.publishing.service.gov.uk/government/uploads and
-www.gov.uk/government/uploads, even for content that has been migrated to the
+Whitehall is a standalone publishing app that manages a type of asset called "attachments". It serves attachments, via both
+`assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data` and
+`www.gov.uk/government/uploads/government/uploads/system/uploads/attachment_data`, even for content that has been migrated to the
```
floehopper: Since https://github.com/alphagov/whitehall/pull/3918 requests for attachments on `www.gov.uk` are now redirected to `assets.publishing.service.gov.uk`, so it might be simpler not to mention `www.gov.uk` here...?

PR #864, opened 2018-04-05
Document issues I encountered following the getting started docs

barrucadu:
~Also resurrects the `troubleshooting-vagrant` page, which was referred to on the `get-started` page, but didn't exist.~

~Alternatively, that page could remain as-is and I could put the extra information directly in the `get-started` page.  This might be better, as I'm not sure an issue with VirtualBox installation really belongs on a troubleshooting page.~

The new troubleshooting stuff is scattered around the existing pages, and I've added a couple of extra links to the getting started page to aid discoverability.

### DIFF
--- a/source/manual/replicate-app-data-locally.html.md
+++ b/source/manual/replicate-app-data-locally.html.md
@@ -128,6 +128,12 @@ Then follow the instructions above for importing using the `-s` flag.
 
 See [running out of disk space in development](/manual/development-disk-space.html).
 
+## If you get a curl error when restoring Elasticsearch data
+
+Check the service is running:
+
+    dev$ sudo service elasticsearch-development.development start
+
 ## Can't take a write lock while out of disk space (in MongoDB)
 
 You may see such an error message which will prevent you from creating or even dropping collections. So you won't be able to replicate the latest data.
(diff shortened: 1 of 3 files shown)


source/manual/replicate-app-data-locally.html.md
```diff
@@ -126,7 +126,13 @@ Then follow the instructions above for importing using the `-s` flag.
 
 ## If you're running out of disk space
 
-See [running out of disk space in development](/manual/development-disk-space.html).
+See [troubleshooting Vagrant](/manual/troubleshooting-vagrant.html).
```
tijmenb: I think we split up `troubleshooting-vagrant.html` because it became too large (https://github.com/alphagov/govuk-developer-docs/pull/462) so I wonder if it's better to create a new "VirtualBox installation fails on High Sierra" page. 

We should probably redirect `/manual/troubleshooting-vagrant.html` somewhere appropriate. Where was it linked from?

barrucadu: It was only linked from `get-started`, so it was a bit irrelevant as people probably won't run out of space while first installing everything.

PR #866, opened 2018-04-06
Jenkins troubleshooting documentation

kevindew:
Off the back of the issue we've had with Jenkins over the last few
weeks, I've documented some of the information we've gleaned and a
process to clean up the situation that was causing us problems.

### DIFF
--- a/source/manual/troubleshooting-jenkins-performance.html.md
+++ b/source/manual/troubleshooting-jenkins-performance.html.md
@@ -1,0 +1,47 @@
+---
+owner_slack: "#2ndline"
+title: Troubleshooting CI Jenkins Performance
+parent: "/manual.html"
+layout: manual_layout
+section: Testing
+last_reviewed_on: 2018-04-06
+review_in: 1 year
+---
+
+There are many reasons Jenkins could perform poorly, this document attempts to
+outline some common approaches to diagnosing/monitoring and outlines steps to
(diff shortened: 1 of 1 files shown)


source/manual/troubleshooting-jenkins-performance.html.md
```diff
@@ -0,0 +1,47 @@
...
+section: Testing
+last_reviewed_on: 2018-04-06
+review_in: 1 year
+---
+
+There are many reasons Jenkins could perform poorly, this document attempts to
+outline some common approaches to diagnosing/monitoring and outlines steps to
+resolve common issues.
+
+A useful link is [/monitoring][monitoring] which shows a variety of diagnostic
```
cbaines: Really interesting! This I did not know.

source/manual/troubleshooting-jenkins-performance.html.md
```diff
@@ -0,0 +1,47 @@
...
+information. This can be used to identify which aspects of Jenkins
+performance/usage have changed over periods of time.
+
+## There are resources Jenkins was not able to dispose automatically
+
+This is an error that can be seen on the [/manage][manage] page in Jenkins and
```
cbaines: I can't see this, but I guess that could be because there are no files Jenkins can't delete at the moment.

PR #873, opened 2018-04-12
Add whitehall url to getting started guide

MahmudH:
- Include url in step 8 of guide

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -229,6 +229,12 @@ If you don't need an optional dependency, you can pass the `-w` option:
 
 If these `bowl` commands fail, try the troubleshooting guide on [how to fix a broken bowl][bowl-error].
 
+Now visit this url once the app is running:
+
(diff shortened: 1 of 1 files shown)


source/manual/get-started.html.md
```diff
@@ -229,6 +229,12 @@ If you don't need an optional dependency, you can pass the `-w` option:
 
 If these `bowl` commands fail, try the troubleshooting guide on [how to fix a broken bowl][bowl-error].
 
+Now visit this url once the app is running:
```
cbaines: I'd suggest capitalising "url" here, as it's an initialism.

source/manual/get-started.html.md
```diff
@@ -229,6 +229,12 @@ If you don't need an optional dependency, you can pass the `-w` option:
...
+
+`http://whitehall-admin.dev.gov.uk/`
+
+You should be able to see whitehall.
```
cbaines: I'd also suggest capitalising the "w" of "whitehall", as this is the name of the application.

PR #886, opened 2018-04-17
Fix ssh instructions in Emergency banner docs

leenagupte:
Trello: https://trello.com/c/eyPmZqho/7-emergency-publishing-drill

### DIFF
--- a/source/manual/emergency-publishing.html.md
+++ b/source/manual/emergency-publishing.html.md
@@ -208,19 +208,26 @@ various points in our stack as well as locally in your browser. Things to try:
 
 You can manually check whether the data has been stored in Redis by the Jenkins job on one of the frontend machines.
 
-From your development machine SSH into a frontend box appropriate to the environment you want to check
+1) From your development machine SSH into a frontend box appropriate to the environment you want to check.
+
+For staging or production:
+```
+ssh frontend-1.frontend.staging
+```
+
+for integration:
 
 ```
-ssh frontend1.frontend.integration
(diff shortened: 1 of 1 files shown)


source/manual/emergency-publishing.html.md
```diff
@@ -211,7 +211,7 @@ You can manually check whether the data has been stored in Redis by the Jenkins
 From your development machine SSH into a frontend box appropriate to the environment you want to check
 
 ```
-ssh frontend1.frontend.integration
+ssh frontend-1.frontend.integration
```
cbaines: Well... this won't work either, given the changes to AWS?

cbaines: Or rather, given the move to AWS.

leenagupte: @cbaines True. I'll change the original command to point to staging and see if i can figure out how to run it on integration.

bevanloon: @leenagupte - if you want it to point specifically to integration, one should be able to do something like 

```
ssh integration -At "ssh -q \`govuk_node_list --single-node -c frontend\`"
```

to get straight on to an integration box

Or you could go for `govukcli` https://github.com/alphagov/govuk-aws/blob/master/doc/guides/connecting-to-the-environment.md

leenagupte: @bevanloon @cbaines I've updated the command for integration.

PR #896, opened 2018-04-27
Add docs about checking if a process is running

thomasleese:
This will be attached to the various Icinga alerts we have about checking that a process is running as part of making sure that all alerts have documentation associated with them.

[Trello Card](https://trello.com/c/x9gjfwqn/117-find-out-how-many-alerts-do-not-point-to-documentation-on-how-to-fix-the-problem-2)

### DIFF
--- a/source/manual/alerts/check-process-running.html.md
+++ b/source/manual/alerts/check-process-running.html.md
@@ -1,0 +1,43 @@
+---
+owner_slack: "#2ndline"
+title: Check process running
+section: Icinga alerts
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2018-04-27
+review_in: 6 months
+---
+
+This alert means that a process which should be running is not.
+
+It's likely that this process corresponds to a service, which you can check by looking in `/etc/init` or try running:
+
+```bash
+$ sudo service <process> status
+```
+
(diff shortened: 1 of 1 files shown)


source/manual/alerts/check-process-running.html.md
```diff
@@ -0,0 +1,37 @@
...
+last_reviewed_on: 2018-04-27
+review_in: 6 months
+---
+
+This alert means that a process which should be running is not.
+
+You can check the status of the service to confirm the alert by using:
+
+```bash
+$ sudo service --status-all | grep <process>
```
kevindew: is this better than running `sudo service <process> status`?

source/manual/alerts/check-process-running.html.md
```diff
@@ -0,0 +1,37 @@
...
+
+```bash
+sudo service <process> start
+```
+
+If the process is referring to a GOV.UK application, it may be necessary to also restart the Procfile worker:
+
+```bash
+sudo service <process> restart
+sudo service <process>-procfile-worker restart
```
kevindew: should this be more like <app-name>-procfile-worker?

thomasleese:
@kevindew Thanks, I've tried to clarify the wording between service and process, and also addressed your other comments.

PR #906, opened 2018-05-03
Update howto-ssh-to-machines-in-aws.html.md

Davidslv:


source/manual/howto-ssh-to-machines-in-aws.html.md
```diff
@@ -76,3 +76,9 @@ the traditional way:
 You may also use the node class to select a random node:
 
 `govukcli ssh calculators_frontend`
+
+### Troubleshooting
+
+Sometimes you might try to ssh into a server and nothing happens, double check that you added the key into the keychain like so: `ssh-add -K ~/.ssh/id_rsa`.
```
MatMoore: When you say nothing happens, do you mean you get "permission denied (private key)" or something? If so it would be good to include that output here, so the troubleshooting is like 'if X then try Y'

Davidslv: @MatMoore that's what happened before doing `ssh-add -K ~/.ssh/id_rsa`

screen shot 2018-05-08 at 10 15 06

PR #911, opened 2018-05-04
How to monitor rabbitmq on integration/dev

MatMoore:
The previous command doesn't work any more since we moved to AWS.

The ssh-fu is based on what we suggest in the sidekiq monitoring page.
If there's a simpler way to do this please let me know and I'll
update all the examples.

The user journey is still broken because you can't get the password
for integration without getting full production access.
Devs do not get access to govuk-secrets on day 1.

### DIFF
--- a/source/manual/rabbitmq.html.md
+++ b/source/manual/rabbitmq.html.md
@@ -35,18 +35,22 @@ in the `publishing-api` app.
 
 1.  Create an SSH tunnel to access the web control panel
 
-    For example for integration:
+    For Carenza environments:
 
-    > ssh rabbitmq-1.backend.integration -L 15672:127.0.0.1:15672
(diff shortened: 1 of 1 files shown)


source/manual/rabbitmq.html.md
```diff
@@ -35,16 +35,20 @@ in the `publishing-api` app.
...
 
 2.  Log in to the web control panel
 
     Point your browser at <http://127.0.0.1:15672>
 
     The username is root. The password you can obtain from the govuk-secrets
-    repo. Look for govuk\_rabbitmq::root\_password in the file for the
+    repo if you have access. Look for govuk\_rabbitmq::root\_password in the file for the
     relevant environment in
     <https://github.com/alphagov/govuk-secrets/tree/master/puppet/hieradata>
```
issyl0: You'll probably want to update this to be `hieradata_aws` as it's integration.

PR #927, opened 2018-05-17
Update mongod-replication-lag.html.md

Davidslv:


source/manual/alerts/mongod-replication-lag.html.md
```diff
@@ -13,7 +13,7 @@ review_in: 6 months
 There is a Fabric task to show various MongoDB replication status
 information.:
 
-    fab <environment> -H mongo-?.backend mongo.status
+    fab <environment> -H api-mongo-[n].api mongo.status
```
thomasleese: Would be good to wrap these in code blocks.

thomasleese: Oops... I just realise they are because they're tabbed in!

PR #939, opened 2018-05-18
A guide to removing users from Puppet

kevindew:
This is an attempt to clarrify the process that should be used as
different developers seem to approach this in different ways and if you
log onto any of our machines you'll see many home directories for staff
that have left.

A better alternative would be that the govuk_user module removed every
home directory of users who aren't available as this would simplify
the process.

### DIFF
--- a/source/manual/removing-a-user-from-puppet.html.md
+++ b/source/manual/removing-a-user-from-puppet.html.md
@@ -1,0 +1,37 @@
+---
+owner_slack: "#govuk-2ndline"
+title: Removing a user from Puppet
+parent: "/manual.html"
+layout: manual_layout
+section: 2nd line
+last_reviewed_on: 2018-05-18
+review_in: 12 months
+---
+
+Removing a user from our infrastructure via Puppet is a 2 change process that
+requires a deploy in the middle. The first change ensures that when puppet
+runs the users home directory is removed, the second change removes the
+user from puppet itself. If the user is just removed from puppet their files
+will remain on our servers forever more.
+
+1. First find the user manifest in: [modules/users/manifests][manifest-path].
+1. Add an entry to the govuk_user class of `ensure => absent,`. Here is an
+   [example][absent-example].
+1. Once this has been raised as a PR and merged, deploy puppet to all
+   environments.
+1. Create another PR for puppet that:
+  - Removes the user manifest file
+  - Removes the user from [Integration users][integration-users]
+  - Removes the user from [CI users][ci-users]
+1. Create a PR in [GOV.UK secrets][govuk-secrets] that:
+  - Remove the user from [production hieradata][production-hieradata]
+  - Remove the user from [AWS production hieradata][aws-production-hieradata]
+1. Once these have been merged, deploy puppet again to all environments.
+
+[manifest-path]: https://github.com/alphagov/govuk-puppet/tree/master/modules/users/manifests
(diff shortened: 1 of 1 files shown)


source/manual/removing-a-user-from-puppet.html.md
```diff
@@ -0,0 +1,26 @@
...
+runs the users home directory is removed, the second change removes the
+user from puppet itself. If the user is just removed from puppet their files
+will remain on our servers forever more.
+
+1. First find the user manifest in: [modules/users/manifests][manifest-path]
+1. Add an entry to the govuk_user class of `ensure => absent,` -
+   [example][absent-example]
+1. Once this has been raised as a PR and merged, deploy puppet to all
+   environments
+1. Create another PR that removes the user manifest file
```
cbaines: ... and the references to the user in the `hieradata`, both in `govuk-puppet` and `govuk-secrets`.

source/manual/removing-a-user-from-puppet.html.md
```diff
@@ -0,0 +1,26 @@
...
+1. Deploy puppet again
+
+[manifest-path]: https://github.com/alphagov/govuk-puppet/tree/master/modules/users/manifests
+[absent-example]: https://github.com/alphagov/govuk-puppet/pull/7319/files#diff-23436d1831fbc44a8f2bb652b9233d6cR4
```
cbaines: I'd suggest a different example, as the change in this example wouldn't have been very effective, as I believe you need to keep the hieradata when you add `ensure => absent`. Perhaps use  https://github.com/alphagov/govuk-puppet/commit/0757bad41ed577f15c7f5d9e508f55e78c612ddb ?

source/manual/removing-a-user-from-puppet.html.md
```diff
@@ -0,0 +1,26 @@
...
+
+Removing a user from our infrastructure via Puppet is a 2 change process that
+requires a deploy in the middle. The first change ensures that when puppet
+runs the users home directory is removed, the second change removes the
+user from puppet itself. If the user is just removed from puppet their files
+will remain on our servers forever more.
+
+1. First find the user manifest in: [modules/users/manifests][manifest-path]
+1. Add an entry to the govuk_user class of `ensure => absent,` -
+   [example][absent-example]
```
cbaines: Maybe mention here not to remove the hieradata, as I think this is necessary for `ensure => absent` to take effect.

cbaines:
Looks good Kevin, I normally comment on user removal PRs to suggest doing this, but having some documentation to point to will be helpful :+1:

source/manual/removing-a-user-from-puppet.html.md
kevindew: Ah thanks - yes in my head they're distinct steps because of different leaver checklist but it makes sense that this list does not rely on that.

source/manual/removing-a-user-from-puppet.html.md
kevindew: Oh really - I didn't realise it worked like that. I'll update.

kevindew:
@cbaines I've updated this now if you don't mind giving it another look over.

cbaines:
> @cbaines I've updated this now if you don't mind giving it another look over.

Sure, it looks good to me :+1:

PR #940, opened 2018-05-21
Update last reviewed date of useful Kibana queries

barrucadu:
Tried them all.  The publishing API timeouts query didn't give any results, but maybe there just weren't any timeouts in the time period I checked.


Davidslv:
I just used `@fields.error:"TimedOutException"` and there were no results, I wonder if the filter is correct or if there are no errors at the moment

barrucadu:
I found one hit in the last 30 days for `error:TimedOutException`, so it might be that.

PR #955, opened 2018-05-30
Add smokey-loop process dying in high priority docs

thomasleese:
[Trello Card](https://trello.com/c/ufIErt2Y/103-👻-smokey-loop-process-sometimes-dies-in-production-because-phantom-js-segfaults)

### DIFF
--- a/source/manual/alerts/high-priority-tests.html.md
+++ b/source/manual/alerts/high-priority-tests.html.md
@@ -4,11 +4,13 @@ title: Run high priority tests
 parent: "/manual.html"
 layout: manual_layout
 section: Icinga alerts
-last_reviewed_on: 2018-03-21
+last_reviewed_on: 2018-05-30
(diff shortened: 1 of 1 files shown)


source/manual/alerts/high-priority-tests.html.md
```diff
@@ -20,5 +22,20 @@ $ fab $environment class:monitoring app.reload:nginx
 $ fab $environment class:monitoring app.restart:smokey-loop
 ```
 
+### `Traceback (most recent call last):`
+
+If you see this error in Icinga, it may imply that the `smokey-loop` process has died. You can try looking through the
```
barrucadu: I think "may mean" is more straightforward than "may imply"

thomasleese: Amended :)

PR #981, opened 2018-06-07
Update vagrant-dns.html.md

Davidslv:


source/manual/vagrant-dns.html.md
```diff
@@ -60,3 +60,13 @@ If you're still having issues you can try to update the vagrant-dns plugin:
...
 ```
+
+## /etc/hosts
+
+If none of the steps above seem to help, have a look into your `/etc/hosts` file.
+
+For example if you are having trouble accessing pages rendered by the collections app, try adding the following:
+
+```
+10.1.1.254       collections.dev.gov.uk
```
cbaines: This is effectively not using Vagrant DNS as far as I'm aware, but fair enough. I'm also not sure if this IP address is static...

PR #983, opened 2018-06-12
Make sure CKAN env activation comman works

benthorner:


source/manual/data-gov-uk-supporting-ckan.html.md
```diff
@@ -41,7 +41,7 @@ with the server's `virtualenv` an environment for Python that includes all of th
 It should be loaded for you on connection, but if not you can activate it from the command line with:
 
 ```
-. /usr/lib/venv/bin/activate
+. /home/co/ckan/bin/activate
```
andrewgarner: These docs are written for https://github.com/alphagov/datagovuk_ckan_deployment where `/usr/lib/venv/bin/activate` is the correct path.

benthorner:
Ah, I see now.

PR #1023, opened 2018-06-28
Add basic page about Licensify app

timblair:
Because Licensify is a private repo, we need to access the repo list in a different way, and that requires an auth key with permission to access those private repos in the alphagov org (which the current deployment process doesn't have.)

**Update:** Approach changed to just use data from the applications.yml file instead of needing access to private repos.

### DIFF
--- a/app/app_docs.rb
+++ b/app/app_docs.rb
@@ -20,7 +20,7 @@ class AppDocs
   end
 
   def self.topics_on_github
-    pages.reject(&:retired?).flat_map(&:topics).sort.uniq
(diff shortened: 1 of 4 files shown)


timblair:
@tijmenb: I'd appreciate your thoughts on this. It means a hard dependency on a GitHub API key, not just for getting around rate limits.

timblair:
It's better than nothing, although it would be good to have an app page for Licensify, which the hard-coding approach doesn't allow for.

I'll have a think about how we might be able to inject a stub (hard-coded) repo into the list, so it'll generate the right thing but without the need for access to private repos.

timblair:
Approach updated to just use data from the `applications.yml` file instead of needing access to private repos.

PR #1036, opened 2018-07-03
Update GOV.UK architecture diagram

tijmenb:
Changes:

- Content performance manager doesn't talk to content-store anymore 🎉 
- The 3 apps that use the Whitehall organisations API now use collections (because the `/api/organisations` route is now served by that). I think this means that Whitehall doesn't have an API anymore? @sihugh or @rubenarakelyan to confirm.
- Added govuk-content-quality-service
- Update: email-alert-api is no longer talking to GovDelivery
- Added link checker, Licensify, DGU to the "Not pictured" section


sihugh:
We've only migrated the orgs api.  There are still `government`, `worldwide_location` and `worldwide_organisation` apis in Whitehall.  https://github.com/alphagov/whitehall/blob/b361148f2103527b4be04f34d3d6a3ead1ae8b3e/config/routes.rb#L33-L40

I'll be removing the org api bits soon

tijmenb:
@sihugh thanks! Do you know if any of those APIs are used internally? Or externally for that matter?

bevanloon:
Do the registers team still use the whitehall api? I think they might have used the `/organisations` at some point.

sihugh:
Registers (and other people) definitely use the organisations api endpoint.  I'm just doing some digging into the others

bevanloon:
`world-locations` too by the looks: https://github.com/openregister/country-data/blob/master/lists/world/bin/world.py#L6

cbaines:
I'm not sure about the `rummager` - `smartanswers` link, I can't see either app directly talking to the other.

tijmenb:
@cbaines good spot!

PR #1052, opened 2018-07-12
Update team channel names

rubenarakelyan:
* Add `govuk-` prefix to channel names
* Update channel names for taxonomy and data informed content
* Move organisations-publisher to platform health team

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -2,16 +2,16 @@
 
 - github_repo_name: collections-publisher
   type: Publishing apps
-  team: "#taxonomy-and-nav"
+  team: "#govuk-taxonomy"
 
 - github_repo_name: contacts-admin
   type: Publishing apps
   puppet_name: contacts
-  team: "#platform-health"
+  team: "#govuk-platform-health"
 
 - github_repo_name: content-tagger
   type: Publishing apps
-  team: "#taxonomy-and-nav"
+  team: "#govuk-taxonomy"
 
 - github_repo_name: content-publisher
   type: Publishing apps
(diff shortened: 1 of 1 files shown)


data/applications.yml
```diff
@@ -2,230 +2,230 @@
...
 
 - github_repo_name: calendars
   type: Frontend apps
-  team: "#platform-health"
+  team: "#govuk-platform-health"
 
 - github_repo_name: collections
   type: Frontend apps
-  team: "#taxonomy-and-nav"
+  team: "#govuk-taxonomy"
```
leenagupte: should this be #govuk-content-pages ?

data/applications.yml
```diff
@@ -2,230 +2,230 @@
 
 - github_repo_name: collections-publisher
   type: Publishing apps
-  team: "#taxonomy-and-nav"
+  team: "#govuk-taxonomy"
```
leenagupte: should this be #govuk-content-pages ?

data/applications.yml
cbaines: Maybe, I think different people have different opinions of what applications related (or are "owned") by which people (usually product managers) or teams. 

I think this was talked about over the Firebreak week, or at least I remember someone talking about it, but in general, the whole concept seems to have lost relevance since the appearance of the Platform Health team.

data/applications.yml
rubenarakelyan: Potentially - feel free to change it since I don't have context of who owns what apart from platform health owns nearly everything

PR #1087, opened 2018-07-31
"Post a message on the status page"

tijmenb:
WIP - needs THE content designer.

https://govuk-developer-docs-pr-1087.herokuapp.com/manual/post-a-statuspage-message.html


hongoose:
Here's some feedback from Tatiana which gives a nice summary of the other received feedback that we should include in our guidance:

1) I didn't spot the templates until I'd already posted a few updates, and it seems when you go back to edit them you can't use the templates anymore. Probably just something that might happen the first time someone uses status page.

2) I hesitated a bit between assigning it a "degraded performance" or a "partial outage" level. Could be worth considering adding guidance about that in the incident severity level description, or having them match? It seems Priority 1, 2 and 3 broadly match to major outage, partial outage and degraded performance

PR #1088, opened 2018-07-31
Rather Large Documentation Review™

boffbowsh:
I'm leaving scarily soon. This is an update of the dev docs based on what I know about GOV.UK. There's a lot of deletions because things aren't relevant any more.


source/manual/attachments.html.md
```diff
@@ -1,11 +1,13 @@
...
-owner_slack: "#govuk-2ndline"
+owner_slack: "#govuk-platform-health"
 title: Attachment backups
 section: Backups
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2018-05-30
+last_reviewed_on: 2018-07-31
 review_in: 2 months
+next_review_notes: |
```
tijmenb: Excellent addition!

source/manual/github-enterprise.html.md
```diff
@@ -1,11 +0,0 @@
----
```
tijmenb: The goal of this page was to record somewhere that we used `github.gds` at some point, in case there are old emails/trello tickets/github PR mentioning it.

boffbowsh: Restored.

cbaines:
> Remove section on router-data
> 
> This is now properly retired - it's best not to advertise it as an option

What do you mean by properly retired here?

PR #1092, opened 2018-08-02
Add specific docs for mapit high memory

elliotcm:
This is typically being caused by zombie processes from a failed soft restart of the application.

### DIFF
--- a/source/manual/alerts/high-memory-for-application.html.md
+++ b/source/manual/alerts/high-memory-for-application.html.md
@@ -4,7 +4,7 @@ title: High memory for application
 parent: "/manual.html"
 layout: manual_layout
 section: Icinga alerts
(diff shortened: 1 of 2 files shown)


source/manual/alerts/high-memory-for-application.html.md
```diff
@@ -4,7 +4,7 @@ title: High memory for application
 parent: "/manual.html"
 layout: manual_layout
 section: Icinga alerts
-last_reviewed_on: 2017-11-22
+last_reviewed_on: 2018-11-02
```
emmabeynon: Is this supposed to be today's date?

elliotcm: Apparently I can't tell the difference between august and november.  Updated.

PR #1103, opened 2018-08-06
Cover need to redirect the old world news slug

barrucadu:
Just reslugging the world news page makes the new slug available on the site, but doesn't redirect the old slug.

---

[Trello card](https://trello.com/c/iwDCm6tR/353-change-slugs-from-gambia-to-the-gambia?menu=filter&filter=member:michaelswalker)


barrucadu:
Sure, but that can be a separate task.

PR #1118, opened 2018-08-10
Document which apps use advisory locks

barrucadu:
[Trello card](https://trello.com/c/zcbWIyoF/390-add-advisory-locking-to-the-dev-docks)

### DIFF
--- a/source/manual/postgresql-advisory-locks.html.md
+++ b/source/manual/postgresql-advisory-locks.html.md
@@ -1,0 +1,24 @@
+---
+owner_slack: "#govuk-2ndline"
+title: PostgreSQL Advisory Locks
+section: Databases
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2018-08-10
+review_in: 3 months
+---
+
+[Advisory locks][pgdocs] are a feature of PostgreSQL to achieve mutual
+exclusion between different processes.  This is a session-level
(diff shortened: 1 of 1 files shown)


source/manual/postgresql-advisory-locks.html.md
```diff
@@ -0,0 +1,23 @@
...
+
+[Advisory locks][pgdocs] are a feature of PostgreSQL to achieve mutual
+exclusion between different processes.  This is a session-level
+feature: a lock acquired in a transaction will be held until it is
+explicitly released or the session ends, even if the transaction is
+rolled back.
+
+[pgdocs]: https://www.postgresql.org/docs/9.3/static/explicit-locking.html#ADVISORY-LOCKS
+
+## Apps which use advisory locks
```
thomasleese: I wonder if we should include a GitHub search for the advisory locks Gem just in case others get added without being added to this list.

source/manual/postgresql-advisory-locks.html.md
```diff
@@ -0,0 +1,23 @@
...
+
+- The content-store uses an advisory lock to prevent concurrent writes to the content-store in [ContentStoreWriter](https://github.com/alphagov/publishing-api/blob/master/app/clients/content_store_writer.rb).  Not using the lock here can result in overlapping modifications to the same content item.
```
thomasleese: `publishing-api` 😄

PR #1148, opened 2018-08-30
Querying CDN logs

kevindew:
This adds documentation on the process to query the CDN logs via Athena. It intends to explain what is stored in AWS log wise, how you can access it and some example queries so you can use it.

It also updates the existing CDN documentation in light of this change.

### DIFF
--- a/source/manual/cdn.html.md
+++ b/source/manual/cdn.html.md
@@ -4,7 +4,7 @@ title: Our content delivery network (CDN)
 section: CDN & Caching
 layout: manual_layout
 parent: "/manual.html"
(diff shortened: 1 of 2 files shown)


source/manual/querying-cdn-logs.html.md
```diff
@@ -0,0 +1,183 @@
+---
+owner_slack: "#govuk-2ndline"
+title: Querying CDN Logs
```
tijmenb: Perhaps "Query CDN logs" to make it conform to [the styleguide](https://docs.publishing.service.gov.uk/manual/docs-style-guide.html#title).

PR #1212, opened 2018-09-07
Document schema's used on GOV.UK

MahmudH:
Documentation all the schemas from schema.org that have been added to GOV.UK


source/manual/schemas.html.md
```diff
@@ -0,0 +1,41 @@
+# Schemas used on GOV.UK
+Here are a list of schema's being used
```
rubenarakelyan: Nit: "schemas" (since it's a plural)

source/manual/schemas.html.md
```diff
@@ -0,0 +1,41 @@
...
+
+### [HowTo][how-to]
+
+[how-to]: https://schema.org/HowTo
+
+Used in step-by-step, for example, [learn-to-drive-a-car][learn-to-drive-a-car]
```
rubenarakelyan: The text in the first set of square brackets is what will be displayed on the page, so it would be good to make it a human-readable title, like "Learn to Drive a Car" or something similar.

source/manual/schemas.html.md
```diff
@@ -0,0 +1,41 @@
+# Schemas used on GOV.UK
+Here are a list of schema's being used
+
+### [HowTo][how-to]
```
rubenarakelyan: You need to use a second-level heading here (## rather than ###) otherwise you'll get a linter error.

rubenarakelyan: This also applies to the other headings on this page

source/manual/schemas.html.md
rubenarakelyan: This also applies to the other links on the page.

source/manual/schemas.html.md
```diff
@@ -0,0 +1,41 @@
...
+
+[how-to]: https://schema.org/HowTo
```
rubenarakelyan: It would be nice to bundle all the link definitions together, either at the end of each section (preferred) or at the end of the document. This way, they don't break up the flow of a section.

source/manual/schemas.html.md
```diff
@@ -0,0 +1,42 @@
...
+owner_slack: "#govuk-2ndline"
+title: Schemas
+parent: "/manual.html"
+layout: manual_layout
+section: Tools
+last_reviewed_on: 2018-09-10
+review_in: 12 months
+---
+
+# Schemas used on GOV.UK
```
thomasleese: You actually don't need this heading, it'll come from the `title` setting in the block above.

MahmudH: 👍 changes made

PR #1213, opened 2018-09-10
Add extra actions to email alerts not sent

bevanloon:
Add some advice and links to email alert api to help diagnose and fix
travel and drug advice emails not being sent.

Remove a redundant sentence as well.


source/manual/alerts/email-alerts.html.md
```diff
@@ -8,8 +8,7 @@ last_reviewed_on: 2018-08-31
 review_in: 6 months
```
thomasleese: Worth updating the last reviewed on date as well I think.

bevanloon: fixed in 31e375d

PR #1250, opened 2018-10-01
Fix link to sharing pacts

alexmuller:


alexmuller:
@cbaines No problem! Thanks for the useful docs 👍

PR #1266, opened 2018-10-08
Updated guidance on $USERNAME

DilwoarH:


source/manual/get-started.html.md
```diff
@@ -212,6 +212,8 @@ After gaining access to the integration environment, download the data by runnin
 
     mac$ cd ~/govuk/govuk-puppet/development-vm/replication
     mac$ ./replicate-data-local.sh -u $USERNAME -F ../ssh_config -n
+    
+Note: the `$USERNAME` is usually `firstnamelastname`
```
tijmenb: Maybe this can link to [step 3](https://docs.publishing.service.gov.uk/manual/get-started.html#3-create-a-user-in-integration-and-ci) instead - it's the same username as was created there.

barrucadu: An alternative is to just change it to `$USER`, which is a standard shell variable, and the username on your machine is probably going to be your username on integration:

```
bash-3.2$ echo $USERNAME

bash-3.2$ echo $USER
michaelswalker
```

chao-xian: Updated to take on @barrucadu's suggestion. Please can someone 👀 ?

PR #1273, opened 2018-10-10
Tidy up page sections

tijmenb:
This tidies up the page sections for the pages in the manual.

Highlights:

- Split up "AWS" into separate sections
- Add an "Accounts" section for access-related things
- Move a lot out of "Tools" into more appropriate sections

## Preview

https://govuk-developer-docs-pr-1273.herokuapp.com/manual.html

### DIFF
--- a/source/manual/ab-testing.html.md
+++ b/source/manual/ab-testing.html.md
@@ -2,7 +2,7 @@
 title: Run an A/B or multivariate test
 parent: "/manual.html"
 layout: manual_layout
-section: Tools
+section: A/B testing
 owner_slack: "#govuk-2ndline"
 last_reviewed_on: 2018-07-03
 review_in: 6 months
--- a/source/manual/concept-of-stacks.html.md
+++ b/source/manual/concept-of-stacks.html.md
@@ -1,7 +1,7 @@
 ---
 owner_slack: "#govuk-2ndline"
 title: Stacks in AWS
-section: AWS
+section: Architecture
 layout: manual_layout
 parent: "/manual.html"
 last_reviewed_on: 2018-09-03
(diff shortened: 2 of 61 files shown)


source/manual/govuk-in-aws.html.md
```diff
@@ -1,7 +1,7 @@
 ---
 owner_slack: "#govuk-2ndline"
 title: GOV.UK in AWS
-section: AWS
+section: AWS architecture
```
barrucadu: Given that there's an Infrastructure section, would it make sense to call this "AWS Infrastructure"?

source/manual/concept-of-stacks.html.md
```diff
@@ -1,7 +1,7 @@
 ---
 owner_slack: "#govuk-2ndline"
 title: Stacks in AWS
-section: AWS
+section: AWS architecture
```
barrucadu: Same comment as https://github.com/alphagov/govuk-developer-docs/pull/1273/files#r224434650

source/manual/govuk-in-aws.html.md
tijmenb: 🤔 maybe it should move to a "Architecture" section?

source/manual/ab-testing.html.md
```diff
@@ -2,7 +2,7 @@
 title: Run an A/B or multivariate test
 parent: "/manual.html"
 layout: manual_layout
-section: Tools
+section: Development & Reviews
```
barrucadu: I'm not sure this really fits in "Development & Reviews", but it doesn't fit in "Tools" either so I'm not sure what to suggest.

tijmenb: I've given it it's very own category, we should split up the page at some point anyway.

source/manual/govuk-in-aws.html.md
tijmenb: I've done another pass. I've created a new "Architecture" section with these pages in it, and I've moved some stuff from Basics in there too. To keep it focused I've renamed "Infrastructure" to "Infrastructure tasks".

PR #1279, opened 2018-10-11
Retire policy publisher

tijmenb:


data/applications.yml
```diff
@@ -362,3 +358,10 @@
     finance support schemes. In March 2017, it was replaced with a new-style finder
     rendered by [finder-frontend](/apps/finder-frontend.html) and published by
     [specialist-publisher](/apps/specialist-publisher.html).
+
+- github_repo_name: policy-publisher
+  retired: true
+  type: Publishing apps
+  description: |
+    Policy publisher published "policies", which were a kind of taxonomy for
+    government policies. They were merged into the single GOV.UK taxonomy.
```
1pretz1: Maybe use 'topic taxonomy' instead of 'single GOV.UK taxonomy', as at the minute the topic taxonomy is not really the only one and it's a bit more descriptive?

PR #1302, opened 2018-10-18
Refresh the deploy instructions

tijmenb:
A refresh of the page, rewriting it to be more step-by-step and remove information not necessarily relevant to deploying so that we can follow it more easily.

There are no real changes to policy, but I've changed a few things to reflect current working practices.

- Add step to check deployment dashboards
- State that most deploys take place during core hours (between 10 and 4)

## Preview

https://govuk-developer-docs-pr-1302.herokuapp.com/manual/deploying.html

### DIFF
--- a/source/manual/deploying.html.md
+++ b/source/manual/deploying.html.md
@@ -5,107 +5,64 @@ parent: "/manual.html"
 layout: manual_layout
 section: Deployment
 important: true
-last_reviewed_on: 2018-07-31
+last_reviewed_on: 2018-10-18
 review_in: 6 months
 ---
 
-## Introduction
-
-2nd line is responsible for:
-
-- ensuring that software is released to GOV.UK responsibly
-- providing access to deploy software for teams who can't deploy it themselves
-
-As far as possible, teams are responsible for deploying their own work. We
-believe that [regular releases minimise the risk of major problems][regular_releases_reduce_risk]
-and improve recovery time.
-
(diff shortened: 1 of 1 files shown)


source/manual/deploying.html.md
```diff
@@ -5,107 +5,64 @@ parent: "/manual.html"
...
 
-### Security patches
+## 4. Deploy to staging
 
-If you are responding to a security incident, follow the steps to
-[deploy fixes for a security vulnerability](deploy-fixes-for-a-security-vulnerability.html).
+From the [Release application](https://release.publishing.service.gov.uk), click the deploy buttons. If you have production access, this will bring you to the Jenkins job to deploy your change.
 
-### Rollback
+Once deployed to staging, let the application run there for 30 minutes. You will need to:
```
sihugh: This is new (apart from Static).  I suggest we communicate this a little more widely than just in the docs.  Maybe worth a recap at tech fortnightly?

I like the sentiment, particularly when catching up with the dependabot deploys.  I might get a bit impatient if it's just my code though.

tijmenb: How would you phrase the current practice? Only wait until smokey is finished?

sihugh: I understood it as being the same sorts of instructions, only without the time constraint.

sihugh: So you still need to make sure that your changes actually work etc...

tijmenb: I've removed it for now, also because 30 minutes doesn't do much for admin applications - it's only relevant for frontend apps that benefit from traffic replay.

PR #1334, opened 2018-11-02
Review & update post-a-statuspage-message.html.md

AlanGabbianelli:


source/manual/post-a-statuspage-message.html.md
```diff
@@ -4,27 +4,32 @@ title: Post a message on the status page
...
 
-Remember: government publishers are the main audience of the page. They'll want to know what isn't working, and how long it will take to fix. Write in non-technical language. Use the pre-written templates, they've been proofread and can be customised to most situations.
+Remember: government publishers are the main audience of the page.
+They'll want to know what isn't working, and how long it will take to fix. Write in non-technical language.
+Use the pre-written templates, they've been proofread and can be customised to most situations.
 
 This is what the [Statuspage admin interface][man] looks like:
 
-![](manual/images/create-new-incident.png)
+![](images/create-new-incident.png)
```
barrucadu: This doesn't seem right, compare:

* https://docs.publishing.service.gov.uk/manual/images/create-new-incident.png
* https://docs.publishing.service.gov.uk/images/create-new-incident.png

source/manual/post-a-statuspage-message.html.md
```diff
@@ -4,27 +4,32 @@ title: Post a message on the status page
...
 
-## Title of the incident
+## Incident Name
```
chao-xian: I don' think we use Title Case in headings unless the words within themselves are titles, so `Incident name` would be more appropriate.

source/manual/post-a-statuspage-message.html.md
```diff
@@ -66,7 +72,10 @@ Example messages:
 
 ### "Monitoring" status
 
-We've fixed the issue - but are keeping are an eye on it to make sure it's properly resolved. State whether there will be a delay to the fix being visible, ie. "This should start functioning normally again in 20 minutes."
+Example messages:
+
+We've fixed the issue - but are keeping an eye on it to make sure it's properly resolved.
```
chao-xian: This needs to be in a `>` block.

source/manual/post-a-statuspage-message.html.md
AlanGabbianelli: If you click in the image in the currently live version here http://localhost:4567/manual/post-a-statuspage-message.html you can see that it goes to https://docs.staging.publishing.service.gov.uk/manual/manual/images/create-new-incident.png (notice the double `manual/`) resulting in a `404`.
Removing `manual/` from `` fixes the issue.

source/manual/post-a-statuspage-message.html.md
AlanGabbianelli: I see how this looks like Title Case. That wasn't my intention, though. I only followed the names of the sections in https://manage.statuspage.io/pages/xcz7hfscrbjs/incidents/new. They are:
- `Incident Name`
- `Incident Status`
- `Message`

source/manual/post-a-statuspage-message.html.md
AlanGabbianelli: This is not an example message; it's the explanation. I had wrongly added `Example messages:` above it when it should have been below it. I moved it now.

source/manual/post-a-statuspage-message.html.md
chao-xian: Ah ok. I think America defaults to Title Case on everything, but I'm not going to hold this up on that front. Thanks for clarifying!

source/manual/post-a-statuspage-message.html.md
chao-xian: Oh cool!

PR #1346, opened 2018-11-09
Add docs for DGU map previews

steventux:
https://trello.com/c/Go7bVEFZ/615-document-dgu-map-preview-javascript

Map previews on DGU are fairly large client-side javascript applications which rely on external Web Mapping Services and assets, here's some docs to explain.


steventux:
@bevanloon I'm not sure whether this is a preview app specific thing but I've updated them to full URLs.

PR #1356, opened 2018-11-15
Update instructions for migrating between CKAN instances

brucebolt:
We were previously using the API to migrate data from one CKAN to another.  The new method takes a database dump and updates this to reflect the changes to the schema and models.

Trello card: https://trello.com/c/e3qWIvmc/48-implement-new-method-for-importing-data-from-old-ckan-to-new-ckan


brucebolt:
The PostGIS extension is being installed by Puppet when the app is deployed (https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/ckan/db.pp#L38).

PR #1362, opened 2018-11-22
Update docs about who to ask for support

bevanloon:
We tend not to have Tech Leads for objectives at the moment so
changed this to be a Lead Dev on the programme.


source/manual/who-do-i-ask-for-support.html.md
```diff
@@ -33,7 +33,7 @@ The Reliability Engineering team:
 If you and your colleagues can’t resolve a technical issue, problem or question, you should escalate it through, in order:
 
 1. The Technical Lead on the team
-2. The Technical Lead for the team’s objective
+2. The Lead Developer on the programme
```
AlanGabbianelli: Is this always only one?

bevanloon: not always, but it is at the moment.

PR #1371, opened 2018-11-26
Review and update add-a-new-document-type.html.md

AlanGabbianelli:


source/manual/add-a-new-document-type.html.md
```diff
@@ -4,15 +4,17 @@ parent: "/manual.html"
 layout: manual_layout
 section: Publishing
 owner_slack: "#govuk-2ndline"
-last_reviewed_on: 2018-08-17
+last_reviewed_on: 2018-11-26
 review_in: 3 months
 ---
 
-The [document type](https://docs.publishing.service.gov.uk/document-types.html) describes what a page on GOV.UK looks like.
+The [document type](document-type) describes what a page on GOV.UK looks like.
```
thomasleese: To use named links, the name needs to be in square brackets also, like `[document type][document-type]`. You can also do `[govuk-content-schemas]` if the name is the same as the text of the link.

PR #1373, opened 2018-11-27
Remove Policy Publisher from applications.yml

cbaines:
This application has been retired now, so remove it from here, mostly
to stop the deploy lag badger from moaning on Slack.


cbaines:
I've now created https://github.com/alphagov/govuk-developer-docs/pull/1375

PR #1392, opened 2018-12-05
Update dev VM disk space docs

rubenarakelyan:


rubenarakelyan:
Luckily the app still works in the latest OS versions, and it seems to be the only such free app available so I think we need to carry on using it.

PR #1393, opened 2018-12-05
[Do not merge] Add docs explaining how to update the global banner

steventux:
See https://github.com/alphagov/static/pull/1583 this PR adds docs to detail the process of updating and activating a site-side banner.


nickcolley:
This is useful information but are we planning on keeping the global bar around in it's current form?

I was under the impression we'd be removing it after it's served it's purpose.

steventux:
@nickcolley good point, this is also a dupe of static docs here https://github.com/alphagov/static/blob/master/doc/global-banner.md so I'll close this.

PR #1399, opened 2018-12-12
Add instructions for updating Zendesk password for DGU Find

brucebolt:


brucebolt:
I've updated to explain that the username and password come from the first step.

PR #1423, opened 2019-01-03
Document where applications are hosted

tijmenb:
At the moment it's not easy to find in the document whether an app is hosted on Carrenza or AWS, because we're currently migrating apps over.

This PR adds where the application is hosted to the application page, application index pages and CSV export. It should make it easier to find the location of the apps, as well as automating some parts of the docs. Source of the apps is [this Trello ticket on 2nd line](https://trello.com/c/49OGPxEk/314-%F0%9F%94%AD-where-are-our-apps-apps-are-moving-from-carrenza-to-aws).

- Add hosting to app page
- Add list of providers and number of apps to application home
- Add hosting to CSV and JSON export for reuse
- Fix SSH instructions

## Screenies

screen shot 2019-01-03 at 12 51 19
screen shot 2019-01-03 at 12 50 59
screen shot 2019-01-03 at 12 50 55

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -3,57 +3,70 @@
 - github_repo_name: collections-publisher
   type: Publishing apps
   team: "#govuk-platform-health"
+  production_hosted_on: carrenza
 
 - github_repo_name: contacts-admin
   type: Publishing apps
   puppet_name: contacts
   team: "#govuk-platform-health"
+  production_hosted_on: carrenza
 
 - github_repo_name: content-tagger
   type: Publishing apps
   team: "#govuk-platform-health"
+  production_hosted_on: carrenza
 
 - github_repo_name: content-publisher
   type: Publishing apps
   team: "#govuk-pub-workflow"
+  production_hosted_on: carrenza
 
 - github_repo_name: local-links-manager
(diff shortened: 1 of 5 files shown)


data/applications.yml
```diff
@@ -3,250 +3,305 @@
...
 
 - github_repo_name: datagovuk_publish
   type: data.gov.uk apps
   team: "#govuk-platform-health"
+  production_hosted_on: paas
 
 - github_repo_name: ckanext-datagovuk
   type: data.gov.uk apps
   team: "#govuk-platform-health"
+  production_hosted_on: paas
```
thomasleese: I think this is actually in AWS (@brucebolt would know for sure).

thomasleese:
This is a good idea! It would be nice if we could keep it up to date with the data in https://release.publishing.service.gov.uk/applications (probably if Release app could be converted to use this as an API) but perhaps that's not worth it for the moment.

data/applications.yml
brucebolt: Yes, it's on AWS

tijmenb:
```suggestion
  production_hosted_on: aws
```

PR #1514, opened 2019-02-08
Move integration access out of "Get started" guide

tijmenb:
Plot twist! Unless I'm mistaken, since we moved the data sync to S3 (https://github.com/alphagov/govuk-puppet/pull/7209), new starters don't actually need integration SSH access to be able to replicate data to the VM. They probably won't need integration access at all for the first few months.

This commit moves the docs about getting integration access into a separate page, and updates some old references to SSH'ing into integration.

Should make life a little easier for new joiners.

### DIFF
--- a/source/manual/get-ssh-access.html.md
+++ b/source/manual/get-ssh-access.html.md
@@ -1,0 +1,62 @@
+---
+owner_slack: "#govuk-2ndline"
+title: Get SSH access to integration
+layout: manual_layout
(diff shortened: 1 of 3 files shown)


brucebolt:
Is it possible to access the S3 buckets in integration without integration access?

rubenarakelyan:
You don't need SSH access but you do need AWS access

PR #1540, opened 2019-02-15
Add details about when to dismiss stale PR reviews

rubenarakelyan:
This commit adds a new rule about dismissing stale PR reviews when making changes before merging, unless they are minor changes and the author is confident there are no side-effects.

It also adds two example of the new rule at work.


rubenarakelyan:
Just to clarify, this has always been the process but it's just not been clearly documented until now.

PR #1543, opened 2019-02-18
Describe the GOV.UK development/deployment pipeline

tijmenb:
We've been talking about replatforming GOV.UK and what that would look like for the development experience. This is an attempt to document what  our current pipeline looks like, so we're all on the same page.

### DIFF
--- a/source/manual/development-pipeline.html.md
+++ b/source/manual/development-pipeline.html.md
@@ -1,0 +1,46 @@
+---
+title: The development and deployment pipeline
+parent: "/manual.html"
+layout: manual_layout
+section: Deployment
+type: learn
+owner_slack: "#govuk-2ndline"
+last_reviewed_on: 2019-02-18
+review_in: 6 months
+---
+
+Our development and deployment pipeline looks like this:
+
+<img src="https://docs.google.com/drawings/d/e/2PACX-1vScz6LD_vuXiLH8Nug-29qil0U1m-0p9axgTDAR9_pXRsyTes1sDix45vVQtgjJ4q_msUKDU6fxXuha/pub?w=1289&amp;h=178">
+
+## 1. Local development
+
+Developers develop the code on their laptop. Most of the time, they [use the Development VM](/manual/get-started.html) for this.
+
+Once the tests pass locally you push the code to GitHub and open a Pull Request. The Pull Request [should conform to the guidelines in the styleguides](https://github.com/alphagov/styleguides/blob/master/pull-requests.md).
+
+## 2. GitHub Pull Request
+
+Once the Pull Request is opened, a job starts running on Jenkins, our Continuous Integration tool (CI). For a detailed look, you should [read about the CI infrastructure](/manual/jenkins-ci.html).
+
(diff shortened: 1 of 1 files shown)


source/manual/development-pipeline.html.md
```diff
@@ -0,0 +1,60 @@
...
+
+We've got [guidelines on merging of Pull Requests](/manual/merge-pr.html).
+
+Only when the tests pass and the code has been approved the Pull Request can be merged, since we've [configured GitHub to prevent merges](/manual/configure-github-repo.html) otherwise.
+
+## 3. Master
+
+Code that is merged to `master` is tested again on CI. If the tests on `master` pass, Jenkins pushes a git tag to GitHub - what we call the "release tag", which looks something like `release_312`.
+
+After that, the CI Jenkins sends a message to the Deploy Jenkins to deploy the release tag.
```
timblair: Worth mentioning that "deploy the release tag" is just to integration.

source/manual/development-pipeline.html.md
```diff
@@ -0,0 +1,60 @@
...
+
+Once on integration, it can be [tested out by team members](/manual/deploying.html#1-test-on-integration). Since there's often a lot going on in Integration, it's often broken.
+
+If everything checks out, you can [deploy to staging](/manual/deploying.html#4-deploy-to-staging).
+
+## 5. Staging
+
+Once on staging the [application is monitored](/manual/deploying.html#4-deploy-to-staging).
+
+An important feature of staging is the traffic replay - we use a tool called [GoReplay](https://goreplay.org/) to replay all production GET requests to staging. This means that any bugs in the frontend code might surface.
```
timblair: We reply 10% traffic to integration too, I believe. Don't know if that's worth mentioning or not.

timblair:
A couple of comments, only one which probably requires any change (and only minor at that). Looks like a great overview :+1:

source/manual/development-pipeline.html.md
```diff
@@ -0,0 +1,60 @@
...
+
+If all's well, we can [deploy to production](/manual/deploying.html#5-deploy-to-production)
```
cbaines: It's maybe a little aspirational, but looking to see if the Smokey job run on the Staging Deploy Jenkins may be useful.

Also, checking Icinga in Staging to see if there are any alerts that potentially relate to the app. This can be useful for Ruby upgrades for example.

cbaines: Ah, I see that this is mentioned on the `/manual/deploying.html` page, which makes me unsure about the overlap between that existing page, and this new page?

AlanGabbianelli:
I like how simple and clear this is.
I see that points 4, 5 and 6 reference https://docs.publishing.service.gov.uk/manual/deploying.html.
Would it be worth adding points 1, 2 and 3 there as well, basically merging these two docs?
It looks like the info in these two docs could live well together in one page.

tijmenb:
I've replaced the text in the deploy steps with a link to the deployment page, because it otherwise risks duplicating that. Thanks @cbaines and @AlanGabbianelli !

PR #1545, opened 2019-02-19
There is only one Elasticsearch cluster for GOV.UK

brucebolt:
Remove text implying there are multiple Elasticsearch clusters, when we only have the one.

### DIFF
--- a/source/manual/alerts/elasticsearch-cluster-health.html.md
+++ b/source/manual/alerts/elasticsearch-cluster-health.html.md
@@ -36,8 +36,8 @@ to help you diagnose any problems.
 
 #### Find hosts in an Elasticsearch cluster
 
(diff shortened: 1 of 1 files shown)


source/manual/alerts/elasticsearch-cluster-health.html.md
```diff
@@ -36,8 +36,7 @@ to help you diagnose any problems.
 
 #### Find hosts in an Elasticsearch cluster
 
-We use different Elasticsearch clusters for different applications. For example
-the `rummager-elasticsearch` cluster powers the GOV.UK search API.
+The `rummager-elasticsearch` cluster powers the GOV.UK search API.
```
barrucadu: It also powers the licence-finder sector name lookup

PR #1550, opened 2019-02-22
Bump "Name a new application or gem"

tijmenb:
It's probably good for the next 12 months. Added a sentence to make clear that these rules should be broken if there's a good reason to.

### DIFF
--- a/source/manual/naming.html.md
+++ b/source/manual/naming.html.md
@@ -63,7 +63,7 @@ Good:
 
 Not so good:
 
-- rummager (became be search-api)
+- rummager (became search-api)
 
 ### Admin applications
 
--- a/source/manual/naming.html.md
+++ b/source/manual/naming.html.md
@@ -4,11 +4,13 @@ title: Name a new application or gem
 section: Applications
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2018-08-20
-review_in: 6 months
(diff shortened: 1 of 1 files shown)


source/manual/naming.html.md
```diff
@@ -61,7 +63,7 @@ Good:
 
 Not so good:
 
-- rummager (should be search-api)
+- rummager (became be search-api)
```
tijmenb:
```suggestion
- rummager (became search-api)
```

source/manual/naming.html.md
```diff
@@ -4,12 +4,14 @@ title: Name a new application or gem
...
 parent: "/manual.html"
-last_reviewed_on: 2018-08-20
-review_in: 6 months
+last_reviewed_on: 2019-02-22
+review_in: 12 months
 ---
 
 This describes how you should name an application or gem on GOV.UK. It was first proposed in [RFC 63](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-063-naming-new-apps-gems-on-gov-uk.md).
 
+Of course, break any of these rules sooner than [say anything outright barbarous](https://en.wikipedia.org/wiki/Politics_and_the_English_Language#Remedy_of_Six_Rules).
```
rubenarakelyan: This is *so* intellectual

tijmenb: 👨‍🏫

PR #1572, opened 2019-03-01
Update Concourse docs

rubenarakelyan:
Split the how-to from the learning docs, and add docs on applying config.

### DIFF
--- a/source/manual/create-a-concourse-pipeline.html.md
+++ b/source/manual/create-a-concourse-pipeline.html.md
@@ -1,0 +1,26 @@
+---
+owner_slack: "#govuk-2ndline"
+title: Create a Concourse pipeline
+section: Infrastructure
+layout: manual_layout
+parent: "/manual.html"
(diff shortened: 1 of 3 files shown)


source/manual/create-a-concourse-pipeline.html.md
```diff
@@ -0,0 +1,26 @@
+---
+owner_slack: "#govuk-2ndline"
+title: Create a Concourse pipeline
+section: Infrastructure
```
thomasleese: I wonder if we should create a separate category for Concourse because then you get a nice box at the bottom of each of these pages linking to related Concourse documentation... at the same time I can see the benefits of grouping it with infrastructure as well.

rubenarakelyan: Yup, agreed but I think we should do this once we have more stuff in Concourse. At the moment it feels like infrastructure but once CI etc is running in there I think it will need a separate section.

PR #1579, opened 2019-03-04
Reviewed and looks upto date.

1pretz1:
The date is slightly staggered due to a big influx of docs to be
reviewed around this time.

### DIFF
--- a/source/manual/fetching-packages.html.md
+++ b/source/manual/fetching-packages.html.md
@@ -4,7 +4,7 @@ title: Problems provisioning and fetching packages in VM
 section: Development VM
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2018-09-04
+last_reviewed_on: 2019-03-12
 review_in: 6 months
 ---
 


source/manual/fetching-packages.html.md
```diff
@@ -4,7 +4,7 @@ title: Problems provisioning and fetching packages in VM
 section: Development VM
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2018-09-04
+last_reviewed_on: 2018-03-12
```
bilbof: Should this be todays date?

bilbof: It looks like last_reviewed_on should be today, and review_in should be something like `6 months`. https://alphagov.github.io/tech-docs-manual/#frontmatter

1pretz1: Yep, you're correct! I've tried to stagger the dates by +- a week due to a massive amount of docs (80+) all needing reviewing on one day. This should alleviate the problem for the next people on second line in 6 months!

PR #1600, opened 2019-03-06
Add information about how to restart CKAN harvesters

thomasleese:
It seems to be crashing fairly regularly so I've added a section outlining what to do in that case to get harvesting up and running again.

[Trello Card](https://trello.com/c/07jTJBX5/823-make-ckan-harvesting-restart-when-it-fails)

### DIFF
--- a/source/manual/data-gov-uk-troubleshooting.html.md
+++ b/source/manual/data-gov-uk-troubleshooting.html.md
@@ -4,7 +4,7 @@ title: Troubleshoot data.gov.uk
 section: data.gov.uk
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2019-01-02
+last_reviewed_on: 2019-03-06
 review_in: 6 months
 ---
 [find]: apps/datagovuk_find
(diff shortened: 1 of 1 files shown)


source/manual/data-gov-uk-troubleshooting.html.md
```diff
@@ -68,3 +68,33 @@ If you see a lot of tracebacks, it might be necessary to restart Celery.
...
+## Harvesters not processing or seem stuck
+
+The harvesting process runs as a single threaded program, if any harvesting
+process crashes by raising an exception, it will take out the entire process.
+We have configured Upstart to restart the process automatically, but if the
+service keeps crashing, Upstart will decide it's unhealthy and stop that after
+a while.
+
+You can check whether the process is still running by checking if entries are
+still being written to the log file:
```
bevanloon: Could we pop the machine class we should be sshing into in this sentence just to reduce mental overhead a bit. It's implied in the fab script down the bottom, but nice to have up here as well.

thomasleese: Good idea, I've added a bit about that to this section and the section above.

PR #1605, opened 2019-03-08
Add information on automatic deployments

AlanGabbianelli:


source/manual/data-gov-uk-deployment.html.md
```diff
@@ -32,9 +35,16 @@ Heroku has a pipeline for each of [Publish][publish-heroku] and [Find][find-hero
 
 Each repo has a `Procfile` and an `app.json` file, which help to specify how the app is deployed. The environment variables ('Config Vars') are then set via the website, both for the permanent app instance, and the review app template.
 
-### PaaS Staging and Production Environment
+### PaaS Staging and Production Environments
 
-[Publish] and [Find] are provisioned on [GOV.UK PaaS][paas]. Each app repo contains a set of manifests that specify the container settings for when it's deployed. You can deploy an app manually as follows.
+[Publish] and [Find] are provisioned on [GOV.UK PaaS][paas]. Each app repo
+contains a set of manifests that specify the container settings for when it's
+deployed. Both apps are automatically deployed to staging and production. They
```
thomasleese: > Both apps are automatically deployed to staging and production.

I wonder if there is another way of writing this because, if I read that quickly, it sounds to me like it gets deployed to both environments whenever something gets committed to master. I wonder if perhaps something like this might be better:

```
The app is automatically deployed to staging whenever a PR gets merged into master, you can check the Travis logs of the `master` build to see progress.

To promote this build to production, you can create a release in GitHub with an appropriate version number which will trigger a release.

This behaviour is defined in...
```

PR #1606, opened 2019-03-08
Add instructions for accessing *.dev.gov.uk after using GDS VPN

brucebolt:
The GDS VPN modifies your Mac's routing table to direct traffic to *.dev.gov.uk through the VPN.  These instructions allow you to revert that without restarting the machine.

Networking team will soon add this as a script on VPN disconnect, but this solves the problem in the meantime.

### DIFF
--- a/source/manual/vm-vpn-issues.html.md
+++ b/source/manual/vm-vpn-issues.html.md
@@ -1,0 +1,22 @@
+---
+owner_slack: "#govuk-2ndline"
+title: Acccess the VM after using the VPN
+section: Development VM
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2019-03-08
+review_in: 6 months
+---
+
+## Using the GDS VPN
+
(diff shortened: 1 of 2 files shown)


source/manual/vm-vpn-issues.html.md
```diff
@@ -0,0 +1,22 @@
...
+## Using the GDS VPN
+
+Using Cisco AnyConnect will prevent traffic to *.dev.gov.uk from the host (i.e.
+your machine) reaching the guest (i.e. the VM).
+
+After disconnecting from the VPN, the routing can be restored using the following:
+
+```
+sudo ipconfig vboxnet0 delete 10.1.1.1
+sudo ipconfig set vboxnet0 manual 10.1.1.1
```
leenagupte: Should we add `mac$` as the prompt so it's clear where to run these commands from?

brucebolt: Added this now

PR #1613, opened 2019-03-12
Update emergency banner links following move to AWS

thomasleese:
The job now exists in AWS so we need to update the links to point to the right place.

Depends on https://github.com/alphagov/govuk-puppet/pull/8832 being deployed.

### DIFF
--- a/source/manual/emergency-publishing.html.md
+++ b/source/manual/emergency-publishing.html.md
@@ -5,7 +5,7 @@ parent: "/manual.html"
 layout: manual_layout
 section: Publishing
 important: true
-last_reviewed_on: 2019-01-28
+last_reviewed_on: 2019-03-12
 review_in: 3 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manual/emergency-publishing.html.md
```diff
@@ -38,8 +38,8 @@ The data for the emergency banner is stored in Redis. Jenkins is used to set the
 1. Go to the Jenkins task:
 
   - [Deploy the emergency banner on Integration](https://deploy.integration.publishing.service.gov.uk/job/deploy-emergency-banner/)
-  - [Deploy the emergency banner on Staging](https://deploy.staging.publishing.service.gov.uk/job/deploy-emergency-banner/)
-  - [⚠️ Deploy the emergency banner on Production ⚠️](https://deploy.publishing.service.gov.uk/job/deploy-emergency-banner/)
+  - [Deploy the emergency banner on Staging](https://deploy.blue.staging.govuk.digital/job/deploy-emergency-banner/)
```
benthorner: Just a minor thing, but I think we're favouring `deploy.staging.govuk.digital` now.

thomasleese: I didn't even know that existed! When I try it in my browser I get connection refused, but it could be because I'm on the VPN?

benthorner: As with the `.blue` version, you need to go to `https` directly :-|.

thomasleese: Ah yep! I'll update the links.

PR #1616, opened 2019-03-13
Update rummager replay docs

bevanloon:
Update the rummager replay docs.

Running things at 60x speed seems to cause lots of lost connections. Drop this down to 10x which seems more reliable.

Make sure we are explicit about running the replay on all 3 servers.

Also tighten up some language to point out that you `must` make a copy of the logs.

### DIFF
--- a/source/manual/rummager-traffic-replay.html.md
+++ b/source/manual/rummager-traffic-replay.html.md
@@ -34,7 +34,10 @@ These indexes are populated by whitehall calling an HTTP API in Rummager.
 We have also setup [Gor][gor] logging for `POST` and `GET` requests so that we
 can replay the traffic.
 
-The logs are stored on the rummager servers (1 file per server) location at:
+The logs are stored on the rummager servers. You will need to run the replay on
+each server.
+
+The location of the logs is:
 
 ```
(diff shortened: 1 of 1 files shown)


source/manual/rummager-traffic-replay.html.md
```diff
@@ -34,21 +34,25 @@ These indexes are populated by whitehall calling an HTTP API in Rummager.
...
+
+The following command can be used to run the restore:
 
 ```bash
-$ sudo goreplay -input-file "20171031.log|6000%" -stats -output-http-stats -output-http "http://localhost:3009/|6000%" -verbose
+$ sudo goreplay -input-file "20171031.log|1000%" -stats -output-http-stats -output-http "http://localhost:3009/|6000%" -verbose
 ```
 
-This runs the restore at 60x the speed it was saved so each hour of logs takes
+This runs the restore at 10x the speed it was saved so each hour of logs takes
```
timblair: The calculation here is no longer correct (each hour will take 6 minutes to process).

bevanloon: 👍 thanks Tim. Updated.

PR #1637, opened 2019-03-19
Update the docs for publishing a module to Puppet Forge

issyl0:
- We did this for the first time in _a while_, and the new version
  changed the build process.


issyl0:
This deployment process changed for [puppet-aptly](https://github.com/gds-operations/puppet-aptly), the module that was updated to use PDK which we just released a new version of.

Thinking about it, the old build instructions should still apply to modules that _aren't_ PDKs? Should we include both?

issyl0:
The instructions now state the two different approaches because _most_ of the modules _aren't_ PDKs... _yet_.

PR #1652, opened 2019-03-27
What's up with search?

barrucadu:
We'll use this, in addition to dropping messages into `#govuk-2ndline`, to update on the state of the migration.

---

[Trello card](https://trello.com/c/8B0HvaDY/100-prepare-some-documentation-while-the-migration-is-happening)

### DIFF
--- a/source/manual/business-readiness-update-content.html.md
+++ b/source/manual/business-readiness-update-content.html.md
@@ -28,7 +28,7 @@ The process of converting the spreadsheet into a CSV involves removing a header 
 6. Run the `tag_metadata` rake task in rummager/search-api to index the contents of the new CSV, it should take 2 to 3 minutes:
     * [integration][tag_metadata_integration]
     * [staging-aws (search-api)][tag_metadata_staging_aws]
-    * [production][tag_metadata_production]
+    * [production (rummager)][tag_metadata_production_r] *and* [production (search-api)][tag_metadata_production_sa]
 
     ![rake_task](images/rake.png)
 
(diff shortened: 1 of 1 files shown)


source/manual/whats-up-with-search.html.md
```diff
@@ -0,0 +1,49 @@
...
+
+We (`#govuk-search-perf`) are upgrading GOV.UK's elasticsearch from version 2 to 5.  We're now rolling the change out to production.
+
+## What you need to know
+
+1. Alerts about search-api in AWS production are now important
+
+1. Please continue making changes to rummager, and then merging rummager `master` into search-api.  [Like this](https://github.com/alphagov/search-api/pull/28).
+
+1. If you want to update a content item *published* by rummager (eg, some finders, including the business readiness finder), you'll need to publish the ner version from rummager.  We'll be changing the `publishing_app` field of relevant content items later.
```
brucebolt:
```suggestion
1. If you want to update a content item *published* by rummager (eg, some finders, including the business readiness finder), you'll need to publish the new version from rummager.  We'll be changing the `publishing_app` field of relevant content items later.
```

barrucadu:
Given that the new page is solely of use to 2ndline, it could all be on a card.  I don't really mind either way.

brucebolt:
My thought is that a card is probably easier to update (since we don't have to get reviews on it)

barrucadu:
Ok, I'll remove the new page and just leave a note about deploying the business readiness finder

PR #1657, opened 2019-03-29
Update guidance on merging PRs to include Dependabot

karlbaker02:
This PR updates the guidance on merging PRs to include Dependabot specific advice, in particular to recognise Dependabot as an external contributor and to ensure two approvals when necessary.

This commit also updates the guidance on deploying changes after they have been merged to emphasise their deployment to production as soon as possible, practicing continuous deployment and ensuring that we can release our apps at any time without worry of a backlog of unreleased changes.

### DIFF
--- a/source/manual/merge-pr.html.md
+++ b/source/manual/merge-pr.html.md
@@ -4,7 +4,7 @@ title: Merge a Pull Request
 section: GitHub
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2019-02-15
+last_reviewed_on: 2019-08-28
 review_in: 6 months
 ---
 
@@ -15,6 +15,8 @@ There are five rules for reviewing and merging PRs, which apply to all applicati
 3. If a branch is force-pushed or rebased after a review on its PR, the author must dismiss the stale review and ask for a new one, unless the change is a rebase on top of a small piece of work and the author is confident there are no side effects.
 4. The GitHub review UI should be used to mark a PR as approved or requiring changes.
 5. The GitHub UI should be used to merge the PR. This ensures the PR number is added to the merge commit.
+
+Once a PR is merged, you should deploy your changes at the earliest convenience to ensure unreleased changes do not back up and to keep our applications deployable - regardless of the perceived size of the PR merged (including [Dependabot](#Dependabot) PRs). [Deploying](/manual/deploying.html) should be done in the regular way, taking the merged changes all the way through to production.
 
 ## Example scenarios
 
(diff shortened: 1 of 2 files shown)


source/manual/merge-pr.html.md
```diff
@@ -64,6 +66,14 @@ git push --set-upstream origin thomasjhughes-patch-1
 
 Change `thomasjhughes` to the GitHub username of the contributor and `patch-1` to the name of the branch in their fork.
 
+#### Dependabot
+
+Dependabot raises PRs whenever it sees new versions of gems available that are required by our applications.
+
+Dependabot is an external contributor and is therefore be subject to the same due diligence checks set out above as any other external contributor would have to go through, requiring two people from GDS to approve the PR before it can be merged. Particular attention should be paid to the changelog(s) of the upgraded gem(s) to ensure that there no unintended side-effects are introduced.
```
rubenarakelyan: "that there no unintended" - should be "that no unintended"

rubenarakelyan: Also, "is therefore be subject" should be "is therefore subject"

karlbaker02: Thanks, I don't know what happened to me here!

source/manual/merge-pr.html.md
```diff
@@ -64,6 +66,14 @@ git push --set-upstream origin thomasjhughes-patch-1
 
 Change `thomasjhughes` to the GitHub username of the contributor and `patch-1` to the name of the branch in their fork.
 
+#### Dependabot
+
+Dependabot raises PRs whenever it sees new versions of gems available that are required by our applications.
+
+Dependabot is an external contributor and is therefore subject to the same due diligence checks set out above as any other external contributor would have to go through, requiring two people from GDS to approve the PR before it can be merged. Particular attention should be paid to the changelog(s) of the upgraded gem(s) to ensure that no unintended side-effects are introduced.
+
+The one exception to the number of GDS people required to review the PR is if the gems being upgraded consist of only GOV.UK-owned gems or gems in the `test` block of the Gemfile - see [Manage Ruby dependencies with Dependabot](/manual/manage-ruby-dependencies.html) for more information.
```
cbaines: So I've been merging Dependabot Pull Requests for GOV.UK gems, without waiting for a second review. But in many of these cases, Dependabot has also changed the Gemfile.lock to update other gems as well.

Seen as you're updating the documentation, it might be good to try and clarify what should be done with Pull Requests relating to internal gems, with external updates mixed in? As a strict reading of this paragraph, seems at least to disagree with what I and others are doing.

karlbaker02: I think in this case we would _have_ to follow the process of requiring two reviews, regardless of the fact that internal gems have also been upgraded, because of the presence of the external gems that have also been upgraded. I'm happy to elaborate this in the above if it isn't clear.

cbaines: Given that I think that differs from what I'm doing at the moment, I think it would be worth being very explicit.

karlbaker02: Updated to provide more clarity.

cbaines: Great, thanks :+1:

PR #1658, opened 2019-03-29
Document how to tag content with facet values

steventux:
https://trello.com/c/KYvbTPDh/316-document-how-to-import-a-facet-group-and-tag-data-by-csv-in-content-tagger


source/manual/tag-content-with-facet-values.html.md
```diff
@@ -0,0 +1,60 @@
...
+2. Click on the facet group you'd like to use for tagging content.
+3. Enter the page path you'd like to edit.
+4. Type to search available facets in the "Facets" select field and assign the appropriate values.
+5. Click "Update facet values". The changes will patch links in the Publishing API and take effect as soon as they reach the content-store.
+
+
+\* Adjust the deploy jenkins environment and data file path accordingly.
+
+
+### How to bulk tag business readiness content to facet values
```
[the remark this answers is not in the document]
steventux: Theoretically yes though I wouldn't be too confident in documenting that.
The tagging works on a string match between the facet value content item and the tag value string in the spreadsheet, so there's nothing finder specific about it.

steventux:
@leenagupte I've split these up into two distinct manual pages, makes more sense to me :+1: 
I've not added screenshots of facet group drafting/publishing Jenkins jobs but have documented the tasks and args.
Also added a screenshot of content-tagger facet tagging for context.

source/manual/tag-content-with-facet-values.html.md
```diff
@@ -0,0 +1,33 @@
...
+type: learn
+owner_slack: "#govuk-platform-health"
+last_reviewed_on: 2019-03-2r9
+review_in: 3 months
+related_applications: [content-tagger]
+---
+
+Content tagger is the central place for tagging content with facet values.
+
+### How to edit facet values for a single page
```
leenagupte: Not a blocker, but I think we should add a note that this should be performed by the content team.

steventux: Done

steventux:
@leenagupte I've added https://github.com/alphagov/govuk-developer-docs/pull/1658/commits/d85f2578e4dd632c402d4631e5589aa7c6773b08 as this updates current tagging documentation to point to the new docs.

PR #1664, opened 2019-04-05
Add docs on connecting to the contracts archive finder

bevanloon:


bevanloon:
@thomasleese I've added that in df6fafc

PR #1670, opened 2019-04-09
Update business readiness rake task screenshots

barrucadu:


leenagupte:
@barrucadu The `tag_metadata` task is still being used, so I think we should keep the screenshot.

barrucadu:
The screenshot is still there, github just displays the diff a bit unhelpfully - click "View file"

PR #1677, opened 2019-04-12
Update NFS errors doc

bilbof:
Links to the documentation on how to fix this issue without rebooting your machine.


source/manual/nfs-errors-vm.html.md
```diff
@@ -20,7 +20,8 @@ This will use a less efficient method of sharing folders, but it will work with
 the VPN.
 
 Once you disconnect from the VPN, you will need to restart your computer before
-you're able to use Vagrant with NFS again.
+you're able to use Vagrant with NFS again. Alternatively, after using the VPN
```
thomasleese: I reckon we can just remove the line about restarting, since it doesn't seem necessary anymore.

PR #1696, opened 2019-04-17
adds ruby / bundler / rbenv instructions

ChrisBAshton:
I found that I couldn't complete the "Import production data" step without installing bundles, which required bundler, which required a version of ruby that isn't locked down

Initially went into detail around installing `rbenv` via the `rbenv-installer` but we can do it through Homebrew more easily. Have left the debug instructions in the PR.

Also tweaked our linter to allow `#` lines in code blocks.

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -57,13 +57,12 @@ First, install:
 * [VirtualBox][]
 * [Vagrant][]
 * The vagrant-dns plugin (`vagrant plugin install vagrant-dns`)
-* [rbenv][] (you may find it easier to [install via rbenv-installer][] than via Homebrew due to Xcode restrictions)
+* [rbenv][] (via Homebrew - `brew install rbenv` - you can install Homebrew via the Managed Software Center)
 
 [git-scm]: https://git-scm.com/downloads
 [VirtualBox]: https://www.virtualbox.org/
 [Vagrant]: https://www.vagrantup.com/downloads.html
 [rbenv]: https://github.com/rbenv/rbenv
-[install via rbenv-installer]: https://github.com/rbenv/rbenv-installer#rbenv-installer
 
 Starting with High Sierra 10.13, kernel extensions must be approved by
 the user (see [this Apple technical note][kext]).  This causes the
@@ -116,7 +115,7 @@ Now verify that rbenv is working as expected:
 
     mac$ cd ~/govuk/govuk-puppet
     mac$ rbenv versions
-    
+
 This should output something like:
 
 ```
@@ -135,13 +134,12 @@ This should output something like:
 If, instead, this outputs `/usr/bin/ruby`, then you'll need to update your `~/.bash_profile` to have `ruby` properly overridden in your `PATH`:
 
 ```bash
-export PATH="$PATH:~/.rbenv"
+# force load rbenv to ensure correct version of ruby used
+export PATH="$PATH:~/.rbenv/shims"
 if which rbenv > /dev/null; then eval "$(rbenv init -)"; fi
 ```
 
-Don't forget to `source ~/.bash_profile` to update the `PATH` for your current terminal session.
-
-This Ruby setup is a pre-requisite for the "[Import production data](#6-import-production-data)" step.
+This Ruby setup is a pre-requisite for the "[Import production data](#7-import-production-data)" step.
 
 ## 4. Boot your VM
 
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -57,10 +57,13 @@ First, install:
 * [VirtualBox][]
 * [Vagrant][]
 * The vagrant-dns plugin (`vagrant plugin install vagrant-dns`)
+* [rbenv][] (you may find it easier to [install via rbenv-installer][] than via Homebrew due to Xcode restrictions)
 
 [git-scm]: https://git-scm.com/downloads
 [VirtualBox]: https://www.virtualbox.org/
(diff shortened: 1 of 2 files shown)


source/manual/get-started.html.md
```diff
@@ -57,10 +57,13 @@ First, install:
 * [VirtualBox][]
 * [Vagrant][]
 * The vagrant-dns plugin (`vagrant plugin install vagrant-dns`)
+* [rbenv][] (you may find it easier to [install via rbenv-installer][] than via Homebrew due to Xcode restrictions)
```
chao-xian: Are the last set of square brackets needed?

source/manual/get-started.html.md
```diff
@@ -105,6 +108,41 @@ Some processes are dependent on your local gov.uk setup being in `~/govuk`:
     mac$ cd ~/govuk
     mac$ git clone git@github.com:alphagov/govuk-puppet.git
 
+### 3.1 Ensure your Ruby setup works
+
+You'll want to install the [specific version of Ruby in govuk-puppet][https://github.com/alphagov/govuk-puppet/blob/master/.ruby-version#L1] (currently `rbenv install 1.9.3-p550`), as the default system Ruby is too locked-down.
```
chao-xian: Probably simpler not to mention the current version as this will go out of date very quickly.

source/manual/get-started.html.md
```diff
@@ -105,6 +108,41 @@ Some processes are dependent on your local gov.uk setup being in `~/govuk`:
...
+
+Now verify that rbenv is working as expected:
+
+    mac$ cd ~/govuk/govuk-puppet
+    mac$ rbenv versions
+    
+This should output something like:
```
chao-xian: I wonder if this is a bit of repetition from the rbenv docs?

source/manual/get-started.html.md
ChrisBAshton: Do you mean are they needed for the markdown to be valid, or do you mean we should trust the developer to go and find the right link for themselves?

source/manual/get-started.html.md
ChrisBAshton: Getting a working rbenv setup is surprisingly fiddly (someone on Slack mentioned they've never managed to get theirs working!) so I think it's useful to provide sanity checks to confirm it is set up as expected.

Agreed that this makes the whole page a bit Ruby-centric though - is it worth splitting out into a separate page I wonder?

source/manual/get-started.html.md
ChrisBAshton: Yep, it's a violation of DRY and doesn't feel like a good thing to hard-code here. On the other hand, `.ruby-version` hasn't been updated for 3 years... 🤔 and I could see people misinterpreting the docs and running `rbenv install 1.9.3` (without the 'patch' bit) and finding it doesn't work!

source/manual/get-started.html.md
chao-xian: Oh no sorry I parsed it wrong late in the day.

source/manual/get-started.html.md
chao-xian: Again, sorry it was late in the day and I hadn't grokked that this is for puppet!

source/manual/get-started.html.md
chao-xian: No let's keep it all in one place. The next person to do this can refine it if they wish :)

ChrisBAshton:
Actually can we hold off merging this for a sec? I'm not sure the bash_profile instructions are working (got `rbenv is not defined` this morning 😫) - and also I've just noticed Homebrew is in the Managed Software Center, so that may be an easier route!

ChrisBAshton:
Should be good for a re-review 😄 thanks for your time @chao-xian !

ChrisBAshton:
Thanks Kelvin! I wanted that commit to fail so that I could verify that https://github.com/alphagov/govuk-developer-docs/pull/1696/commits/e5575bd49a4bf9378c93fec48edbd52599206cf0 fixes it correctly 😄 

I'll go ahead and merge now it has a ✅ !

spec/pages/manual_page_spec.rb
```diff
@@ -33,8 +33,8 @@
       source/manual/howto-merge-a-pull-request-from-an-external-contributor.html.md
     ])
       it "doesn't use H1 tags (the page title is already an H1)" do
-        expect(raw).not_to match(/\n#\s/), "This page contains an unnecessary H1." \
-          " This may have been triggered because of a # comment in a code block"
+        raw_without_code_blocks = raw.remove(/```[a-z]*\n[\s\S]*?\n```/) # want to allow `#` (comments) in code blocks
```
ChrisBAshton: FYI regex taken from https://coderwall.com/p/r6b4xg/regex-to-match-github-s-markdown-code-blocks

source/manual/get-started.html.md
```diff
@@ -105,6 +107,40 @@ Some processes are dependent on your local gov.uk setup being in `~/govuk`:
...
+/Users/yourname/.rbenv/shims/ruby
+```
+
+If, instead, this outputs `/usr/bin/ruby`, then you'll need to update your `~/.bash_profile` to have `ruby` properly overridden in your `PATH`:
+
+```bash
+# force load rbenv to ensure correct version of ruby used
+export PATH="$PATH:~/.rbenv/shims"
+if which rbenv > /dev/null; then eval "$(rbenv init -)"; fi
+```
```
ChrisBAshton: These instructions _look_ like overkill given we installed with Homebrew, but I had to follow these steps myself to get the setup to work persistently!

PR #1718, opened 2019-04-29
Add a doc on changing a slug and creating a redirect in Whitehall

AgaDufrat:
It is a common 2nd line task, which currently is not documented
in the docs.

### DIFF
--- a/source/manual/howto-change-slug-and-create-redirect.html.md
+++ b/source/manual/howto-change-slug-and-create-redirect.html.md
@@ -1,0 +1,27 @@
+---
+owner_slack: "#govuk-2ndline"
+title: Change a slug and create redirect in Whitehall
+section: Publishing
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2019-04-29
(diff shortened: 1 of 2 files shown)


source/manual/howto-change-slug-and-create-redirect.html.md
```diff
@@ -0,0 +1,27 @@
...
+parent: "/manual.html"
+last_reviewed_on: 2019-04-29
+review_in: 6 months
+---
+
+### Change a slug and create redirect in Whitehall
+
+##### Whitehall
+
+A Rake Tasks exist in Whitehall to quickly change the slug of a Document. It also reindexes the Document with it's new slug, republishes the document to Publishing API, which automatically handles the redirect.
```
brucebolt:
```suggestion
A Rake Task exists in Whitehall to quickly change the slug of a Document. It also reindexes the Document with it's new slug, republishes the document to Publishing API, which automatically handles the redirect.
```

brucebolt:
Would you be able to squash these two commits into one?

PR #1723, opened 2019-05-01
add ssh key to keychain persistently through bash_profile

ChrisBAshton:
Keep having to remember to run this command when I restart my machine.

### DIFF
--- a/source/manual/get-ssh-access.html.md
+++ b/source/manual/get-ssh-access.html.md
@@ -26,6 +26,8 @@ To create a new account, start by creating an SSH key at least 4096 bits long. F
 Import the SSH key into your keychain.
 
     mac$ /usr/bin/ssh-add -K ~/.ssh/alphagov
+
+You'll also want to add the above line into your `~/.bash_profile` so that it is persistent between restarts.
     
 Create a public key from this:
 


source/manual/get-ssh-access.html.md
```diff
@@ -26,6 +26,8 @@ To create a new account, start by creating an SSH key at least 4096 bits long. F
 Import the SSH key into your keychain.
 
     mac$ /usr/bin/ssh-add -K ~/.ssh/alphagov
+
+You'll also want to add the above line into your `~/.bash_profile` so that it is persistent between restarts.
```
thomasleese: I haven't needed to run this command more than once, and my understanding was that adding it to Keychain means that it becomes persisted. Do you know why you're having to run it regularly?

bilbof: If you restart your computer you'll need to run it again.

ChrisBAshton: Not sure tbh - I definitely called this command a couple of weeks ago:

```sh
/usr/bin/ssh-add -K ~/.ssh/alphagov
```

Looking at https://unix.stackexchange.com/questions/140075/ssh-add-is-not-persistent-between-reboots/140077 and https://superuser.com/a/1155833, looks like from Sierra onwards you have to do a bit of work to get the SSH keys loaded between reboots. Are you on an older Mac OS?

I know @brucebolt has been having the same issues as me, so this feels like a good instruction to add to the docs.

PR #1767, opened 2019-05-28
Describes storing commands needed between VM destroys

ChrisBAshton:
Occasionally, you just cannot avoid running `vagrant destroy` on a broken VM.

When that happens, you need to set up a few things from scratch:

- Git global configuration
- Symlink for `govukcli`
- Any custom aliases or theming you like to apply to your terminal 

Encouraging the idea of 'throwaway VMs', we should make it easier to get a VM
back up and running after a destroy. This is a good first step for that.

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -161,10 +161,16 @@ Once your VM is running, you can SSH into it with:
 
 ### Set your Git username and email
 
-You can assign your name and email to commits on the VM:
+You can associate your git credentials with commits on the VM by setting your global configs on the dev machine.
 
     dev$ git config --global user.email "friendly.giraffe@digital.cabinet-office.gov.uk"
     dev$ git config --global user.name "Friendly Giraffe"
+
+It's worth noting at this point that if you ever need to destroy and rebuild your machine, you'll need to follow these steps again. Therefore, it's probably worth creating a file on your host machine (e.g. `~/.bash_profile_vm`) and writing these commands in there so that you can easily transfer them into your VM when needed.
+
+    mac$ more ~/.bash_profile_vm | pbcopy
+    dev$ vi ~/.bash_profile # then paste the contents in
+    dev$ source ~/.bash_profile # load it for the current session (or 'exit' and 'vagrant ssh' again)
 
 ### Config for ZSH users
 
@@ -203,6 +209,13 @@ If installing the Python dependencies for fabric-scripts fails, your version of 
 👉 First, [set up your AWS account](/manual/set-up-aws-account.html)
 
 👉 Then, [set up your CLI access for AWS](/manual/aws-cli-access.html)
+
+NB, this is another candidate for adding to your `~/.bash_profile_vm`:
+
+```sh
+# Set up govukcli alias
+sudo ln -sf /var/govuk/govuk-aws/tools/govukcli /usr/local/bin/govukcli
+```
 
 ## 7. Import production data
 


source/manual/get-started.html.md
```diff
@@ -161,11 +161,17 @@ Once your VM is running, you can SSH into it with:
 
 ### Set your Git username and email
 
-You can assign your name and email to commits on the VM:
+You can associate your GitHub account with commits on the VM by setting your git global configs on the dev machine.
```
brucebolt: I don't think this wording is 100% accurate.  You're only setting an arbitrary name/email, which may or may not be linked to a GitHub account.

ChrisBAshton: This is true 👍 Will fix

source/manual/get-started.html.md
```diff
@@ -161,11 +161,17 @@ Once your VM is running, you can SSH into it with:
...
+You can associate your git credentials with commits on the VM by setting your global configs on the dev machine.
 
     dev$ git config --global user.email "friendly.giraffe@digital.cabinet-office.gov.uk"
     dev$ git config --global user.name "Friendly Giraffe"
 
+It's worth noting at this point that if you ever need to destroy and rebuild your machine, you'll need to follow these steps again. Therefore, it's probably worth creating a file on your host machine (e.g. `~/.bash_profile_vm`) and writing these commands in there so that you can easily transfer them into your VM when needed.
+
+    mac$ more ~/.bash_profile_vm | pbcopy
+    dev$ vi ~/.bash_profile # then paste the contents in
+    dev$ source ~/.bash_profile # load it for the current session (or 'exit' and 'vagrant ssh' again)
```
cbaines: Not quite related, but have you considered trying to use Vagrant to manage some of this configuration? You can customise the configuration (e.g. [1]) like memory usage, and I can see being able to specify the Git user details be useful.

1: https://github.com/alphagov/govuk-puppet/blob/master/development-vm/Vagrantfile.localconfig.example

source/manual/get-started.html.md
```diff
@@ -204,6 +210,13 @@ If installing the Python dependencies for fabric-scripts fails, your version of
 
 👉 Then, [set up your CLI access for AWS](/manual/aws-cli-access.html)
 
+NB, this is another candidate for adding to your `~/.bash_profile_vm`:
+
+```sh
+# Set up govukcli alias
+sudo ln -sf /var/govuk/govuk-aws/tools/govukcli /usr/local/bin/govukcli
+```
```
cbaines: How about just adding this to the PATH by default (through govuk-puppet)?

ChrisBAshton: I did try doing this but couldn't figure out where to put it - could you give any pointers please? :)

source/manual/get-started.html.md
ChrisBAshton: Thanks Chris - I'll take a look.

source/manual/get-started.html.md
cbaines: Maybe do something similar to https://github.com/alphagov/govuk-puppet/commit/eb4135413c87844f85c3fa1abee68bf595991797

ChrisBAshton: Raised https://github.com/alphagov/govuk-puppet/pull/9209.

source/manual/get-started.html.md
ChrisBAshton: Raised https://github.com/alphagov/govuk-puppet/pull/9212

PR #1781, opened 2019-06-04
Populate app dependencies in applications.yml and generate PlantUML C4-style diagrams

VitalieMogoreanu:
Added a couple of scripts to make it easier for us to have diagrams in our documentation:

- apps/app_diagrams_generator.rb - generates PlantUML C4-style diagrams (RicardoNiepel/C4-PlantUML) from the content of applications.yml file. It uses application type as a system and application repository as component in C4 model terminology (c4model.com). It also supports a new applications.yml actors attribute to generate C4 Persons

- apps/app_data_dependencies.rb - can be run manually to populate app dependencies in applications.yml from plek references in source code. It assumes govuk-developer-docs is in the same folder as all the other govuk repositories

C4 Model: https://c4model.com/

Why we prefer PlantUML to other tools:
https://docs.google.com/document/d/1nYriTypVW1Yg6GeSFuUg8Kr5eF_WhfRwZXkw37bJq1o

C4 PlantUML: https://github.com/RicardoNiepel/C4-PlantUML


app/app_diagrams_generator.rb
```diff
@@ -0,0 +1,172 @@
+require 'yaml'
+
+class AppDiagramGenerator
+  def initialize
+    @repository_path = Dir.getwd
+    @application_data_load_path = File.join(@repository_path, "data/applications.yml")
+    @application_diagrams_save_path = File.join(@repository_path, "diagrams/")
```
erino: You'll need to create the directory otherwise the script will file at line 52. The following should do it

```ruby
FileUtils.mkdir_p(File.join(@repository_path, "diagrams/"))
```

VitalieMogoreanu: Thanks, I've added this

diagrams/app_data_dependencies.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    puts "Saving application data to #{@application_data_save_path}"
+    File.write(@application_data_save_path, @all_applications.to_yaml)
+
+    puts "Plek match plekRefCount: #{@plek_ref_count}"
+  end
+end
+
+g = AppDependencyGenerator.new
+g.update_all_app_dependencies
```
tijmenb: You could add this to the `Rakefile`, and move this file to `/lib`. That would also allow us to document the task to explain what it does.

diagrams/app_data_dependencies.rb
```diff
@@ -0,0 +1,100 @@
+require 'yaml'
+
+class AppDependencyGenerator
```
tijmenb: I'm not sure how useful this is once we've done the initial update of `applications.yml`. Maybe it can output (not write) all the dependencies it thinks are missing?

diagrams/app_diagrams_generator.rb
```diff
@@ -0,0 +1,219 @@
+require 'yaml'
+require 'fileutils'
+
+class AppDiagramGenerator
+  def initialize
+    @repository_path = Dir.getwd
```
tijmenb: This doesn't need to be an instance variable. I think only calling `YAML.load_file("data/applications.yml")` would also work, we don't need to remember the path I think.

diagrams/app_diagrams_generator.rb
```diff
@@ -0,0 +1,219 @@
...
+    init_system_dependencies
+    generate_system_context_diagram
+
+    @applications_by_type.keys.each do |key|
+      generate_container_diagram(key)
+    end
+  end
+end
+
+g = AppDiagramGenerator.new
```
tijmenb: This can also be moved to a Rake task.

diagrams/app_diagrams_generator.rb
```diff
@@ -0,0 +1,219 @@
+require 'yaml'
+require 'fileutils'
+
+class AppDiagramGenerator
```
tijmenb: There's a lot going on in this class, it would be useful to have a test that exercises it all, to avoid it from breaking in the future.

diagrams/app_diagrams_generator.rb
```diff
@@ -0,0 +1,219 @@
...
+          apps.map { |app| app["actors"] }
+          .flatten
+          .reject(&:nil?)
+          .uniq
+        ]
+      }.to_h
+  end
+
+  def get_valid_alias(title)
+    title.tr(".", "_").tr(" ", "_").tr("-", "_")
```
tijmenb: We use activesupport in the docs, so you could replace this with `title.underscore` for this.

app/app_data_dependencies.rb
```diff
@@ -0,0 +1,113 @@
+require 'yaml'
```
tijmenb: Note that you've [not set your git email](https://github.com/alphagov/govuk-developer-docs/commit/1c68be66ea0062a9bcdcc4d101724fd040459e0b.patch), which is causing the commits not to be associated with your user.

diagrams/app_data_dependencies.rb
VitalieMogoreanu: moved file to /app and removed these lines

diagrams/app_data_dependencies.rb
VitalieMogoreanu: commented out code in this file

diagrams/app_diagrams_generator.rb
VitalieMogoreanu: added some tests

diagrams/app_diagrams_generator.rb
VitalieMogoreanu: replaced

Gemfile
```diff
@@ -5,17 +5,16 @@ ruby File.read('.ruby-version').chomp
 gem 'capybara'
 gem 'rake'
 gem 'rspec', '~> 3.8'
-gem 'webmock', '~> 3.6'
```
tijmenb: Are the changes to the Gemfile intentional? I don't think we can drop middleman as a dependency?

app/app_data_dependencies.rb
```diff
@@ -0,0 +1,100 @@
+# require 'yaml'
```
tijmenb: We don't commit commented out code, this should be removed.

README.md
```diff
@@ -63,13 +63,9 @@ NO_CONTRACTS=true bundle exec middleman build
 
 This will create a bunch of static files in `/build`.
 
-### Testing
-
-`bundle exec rake`
-
 ### Deployment
 
-This project is re-deployed by a production Jenkins task every hour between 9am and 7pm (to pick up external
```
tijmenb: Why the change here?

README.md
```diff
@@ -63,13 +63,9 @@ NO_CONTRACTS=true bundle exec middleman build
 
 This will create a bunch of static files in `/build`.
 
-### Testing
```
tijmenb: This should stay I think.

app/external_doc.rb
```diff
@@ -119,7 +118,11 @@ def call
 
       doc.css('h1, h2, h3, h4, h5, h6').each do |node|
         text = node.text
-        id = StringToId.convert(text)
+        id = text
```
tijmenb: This now duplicates the code in `DeveloperDocsRenderer`? What's the problem with this?

app/github_repo_fetcher.rb
```diff
@@ -35,7 +35,6 @@ def client
         builder.response :logger
         builder.use Faraday::HttpCache, serializer: Marshal, shared_cache: false
         builder.use Octokit::Response::RaiseError
-        builder.use Faraday::Request::Retry, exceptions: Faraday::Request::Retry::DEFAULT_EXCEPTIONS + [Octokit::ServerError]
```
tijmenb: Why remove this?

config.rb
```diff
@@ -48,10 +48,6 @@ def related_things
     @related_things ||= RelatedThings.new(manual, current_page)
   end
 
-  def section_url
```
tijmenb: Why remove this?

PR #1795, opened 2019-06-10
Change section to '2nd line'

ChrisBAshton:
Previous section was 'data.gov.uk', which is applicable as these are data.gov.uk
requests, but I think the section is better suited to '2nd line'. When visiting
https://docs.publishing.service.gov.uk/manual.html, I'd expect to see this page
under 2nd line.

I hope we can add more 'Common 2nd line' type pages for things like alerts, tech
support, content changes etc.


brucebolt:
I'm not keen on moving this out of the DGU section.  If a request comes in about data.gov.uk, I'd probably go to the data.gov.uk section first, rather than the 2nd line section (which appears to be stuff about being on 2nd line, e.g. holiday cover and on-call).

PR #1798, opened 2019-06-11
Document redirect 404 issue and workaround

ChrisBAshton:
In Zendesk ticket 3712180 I had to process a redirect. I followed these instructions, initially
on integration, and they worked perfectly. When I followed the same instructions on production,
the redirect was applied but the new URL would 404.

Working with @cbaines we saw that the `downstream_low` queue in publishing-api was quite full,
and suspect that the rake task had put it in the downstream_low queue. In the meantime we ran
a rake task to bump it to high priority to process the content change more quickly. I feel it is worth
documenting the workaround in the same instructions that led to the issue.

### DIFF
--- a/source/manual/howto-change-slug-and-create-redirect.html.md
+++ b/source/manual/howto-change-slug-and-create-redirect.html.md
@@ -10,9 +10,9 @@ review_in: 6 months
 
 ### Change a slug and create redirect in Whitehall
 
-##### Whitehall
+#### Whitehall
 
-A Rake Tasks exists in Whitehall to quickly change the slug of a Document. It also reindexes the Document with it's new slug, republishes the document to Publishing API, which automatically handles the redirect.
+A Rake Tasks exists in Whitehall to quickly change the slug of a Document. It also reindexes the Document with its new slug, republishes the document to Publishing API, which automatically handles the redirect.
 
 The task takes the Document's old slug and the Document's new slug.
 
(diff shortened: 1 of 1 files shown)


source/manual/howto-change-slug-and-create-redirect.html.md
```diff
@@ -25,3 +25,13 @@ $ bundle exec 'reslug:document[old_slug,new_slug]'
 
 > This rake task also works for Person, Role, PolicyGroup, Organisation, WorldOrganisation and WorldLocation.
 For example: `reslug:world_location[old_slug,new_slug]`
+
+#### Issues
+
+If you run the task and find the redirect has worked but the new location returns a 404, it's likely because there's a queue buildup in publishing-api ([check in Grafana](https://grafana.publishing.service.gov.uk/dashboard/file/sidekiq.json?refresh=1m&orgId=1&var-Application=publishing-api&var-Queues=All&from=now-30m&to=now).
```
cbaines: I guess I'm still not confident enough about the low/high queue priority being an issue here. I'd prefer to investigate further get more confidence before writing it down (as there's a tendency to believe the stuff in the docs to be true).

kevindew: I've seen this problem before.

The problem lies in: 

- Whitehall publishes to Publishing API as a republish which is low priority
- Publishing API runs a nested command to create the redirect
- nested command (creating redirect) is not a redirect and has high priority and is published first
- the previous command languishes in the low priority queue.

I'd guess the root of it is that Publishing API should do nested commands with same priority as the initial one, but since this is all based on update_type it's quite hard to unpick.

ChrisBAshton: Thanks Kevin - I'll rewrite the PR, as your summary is more accurate 👍

ChrisBAshton: Done - please take a look :)

cbaines: :+1:

PR #1810, opened 2019-06-17
Pull in content store README

ChrisBAshton:
Pulls in content-store README:

In order to keep the information in one place and reduce duplication,
and to keep the documentation closer to the code, I've followed the
style of the [step by step docs](https://github.com/alphagov/govuk-developer-docs/blob/master/source/manual/side-by-side-browser.html.md.erb)
and pull in an external README instead.

I've decided to keep the 'limitations' part separate for the time being
as it covers areas outside of the content-store itself and provides a
bit of context more than anything.

Dependent on: https://github.com/alphagov/content-store/pull/579

### DIFF
--- a/source/apis/content-store.html.md.erb
+++ b/source/apis/content-store.html.md.erb
@@ -4,23 +4,19 @@ title: Content Store
 parent: /apis.html
 ---
 
-The [Content Store][content-store-repo] is the store for all published content on
-GOV.UK. When given a path, it responds with the content that should be displayed
-on that path.
+<%= ExternalDoc.fetch(repository: 'alphagov/content-store', path: 'README.md') %>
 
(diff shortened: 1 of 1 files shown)


thomasleese:
I wonder if this documentation is better placed in the [`README.md` for Content Store](https://github.com/alphagov/content-store) and then we can embed it using `ExternalDoc.fetch`, similar to [what was done for the side by side docs](https://github.com/alphagov/govuk-developer-docs/blob/master/source/manual/side-by-side-browser.html.md.erb).

ChrisBAshton:
Hmm, I like the idea 👍is it bad that it pulls in things like [license information](https://github.com/alphagov/content-store#licence)? I guess not!

PR #1811, opened 2019-06-17
Architectural deep-dive document

ChrisBAshton:
Explaining the [architectural overview diagram](https://docs.publishing.service.gov.uk/manual/architecture.html) in words and linking to the relevant repositories.
Rendered version 👉 : https://github.com/alphagov/govuk-developer-docs/blob/architecture-deep-dive/source/manual/architecture-deep-dive.html.md

The aim is for a self-contained document that explains how everything works, offering enough detail to be read in isolation (not just a glossary of links) but without being too prescriptive about the 'current' state of things, where it could fall out of date. This will be a useful document to new joiners, as well as a useful reference for existing devs if they lack knowledge in any part of the stack.

Trello card: https://trello.com/c/Xfq04i3u/1531-write-architectural-deep-dive-document


source/manual/architecture-deep-dive.html.md
```diff
@@ -400,33 +400,41 @@ and processes.
...
-latest versions of govuk-puppet and govuk-secrets to itself (this is why
-you must wait 30 minutes between [deploying Puppet][] between environments).
-The puppet agents, on the other hand, compare their configuration with the
+application offline on that machine. The load balancer would begin serving
+traffic from different nodes of the same class: for this reason it is
+important that all machines are updated at different times. Therefore, each
+instance [runs puppet every half hour][puppet-cronjob], with the puppet
+agents configured to run after the puppetmaster at randomised times.
+
+What actually happens on a puppet run? The puppetmaster copies the latest
```
sengi: The copying of govuk-puppet and govuk-secrets to the Puppetmaster happens when you run the Deploy Puppet job on Jenkins - I wouldn't say that's part of the "puppet run" (puppet agent run) as such.

source/manual/architecture-deep-dive.html.md
```diff
@@ -400,33 +400,41 @@ and processes.
 
 Puppet runs in a master/agent setup. There is a single "puppetmaster" running
 on its own class of machine, whereas the agents run on all the other machines
-(irrespective of class). Each puppet agent is responsible for sending alerts
-and logs to our various [monitoring systems][], and for running health check
```
sengi: Do these reference links actually work with the empty brackets [] ? Usually those are omitted.

source/manual/architecture-deep-dive.html.md
ChrisBAshton: Thanks, I'll amend this. So what does the 'puppet agent run' on the puppetmaster machine do? Does it just detect that it's the master and skip itself?

source/manual/architecture-deep-dive.html.md
ChrisBAshton: They work, but it looks like they're optional. I've now removed them 👍

source/manual/architecture-deep-dive.html.md
sengi: The Puppet agent works in just the same way on the master as it does on all the other machines. That is, it checks for differences between what it sees and what the Puppetmaster (which just happens to be itself) says should be there. Just like you've described at the end of the paragraph :) Or in other words, Puppet manages the Puppetmaster just like any other machine.

ChrisBAshton: Thanks! Have amended.

chao-xian:
Given how excellent your slides were from your Show and Tell, it would be great to illustrate the doc with the diagrams from the slides. It would break it up nicely too, and help people to absorb the vast information.

ChrisBAshton:
Thanks @chao-xian - that certainly opens my slides up to more scrutiny, but I think it'll be worth it! I've selected some key ones, do let me know what you think.

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,462 @@
...
+![](images/nginx-routing.png)
+
+External requests are routed to a 'cache' machine, which uses Varnish for its
+caching. If Varnish has the route response in its cache, it returns that,
+otherwise it forwards the request to [Nginx], a web server running on the
+machine. Nginx is configured with govuk-puppet.
+
+Nginx proxies most requests to [router], which is a GOV.UK maintained
+application running on the cache machines. However, Nginx does proxy some
+routes directly to other apps, such as image URLs routing to asset-manager.
```
cbaines: I believe the requests are processed in this order on the cache machine: NGinx -> Varnish -> Router (at least that's what it looks like if you look at the NGinx and Varnish configuration).

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,462 @@
...
+[govuk-aws]: https://github.com/alphagov/govuk-aws
+[Nginx]: https://www.nginx.com/
+[router]: https://github.com/alphagov/router
+[Terraform]: https://www.terraform.io/
+
+#### Routing via 'router'
+
+'router' is a reverse proxy app written in Go. It stores all known routes
+in memory using a [prefix trie]. The routes are loaded into the trie from
+[redis] at startup. Matched routes are proxied through to the relevant
```
cbaines: I believe the routes come from the MongoDB database shared with the Router API.

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,462 @@
...
+rendering app based on the `backend_id` of the route, which is derived from
+the `rendering_app` field in the corresponding content item in the content
+store (which we'll cover later). For example, if the route has a `backend_id`
+of `frontend`, it will forward the request to the [frontend] application.
+
+When routes are deleted, they're marked as `gone` in router, and router
+returns a 410 Gone response. When routes are changed (which can happen when a
+document's title is changed, affecting its slug), it is marked as `redirect`
+in router, serving a 301 Moved Permanently response. It's worth noting the
```
cbaines: I'm not sure about this bit. The Publishing API handles setting up redirects automatically when content is moved around, not the Router. It just listens to what the Content Store says.

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,462 @@
...
+[short-url-manager] app, whose sole purpose is to allow the creation of short,
+memorable URLs that redirect to longer URLs, often as part of a media campaign.
+It publishes routes to the content store (via the publishing API, which we'll
+also cover later) like other apps publish pages.
+
+Routes are added and updated in real time via [router-api]. The content
+store talks to router-api [directly][register-route] whenever there is a new
+piece of content router needs to know about. router-api stores this information
+in its own database and then sends these routes to router.
```
cbaines: It would be more correct to say "The Router API stores this information in it's database, and then sends requests to the instances of the Router to reload the routes from the database".

One interesting property here is that because there are multiple router processes, one for each cache machine, the Router API has to contact each of them to tell them to reload the routes. This means the Router API has to know about where to find each individual cache machine instance

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,462 @@
...
+
+At this point it's probably worth summarising what "content" is. Almost every
+piece of content on GOV.UK lives in a database called the "[content store]",
+which stores only the latest "edition" of that content. Internally the content
+is referred to as a "document", even if it is not itself a document.
+
+Content is published to the content store via the [publishing API], which
+stores all of the editions of the document, and performs validation checks
+whenever it receives a new edition. Every piece of content has a `document_type`
+corresponding to a particular JSON schema defined in [govuk-content-schemas].
```
cbaines: The relevant schema is decided by the `schema_name`, not the `document_type`.

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,462 @@
...
+- Production
+- Staging
+- Integration
+
+Data is copied from Production to Staging - and from Staging to Integration - every
+24 hours via [automated Jenkins jobs][copy-data-to-staging]. This way our
+environments are always roughly in sync, although it's worth noting that email
+addresses are anonymised and access-limited documents are obfuscated before data is
+copied. The data copying is partly configured in [env-sync-and-backup], however
+we're gradually [configuring this in govuk-puppet][govuk-env-sync] in the move to AWS.
```
cbaines: Given most data syncing now happens through govuk-env-sync, it might be worth flipping the relevance in this paragraph. I realise this was probably correct when you wrote it a while ago.

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,462 @@
...
+As discussed in the routing section, we have different 'classes' of machines
+running in the cloud; to recap, the "cache" machines run the "router"
+application. These classes are configured in [govuk-puppet], which uses
+[puppet] under the hood: tooling which configures resources such as files
+and processes.
+
+Puppet runs in a master/agent setup. There is a single "puppetmaster" running
+on its own class of machine, whereas the agents run on all the other machines
+(irrespective of class). Each puppet agent is responsible for running
+[health check] tests against the GOV.UK applications running on the node,
```
cbaines: I don't think Puppet is responsible for running healthchecks. It can however configure Icinga or other services to check things.

cbaines:
Thanks for writing this Chris, I've made some comments now in the text. I can see this longer form document being useful to people to get up to speed with how GOV.UK works.

One general comment, I at least like consistently capitalising the apps, so Content Store or Publishing API if the application is being referred to. Of course if you're referring to the Git repository, it would be content-store.

source/manual/architecture-deep-dive.html.md
ChrisBAshton: Yes, when a publishing app publishes a new edition with a new slug, Publishing API is aware the route for the content has changed and makes an [additional request to content store to set up a redirect](https://github.com/alphagov/publishing-api/blob/1b8e328540ab96759807787ab87dcb33bbcc72e3/spec/integration/put_content/content_with_a_previously_published_edition_spec.rb#L99-L132).

Content store [registers the redirect with router-api](https://github.com/alphagov/content-store/blob/08f02f990e621c9d2fd473e12a70a6805ddd8dcb/app/models/route_set.rb#L61-L64), which [stores the redirect in MongoDB](https://github.com/alphagov/router-api/blob/96e2838cf68158be7a3f4375e68b9a8c93991b68/app/models/route.rb). This route gets loaded into Router with a `redirect` handler instead of a `backend_id`, I think.

Does that sound right? I haven't been overly explicit about how it all works because I've tried introducing each concept gradually, but I don't think I've written anything here that contradicts that? 🤔

source/manual/architecture-deep-dive.html.md
ChrisBAshton: Thanks Chris. Instead of that, should it be this?

> The Router API stores this information in it's database, and then sends requests to the Router on every 'cache' machine to reload the route.

What you say about router reloading routes from the database contradicts what I've heard from others. My understanding is that it is _indirectly_ loaded from the database at startup, i.e. router-api loads routes from MongoDB and pushes the routes to Router; Router doesn't know about the MongoDB. And then if routes are added/changed at runtime, they're individually added/overwritten via router-api. I may have this wrong though.

source/manual/architecture-deep-dive.html.md
ChrisBAshton: Oops, good spot! 👏

source/manual/architecture-deep-dive.html.md
ChrisBAshton: This is the healthcheck I've linked to: https://github.com/alphagov/govuk-puppet/blob/0b20c7efe7e0c3855d4821e55a914ab577d3b84e/modules/govuk_containers/manifests/app.pp

Perhaps I phrased it poorly. Puppet doesn't define the healthcheck (that's done by the [app itself](https://github.com/alphagov/content-publisher/blob/5b968f1bbfd4fa7e577ea535bb2dc23fcf0c99b8/spec/requests/healthcheck_spec.rb)), but it does appear to be responsible for pinging the healthcheck and triggering an Icinga alert if it fails. How should I rephrase it to be less ambiguous?

source/manual/architecture-deep-dive.html.md
cbaines: > Thanks Chris. Instead of that, should it be this?
> 
> > The Router API stores this information in it's database, and then sends requests to the Router on every 'cache' machine to reload the route.

As far as I know, the Router can only be asked to reload all the routing information, not a specific route.

> What you say about router reloading routes from the database contradicts what I've heard from others. My understanding is that it is _indirectly_ loaded from the database at startup, i.e. router-api loads routes from MongoDB and pushes the routes to Router; Router doesn't know about the MongoDB. And then if routes are added/changed at runtime, they're individually added/overwritten via router-api. I may have this wrong though.

As I understand it, the Router and Router API both access a MongoDB database. The Router API reads and writes, and I think the Router only reads. When the Router API changes something in the database, it sends an empty POST request [1] to each of the Router processes, triggering to reload all the routing information from that shared MongoDB database.

1: https://github.com/alphagov/router-api/blob/master/lib/router_reloader.rb#L45-L48

source/manual/architecture-deep-dive.html.md
cbaines: Yep, that sounds right.

However, given this paragraph and the one prior to it are just about the Router, and the things it does. I would read the bit below as implying that this is a function of the Router, it watches for changes in routes, and sets up redirects.

> When routes are changed (which can happen when a document's title is changed, affecting its slug), it is marked as `redirect` in router, serving a 301 Moved Permanently response.

Given this section is about the router specifically, maybe it's best to bring this up when the Publishing API is mentioned instead?

source/manual/architecture-deep-dive.html.md
cbaines: That `govuk_containers` module isn't a great example, as I'm not sure if we're actually using that Puppet configuration anywhere.

Even so, what's happening there as far as I can see is that Puppet is generating configuration for Icinga to do the checking. The `@@icinga::check` bit relates to this thing [1], I believe the `@@` bit means that the configuration file(s) generated are not ending up on this machine, but somewhere else instead, which in this case will be the monitoring machine that runs Icinga.

1: https://github.com/alphagov/govuk-puppet/blob/master/modules/icinga/manifests/check.pp#L61
2: https://puppet.com/docs/puppet/3.8/lang_exported.html

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,462 @@
...
+title: Architectural deep-dive of GOV.UK
+section: Applications
+type: learn
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2020-03-03
+review_in: 3 months
+---
+
+We can cover most of the GOV.UK architecture by asking ourselves three questions:
```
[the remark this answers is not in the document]
ChrisBAshton: Changing to `can cover a lot of GOV.UK`. 👍 Hopefully this document can grow in the future to cover some of those topics I've neglected! I already cover assets briefly, and could cover more of the email alert architecture.
I have zero experience of smartanswers so not quite sure what's different about them, although each app is bound to have its own quirks and I wouldn't want this document to attempt to capture all of them.

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,462 @@
...
+
+### DNS
+
+![](images/dns.png)
+
+The browser queries a local DNS server to turn the domain name into an IP address.
+The local DNS server might be able to answer immediately from its cache. If not,
+it will query the authoritative name servers for the domain.
+
+`gov.uk.` is a country-code second-level domain or ccSLD (like `co.uk.`) and its
```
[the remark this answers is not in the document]
ChrisBAshton: Nice resource 👍

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,462 @@
...
+authoritative servers are hosted by [Jisc]. Unusually for a ccSLD, `gov.uk` is
+also a website, and hosts the redirect from `gov.uk` to `www.gov.uk`. The records
+for these two domains are within the `gov.uk` second-level zone hosted by Jisc.
+
+GDS hosts several third-level domains under `gov.uk` on [Amazon Route53], for
```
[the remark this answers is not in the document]
ChrisBAshton: I've now included a link to the dns.html at the end of this section, if users wish to read up on it further 👍 but I do want the document to be a reasonably comprehensive reference in its own right, so that users don't _have_ to link off to other pages to get the full picture.
Things like our DNS resolution are unlikely to change often enough for maintenance to be an issue.

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,462 @@
...
+Fastly uses [Varnish] for caching, with a default cache time of 1 hour. (Read
+["Our content delivery network"][our-cdn] for more information). If Fastly
+doesn't have a page in its cache, it fetches the page from origin.
+
+Caches can be purged using the [cache-clearing-service], which tells Fastly to
+soft-purge (i.e. only remove the cached version once it has received the new
+version from origin). This cache clearing service is triggered automatically
+when pages are updated - more on that later.
+
+[cache-clearing-service]: https://github.com/alphagov/cache-clearing-service
```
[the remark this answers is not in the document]
ChrisBAshton: Will add that as a second link. I still want to link to the cache-clearing-service though. I could potentially do that later, in the `Downstream Sidekiq background processing triggered by publishing` section, or I could just have both links in this section.

source/manual/architecture-deep-dive.html.md
ChrisBAshton: That makes more sense to me. I'll need to update my slides too 👍

source/manual/architecture-deep-dive.html.md
ChrisBAshton: I've added a line about this now.

source/manual/architecture-deep-dive.html.md
ChrisBAshton: Have now rewritten this to be clearer 👍

source/manual/architecture-deep-dive.html.md
ChrisBAshton: Thanks Chris. I've updated the docs accordingly 👍

source/manual/architecture-deep-dive.html.md
ChrisBAshton: I've done some rewording/clarifying. Happy to swap out the govuk_containers module for a better example if you have one 😄 but hopefully what I have now is a general overview of how it all pieces together.

ChrisBAshton:
Thank you to everyone who has reviewed so far 🙇 I've now updated the docs to reflect code review comments. Please re-review when you get a chance - I'm hoping to get a few ✅from people before merging this in. Also happy to make some more tweaks.

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,488 @@
...
+How many machines are allocated to a class - and how big those machines are -
+is configured using [Terraform], in [govuk-aws]. What runs on each machine
+class is configured in govuk-puppet, a file and process management system we'll
+cover in more detail later.
+
+![](images/nginx-routing.png)
+
+External requests are routed to a 'cache' machine, where the request is
+received by an [Nginx] web server running on the machine. Nginx uses
+Varnish for its caching: if Varnish has the route response in its cache,
```
cbaines, line 121: The "Nginx uses Varnish for its caching" bit is misleading. NGinx can use the file system for caching, and maybe can use Varnish, but the configuration we have just reverse proxies most requests to Varnish [1].

1: https://github.com/alphagov/govuk-puppet/blob/master/modules/router/templates/router_include.conf.erb

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,488 @@
...
+that is returned, otherwise Nginx handles the route. Nginx is configured
+with govuk-puppet.
+
+Nginx proxies most requests to [router], which is a GOV.UK maintained
```
cbaines, line 125: As with the comment above, NGinx reverse proxies to Varnish, Varnish then reverse proxies to the Router.

source/manual/architecture-deep-dive.html.md
cbaines: I think Redis is still mentioned as what the Router uses.

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,488 @@
...
+[Nginx]: https://www.nginx.com/
+[router]: https://github.com/alphagov/router
+[Terraform]: https://www.terraform.io/
+
+#### Routing via 'router'
+
+'Router' is a reverse proxy app written in Go. It is designed to be fast,
+storing all known routes in memory using a [prefix trie], which it loads
+from [Redis] - an in-memory database - at startup. Redis is populated by
+reading from a MongoDB database of all known routes.
```
cbaines, line 142: I'm pretty sure it's just MongoDB that's involved. The Router reads the data from MongoDB, in to memory.

source/manual/architecture-deep-dive.html.md
```diff
@@ -0,0 +1,488 @@
...
+
+The MongoDB database is written to via [router-api] whenever a route is
+added or changed: the Content Store (which we'll cover later) talks to
+Router API [directly][register-route] to do this. Router API then
+[sends POST requests to every instance of Router][router-api-updating-router],
+whereupon each [Router then reloads all routes][router-reloading-routes]
+from the MongoDB.
```
cbaines, line 149: Either "from the MongoDB database" or "from MongoDB" would read better here.

source/manual/architecture-deep-dive.html.md
ChrisBAshton: Oh, I thought Router loaded routes into its prefix trie from Redis at startup - and that Redis was populated via Router API and the MongoDB. This always did seem unnecessarily complicated!

I see there's actually no mention of Redis anywhere in the router repository, but it is mentioned in [configuration in router-api](https://github.com/alphagov/router-api/blob/dae159a1852fa12cc3d7654af2fd53162b2bd6fa/config/cable.yml#L7-L10). I'm now not exactly sure where it fits in. It looks like that config is used by [Action Cable](https://guides.rubyonrails.org/action_cable_overview.html) for building real-time features, so Redis is potentially used as a temporary store when routes are updated at runtime? That seems like such a transitory implementation detail that it should just be omitted entirely from this overview. What do you think?

source/manual/architecture-deep-dive.html.md
ChrisBAshton: Thanks - I've now updated this.

source/manual/architecture-deep-dive.html.md
ChrisBAshton: Thanks - I've now updated and linked to that file.

source/manual/architecture-deep-dive.html.md
ChrisBAshton: Continuing the conversation in https://github.com/alphagov/govuk-developer-docs/pull/1811#discussion_r407050257.

source/manual/architecture-deep-dive.html.md
cbaines: > I see there's actually no mention of Redis anywhere in the router repository, but it is mentioned in [configuration in router-api](https://github.com/alphagov/router-api/blob/dae159a1852fa12cc3d7654af2fd53162b2bd6fa/config/cable.yml#L7-L10). I'm now not exactly sure where it fits in. It looks like that config is used by [Action Cable](https://guides.rubyonrails.org/action_cable_overview.html) for building real-time features, so Redis is potentially used as a temporary store when routes are updated at runtime? That seems like such a transitory implementation detail that it should just be omitted entirely from this overview. What do you think?

I'm pretty confident that we're not using Redis with the Router API. While it's mentioned in the Action Cable configuration, I don't think we're using Action Cable in the Router API.

ChrisBAshton: Thanks - I've now removed references to Redis.

PR #1813, opened 2019-06-18
Adds instructions for deploying branches

ChrisBAshton:
Don't think these instructions are covered elsewhere, and this
is the page that looked most promising when I was searching
the dev docs.

### DIFF
--- a/source/manual/deploying.html.md
+++ b/source/manual/deploying.html.md
@@ -56,7 +56,7 @@ Go to the [Release application](https://release.publishing.service.gov.uk) and f
 - Check what you will be deploying. If there's other people's code to deploy, ask them whether they're okay for the changes to go out.
 - Check if there's a deploy note for the application, to see if there are any special instructions or reasons not to deploy. Individual app deploy freezes are usually announced here.
 - Check in `#govuk-deploy` if someone's already in the process of deploying the app. This is particularly important for Puppet where there is a delay of 30 minutes between staging and production deployments. Deployments attributed to "Jenkins" are AWS automated deployments and can be ignored for this purpose.
-- Announce in `#govuk-deploy` of your intention to deploy, and in `#govuk-2ndline` if you anticipate your release causing any issues.
+- Announce in `#govuk-2ndline` if you anticipate your release causing any issues.
 
 ### 4. Deploy to staging
 
--- a/source/manual/deploying.html.md
+++ b/source/manual/deploying.html.md
@@ -5,7 +5,7 @@ parent: "/manual.html"
 layout: manual_layout
 section: Deployment
 important: true
-last_reviewed_on: 2019-06-27
+last_reviewed_on: 2019-07-31
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manual/deploying.html.md
```diff
@@ -17,27 +17,43 @@ If you are responding to a security incident, follow the steps to [deploy fixes
...
+## Deploying a branch
+
+It can be useful to test your branch in a non-local environment before it is merged to master. Note that you should only do this on Integration, unless you have a good reason to deploy to a different environment.
+
+Go to the `Deploy_App` job in Jenkins and click 'Build with Parameters':
+
+- `TARGET_APPLICATION` - choose the name of the repository you want to deploy
+- `DEPLOY_TASK` - usually 'deploy' is most appropriate
+- `TAG` - put the name of your branch
+- Un-check the `NOTIFY_RELEASE_APP` checkbox, as we don't want to notify Release about our branch
```
barrucadu: I think it's good to leave this checked, as otherwise release doesn't accurately reflect what the current state of things is.

ChrisBAshton: Ahh fair enough, I was worried this would 'contaminate' Release and risk people accidentally promoting branches through the environments. Will change it now.

brucebolt:
Do we need full instructions for deploying a branch?  The only difference between deploying a release and a branch is the `TAG` value will be different.

ChrisBAshton:
Deploying a release is done via the release app GOV.UK application. Deploying a branch is done via Jenkins CI.

I'm aware they're doing the same thing under the hood but don't think we have the Jenkins CI workflow documented at the moment.

AgaDufrat:
This is very straight forward once you know how to do it. I had no idea how to go about this when I had to do it for the first time. Managed to figure it out by checking #govuk-deploy channel to see who deployed a branch recently. Then looked at their branch on GitHub and recent Jenkins builds to see how they did it 😂 
I would however suggest adding a note about checking #govuk-deploy channel and announcing it there. It is mentioned here (last bullet point): https://github.com/alphagov/govuk-developer-docs/blob/master/source/manual/deploying.html.md#2-check-the-context

brucebolt:
Do we need an optional `Deploy to integration` button in the release app?  These buttons just take you to Jenkins in the relevant environment, with the form pre-filled.  This route would allow you to deploy a branch by changing the tag value.

ChrisBAshton:
I like the idea of a `Deploy to Integration` button, we could simplify these added instructions then. 👍  

But AFAIK you still have to click on a release tag before clicking "Deploy to X", which is confusing UX if you want to deploy a branch.

ChrisBAshton:
Have re-worded the Slack advice. Think in the absence of a "Deploy to integration" button we should prioritise merging this first, and if we add a button in future we can simplify the instructions as we'd have a UI where instructions aren't as necessary.

Please give a re-review 👀🙂

source/manual/deploying.html.md
```diff
@@ -17,27 +17,44 @@ If you are responding to a security incident, follow the steps to [deploy fixes
...
 
-## 3. Check your changes
+### 3. Check your changes
 
 Go to the [Release application](https://release.publishing.service.gov.uk) and find the application you want to deploy. Then select the release tag you want to deploy.
 
 - Check what you will be deploying. If there's other people's code to deploy, ask them whether they're okay for the changes to go out.
 - Check if there's a deploy note for the application, to see if there are any special instructions or reasons not to deploy. Individual app deploy freezes are usually announced here.
 - Check in `#govuk-deploy` if someone's already in the process of deploying the app. This is particularly important for Puppet where there is a delay of 30 minutes between staging and production deployments. Deployments attributed to "Jenkins" are AWS automated deployments and can be ignored for this purpose.
+- Announce in `#govuk-deploy` of your intention to deploy, and in `#govuk-2ndline` if you anticipate your release causing any issues.
```
tijmenb: I don't think we have to announce by default.
 
```suggestion
- Announce in `#govuk-2ndline` if you anticipate your release causing any issues.
```

source/manual/deploying.html.md
```diff
@@ -17,27 +17,44 @@ If you are responding to a security incident, follow the steps to [deploy fixes
 [reg]: https://gds.blog.gov.uk/2012/11/02/regular-releases-reduce-risk
 [sec]: /manual/deploy-fixes-for-a-security-vulnerability.html
 
-## 1. Test on integration
+## Deploying a branch
+
+It can be useful to test your branch in a non-local environment before it is merged to master. Note that you should only do this on Integration, unless you have a good reason to deploy to a different environment.
+
+Announce your deployment in `#govuk-deploy` as a minimum, so that people are aware a non-master branch is being built. If you anticipate your release causing problems, it's worth announcing in `#govuk-2ndline` too.
```
thomasleese: I don't think this is necessary all the time.

source/manual/deploying.html.md
```diff
@@ -17,27 +17,44 @@ If you are responding to a security incident, follow the steps to [deploy fixes
 [reg]: https://gds.blog.gov.uk/2012/11/02/regular-releases-reduce-risk
 [sec]: /manual/deploy-fixes-for-a-security-vulnerability.html
 
-## 1. Test on integration
+## Deploying a branch
+
+It can be useful to test your branch in a non-local environment before it is merged to master. Note that you should only do this on Integration, unless you have a good reason to deploy to a different environment.
```
thomasleese: It might also be worth adding something about checking whether there already is a branch deployed and asking the person who deployed it if they are finished.

thomasleese: We could also put a warning that merging to master will automatically deploy to integration anyway, and if you've got a branch deployed this could get overwritten at any time.

source/manual/deploying.html.md
ChrisBAshton: Can you elaborate? Would be good for the docs to provide clear expectations of when an announcement in #govuk-deploy is necessary.

source/manual/deploying.html.md
ChrisBAshton: Have now added a preamble before this sentence which should cover it 👍

ChrisBAshton: Have now added a warning just after the opening paragraph of this section 👍

source/manual/deploying.html.md
thomasleese: I don't think it's necessary to announce integration deploys in #govuk-deploy. Generally I think there is an expectation that integration is running new/different code compared to staging/production so developers should expect it to not necessarily be working the same as production. If there is a reason we need to keep integration on master, it'll often be made clear via an email to developers or in #govuk-2ndline (an example of this is when content designers are doing training).

I imagine there are still times when announcing to #govuk-2ndline is useful (if you anticipate that the deploy will cause problems as you've already written) but I don't think requiring developers to announce in #govuk-deploy all the time is necessary.

thomasleese: The paragraph below 

> Announce your deployment in `#govuk-2ndline` if it’s potentially problematic, or in `#govuk-deploy` if it involves deploying something other than a release tag (like a branch).

which has been removed in this commit, I think refers specifically to deploying a branch to staging/production and in that case I think it is useful to announce in #govuk-deploy.

ChrisBAshton: Cool, think that can be summed up by:

> If you are deploying to staging or production, announce your deployment in `#govuk-deploy` so that people are aware a non-master branch is being built. It is not necessary to announce if you're deploying to Integration.

Have now pushed up a new commit 👍

source/manual/deploying.html.md
```diff
@@ -17,27 +17,48 @@ If you are responding to a security incident, follow the steps to [deploy fixes
...
+
+As you'll be deploying to Integration, note that your deployment could get overwritten at any time as apps on Integration are automatically overwritten by their master branches when PRs are merged.
+
+Before deploying anything, check whether there is already a non-master branch deployed by looking at the Release application on production. (The Release app on staging and integration won't tell you, as they're synced from production overnight and aren't otherwise updated). If a non-master branch has been deployed, try and ask the person who deployed it if they're finished.
```
thomasleese: If you don't yet have production access, another way to find this out is to [look at the build history in Jenkins](https://deploy.integration.publishing.service.gov.uk/job/Deploy_App/).

ChrisBAshton: Thanks, I'll add that in 👍

PR #1818, opened 2019-06-21
Updates dependencies & fixes tests

### LINKED ISSUE #1807: Unable to update without causing YAML problems
To update a dependency to latest version, we should run `./bin/update`, but this updates _all_ dependencies to their latest versions. And one of these updates causes the YAML parsing to fail:

```
YAML Exception parsing /Users/christopherashton/govuk/govuk-developer-docs/source/manual/fall-back-to-mirror.html.md: Tried to load unspecified class: Date
YAML Exception parsing /Users/christopherashton/govuk/govuk-developer-docs/source/manual/ram-vm.html.md: Tried to load unspecified class: Date
YAML Exception parsing /Users/christopherashton/govuk/govuk-developer-docs/source/manual/fix-signature-expired-errors.html.md: Tried to load unspecified class: Date
YAML Exception parsing /Users/christopherashton/govuk/govuk-developer-docs/source/manual/frontend-architecture.html.md: Tried to load unspecified class: Date
YAML Exception parsing /Users/christopherashton/govuk/govuk-developer-docs/source/manual/receiving-emails-from-email-alert-api-in-integration-and-staging.html.md: Tried to load unspecified class: Date
YAML Exception parsing /Users/christopherashton/govuk/govuk-developer-docs/source/manual/github-trello-poster.html.md: Tried to load unspecified class: Date
YAML Exception parsing /Users/christopherashton/govuk/govuk-developer-docs/source/manual/resync-mongo-db.html.md: Tried to load unspecified class: Date
YAML Exception parsing /Users/christopherashton/govuk/govuk-developer-docs/source/manual/development-pipeline.html.md: Tried to load unspecified class: Date
```

This isn't a silent failure; it means a lot of content is lost from the Manual page:

Screen Shot 2019-06-14 at 09 06 54

We should fix up our usage of YAML parsing so that we can continue to update problem-free.

ChrisBAshton:
- Updates dependencies via `bundle update`. I had to specify a minimum version for Middleman otherwise it would default to a release candidate version for some reason, and that caused a domino effect of other issues! I also had to specify a maximum version of html-pipeline for the same reason as https://github.com/alphagov/govuk-developer-docs/commit/e9541eb4d01d1fb4b883041ca211613086618641.
- Updates tests, including fixing fixture data which contained YML content (and was referred to as YML in the tests) but was saved as a `.json` file, and ensures we stub the right URLs. `announcements` are no longer a thing, so I now test against `redirect` (wanted to retain the ability to test applications that have no dependencies, as was the case with `announcements`).
- Fixes #1807


app/document_types.rb
```diff
@@ -50,7 +50,7 @@ def url
 
     def rendering_apps
       entry = DocumentTypes.rendering_apps_from_content_store.find { |el| el[:name] == name }
-      entry.to_h[:apps].to_a.compact
+      entry.nil? ? [] : entry.to_h[:apps].to_a.compact
```
tijmenb: `nils` are already handled by the `.to_h`, so I don't think this is necessary?

```
irb(main):003:0> nil.to_h[:apps].to_a.compact
=> []
```

ChrisBAshton: Good spot, thanks! Not sure why I needed to add it in the first place as removing it seems to work fine 👍

ChrisBAshton:
Have raised https://github.com/alphagov/govuk-developer-docs/issues/1821 👍

PR #1845, opened 2019-07-08
Add guide on how to set up frontend dependencies

chao-xian:
For local development, it's a slightly convoluted process to get
things hooked up to work for local testing of the frontend apps.

# TODO

Add the same for `govuk-docker`

### DIFF
--- a/source/manual/local-frontend-development.html.md
+++ b/source/manual/local-frontend-development.html.md
@@ -1,0 +1,83 @@
+---
+title: Local frontend development
+parent: "/manual.html"
+layout: manual_layout
+section: Frontend
+type: learn
+owner_slack: "#govuk-2ndline"
+last_reviewed_on: 2019-07-08
+review_in: 6 months
+---
+
+Make the following changes to your setup if you are making changes to frontend apps and need to view those changes locally.
+
+In the instructions below we are running the [frontend] app, but this would work for any of the other frontend apps such as [government-frontend].
+
+## Running just a frontend app against live data
+
+```shell
+cd /var/govuk/frontend
+./startup.sh --live
+```
+
+## Running a frontend app against changes in `govuk_publishing_components`
+
+Update the Gemfile for the frontend app to point to your local `govuk_publishing_components` gem:
+
+```ruby
+gem 'govuk_publishing_components', path: '../govuk_publishing_components'
+```
+
+Then run your frontend app again pointing to live:
+
+```shell
+./startup.sh --live
+```
+
+## Running a frontend app against changes in `static`
+
+Make changes to the Gemfile of the frontend app you're working on as per [changes to govuk_publishing_components above](#running-a-frontend-app-against-changes-in-govuk_publishing_components), then set in `development.rb`:
+
+```ruby
+...
+config.assets.debug = false
+...
+```
+
+If you're making changes to the [govuk_publishing_components] gem that affect [static] and are likely things across the whole site, e.g: cookie banner (and survey banner possibly), update the Gemfile of [static] to point to the gem:
+
+```ruby
+...
+gem 'govuk_publishing_components', path: '../govuk_publishing_components'
+...
+```
+
+Run up static:
+
+```shell
+cd /var/govuk/static
+govuk_setenv static ./startup.sh --live
+```
+
+Run up frontend against local static (the `--live` flag means the rest runs against live data):
+
+```shell
+PLEK_SERVICE_STATIC_URI=localhost:3013 ./startup.sh --live
+```
+
+And you can see frontend running up locally with the right local dependencies:
+
+[http://frontend.dev.gov.uk/](http://frontend.dev.gov.uk/)
+
+> Note: You might sometimes find that you have something running on a port still, which stops you from starting up an app. To kill a process running on a port:
+
+```shell
+# replace [port number]
+sudo fuser -k [port number]/tcp
+```
+
+[govuk_app_config]: https://github.com/alphagov/govuk_app_config
+[frontend]: https://github.com/alphagov/frontend
+[static]: https://github.com/alphagov/static
+[govuk_publishing_components]: https://github.com/alphagov/govuk_publishing_components
+[government-frontend]: https://github.com/alphagov/government-frontend


source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,86 @@
...
+  GOVUK_DOMAINS = [
+    '*.publishing.service.gov.uk',
+    # "*.#{ENV['GOVUK_APP_DOMAIN_EXTERNAL'] || ENV['GOVUK_APP_DOMAIN'] || 'dev.gov.uk'}"
+     '*.dev.gov.uk'
+  ].uniq.freeze
+```
+
+You need to change the frontend app you are working on to use the modified [govuk_app_config]. For example in [frontend] update the Gemfile:
+
+```ruyby
```
injms:
```suggestion
```ruby
```

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,86 @@
...
+parent: "/manual.html"
+layout: manual_layout
+section: Frontend
+type: learn
+owner_slack: "#govuk-2ndline"
+last_reviewed_on: 2019-07-08
+review_in: 6 months
+---
+
+Make the following changes to your local setup:
```
vanitabarrett: This is only relevant if you're trying to run static locally too. You shouldn't need to do this if you just want to run a frontend app (e.g: government-frontend) pointing at live static.

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,86 @@
...
+
+In [govuk_app_config], add your local domain to the allowed CORS domains:
+
+```
+  GOVUK_DOMAINS = [
+    '*.publishing.service.gov.uk',
+    # "*.#{ENV['GOVUK_APP_DOMAIN_EXTERNAL'] || ENV['GOVUK_APP_DOMAIN'] || 'dev.gov.uk'}"
```
vanitabarrett: I think it would be clearer to remove this line, as it's not immediately obvious that it's being commented out here?

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,86 @@
...
+     '*.dev.gov.uk'
+  ].uniq.freeze
+```
+
+You need to change the frontend app you are working on to use the modified [govuk_app_config]. For example in [frontend] update the Gemfile:
```
vanitabarrett: Again, this is only relevant when trying to run a frontend app against local static. Perhaps we could split this guide into sections? E.g:

1. Running a frontend app against live: `./startup.sh --live`
2. Running a frontend app against local changes in govuk_publishing_components
3. Running a frontend app against local static

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,86 @@
...
+
+```ruyby
+...
+gem 'govuk_app_config', path: '/var/govuk/govuk_app_config'
+...
+gem 'govuk_publishing_components', path: '/var/govuk/govuk_publishing_components'
```
vanitabarrett: See comment above about splitting this guide into sections. This change may be confusing, as it's not related to the `govuk_app_config` change.

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,86 @@
...
+...
+```
+
+and set in `development.rb`:
```
vanitabarrett: Technically, this is not __required__ to run a frontend app locally. But in practice, running without this tends to be so slow that it times out a lot. I think we need to make that clear.

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,86 @@
...
+
+[http://frontend.dev.gov.uk/](http://frontend.dev.gov.uk/)
+
+> Note: You might sometimes find that you have something running on a port still, which stops you from starting up an app. To kill a process running on a port:
+
+```shell
+$dev sudo fuser -k 3013/tcp
+```
+
+To do the same to run up [government-frontend]
```
vanitabarrett: I don't think we need to mention `government-frontend` as well. It's the same for all frontend apps, so we just need to make that clear.

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,86 @@
...
+
+and set in `development.rb`:
+
+```ruby
+...
+config.assets.debug = false
+...
+```
+
+Update the Gemfile in [static] to point to [govuk_publishing_components] gem:
```
vanitabarrett: This is only relevant if the change you are making to the gem affects static. Most changes shouldn't as they just get pulled in by the frontend app you're running, so it would be enough for most people to update the gem in the frontend app only. It's likely to only be things that we have across the whole site, e.g: cookie banner (and survey banner possibly).

This should be helped if we split the guide up into sections as mentioned in another comment.

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,86 @@
...
+
+```ruby
+...
+gem 'govuk_publishing_components', path: '/var/govuk/govuk_publishing_components'
```
vanitabarrett: Did you manage to get this working? I seem to remember we tried this and it only worked with a relative path, e.g: `../govuk_publishing_components`

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+And you can see frontend running up locally with the right local dependencies:
+
+[http://frontend.dev.gov.uk/](http://frontend.dev.gov.uk/)
+
+> Note: You might sometimes find that you have something running on a port still, which stops you from starting up an app. To kill a process running on a port:
+
+```shell
+$dev sudo fuser -k 3013/tcp
```
vanitabarrett: We should make clear that `3013` needs to be replaced here with the port you are trying to kill.

tijmenb:
Hey, this doesn't seem right. There's a bunch of things we really shouldn't have to do, let's catch up about this offline!

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+Run up static in the vm:
+
+```shell
+$dev cd /var/govuk/static
+$dev govuk_setenv static ./startup.sh --live
+```
+
+Run up frontend in the vm:
```
vanitabarrett: Using 'frontend' as an example could be confusing for new starters who may assume we only have one frontend app called Frontend. Probably worth making clear that this is an example, but the instructions should apply to other frontend apps too (I think there may be a list of Frontend apps somewhere)

source/manual/local-frontend-development.html.md
vanitabarrett: Also, we should probably make clear this is only a temporary workaround. This should be fixed within `govuk_app_config` really.

chao-xian: Fixed in https://github.com/alphagov/govuk_app_config/pull/100

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,84 @@
...
+section: Frontend
+type: learn
+owner_slack: "#govuk-2ndline"
+last_reviewed_on: 2019-07-08
+review_in: 6 months
+---
+
+Make the following changes to your setup if you are making changes to frontend apps and need to view those changes locally.
+
+In the instructions below we are running the [frontend] app, but would work for any of the other frontend apps such as [government-frontend].
```
vanitabarrett:
```suggestion
In the instructions below we are running the [frontend] app, but this would work for any of the other frontend apps such as [government-frontend].
```

vanitabarrett:
```suggestion
In the instructions below we are running the [frontend] app, but this would work for any of the other frontend apps such as [government-frontend].
```

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Running just a frontend app against live data:
```
vanitabarrett:
```suggestion
## Running just a frontend app against live data
```

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,84 @@
...
+
+```shell
+$dev cd /var/govuk/frontend
```
vanitabarrett: I don't think we need the `$dev`? Someone could technically run this outside the VM, it's not specific.

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,84 @@
...
+
+Then run your frontend app again pointing to live:
+
+```shell
+$dev ./startup.sh --live
+```
+
+## Running a frontend app against changes in `static`
+
+This is only relevant if the change you are making to the [govuk_publishing_components] gem affects [static]. It's likely to only be things that we have across the whole site, e.g: cookie banner (and survey banner possibly).
```
vanitabarrett:
```suggestion
This is only relevant if the change you are making to the [govuk_publishing_components] gem affects [static]. It's likely to only be things that we have across the whole site, e.g: cookie banner
```

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,84 @@
...
+
+Make changes to the Gemfile of the frontend app as per above, then set in `development.rb`:
```
vanitabarrett: It's not clear here what changes you are referring to..? Do you mean the `govuk_publishing_components` change? Can we add a link to the relevant heading?

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,84 @@
...
+```ruby
+...
+gem 'govuk_publishing_components', path: '../govuk_publishing_components'
+...
+```
+
+Run up static in the vm:
+
+```shell
+$dev cd /var/govuk/static
```
vanitabarrett: Again, don't think we need `$dev`

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,84 @@
...
+$dev govuk_setenv static ./startup.sh --live
+```
+
+Run up frontend in the vm:
+```shell
+$dev PLEK_SERVICE_STATIC_URI=localhost:3013 ./startup.sh --live
```
vanitabarrett: May be worth explaining in the line above that this runs everything as live except for static.

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+Run up static in the vm:
+
+```shell
+$dev cd /var/govuk/static
+$dev govuk_setenv static ./startup.sh --live
+```
+
+Run up frontend in the vm:
```
vanitabarrett: Not necessarily "in the VM"

source/manual/local-frontend-development.html.md
```diff
@@ -0,0 +1,84 @@
...
+
+Update the Gemfile in [static] to point to [govuk_publishing_components] gem:
+
+```ruby
+...
+gem 'govuk_publishing_components', path: '../govuk_publishing_components'
+...
+```
+
+Run up static in the vm:
```
vanitabarrett: Not necessarily "in the VM"

chao-xian:
[(1) Write up intro to frontend](https://trello.com/c/JpkeIZNJ/6-1-write-up-intro-to-frontend)

PR #1847, opened 2019-07-09
Add link to Mainstream Publisher documentation

ChrisBAshton:


source/manual/howto-upload-an-asset-to-asset-manager.html.md
```diff
@@ -8,7 +8,7 @@ last_reviewed_on: 2019-05-29
 review_in: 6 months
 ---
 
-Some publishing apps such as Mainstream Publisher do not provide the facility for editors to upload
+Some publishing apps such as [Mainstream Publisher](https://docs.publishing.service.gov.uk/apps/publisher.html) do not provide the facility for editors to upload
```
tijmenb: Oh, a relative link would be better here I think.

PR #1856, opened 2019-07-12
Add page about removing a URL redirect from Transition

MuriloDalRi:
[trello](https://trello.com/c/RVP3vVZ3/1113-remove-redirected-urls-from-transition-tool)


jennyd:
[`rake import:revert_entirely_unsafe[site_abbr]`](https://github.com/alphagov/transition/blob/4a13421139095842364be0f7fa597bea55b559fc/lib/tasks/import/revert_entirely_unsafe.rake) does this - it asks for confirmation before deleting a site and all data associated with it.

PR #1857, opened 2019-07-12
Change finder-frontend ownership to Search and Nav

stevenjmesser:
Ran this past Simon and Bevan but we feel it's only fair that Search & Nav look after finder-frontend, since we're the ones tinkering with it.

There's probably other PRs for me to raise off the back of this. Seal? Anywhere else?

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -471,7 +471,7 @@
       description: ''
 - github_repo_name: finder-frontend
   type: Frontend apps
-  team: "#govuk-platform-health"
+  team: "#govuk-searchandnav"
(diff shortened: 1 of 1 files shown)


barrucadu:
The seal doesn't appear to know about app ownership: https://github.com/alphagov/seal/blob/master/config/alphagov.yml
I guess it gets that from this file?

stevenjmesser:
Phew!

tijmenb:
@barrucadu Seal only knows about people, not about apps & repos. Any pull request from a team member will show up on the team's seakl.

PR #1859, opened 2019-07-12
Add Fix For Issue 1827

### LINKED ISSUE #1827: Broken SSH command on several applications
See https://docs.publishing.service.gov.uk/apps/smart-answers.html

> ssh -A -t jumpbox.integration.publishing.service.gov.uk 'ssh `govuk_node_list -c Unknown - have you configured and merged your app in govuk-puppet/hieradata_aws/common.yaml --single-node`'

^ This doesn't look right!

See also https://docs.publishing.service.gov.uk/apps/transition.html#ssh-access-aws

`smartanswers` has node classes on `calculators_frontend` and `draft_frontend`, so something should appear in this command. I have a hunch the problem is down to [hieradata_aws/common.yaml](https://raw.githubusercontent.com/alphagov/govuk-puppet/master/hieradata_aws/common.yaml) naming `smartanswers` rather than `smart-answers` in places...?

Either way, we should probably avoid rendering broken SSH commands in cases like these, before fixing the smartanswers problem specifically.

camdesgov:
Some of the SSH commands in the Developer Documentation are broken.
The common theme seems to be that they all have a different puppet_name
and app_name. The fix ensures the code checks for both variables before
rendering an error.

More information on this issue can be found here:
https://github.com/alphagov/govuk-developer-docs/issues/1827

Fixes #1827

### DIFF
--- a/app/app_docs.rb
+++ b/app/app_docs.rb
@@ -55,7 +55,7 @@ class AppDocs
 
     def aws_puppet_class
       AppDocs.aws_machines.each do |puppet_class, keys|
-        if keys["apps"].include?(app_name)
+        if keys["apps"].include?(app_name) || keys["apps"].include?(puppet_name)
           return puppet_class
         end
       end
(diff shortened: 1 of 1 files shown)


issyl0:
GitHub tip: If you put "Fixes #issuenumber" in the PR text, it'll close the linked issue when this gets merged. :muscle:

camdesgov:
Good tip! I've added this. Thanks Issy!

PR #1861, opened 2019-07-15
Add caching detail

bilbof:
This addition relates to the following PR:
https://github.com/alphagov/govuk-cdn-config/pull/160

### DIFF
--- a/source/manual/cdn.html.md
+++ b/source/manual/cdn.html.md
@@ -52,7 +52,16 @@ We set a default TTL of 5000s on cached objects. This means that pages such as t
 
 We also set a grace period of 24 hours. So if the homepage server is down, we'll continue to serve a stale homepage for 24 hours.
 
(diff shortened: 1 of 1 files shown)


source/manual/cdn.html.md
```diff
@@ -52,6 +52,10 @@ We set a default TTL of 5000s on cached objects. This means that pages such as t
 
 We also set a grace period of 24 hours. So if the homepage server is down, we'll continue to serve a stale homepage for 24 hours.
 
+We will cache any request that returns a 404 or 405 status for the default TTL.
```
ChrisBAshton: Isn't this "We will cache any non-GET request"?" (https://github.com/alphagov/govuk-cdn-config/pull/160/files#diff-c789eaeb1ab164ae6d3fb552b5b855d7R381)

PR #1878, opened 2019-07-22
Remove outstanding-security-updates manual

benthorner:
This seems to be covered by the 'rebooting-machines' manual.

Reference: https://github.com/alphagov/govuk-puppet/pull/9390


benthorner:
@thomasleese I've amended the generic reboot manual a little, as discussed.

PR #1879, opened 2019-07-22
Review outstanding docs

thomasleese:
This reviews all the documentation pages which are out of date. See individual commits for more details.

I've staggered the review dates around to prevent these all coming back for review again on the same day.


source/manual/sidekiq.html.md
```diff
@@ -0,0 +1,34 @@
+---
+owner_slack: "#govuk-2ndline"
```
benthorner: Should we merge the 'monitor' manual in here too?

source/manual/howto-switch-off-app.html.md
```diff
@@ -4,7 +4,7 @@ title: Switch an app off temporarily
 layout: manual_layout
```
benthorner: Is this one actually useful?

source/manual/which-gem-to-use.html.md
```diff
@@ -1,25 +1,28 @@
 ---
 owner_slack: "#govuk-developers"
```
benthorner: Is this one useful? I didn't actually know it exists, so it seems more hopeful than practical ;-).

source/manual/howto-switch-off-app.html.md
thomasleese: I think it's easy to forget to disable puppet, but whether or not people would actually look at this page, I'm not sure.

source/manual/sidekiq.html.md
thomasleese: Good idea, I've pushed up a commit that does that.

source/manual/which-gem-to-use.html.md
thomasleese: I'm not sure if we want to remove this page as it might useful for beginners, but I think we could increase the review time to at least a year so it doesn't come up as often.

source/manual/howto-switch-off-app.html.md
benthorner: Yeah, that's why I'm wondering if we really need it.

source/manual/sidekiq.html.md
benthorner: Great - thanks :-).

PR #1881, opened 2019-07-23
Add developer docs for related links

karlbaker02:
This PR adds developer docs for related links - specifically the new process which uses machine learning to generate links. This documentation details how the related links are generated, how they're ingested, the rules that determine when they're shown, and what to do if there's a problem with the generated links.

Trello: https://trello.com/c/ivIVYQVT

### DIFF
--- a/source/manual/related-links-on-govuk.html.md
+++ b/source/manual/related-links-on-govuk.html.md
@@ -1,0 +1,69 @@
+---
+title: Related links on GOV.UK
+parent: "/manual.html"
+layout: manual_layout
+type: learn
+section: Related links
+owner_slack: "#govuk-developers"
+last_reviewed_on: 2019-07-22
+review_in: 6 months
+---
+
+## Overview
+
+Since July 2019, related links on GOV.UK have been supplemented by links generated through an automated machine learning pipeline. The first job in this pipeline uses data on how pages are linked together (the structural network) and analytics from users' behaviour on the site (the functional network) to create a set of most relevant related links for a subset of pages.
+
+Once these links have been created, a second job ingests the links via the Publishing API to be shown to users navigating the site.
+
+The code for the process of generating and ingesting related links can be found in the [govuk-related-links-recommender repository on GitHub](https://github.com/alphagov/govuk-related-links-recommender).
+
+## When we show the new related links
+
+Machine learning generated related links are created exclusively for Whitehall content, with a number of measures in place to ensure that we always show manually curated links set by publishers.
+
+- Related links are only generated for a number of content types, with exclusions in place for content that should not be linked _from_, linked _to_, or both. These exclusions can be found at https://github.com/alphagov/govuk-related-links-recommender/tree/master/src/config.
+- Generated related links are stored within the `suggested_ordered_related_items` property of a content item, while all existing and manually curated links continue to be stored within the `ordered_related_items` property.
+- If links exist with `ordered_related_items` for a content item, we do not show generated related links.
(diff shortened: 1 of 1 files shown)


source/manual/related-links-on-govuk.md
```diff
@@ -0,0 +1,72 @@
...
+parent: "/manual.html"
+layout: manual_layout
+type: learn
+section: Related links
+owner_slack: "#govuk-developers"
+last_reviewed_on: 2019-07-22
+review_in: 6 months
+---
+
+This is technical documentation for the [GOV.UK](https://www.gov.uk/) team in the [Government
```
barrucadu: Everything in this repository is tech docs for GOV.UK

source/manual/related-links-on-govuk.md
```diff
@@ -0,0 +1,72 @@
...
+
+Currently, this job is scheduled to run once every three weeks via a `time` resource attached to the production `run-generation-production` step.
+
+### `link-ingestion`
+
+This job runs the process of ingesting related links via the Publishing API.
+
+Links are retrieved from the environment-specific S3 bucket mentioned previously and then fed (in batches) to the Publishing API. There is a 20 minute delay between batches.
+
+Currently, this job is triggered manually after reviewing a spreadsheet of the generated related links; it will be scheduled to run automatically in the near future.
```
barrucadu: Can we nail down "in the near future" a bit more precisely?  Statements like that tend to hang around.

source/manual/related-links-on-govuk.md
```diff
@@ -0,0 +1,72 @@
...
+  - `MACHINE_CLASS: publishing_api`
+  - `RAKE_TASK: content:reset_related_links_for_pages['CONTENT_ID’]`, where `CONTENT_ID` is the content id of the page to remove suggested links from
+
+### Links for certain document types
+
+To exclude a document type or page from having links generated _to_ it, _from_ it, or both, [update the exclusions in the GitHub repository](https://github.com/alphagov/govuk-related-links-recommender/tree/master/src/config). Note: this will take effect from the _next_ time the related links generation process is run, and will not affect current links.
+
+### Suspending all suggested related links
+
+To immediately and temporarily stop showing all suggested related links (for example, if a problem has led to bad links being generated for a large percentage of content / high traffic content), you will need to update the CDN to set the `Govuk-Use-Recommended-Related-Links` header to `False` and then [deploy the CDN](https://deploy.publishing.service.gov.uk/job/Deploy_CDN/) for `vhost: www`.
```
barrucadu: Will we also need to invalidate the cache, or is the value of the `Govuk-Use-Recommended-Related-Links` header taken into account in the caching?

source/manual/related-links-on-govuk.md
karlbaker02: Fair. I was using another doc as a template and it had this, which means it is also incorrect. I'll remove 👍

source/manual/related-links-on-govuk.md
karlbaker02: One of our missions at the moment is to monitor related links, so by the end of September; I'll add that in.

source/manual/related-links-on-govuk.md
karlbaker02: We set the `Vary` header to use `Govuk-Use-Recommended-Related-Links`, so we will get back different results from origin when we change the header. I was under the belief that if the CDN doesn't have a request with a particular header set when it receives a request, it will forward it on to origin - is there anywhere that we control this more finely?

PR #1895, opened 2019-07-30
Revert "Remove "Find hardcoded markup in GovSpeak""

chao-xian:
Reverts alphagov/govuk-developer-docs#1461

I have a need, and also @vanitabarrett said this would have been useful a couple of weeks ago for the YouTube embed work she recently did.


vanitabarrett:
This may need some updating. For example: https://govuk-static.herokuapp.com/component-guide/govspeak no longer works.

PR #1902, opened 2019-08-01
Iterate through each file all at once rather than once per commit

v1v:
This is a proposal to run with `fail-at-the-end` rather than `fail-fast` approach.


thomasleese:
Thanks for your contribution!

PR #1903, opened 2019-08-01
Use pre-commit within the pipeline

v1v:
https://pre-commit.com/#usage-in-continuous-integration provides a way to run the same command in the CI therefore it's not required a different approach to run the same script but using the pre-commit wrapper itself.

Besides, this implementation will only monitor the changes for that particular git commit. In other words, it does not require to evaluate all the files every time but only the ones which have changed in the PR itself.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,7 +40,10 @@ node {
     }
 
     stage("Lint documentation") {
-      sh "vale --glob='*.md' ."
+      sh '''
+        curl https://pre-commit.com/install-local.py | python -
+        git diff-tree --no-commit-id --name-only -r $(git rev-parse HEAD) | xargs pre-commit run --files
+      '''
     }
 
     stage("Tests") {


thomasleese:
Thanks for your contribution and clear commit message!

Generally, our linting approach on GOV.UK is that all the files get linted on each build. We've had problems in the past where an upgrade to the linter has changed the results for all our files, and we only see the problems come up in subsequent unrelated PRs which takes up time to fix.

I'm also not sure about having to install `pre-commit` on each build as that will slow it down, and there might be a better way of doing this by installing it on each of our CI machines. On GOV.UK [we would use Puppet to do this](https://github.com/alphagov/govuk-puppet).

I think for these reasons I'm going to close this PR for now as I don't think we're likely to accept it, but thanks again for taking the time to make this contribution.

v1v:
No problem at all, some feedback for the first reason it will work as simple as using:

```
pre-commit run --all-files
```

Regarding the pre-commit installation doesn't take longer more than 10 seconds, although in my case, there are genuine reasons to use the same homogeneous approach locally and in the CI. I thought it could be a similar case for this repo though.

Cheers

PR #1927, opened 2019-08-12
Docs are fine. Update review date

deanwilson:


cbaines:
On a separate note, #govuk-2ndline has switched to using the PagerDuty integration for Slack, rather than going via Hubot, so I'm not sure what Hubot is being used for now (if anything).

tijmenb:
Hubot is still used for the Seal, Dependapanda, and a (perhaps no longer working) feature to ask the docs who owns what app:

https://github.com/alphagov/gds-hubot/tree/master/scripts

PR #1940, opened 2019-08-16
Add documentation about what to do with data sync Icinga alerts

oscarwyatt:
The alert provides little information on what to do, this may help

### DIFF
--- a/source/manual/alerts/data-sync.html.md
+++ b/source/manual/alerts/data-sync.html.md
@@ -24,6 +24,8 @@ Check the output of the production Jenkins job to see which part of the sync fai
 > to back to their staging equivalents.  This means that any deploys to staging
 > that rely on GDS API Adapters are likely to fail due to authentication
 > failures, as well as Smokey tests that attempt to use Signon.
+>
+>  Running the [data_sync_complete](https://deploy.staging.publishing.service.gov.uk/job/Data_Sync_Complete/) job may fix this.
 
 The Jenkins jobs included in the sync are:
 
--- a/source/manual/govuk-env-sync.html.md
+++ b/source/manual/govuk-env-sync.html.md
@@ -120,3 +120,5 @@ Wherein we are
 
 ### Icinga checks
 For every `govuk_env_sync::task:`, a passive Icinga alert `GOV.UK environment sync <title>` is created. They are updated on exit of the sync script.
+
+If you get an Icinga alert about a failing task, investigate whether there were any new problems as it does occasionally fail. If not, it can be acked for 24 hours and left to run overnight again.


source/manual/govuk-env-sync.html.md
```diff
@@ -120,3 +120,7 @@ Wherein we are
 
 ### Icinga checks
 For every `govuk_env_sync::task:`, a passive Icinga alert `GOV.UK environment sync <title>` is created. They are updated on exit of the sync script.
+
+If you get an Icinga alert about a failing job, investigate whether there were any new problems as it does occasionally fail. If not, it can be acked for 24 hours and left to run overnight again.
+
+A failure may mean that links to apps in Signon in staging link to production apps. Running the [data_sync_complete](https://deploy.staging.publishing.service.gov.uk/job/Data_Sync_Complete/) job may fix this.
```
thomasleese: This line is probably better suited to this page https://docs.publishing.service.gov.uk/manual/alerts/data-sync.html because it's not related to the new `govuk_env_sync` system.

oscarwyatt: Ah right, I've updated that now

source/manual/alerts/data-sync.html.md
```diff
@@ -25,6 +25,8 @@ Check the output of the production Jenkins job to see which part of the sync fai
 > that rely on GDS API Adapters are likely to fail due to authentication
 > failures, as well as Smokey tests that attempt to use Signon.
 
+A failure may mean that links to apps in Signon in staging link to production apps. Running the [data_sync_complete](https://deploy.staging.publishing.service.gov.uk/job/Data_Sync_Complete/) job may fix this.
```
cbaines: This line seems to contain the same information as the above warning.

If you do want to make a change here, it looks like it would be worth consolidating the information.

oscarwyatt: Excellent point, hows this?

source/manual/govuk-env-sync.html.md
```diff
@@ -120,3 +120,5 @@ Wherein we are
 
 ### Icinga checks
 For every `govuk_env_sync::task:`, a passive Icinga alert `GOV.UK environment sync <title>` is created. They are updated on exit of the sync script.
+
+If you get an Icinga alert about a failing job, investigate whether there were any new problems as it does occasionally fail. If not, it can be acked for 24 hours and left to run overnight again.
```
cbaines: I'd perhaps say `task` here, rather than `job`, firstly because that's the language the `govuk_env_sync` module uses, but also to avoid confusion with the data sync Jenkins jobs.

oscarwyatt: 👍

source/manual/alerts/data-sync.html.md
cbaines: That looks fine now :+1:

PR #1945, opened 2019-08-19
Add instructions for responding to 'withdrawn content' requests

ChrisBAshton:


source/manual/data-gov-uk-2nd-line.html.md
```diff
@@ -92,6 +92,17 @@ This is a generic response for such cases:
...
 > I hope that helps. I’ll close this ticket now.
 
+### Accessing withdrawn content
+Sometimes we get queries from users asking for data that is no longer publicly available, e.g. a withdrawn dataset.
+
+#### If the withdrawn content was a dataset
+Ask the user to email the organisation that originally published the dataset. You can find an email address for the organisation on CKAN.
+
+#### If the withdrawn content was a publication
+If the content was published under http://data.gov.uk/library, this was part of the former Drupal-based data.gov.uk publishing application, which in the process of moving to the new CKAN is no longer being hosted on DGU.
```
brucebolt:
```suggestion
If the content was published under http://data.gov.uk/library, this was part of the former Drupal-based data.gov.uk publishing application, which is no longer being hosted on DGU.
```
I'd recommend avoiding using words like "new" or "old" when referring to CKAN, since it isn't a clear way of describing the system (e.g. if we upgrade to CKAN 2.8, then the current 2.7.4 instance will also become the old CKAN).

ChrisBAshton: 👍

brucebolt:
Could you squash the two commits into one (since the second commit is solely updating a change in the first) then I'll approve.

PR #1946, opened 2019-08-19
Update documentation on how to remove duplicate search results

MahmudH:
Search admin removes the result temporarily so a permanent solution is added to the docs

Trello: https://trello.com/c/DauneS2z

### DIFF
--- a/source/manual/rename-a-country.html.md
+++ b/source/manual/rename-a-country.html.md
@@ -126,9 +126,6 @@ The country's subscription list(s) `title` and `slug` needs to be updated, such 
 
 ### 6. Remove duplicate search results
 
-If the old country slug is still appearing in the search results, this can be removed through the [Search Admin](https://search-admin.integration.publishing.service.gov.uk/results) tool:
-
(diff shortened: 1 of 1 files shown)


cbaines:
Could you rephrase the first line of the commit message to say what the commit does:

"Removing duplicate search results updated"

as at the moment, it's a little difficult to guess what the commit is doing from reading the commit message.

source/manual/rename-a-country.html.md
```diff
@@ -130,5 +130,6 @@ If the old country slug is still appearing in the search results, this can be re
 
 1. Navigate to the `Search results` page and enter the path to be removed
 2. Remove result, it should now disappear from the listing.
```
cbaines: Unless a specific reason for keeping the documentation regarding using the Search Admin is around, I'd suggest not mentioning it, and just mentioning the Whitehall approach which you're adding in.

MahmudH: I'm happy with that, I'll update it now 👍

PR #1948, opened 2019-08-20
Document the new subscriptions count rake task

ChrisBAshton:
This was added in https://github.com/alphagov/email-alert-api/pull/887


thomasleese:
I reckon we can probably just update this page https://docs.publishing.service.gov.uk/manual/email-alert-api-analytics.html#how-many-subscribers-does-a-list-have with the new Rake task rather than creating a new card.

ChrisBAshton:
Good shout 👍

PR #1950, opened 2019-08-20
Clarify: base path rake task needs additional action to publish

ChrisBAshton:


source/manual/change-base-path-in-specialist-publisher.html.md
```diff
@@ -14,6 +14,7 @@ You can use the Jenkins rake task runner to run the task:
 
 [`base_path:edit[content_id,/new_base_path]`](https://deploy.publishing.service.gov.uk/job/run-rake-task/parambuild/?TARGET_APPLICATION=specialist-publisher&MACHINE_CLASS=backend&RAKE_TASK=base_path:edit[content_id,/new_base_path])
 
-> **Note**
->
+**NOTE**: the rake task will create a draft with the new slug, so requires you to hit 'Publish' in the Specialist
```
thomasleese: Small thing, I don't think it creates one specifically, I think it just edits the latest draft (and only creates one if necessary).

ChrisBAshton:
That would be nice! Perhaps one for the next 2nd line shift 🙂

PR #1952, opened 2019-08-22
Update documentation on falling back to mirrors

brucebolt:
Having needed to update a file on the mirror machines, I've made some changes to the documentation to (hopefully) make this process clearer, add a missing step (update Google mirror) and reduce duplication on the page.

### DIFF
--- a/source/manual/fall-back-to-mirror.html.md
+++ b/source/manual/fall-back-to-mirror.html.md
@@ -4,7 +4,7 @@ section: Deployment
 layout: manual_layout
 parent: "/manual.html"
 owner_slack: "#govuk-2ndline"
-last_reviewed_on: 2019-06-28
+last_reviewed_on: 2019-08-22
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manual/fall-back-to-mirror.html.md
```diff
@@ -94,9 +91,9 @@ HTML, it's hard to make broad changes to the site (like putting a banner on ever
 
 You'll be notified by the escalation on-call contact that you need to edit the site.
 
-1. If you're at home, connect to the [VPN][gds-vpn]
+1. If you're at home, connect to the [VPN][gds-vpn].
 
-2. ssh the Mirrorer machine by running:
+1. ssh the Mirrorer machine by running:
```
sihugh: > ssh _to_ the

?

brucebolt: Good spot :+1:

source/manual/fall-back-to-mirror.html.md
```diff
@@ -94,9 +91,9 @@ HTML, it's hard to make broad changes to the site (like putting a banner on ever
 
 You'll be notified by the escalation on-call contact that you need to edit the site.
 
-1. If you're at home, connect to the [VPN][gds-vpn]
+1. If you're at home, connect to the [VPN][gds-vpn].
```
Tetrino: These are all numbered 1, deliberate?

sihugh: Yep, they'll get automagically worked out

brucebolt: Yes, it's a feature in markdown where you number each item as `1.` then it gets numbered correctly when rendered.  Reduces the likelihood of the list getting out of order when items are added/removed.

PR #1973, opened 2019-09-03
Fix formatting of code blocks in Carrenza VPN docs.

sengi:
GOV.UK developer docs handles Markdown code blocks subtly differently to the Github preview, which led to my not noticing that I'd broken the formatting here previously.

Inserting a blank line before the code block fixes the problem.


sengi:
Sorry, realised I'd broken the list numbering. Should be fixed now (by indenting the code blocks so that they don't start a new list every time).

Not sure what the local etiquette is for post-review changes but by all means take another look if you like. Thanks for the quick review!

PR #1981, opened 2019-09-05
Create documentation for local replication of data with Docker

Tetrino:
We developed a means to replicate data easily into a Docker container
for use with Whitehall, which is especially beneficial due to the
weight of the tool and amount of data present.

Before this commit there was no documentation on the process, and thus
this commit adds both this documentation and a relevant link from the
existing data replication page.

### DIFF
--- a/source/manual/replicate-app-data-locally.html.md
+++ b/source/manual/replicate-app-data-locally.html.md
@@ -8,6 +8,10 @@ last_reviewed_on: 2019-04-25
 review_in: 6 months
 ---
 
+> If you're using Docker, you may wish to consider [this
+ guide](/manual/replicating-whitehall-data-with-docker.html) if importing
+ Whitehall data.
+
 Dumps are generated from production data in the early hours each day, and can
 then be downloaded.  You'll use the
 [replicate-data-local.sh script](https://github.com/alphagov/govuk-puppet/blob/master/development-vm/replication/replicate-data-local.sh)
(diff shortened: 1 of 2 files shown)


kevindew:
Should we look at linking directly to the [govuk-docker readme](https://github.com/alphagov/govuk-docker#how-to-replicate-data-locally) where this is documented? It looks like we've got the same task documented in two places with slightly different processes which may increase confusion for anyone new.

If we need this doc in both govuk-developer-docs and govuk-docker repo then we may want to export an external file directly into govuk-developer-docs (as per https://github.com/alphagov/govuk-developer-docs/blob/aba17b055854dd4f3f7f8fa02b70e0ab50a85632/source/manual/email-alert-api-analytics.html.md.erb#L10) and do some splitting up of files in govuk-docker docs. Though I'd suggest trying to keep the govuk-docker docs brief in govuk-developer-docs and just link to topics (as per https://docs.publishing.service.gov.uk/manual/publishing-e2e-tests.html) that'd be the lowest risk of fragmentation/divergence.

Tetrino:
@kevindew When I originally wrote it I was linking it directly to the existing data replication guidance because that is where many of the new starters are pointed towards and in my original version, there wasn't originally any guidance on how to run the app - I felt that was best left in the readme.

With the mutation Kelvin added, it does muddy the waters a bit, and I am concerned we could overload this document.

Tetrino:
I think we should remove the running steps as they are adequately covered in the govuk-docker readme, keep the location where it is due to its links to the setup documentation in the rest of the docks, and link to this document in the readme. Unsure if that is sensible?

chao-xian:
@kevindew I think we should have at least a signpost from the dev docs as the main entry point for devs, and make it easier for new starters. I wasn't even aware of that addition to the Docker README

kevindew:
Sounds good 👍

PR #1989, opened 2019-09-11
Remove redundant manuals

benthorner:
This is a suggestion to remove a couple of manuals, one of which is pending review. Please see the commits for a suggestion of why we don't need them - unless we do!


kevindew:
If I understand correctly this will mean there are no references to end-to-end tests in the developer docs. This doesn't seem ideal since it's something that runs on so many apps and is frequently a question point for developers.

I don't think we need the level of specific detail that we have in the documentation here but I think it'd be useful to briefly summarise the tests purpose and role with links direct to e2e documentation for help.

kevindew:
I wrote a quick pass at what the e2e content could be:

> Most GOV.UK applications are run against a [end-to-end test suite](https://github.com/alphagov/publishing-e2e-tests) as part of their build process. This is done to determine if a change in one application has a negative impact on a dependant application and to test the changed application is functioning correctly in a quasi-production environment.
> 
> There is specific documentation available on:
> - [what belongs in the test suite](https://github.com/alphagov/publishing-e2e-tests/blob/master/docs/what-belongs-in-these-tests.md)
> - [how to debug a failing test in jenkins](https://github.com/alphagov/publishing-e2e-tests/blob/master/docs/debugging-failures.md)
> - [how to resolve breaking changes](https://github.com/alphagov/publishing-e2e-tests/blob/master/docs/breaking-app-change.md)
> - [how to handle intermittently failing (flakey) tests](https://github.com/alphagov/publishing-e2e-tests/blob/master/CONTRIBUTING.md#dealing-with-flaky-tests)
> - [how to add new tests/applications](https://github.com/alphagov/publishing-e2e-tests/blob/master/CONTRIBUTING.md)

benthorner:
@kevindew thanks for the suggestion. I've pushed another commit with that content. Do you think we're good to merge now?

I notice that smokey doesn't have any presence either, and was wondering if longterm we should instead represent both as another 'app' group?

benthorner:
@kevindew agreed - they're not 'applications'. It's just they're all listed in a `applications.yaml` file!

PR #1992, opened 2019-09-13
New documentation in regards to errors seen in Integration

Tetrino:
We saw some odd errors in Integration that had no documentation that really applied to
the problem at hand. After RE finding solutions for both, I wrote these pages to shed
light on them in the rare case that either happen again.

### DIFF
--- a/source/manual/alerts/smokey-json-older-than-30m.html.md
+++ b/source/manual/alerts/smokey-json-older-than-30m.html.md
@@ -1,0 +1,49 @@
+---
+owner_slack: "#govuk-2ndline"
+title: smokey.json older than 30m
+section: Development VM
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2019-09-12
+review_in: 6 months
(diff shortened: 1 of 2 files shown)


source/manual/alerts/smokey-json-older-than-30m.html.md
```diff
@@ -0,0 +1,38 @@
...
+
+There can be the rare case wherein Smokey has failed to correctly run. While the
+causes can be various and ultimately unknown, the symptoms include a large number
+of Unknown errors in Icinga on the `monitoring` box.
+
+The easiest way to clear these issues is ensure all relevant Smokey tasks are
+killed:
+
+```shell
+#if killall doesn't work kill -15 (or -9) the processes directly
```
vanitabarrett: We might need an example on how to do this / what the command would look like.

Tetrino: This should clear it up.

PR #1998, opened 2019-09-16
Review auto scaling groups

AlanGabbianelli:
Add step to tell 2nd line about the manual scaling up/down

### DIFF
--- a/source/manual/auto-scaling-groups.html.md
+++ b/source/manual/auto-scaling-groups.html.md
@@ -4,7 +4,7 @@ title: Auto Scaling Groups
 section: Infrastructure
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2019-09-11
+last_reviewed_on: 2019-09-16
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manual/auto-scaling-groups.html.md
```diff
@@ -28,41 +28,45 @@ the load.
 > ensure the number doesn't get put back to the old value if Terraform gets
 > deployed.
 
+1. Scaling up/down machines in AWS will trigger [Icinga alerts](icinga) so let
+   developers in`#govuk-2ndline` know you are about to do this.
+
 1. [Access the AWS Console][aws-console] and [go to the EC2 service][ec2-home].
 
-2. Select "Auto Scaling Groups" from the bottom of the menu on the left hand
+1. Select "Auto Scaling Groups" from the bottom of the menu on the left hand
```
vanitabarrett: It looks like the numbering is using `1` for every step?

AlanGabbianelli: Yes, using `1` for every step will make the numbers change automatically. So if we need to add an extra intermediate step in the future, we don't have to change them all but we can simply add it using a `1`.

vanitabarrett: I had no idea! Thanks @AlanGabbianelli .

PR #2006, opened 2019-09-17
Review Unicorn herder and added amended command for Integration servers

Tetrino:


source/manual/alerts/unicorn-herder.html.md
```diff
@@ -24,7 +24,8 @@ Configure and activate your fabric environment as described in the
 Then run the `vm.bodge_unicorn` task for the affected machine and application:
 
 ```bash
-fab $environment -H whitehall-backend-1.backend vm.bodge_unicorn:whitehall
+# Replace the machine_ip with that of the machine that is showing the error in Icinga
+fab $environment -H $machine_ip vm.bodge_unicorn:whitehall
```
vanitabarrett: I think you'd also have to replace `whitehall` with the relevant app name

Tetrino: Good catch.

source/manual/alerts/unicorn-herder.html.md
```diff
@@ -24,7 +24,9 @@ Configure and activate your fabric environment as described in the
 Then run the `vm.bodge_unicorn` task for the affected machine and application:
 
 ```bash
-fab $environment -H whitehall-backend-1.backend vm.bodge_unicorn:whitehall
+# Replace the machine_ip with that of the machine that is showing the error in Icinga
```
thomasleese: Should we add some examples as well, people may not know to turn the IP address in Icinga to something like `ip-10-1-4-159.eu-west-1.compute.internal`.

Tetrino: Like this?

PR #2007, opened 2019-09-17
Remove page on whitehall and docker replication

chao-xian:
Much of this information was already on the README for govuk-docker,
which we weren't aware of previously. To avoid duplication, we're
removing the recently added page and point to the README instead.


chao-xian:
Also moved some of the info here to the [README in govuk-docker](https://github.com/alphagov/govuk-docker/pull/182)

PR #2021, opened 2019-09-26
Add documentation for the rabbitmq federation.

rtrinque:
During the migration to AWS we setup a federation between the
rabbitmq cluster in Carrenza and the one in AWS.


source/manual/rabbitmq.html.md
```diff
@@ -17,6 +17,20 @@ which listen to the exchanges, and gather the messages sent to the
 exchanges together. Applications then run "consumers" which receive
 messages from the queues.
 
+While the migration of gov.uk to AWS in in progress we actually run
```
sengi: typo: "in in" -> "is in"

PR #2024, opened 2019-09-30
Add short description of AKA domains

sihugh:
They are necessary if you want to use the [side-by-side browser](https://github.com/alphagov/side-by-side-browser) when transitioning a site.

### DIFF
--- a/source/manual/transition-a-site.html.md
+++ b/source/manual/transition-a-site.html.md
@@ -4,7 +4,7 @@ title: Transition a site to GOV.UK
 section: Transition
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2019-10-08
+last_reviewed_on: 2019-10-30
 review_in: 6 months
 related_applications: [bouncer, transition]
 ---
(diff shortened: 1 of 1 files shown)


source/manual/transition-a-site.html.md
```diff
@@ -34,7 +34,22 @@ The content team might ask a developer to offer the department some guidance aro
 
 Follow the instructions in the [transition-config README][transition-config].
 
-### 3) Get a list of old URLs
+### 3) Consider AKA domains
```
issyl0: To save time in the future, can we make this numbered list an actual numbered list (rather than headings) and use the `1.` Markdown trick?

sihugh: Yeah, that's a good shout

sihugh: Needs a rebase now anyway

barrucadu:
I've resolved the conflict, keeping the reviewed date from `master` - I didn't bump the date to today because I've not actually reviewed it.

PR #2026, opened 2019-09-30
Fix linting errors

sihugh:
Ran `bundle exec govuk-lint-ruby -a app lib spec helpers` to autofix a bunch of stuff

Manually fixed "memoized variable names must match surrounding method names"

Manually uppercased heredoc delimiters

Allowed us to use `YAML.load` rather than `safe_load` because it was breaking tests.  The Yaml sources are trusted, so this should still be OK.


kevindew:
You're a good person @sihugh ⭐️

PR #2047, opened 2019-10-08
Update "Which browsers we support"

alex-ju:
This page was due to be reviewed.

I added reference to GOV.UK Frontend browser support matrix (since it was defined looking at analytics data on GOV.UK) and it reflects the current state of frontend applications that serve pages to the general public.

I also added a paragraph about informing browser support with analytics data for applications that are not serving pages for the general public (e.g. admin apps).

### DIFF
--- a/source/manual/browser-support.html.md
+++ b/source/manual/browser-support.html.md
@@ -5,8 +5,12 @@ layout: manual_layout
 section: Frontend
 type: learn
 owner_slack: "#govuk-frontenders"
-last_reviewed_on: 2018-08-29
+last_reviewed_on: 2019-10-08
 review_in: 12 months
 ---
 
-For browser support, see the Service Manual page about [Designing for different browsers and devices](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).
+GOV.UK [shares the same browser support matrix with GOV.UK Frontend](https://github.com/alphagov/govuk-frontend#browser-support) for pages served to the general public.
+
+If the application you are working on is aimed at internal users rather than the general public you should look at your analytics data and check which browsers your users are using. Then you can then make an informed decision about which browsers to test with.
+
+For more information around browser support, see the Service Manual page about [Designing for different browsers and devices](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).


source/manual/browser-support.html.md
```diff
@@ -5,8 +5,12 @@ layout: manual_layout
 section: Frontend
 type: learn
 owner_slack: "#govuk-frontenders"
-last_reviewed_on: 2018-08-29
+last_reviewed_on: 2019-10-08
 review_in: 12 months
 ---
 
-For browser support, see the Service Manual page about [Designing for different browsers and devices](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).
+GOV.UK [shares the same browser support matrix with GOV.UK Frontend](https://github.com/alphagov/govuk-frontend#browser-support) for pages served to the general public.
```
injms: GOV.UK Frontend supports IE 8 to a functional extent, whilst the Service Manual makes no mention of IE 8. This seems a bit confusing. I was under the impression that GOV.UK browser support was the same as listed in the Service Manual.

source/manual/browser-support.html.md
```diff
@@ -5,8 +5,12 @@ layout: manual_layout
...
+
+If the service you are working on is aimed at internal users rather than the general public you should look at your analytics data and check which browsers your users are using. Then you can then make an informed decision about which browsers to test with.
```
sihugh: This sounds a little off-topic for GOV.UK's internal docs, but I can understand that external developers may be looking to us for guidance.  Is it possible for us to quote something similar from the service manual and provide a link to that so that it feels a bit less like we might be providing advice to a non-GOV.UK audience?

Alternatively, is there a separate list of browsers that we need to support for GOV.UK publishing tools that we might need to consider (and list here)?

sihugh: Maybe it's just the way I read it 🤷‍♂

source/manual/browser-support.html.md
36degrees: The service manual defines the _minimum_ browser matrix that we ask services to test against. Some services may have different needs; for example, we _believe_ there are still services out there that have a higher percentage of traffic from IE8-10, and so they may choose to continue to support it. We want those services to be able to use GOV.UK Frontend, so we continue to support it at a functional level.

source/manual/browser-support.html.md
alex-ju: I think 'service' may not the right word for this. What I was trying to explain is that some GOV.UK applications (e.g. Content Publisher, Collections Publisher) may have different needs than the public-facing apps. For example Content Publisher is aligned with the service manual minimal requirements (as we've checked analytics data from Whitehall and figured that that will suffice) but public-facing apps still have a high number of users in IE8-10 and so we should provide the best possible experience for the general public.

alex-ju: Updated now to use 'application' instead of 'service'. We also have a link to the Service Manual, as suggested right below this paragraph.

`
For more information around browser support, see the Service Manual page about [Designing for different browsers and devices](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).
`

sihugh: Cool, thanks Alex!

source/manual/browser-support.html.md
```diff
@@ -5,8 +5,12 @@ layout: manual_layout
...
 owner_slack: "#govuk-frontenders"
-last_reviewed_on: 2018-08-29
+last_reviewed_on: 2019-10-08
 review_in: 12 months
 ---
 
-For browser support, see the Service Manual page about [Designing for different browsers and devices](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).
+GOV.UK [shares the same browser support matrix with GOV.UK Frontend](https://github.com/alphagov/govuk-frontend#browser-support) for pages served to the general public.
+
+If the application you are working on is aimed at internal users rather than the general public you should look at your analytics data and check which browsers your users are using. Then you can then make an informed decision about which browsers to test with.
```
andysellick: 👍

andysellick: I wonder if this line is valuable as a general point? Look at your traffic and decide which browsers to support (regardless of whether your users are internal or not)

PR #2092, opened 2019-10-31
Correct steps for leavers process

bevanloon:
If we merge the second set of puppet PRs (that remove the user manifest
file) before we've removed the user from govuk-secrets, we risk errors
during a puppet deploy.

Rejig the ordering so that removing the user from govuk-secrets comes
before removing them from puppet. This makes it more explicit and will
encourage us to merge the secrets PR before the puppet PR.

### DIFF
--- a/source/manual/removing-a-user-from-puppet.html.md
+++ b/source/manual/removing-a-user-from-puppet.html.md
@@ -15,19 +15,19 @@ user from Puppet itself. If the user is just removed from Puppet their files
 will remain on our servers forever more.
 
 1. First find the user manifest in: [modules/users/manifests][manifest-path].
-2. Add an entry to the govuk_user class of `ensure => absent`. Here is an
+1. Add an entry to the govuk_user class of `ensure => absent`. Here is an
    [example][absent-example].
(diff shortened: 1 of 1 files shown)


source/manual/removing-a-user-from-puppet.html.md
```diff
@@ -19,14 +19,14 @@ will remain on our servers forever more.
    [example][absent-example].
 3. Once this has been raised as a PR and merged, deploy Puppet to all
    environments.
-4. Create another PR for Puppet that:
+4. Create a PR in [GOV.UK secrets][govuk-secrets] that:
+  - Removes the user from [production hieradata][production-hieradata]. Read [what to do when someone leaves][what-to-do-when-someone-leaves]
+  - Removes the user from [AWS production hieradata][aws-production-hieradata]. Read [what to do when someone leaves][what-to-do-when-someone-leaves]
+5. Create another PR for Puppet that:
```
issyl0: Can we use the Markdown `1.` trick for these numbers?

bevanloon: yeah - let's have a go 2cc6bfe

PR #2094, opened 2019-10-31
Update instructions for working in private

brucebolt:
This makes a number of changes:
- updates "making a repo private" to use a new private mirror of an application's repo (rather than making the repo itself private)
- proposes using this new workflow for security fixes, rather than AWS CodeCommit
- moves AWS CodeCommit instructions into the "when GitHub is unavailable" section


source/manual/github-unavailable.html.md
```diff
@@ -23,9 +23,92 @@ We [mirror all GitHub repositories](repository-mirroring.html) tagged with `govu
...
+1. Push your changes to CodeCommit by running `git push`
+
+1. Tag your changes by running `git tag release_XYZ`, where XYZ is one more that the latest
+   release tag for the application you're working on, as reported by the Release app
+
+1. Push your new tag to CodeCommit by running `git push aws release_XYZ`
+
+#### Deploying the code change
+
+1. Review the pull request on AWS CodeCommit.
```
benthorner: This might be an stupid question, but how does one do this?

source/manual/github-unavailable.html.md
```diff
@@ -23,9 +23,92 @@ We [mirror all GitHub repositories](repository-mirroring.html) tagged with `govu
...
+
+1. Select "git-credential-osxkeychain" from the list.
+
+1. Hit the minus key.
+
+1. Try your terminal commands again.
+
+1. If you are prompted to add the item to keychain, deny.
+
+There is more information in Step 3 in the [AWS
```
benthorner: I'm a little confused here: should this relate to the export access key stuff from earlier?

source/manual/github-unavailable.html.md
brucebolt: I have no idea.  I'll see if I can figure it out in the AWS Console.

PR #2113, opened 2019-11-11
Replace govuk-lint

issyl0:
- The GOV.UK Lint gem is deprecated in favour of using Rubocop directly
  with a set of shared configs.
- See https://github.com/alphagov/govuk-rfcs/pull/ 100 for more context.

https://trello.com/c/CdBFg6yw/1521-replace-govuk-lint-with-rubocop-govuk

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,7 @@
+inherit_gem:
+  rubocop-govuk:
+    - config/default.yml
+
 Style/ClassVars:
   Enabled: false
 
--- a/Gemfile
+++ b/Gemfile
@@ -5,9 +5,9 @@ ruby File.read('.ruby-version').chomp
 gem 'capybara', '~> 3.29.0'
 gem 'rake'
 gem 'rspec', '~> 3.9'
+gem 'rubocop-govuk'
 gem 'webmock', '~> 3.7'
 gem 'activesupport', '~> 5.0'
-gem 'govuk-lint', '~> 4.2'
 gem 'simplecov'
 
 gem 'govuk_tech_docs'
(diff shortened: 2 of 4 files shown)


issyl0:
On my last commit, it's worth posting the important piece of the commit message here:

> - The only thing that the default library Jenkinsfile doesn't do is `vale` linting. We're using a [really
  old](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/manifests/packages/vale.pp#L11) version of Vale anyway (the current is 1.7.1), so I'm not sure of its value any more. If we want it, we can migrate it to be run on Concourse, or make it a pre-commit hook, or somesuch.

Do people agree?

barrucadu:
Text linting is good, I run both `vale` and `proselint` on my blog posts.  Sometimes I do need to ignore them though (not for the dev docs, but that could be because an older version has fewer checks), so I'm less sure of the utility of `vale` being a part of CI which must pass.  And if it is a part of CI which can fail, people won't bother to follow it, hmm...

Text linting helps enforce a standard style, which is good.  But there is much more to a standard style than what these tools provide.  So maybe it's not so useful after all.

issyl0:
I mostly don't want to do the aptly dance to update Vale to the latest version if we can avoid doing it by people using the pre-commit hook setup (which already exists), something in their editors, or another (easier to maintain) Concourse pipeline. Less software for us to maintain! And at the moment with the changes in this PR, if we accept the current changes then we're decreasing functionality.

issyl0:
There's a [branch of govuk-puppet](https://github.com/alphagov/govuk-puppet/tree/rm-vale-linter-package) to remove Vale from our CI infra.

issyl0:
Thanks, Bevan. On that note, we're removing Vale from CI boxes and relying on people having pre-commit installed and working. There's a note in the README. I'll publicise this on Slack, too.

PR #2118, opened 2019-11-13
README: Make Vale install instructions less verbose

issyl0:
- I think most people use Homebrew, and if they don't they can figure
  out installation instructions themselves. The 0.9.0 release is really
  old.


issyl0:
Force-pushed to make them _even shorter_.

PR #2129, opened 2019-11-18
Replace the many AWS access docs with docs on using the GDS CLI

issyl0:
- As part of the working group on developer tooling, and as part of an ongoing approach to standardize tooling and make things more secure and better for new people joining GOV.UK, we are standardising on the GDS CLI for AWS-related tasks (signing into the Console, generating
access keys for Terraform runs, etc).

- It's a CLI that's developed by RE and is secure in its approach to dealing with AWS credentials. It's substantially less code for us to maintain, and doesn't involve copying and pasting credentials into ~/.aws/config and ~/.aws/credentials files as part of an onboarding process. Credential rotation is also easier in the event of a compromise.

- This removes `aws-cli-access` and `aws-console-access` and adds a new page with docs on the `gds-cli`. It also changes `set-up-aws-account` to link to the `gds-cli` page not the two dead pages.

- Use `aws s3 ls` as the `gds-cli` command chaining example as we don't want to encourage people to run Terraform locally, necessarily. And the raw Terraform command doesn't work with our complicated setup anyway.

- Note that people without Yubikeys will have to set `gds config yubikey false`.

- Remove `govuk aws` instructions in favour of `gds-cli` instructions in Terraform deploy.

Related issues and PRs: https://github.com/alphagov/govuk-aws/issues/1159 and https://github.com/alphagov/govuk-aws/pull/1177.
The slides for the talk I'm giving at Tech Fortnightly are [on Google Drive](https://docs.google.com/presentation/d/1ifPCXe9KwqYcjzfP2v7V2iU4BFahteH3wRYdtk3wrr8/edit).

Co-authored-by: Issy Long <issy.long@digital.cabinet-office.gov.uk> 
Co-authored-by: Chris Banks <chris.banks@digital.cabinet-office.gov.uk>


source/manual/gds-cli.html.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Usage for AWS
+
+Work out which [list of users you're part of in govuk-aws-data][govuk-aws-data-users-group].
+
+If you're new to GOV.UK and want to look around in integration:
+    gds aws govuk-integration-readonly -l
+
+If you're a PlatformHealthPowerUser and want to generate tokens for deploying Terraform in production, run:
+    gds aws govuk-production-platformhealth-poweruser -e
```
issyl0:
```suggestion
        gds aws govuk-production-platformhealth-poweruser -e
```

source/manual/gds-cli.html.md
```diff
@@ -0,0 +1,39 @@
...
+
+You can also chain commands, like this one to list S3 buckets in integration:
+
+    gds aws govuk-integration-poweruser aws s3 ls
```
issyl0:
```suggestion
        gds aws govuk-integration-poweruser aws s3 ls
```

source/manual/gds-cli.html.md
```diff
@@ -0,0 +1,39 @@
...
+1. Run `gds config yubikey false` if you type AWS MFA codes from your phone.
+
+If you have `bash-completion` installed and configured, the gds-cli tab completions will work out of the box. They're especially useful for long commands like AWS account names.
+
+## Usage for AWS
+
+Work out which [list of users you're part of in govuk-aws-data][govuk-aws-data-users-group].
+
+If you're new to GOV.UK and want to look around in integration:
+        gds aws govuk-integration-readonly -l
```
issyl0:
```suggestion

        gds aws govuk-integration-readonly -l
```

source/manual/gds-cli.html.md
```diff
@@ -0,0 +1,39 @@
...
+
+If you're a PlatformHealthPowerUser and want to generate tokens for deploying Terraform in production, run:
+        gds aws govuk-production-platformhealth-poweruser -e
```
issyl0:
```suggestion

        gds aws govuk-production-platformhealth-poweruser -e
```

source/manual/deploying-terraform.html.md
```diff
@@ -4,76 +4,82 @@ title: Deploy AWS infrastructure with Terraform
...
-We use [Terraform](https://terraform.io) for configuring the GOV.UK
-infrastructure in AWS.
+We use [Terraform](https://terraform.io) for configuring GOV.UK infrastructure in AWS.
+This document describes:
 
-## 1. Check what you can deploy
+ * How to set up your machine to run Terraform.
+ * How to deploy infrastructure changes using Terraform.
+
+## One-time setup
```
AlanGabbianelli: @issyl0 I think it's worth extracting this section to a separate page as it's not actually about deploying terraform. We can then add link to that page here with something like "if you haven't done this before, read the one-time setup doc page".

issyl0: It *is* though. It's about using our tooling to deploy Terraform, for which you need to do some setup.

issyl0: @sengi What do you think here?

sengi: I think it's a fair comment but I can see arguments for/against either way. I think there's some merit in reducing the number of different documents that a new user has to navigate through in order to complete the setup process. On the other hand, I accept that keeping both the setup steps and the "doing the thing" steps in this document means that the document kinda has two purposes.

I think I have a mild preference for keeping it as-is, but I could be persuaded otherwise if you feel strongly about it.

PR #2136, opened 2019-11-25
Remove references to replicate-data-local.sh

barrucadu:
~**Depends on https://github.com/alphagov/govuk-docker/pull/263**~

This is going away.  Replicating data in the dev VM is now a painful
process, but we're actively working to remove the dev VM so this is
fine.

---

[Trello card](https://trello.com/c/7G8MG8yh/117-review-dev-docs-for-replication-stuff)


issyl0:
I was sure there were more places than this, but this is a good start! Actually, maybe the more places were app READMES (for reasons unknown).

PR #2142, opened 2019-11-28
Update note to say that powerusers can now rotate keys

erino:


source/manual/aws-iam-key-rotation.html.md
```diff
@@ -21,9 +21,9 @@ every 3 months.
 
 > **Note**
 >
-> Rotating AWS IAM keys requires permissions which are available
-> with the `internal-admin` or `admin` roles. If you can't assume
-> one of these roles, you won't be able to rotate keys.
+> Rotating AWS IAM keys requires permissions which are available with the
+> `internal-admin`, `admin`, `platformhealth-poweruser` or `poweruser` roles.
```
issyl0:
```suggestion
> `admin`, `internal-admin`, `platformhealth-poweruser` or `poweruser` roles.
```

For consistency with the order of the role names in other docs and `govuk-aws-data`.

erino: Fixed in 49c7602

PR #2145, opened 2019-11-28
Add follow-on tutorial for Docker

benthorner:
https://trello.com/c/gl5911VV/116-advanced-tutorial-on-docker

This extends the existing 'Intro to Docker' manual to cover additional
concepts that we use in govuk-docker.

👉 [Rendered version](https://github.com/alphagov/govuk-developer-docs/blob/a4b8d97f9143283c93401cf7d84b51eed2e962dc/source/manual/intro-to-docker-even-more.html.md)

### DIFF
--- a/source/manual/intro-to-docker-even-more.html.md
+++ b/source/manual/intro-to-docker-even-more.html.md
@@ -1,0 +1,396 @@
+---
+title: Intro to Docker (Even More)
+section: Docker
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2020-02-03
+review_in: 6 months
+owner_slack: "#govuk-developers"
+---
+
+In the [Intro to Docker tutorial](/manual/intro-to-docker.html) we began with a generic container running the `ruby` image, and finished with a powerful `docker-compose` command to run the [content-publisher][] tests against a Postgres DB.
+
+This tutorial is going to pick up where we left off and introduce some more advanced concepts that we make use of in [govuk-docker][]. You should be able to understand the bulk of the repo after completing this tutorial.
+
+> You will need the following from the [previous tutorial](/manual/intro-to-docker.html):
+>
+>   * The `docker-compose.yml` file from Step 6
+>   * The Dockerfile that we wrote incrementally
+>   * The `.dockerignore` file from Step 3 (optional)
+
+## Step 1: Makefiles
+
+We did some tidy-up work at the end of the last tutorial, but there are still some manual setup commands to run.
+
+```shell
+$mac docker-compose run content-publisher-demo bundle install
+$mac docker-compose run content-publisher-demo bundle exec rake db:setup
+```
+
+To avoid having to remember these commands and type them to setup different projects, we can use a [Makefile][makefile].
+
+```
+content-publisher:
+  docker-compose run content-publisher-demo bundle install
+  docker-compose run content-publisher-demo bundle exec rake db:setup
+```
+
+> Makefiles use strict tabs for indentation, so you may need to adjust your editor if it's inserting spaces automatically.
+
+Now we can run `make content-publisher` and the commands are run for us.
+
+## Step 2: `/govuk`
+
+At the end of the last tutorial we got some, but not all, of the tests passing. The remaining failures should be due to a missing dependency: [govuk-content-schemas][]. The tests are looking for the shemas at `../govuk-content-schemas`, which doesn't exist inside our container, since we only mapped the `content-publisher` directory to `/app`.
+
+* Mount the whole of the `~/govuk` directory
+
+```
+volumes:
+  - bundle:/usr/local/bundle
+  - ~/govuk:/govuk
+```
+
+* Adjust the working directory accordingly
+
+```
+working_dir: /govuk/content-publisher
+```
+
+* Check all of the tests now pass (hopefully!)
+
+```shell
+$mac docker-compose run content-publisher-demo bundle exec rake
+```
+
+## Step 3: rbenv
+
+In the last tutorial we used a `bundle` volume to persist the gems we installed. This works fine initially, but [Bundler][] doesn't cope well if we start using a new version of Ruby.
+
+On GOV.UK we use [rbenv](/manual/ruby.html) to manage different Ruby versions. To support this, we need to build a lower-level base image that has rbenv and isn't tied to a version of Ruby.
+
+* Change the `FROM` entry in our Dockerfile
+
+```
+FROM buildpack-deps
+```
+
+* Install rbenv before we create the 'build' user
+
+```
+RUN git clone https://github.com/sstephenson/rbenv.git /rbenv
+RUN git clone https://github.com/sstephenson/ruby-build.git /rbenv/plugins/ruby-build
+RUN /rbenv/plugins/ruby-build/install.sh
+ENV PATH /rbenv/bin:$PATH
+```
+
+* Add shims for ruby, bundler, etc. to the `PATH`
+
+```
+ENV PATH /home/build/.rbenv/shims:${PATH}
+```
+
+* Tell compose to rebuild the image we're using
+
+```shell
+$mac docker-compose build content-publisher-demo
+```
+
+Rubies and gems will go into `/home/build/.rbenv` by default. We need to make sure this directory is persisted. If we persist the entire `/home/build` directory, we can also benefit from [caching for chromedriver][webdrivers].
+
+* Replace the 'bundle' volume with a 'home' one
+
+```
+volumes:
+  home:
+  postgres:
+```
+
+* Change the service to use the new 'home' volume
+
+```
+volumes:
+  - home:/home/build
+  - ~/govuk:/govuk
+```
+
+* Change our `Makefile` to install Ruby and bundler
+
+```
+content-publisher:
+  docker-compose run content-publisher-demo rbenv install -s
+  docker-compose run content-publisher-demo sh -c 'gem install --conservative --no-document bundler -v $$(grep -A1 "BUNDLED WITH" Gemfile.lock | tail -1)'
+  docker-compose run content-publisher-demo bundle install
+  docker-compose run content-publisher-demo bundle exec rake db:setup
+```
+
+* Get the tests passing now we're running with rbenv
+
+```shell
+# it can take some time to install Ruby!
+$mac make content-publisher
+
+# run the tests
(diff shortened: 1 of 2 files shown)


source/manual/intro-to-docker-even-more.html.md
```diff
@@ -0,0 +1,398 @@
+---
+title: Intro to Docker (Even More)
+section: Docker
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2019-11-06
```
ChrisBAshton: You can update this to Feb 2020!

source/manual/intro-to-docker.html.md
```diff
@@ -353,19 +353,18 @@ $mac docker-compose run content-publisher-demo bundle exec rake
...
-* creating and adding volumes
-* installing extra libraries with a Dockerfile
-* creating a Network
+* **volumes** the repo and gems
+* a **network** to communicate with Postgres
+* **environment variables** for Postgres
+* a **working directory** to reduce typing
 
-These concepts are useful to know when you start to use [govuk-docker], which you should use from this point on with GOV.UK.
+We also explored how to write and build an **image** that has the essential libraries and presets, using a **Dockerfile**. The end result is a single, relatively short command that we can use for day-to-day tasks, such as `bundle exec rake`.
```
ChrisBAshton: Perhaps encapsulate the above changes in a separate commit (and bump the last reviewed date on this file).

Then in the "Even More" commit, you can link to part 2.

source/manual/intro-to-docker-even-more.html.md
```diff
@@ -0,0 +1,398 @@
...
+
+To avoid having to remember these commands and type them for different projects, we can use a [Makefile][makefile].
+
+```
+content-publisher:
+  docker-compose run content-publisher-demo bundle install
+  docker-compose run content-publisher-demo bundle exec rake db:setup
+```
+
+> Makefiles use strict tabs for indentation, so you may need to adjust your editor if its inserting spaces automatically.
```
ChrisBAshton:
```suggestion
> Makefiles use strict tabs for indentation, so you may need to adjust your editor if it is inserting spaces automatically.
```

source/manual/intro-to-docker-even-more.html.md
```diff
@@ -0,0 +1,398 @@
...
+* Adjust the working directory accordingly
+
+```
+working_dir: /govuk/content-publisher
+```
+
+* Check all of the tests now pass (hopefully!)
+
+```shell
+$mac docker-compose run content-publisher-demo bundle exec rake
```
ChrisBAshton: I was a little surprised to come across this syntax again, now that you've encapsulated these types of commands in the Makefile. Then I noticed this command is slightly different to the ones in the Makefile and that therefore we must run it verbosely like this.

May be worth saying that the Makefile is only for one-off setup?

source/manual/intro-to-docker-even-more.html.md
```diff
@@ -0,0 +1,398 @@
...
+
+* Change our `Makefile` to install Ruby and bundler
+
+```
+content-publisher:
+  docker-compose run content-publisher-demo rbenv install -s
+  docker-compose run content-publisher-demo sh -c 'gem install --conservative --no-document bundler -v $$(grep -A1 "BUNDLED WITH" Gemfile.lock | tail -1)'
+  docker-compose run content-publisher-demo bundle install
+  docker-compose run content-publisher-demo bundle exec rake db:setup
+
```
ChrisBAshton:
```suggestion
```

source/manual/intro-to-docker-even-more.html.md
```diff
@@ -0,0 +1,398 @@
...
+* Start with a hacky bunch of command flags
+
+```shell
+$mac docker-compose run -p 80:3000 content-publisher-demo bin/rails s -b 0.0.0.0
+```
+
+> If you see a `Bind for 0.0.0.0:80 failed` error, you need to stop the process on your local machine that's using this port. This is probably GOV.UK Docker, so try doing `govuk-docker down`.
+
+> If you see a `A server is already running` error, you need to add an additional `--restart` flag to tell Rails to cleanup its `tmp/pids` directory, which sometimes fails when containers exit too quickly.
```
ChrisBAshton:
```suggestion
> If you see a `A server is already running` error, you need an additional `--restart` flag to tell Rails to clean up its `tmp/pids` directory, which sometimes fails when containers exit too quickly.
```

source/manual/intro-to-docker-even-more.html.md
```diff
@@ -0,0 +1,398 @@
...
+You should now be able to go to this domain in your browser and see the Content Publisher homepage. After the next section we'll iterate this approach to make it more dynamic and work across multiple apps.
+
+## Step 5: YAML
+
+The `content-publisher-demo` service in our `docker-compose.yml` file is now serving two purposes: running tests and running the app. As we add more environment variables, dependencies, etc., it makes sense to have variants of `content-publisher-demo` that have just the right amount of config for what they need to do.
+
+* Move common config into [an extension][yaml-extension]
+
+```
+# this feature needs a higher version of compose
```
ChrisBAshton: Which feature? Aliasing?

source/manual/intro-to-docker-even-more.html.md
```diff
@@ -0,0 +1,398 @@
...
+    <<: *content-publisher-demo
+    depends_on:
+      - postgres
+    environment:
+      DATABASE_URL: "postgresql://postgres@postgres/content-publisher"
+      HOST: 0.0.0.0
+    command: bin/rails s
+```
+
+> Although the `depends_on` config is the same for both stacks, in general we expect it to vary between stacks.
```
ChrisBAshton:
```suggestion
> Although the `depends_on` config is the same for both stacks in this case, we would normally expect it to vary between stacks.
```

source/manual/intro-to-docker-even-more.html.md
```diff
@@ -0,0 +1,398 @@
...
+
+* Update the Makefile to use the `lite` stack
+
+Each service will have its own image by default, which is based on the name of the service. Compose will build the images automatically when we try to `run` one of the stacks. Although the build time will be fast, because Docker will re-use layers it built previously, the extra 'build' step is unnecessary and we can easily avoid it.
+
```
ChrisBAshton:
```suggestion
```

source/manual/intro-to-docker-even-more.html.md
```diff
@@ -0,0 +1,398 @@
...
+# run the tests
+$mac docker-compose run content-publisher-demo-lite bundle exec rake
+
+# run the app
+$mac docker-compose run -p 80:3000 content-publisher-demo-app
+```
+
+## Step 6: NGINX
+
+In the previous section we used a `-p 80:3000` flag to make Rails, which runs on port 3000, available on our host machine, on port 80. This approach doesn't scale to multiple apps, since only one process can bind to port 80 at a time. We need a single app that can 'proxy' our HTTP requests...an [`nginx-proxy`][nginx-proxy]!
```
ChrisBAshton:
```suggestion
In the previous section we used a `-p 80:3000` flag to make Rails, which runs on port 3000, available on our host machine, on port 80. This approach doesn't scale to multiple apps, since only one process can bind to port 80 at a time. We need a single app that can 'proxy' our HTTP requests... an [`nginx-proxy`][nginx-proxy]!
```

source/manual/intro-to-docker-even-more.html.md
```diff
@@ -0,0 +1,398 @@
...
+That's it! You should now have something very similar to the repo we use to develop on GOV.UK. All that remains is for you to go forth, use it, make it work better, and develop some software :-).
+
+## Step 8: Cleanup
+
+During this tutorial you created or changed a lot of files and state on your system. The following indicate how to cleanup your system e.g. if you want to run through the tutorials again.
+
+* Remove all the extra docker state
+
+```shell
+# ignore any errors, you may need to run multiple times
```
ChrisBAshton: This instruction is a little confusing. Do I need to re-run these commands until no errors are happening anymore?

source/manual/intro-to-docker-even-more.html.md
```diff
@@ -0,0 +1,398 @@
...
+
+[content-publisher]: https://github.com/alphagov/content-publisher
+[makefile]: https://www.gnu.org/software/make/manual/html_node/Introduction.html
+[govuk-content-schemas]: https://github.com/alphagov/govuk-content-schemas
+[govuk-docker]: https://github.com/alphagov/govuk-docker
+[Bundler]: https://bundler.io/
+[nginx-proxy]: https://github.com/jwilder/nginx-proxy
+[dnsmasq]: http://www.thekelleys.org.uk/dnsmasq/doc.html
+[webdrivers]: https://github.com/titusfortner/webdrivers#download-location
+[yaml-extension]: https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd
```
ChrisBAshton: Could you alphabetise this list?

source/manual/intro-to-docker-even-more.html.md
benthorner: I've reworded this slightly - it refers to the bullet this is part of. Any better?

source/manual/intro-to-docker-even-more.html.md
benthorner: I think I see what you mean: we're only using `make` for setup, and not as a wrapper for running commands. I've added the word "setup" into a couple of places in the previous section. Does that help?

source/manual/intro-to-docker-even-more.html.md
benthorner: I've rephrased as a note below the code block - is that better? Re-ordering the commands in the block __should__ mean that you only need to run it once - pretty sure there aren't any cyclic dependencies in there, which was what the old comment was about.

source/manual/intro-to-docker.html.md
benthorner: I've bumped the review dates. I think it makes sense to keep everything in one commit: I can't see a clear motivation to split it up, and keeping everything in one commit has the advantage of making it easy to revert, although I hope we won't need to do that!

source/manual/intro-to-docker-even-more.html.md
ChrisBAshton: That looks better 👍

PR #2149, opened 2019-12-02
Update cert renewal procedure in Fastly

rtrinque:
Fastly has a better interface to import new certs now.

### DIFF
--- a/source/manual/renew-a-tls-certificate.html.md
+++ b/source/manual/renew-a-tls-certificate.html.md
@@ -4,7 +4,7 @@ title: Renew a TLS certificate for GOV.UK
 section: Infrastructure
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2019-11-25
+last_reviewed_on: 2019-12-02
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manual/renew-a-tls-certificate.html.md
```diff
@@ -36,15 +36,9 @@ Credentials for the Fastly dashboard and Zendesk support sites are in the
...
-    renewed) and everything else (the certificate chain) and upload these as
-    separate files.
-12. *For staging and integration only:* Make a note of the reference ID returned
-    by the TLS Uploader. Open a new ticket with
-    [Fastly support](https://fastly.zendesk.com/hc/en-us) quoting the reference
-    ID and the service ID for the Fastly service the certificate is for
-    (such as for "Staging GOV.UK") and asking them to make the certificate live.
-    This process may take up to a week.
+11. *For staging and integration only:*
+    Go to the Fastly interface and then to Configure -> HTTPS and network
```
sengi: This line gets merged together with the next one when it's rendered, so this reads "HTTPS and network Go to TLS certificates and upload your new cert." Can fix that by just adding a full stop. Or could make it a sublist if you prefer.
```suggestion
    Go to the Fastly interface and then to Configure -> HTTPS and network.
```

rtrinque: Ah right thanks, full stop added.

benthorner:
@rtrinque is this still good to go in?

PR #2150, opened 2019-12-02
Clarify how the email alert check works

benthorner:
https://trello.com/c/13JaJ5mQ/203-review-the-email-alert-traffic-dashboard

This clarifies the description of the email alert check to say that it
(currently) works against two mailboxes, not just one.


sihugh:
🚀 🐦 ✉️

PR #2151, opened 2019-12-03
2nd line review

ChrisBAshton:
Updated various email-related docs, tried to remove some duplication.


source/manual/alerts/email-alert-api-app-healthcheck-not-ok.html.md
```diff
@@ -4,28 +4,85 @@ title: email-alert-api app healthcheck not ok
...
+## Check the email alert API
+
+A useful debug step if you're not sure if an email has been sent is to
+check Email Alert API to see if it has created records in its database.
+
+```
+$ govuk_app_console email-alert-api
+> ContentChange.last
+> Email.last
+> DeliveryAttempt.last
```
oscarwyatt: Email records (and I think the others too) have a uuid as a primary key so `Email.last` won't necessarily return the most recently created email, you might be better off with something like `Email.where(created_at: DateTime.now - 15.minutes...DateTime.now).count` or something

ChrisBAshton: Thanks for the review! I agree this is more of a sanity check than a truly reliable test. For example, if another email is coincidentally created around the same time as you're debugging you could get some confusing results!

Not sure how changing to `created_at` would make the sanity check any better, unless I'm missing something 🤔`uuid` should always return the most recently created email, whereas `created_at` can be overridden when the email is created so that it appears to be older than it is.

oscarwyatt: Sorry, I wasn't clear. `.last` finds the record that comes last according to the database primary key. For most apps where a strictly increasing integer is used, that works perfectly. However, for any app that uses a uuid as the primary key, the order will be alphabetical so it won't return the most recent record but the one with the 'highest' alphabetical order, which is unlikely to be the most recently created. Agreed, someone could come along and alter the created_at fields but I think it's a bit unlikely

ChrisBAshton: Ahhh I'm with you. Makes sense! I can edit the doc to that effect.

PR #2152, opened 2019-12-03
Try to consolidate guidance around AWS

benthorner:
Searching for 'AWS' returns quite a lot of content. This tries
to consolidate some of the docs to get people to the information
they need a bit faster.


source/manual/access-aws-console.html.md
```diff
@@ -22,17 +22,23 @@ The (private) GitHub repo is [alphagov/gds-cli](https://github.com/alphagov/gds-
...
 
         gds aws govuk-integration-readonly -l
 
-If you're a PlatformHealthPowerUser and want to generate tokens for deploying Terraform in production, run:
+### Terraform
+
+👉 [Deploy AWS infrastructure with Terraform](/manual/deploying-terraform.html)
 
-        gds aws govuk-production-platformhealth-poweruser -e
+### Manual Hacks
```
issyl0: I dispute the term "hacks": they're not hacks, they're legit usages with the AWS CLI etc.

source/manual/concept-of-stacks.html.md
```diff
@@ -1,44 +0,0 @@
----
```
issyl0: Stacks were a nice idea but ultimately didn't go any further than `blue` and they now just get in the way.

source/manual/access-aws-console.html.md
benthorner: Sure. My comedy didn't quite come off - updated 👍

source/manual/concept-of-stacks.html.md
benthorner: Interestingly, after putting this up I notice we've got 'green' DNS records defined on integration. But I still don't think it affects anything day-to-day.

PR #2154, opened 2019-12-04
Fix onboarding instructions for setting up 'govukcli'

ChrisBAshton:
It was being referred to in 'howto-ssh-to-machines-in-aws.html.md', which
is not in the standard flow of setup instructions for new joiners. Other
instructions expect govukcli to be set up already, which tripped up our
new starter.
We now move the setup instructions closer to the beginning of the
onboarding journey so that the step cannot be missed.

There's also a commit for removing an anchor link to a section that no longer
exists, which I discovered while double-checking my change doesn't affect
other files.

### DIFF
--- a/source/manual/get-ssh-access.html.md
+++ b/source/manual/get-ssh-access.html.md
@@ -56,7 +56,17 @@ Create a pull request with these changes. Once it has been [reviewed by a member
 [integration-hiera]: https://github.com/alphagov/govuk-puppet/blob/master/hieradata/integration.yaml
 [merging]: /manual/merge-pr.html
 
-## 3. Access remote environments
+## 3. Set up govukcli
+
+Clone the `govuk-aws` repository and add a symlink to make `govukcli` executable globally:
+
+```sh
+cd ~/govuk
+git clone https://github.com/alphagov/govuk-aws
+ln -s ~/govuk/govuk-aws/tools/govukcli /usr/local/bin/govukcli
+```
+
+## 4. Access remote environments
 
 Your pull request from earlier will hopefully have been merged by now. It's time to test your access to servers via SSH.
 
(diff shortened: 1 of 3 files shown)


source/manual/howto-ssh-to-machines-in-aws.html.md
```diff
@@ -26,16 +26,7 @@ context, any subsequent `govukcli` commands will be tied to the environment you
...
-
-```sh
-cd ~/govuk
-git clone https://github.com/alphagov/govuk-aws
-ln -s ~/govuk/govuk-aws/tools/govukcli /usr/local/bin/govukcli
-```
-
-You will also need to have followed the [Get SSH access to integration](/manual/get-ssh-access.html)
-instructions.
+Followed the [set up govukcli](/manual/get-ssh-access.html#3-set-up-govukcli) instructions.
```
issyl0:
```suggestion
Follow the [set up govukcli](/manual/get-ssh-access.html#3-set-up-govukcli) instructions.
```

source/manual/get-ssh-access.html.md
```diff
@@ -56,16 +56,25 @@ Create a pull request with these changes. Once it has been [reviewed by a member
...
 [merging]: /manual/merge-pr.html
 
-## 3. Access remote environments
+## 3. Set up govukcli
+
+Clone the `govuk-aws` repository and add a symlink to make `govukcli` executable globally:
+
+```sh
+cd ~/govuk
+git clone https://github.com/alphagov/govuk-aws
```
issyl0:
```suggestion
git clone git@github.com:alphagov/govuk-aws
```

Can we standardize on cloning repos over SSH to make pushing back to them via SSH key easier (rather than having to use access tokens)?

source/manual/howto-ssh-to-machines-in-aws.html.md
ChrisBAshton: Whoops, good spot!

source/manual/get-ssh-access.html.md
issyl0: Hmm, standardize was maybe the wrong word, there's only this occurrence of `git clone https://...` in the actual docs.

ChrisBAshton: Another good suggestion 👍 will do. Thanks!

PR #2155, opened 2019-12-04
Amend guidance on pages linked from Getting Started

issyl0:
- I was inspired by #2154. See commit messages for more info.

### DIFF
--- a/source/manual/get-ssh-access.html.md
+++ b/source/manual/get-ssh-access.html.md
@@ -21,21 +21,7 @@ User accounts in our integration environments are managed in the [govuk-puppet][
     mac$ cd ~/govuk
     mac$ git clone git@github.com:alphagov/govuk-puppet.git
 
-To create a new account, start by creating an SSH key at least 4096 bits long. For example:
-
-    mac$ ssh-keygen -t rsa -b 4096 -C "your_email@example.com" -f ~/.ssh/alphagov
-
-Import the SSH key into your keychain.
-
-    mac$ /usr/bin/ssh-add -K ~/.ssh/alphagov
-
-You'll also want to add the above line into your `~/.bash_profile` so that it is persistent between restarts.
-    
-Create a public key from this:
-
-    mac$ ssh-keygen -y -f ~/.ssh/alphagov > ~/.ssh/alphagov.pub
-
-Now create a user manifest in `~/govuk/govuk-puppet/modules/users/manifests` with your username and the public key you just created. Your username should use the `firstnamelastname.pp` format.
+Now create a user manifest in `~/govuk/govuk-puppet/modules/users/manifests` with your username and the public key you [created when you set up your GitHub account](/manual/github-setup.html). Your username should use the `firstnamelastname.pp` format.
 
 ```
 class users::johnsmith {
--- a/source/manual/github-setup.html.md
+++ b/source/manual/github-setup.html.md
@@ -17,6 +17,8 @@ review_in: 6 months
     $ /usr/bin/ssh-add -K your-private-key
     ```
 
+1. Add the above line into your `~/.bash_profile` or equivalent so that it is persistent between restarts.
+    
 1. Test that it all works by running `ssh -T git@github.com`.
 
 1. While you're here, associate your name and email to your git commits:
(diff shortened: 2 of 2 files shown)


source/manual/github-setup.html.md
```diff
@@ -7,11 +7,11 @@ last_reviewed_on: 2019-10-21
 review_in: 6 months
 ---
 
-1. Set up a [GitHub][] account.
+1. Set up a [GitHub][] account. Existing personal accounts are fine to use.
```
ChrisBAshton: Good to clarify this 👏 thanks

source/manual/github-setup.html.md
```diff
@@ -7,11 +7,11 @@ last_reviewed_on: 2019-10-21
...
 1. Ask your tech lead to add you to the [alphagov organisation][alphagov]. You will have to be added to the [GOV.UK team][govuk-team] to get access to repos & CI. Remember to click accept in the GitHub email invitation.
 1. Ask somebody with access to add your GitHub username to the [user monitoring system][user-reviewer].
-1. [Generate and register an SSH key pair][register-ssh-key] for your Mac for your GitHub account (use a `4096` bit key)
-1. Import the SSH key into your keychain. Once you’ve done this, it’ll be available to the VM you'll install in the next step.
+1. [Generate and register an SSH key pair][register-ssh-key] for your laptop for your GitHub account. You should use a `4096` bit key.
+1. Import the SSH key into your keychain. Once you’ve done this, you can clone repos over SSH in the next steps.
```
ChrisBAshton: There are no more steps in this document, so might I suggest this instead?

```suggestion
1. Import the SSH key into your keychain. Once you’ve done this, you'll be able to clone repos over SSH.
```

source/manual/get-ssh-access.html.md
```diff
@@ -21,21 +21,7 @@ User accounts in our integration environments are managed in the [govuk-puppet][
...
-
-    mac$ /usr/bin/ssh-add -K ~/.ssh/alphagov
-
-You'll also want to add the above line into your `~/.bash_profile` so that it is persistent between restarts.
-    
-Create a public key from this:
-
-    mac$ ssh-keygen -y -f ~/.ssh/alphagov > ~/.ssh/alphagov.pub
-
-Now create a user manifest in `~/govuk/govuk-puppet/modules/users/manifests` with your username and the public key you just created. Your username should use the `firstnamelastname.pp` format.
```
ChrisBAshton: Disclaimer: I think I was the one that added these ssh instructions in the first place. I think there's value in keeping them because GitHub doesn't actually show you how to create a public key from your private one, nor does it have the instruction around making the SSH key import persistent between restarts.

I agree it's probably in the wrong place though - perhaps it could be moved to the same file where the GitHub setup page is linked?

issyl0: I just tried this and the public key was auto-generated by the initial command:

```
$ ssh-keygen -t rsa -b 4096 -C "your_email@example.com" -f ~/.ssh/alphagov
[...]
Your identification has been saved in /Users/isabelllong/.ssh/alphagov.
Your public key has been saved in /Users/isabelllong/.ssh/alphagov.pub.
``

issyl0: Done!

source/manual/github-setup.html.md
issyl0: Totally. Thanks!

source/manual/get-ssh-access.html.md
ChrisBAshton: So it does! Hmmm. Could we refer to the output in the docs, along the lines of "the previous command will generate a private and a public key. You'll need to grab the contents of your public key and save them HERE:".

issyl0: This is already here:

>     ssh_key  => 'this public key will be a few lines long (copy the output from `more ~/.ssh/alphagov.pub`). It should begin with `ssh-rsa AAA` and end with `== john.smith.123@digital.cabinet-office.gov.uk`. You may need to add the email address to the end of your public key manually.',

Line 31.

ChrisBAshton: Ah, perfect. Thanks!

PR #2158, opened 2019-12-04
Tweak manual on deploying Terraform

benthorner:
This tweaks (actually, it's more of a rewrite) the manual around
deploying Terraform projects to be a bit more concise and clearly
highlight the steps to take, with less emphasis on explanation.

This also adds more sections to distinguish between deploing directly
from Jenkins vs. triggering a Jenkins job locally, which requires
a bit more effort to setup for small one-time changes.

👉 [Rendered version](https://github.com/alphagov/govuk-developer-docs/blob/ad27d3b862899e515ff430838c72f61f12ca5ca2/source/manual/deploying-terraform.html.md)

### DIFF
--- a/source/manual/deploying-terraform.html.md
+++ b/source/manual/deploying-terraform.html.md
@@ -1,85 +1,87 @@
 ---
 owner_slack: "#re-govuk"
-title: Deploy AWS infrastructure with Terraform
+title: Deploy Terraform
 section: Deployment
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2019-11-08
-review_in: 3 months
+last_reviewed_on: 2019-12-04
+review_in: 6 months
 ---
 
 We use [Terraform](https://terraform.io) for configuring GOV.UK infrastructure in AWS.
-This document describes:
-
- * How to set up your machine to run Terraform.
- * How to deploy infrastructure changes using Terraform.
 
 ## One-time setup
 
 ### 1. Check that you have sufficient access
 
 Which changes you can deploy depends on the level of access you have
-to our AWS environments. Specifically, the level of access your Amazon Resource Name (ARN) has been given.
+to our AWS environments.
 
-- `govuk-users` (`role_user_user_arns`) can't deploy anything.
-- `govuk-powerusers` (`role_poweruser_user_arns`) and `govuk-platformhealth-powerusers` (`role_platformhealth_poweruser_user_arns`) can deploy everything except IAM (users and policies).
-- `govuk-administrators` (`role_admin_user_arns`) and `govuk-internal-administrators` (`role_internal_admin_user_arns`) can deploy everything including IAM.
+- `govuk-users` can't deploy anything.
+- `*-powerusers` can deploy anything except IAM.
+- `*-administrators` can deploy anything.
 
-You can find which class of user you are (that is, which roles your ARN has been assigned to) in the
-[infra-security project in govuk-aws-data](https://github.com/alphagov/govuk-aws-data/tree/master/data/infra-security).
+You can find which class of user you are [infra-security project in govuk-aws-data](https://github.com/alphagov/govuk-aws-data/tree/master/data/infra-security).
 
-### 2. Install `gds-cli`
+## Deploying on Jenkins
 
-See [instructions for installing `gds-cli`](/manual/access-aws-console.html).
+> **This is not the recommended approach.**
+>
+> It's slightly easier to start with, but you'll quickly want to switch to deploying locally (see the next section) using the `deploy.rb` helper script, to avoid pasting secrets over and again!
 
-### 3. Create a read-only GitHub personal access token
+### 1. Generate an AWS token and key
 
-Create a [Github personal access token](https://github.com/settings/tokens) with the `read:org` scope.
+First make sure you have [`gds-cli`](/manual/access-aws-console.html) installed.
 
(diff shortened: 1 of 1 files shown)


source/manual/deploying-terraform.html.md
```diff
@@ -9,77 +9,78 @@ review_in: 3 months
...
-To follow these instructions you will need:
+**Always `plan` first, check that the output is what you expect, then `apply`.**
 
-* The name of the Terraform project which you want to deploy, for example `app-backend`.
-* The stackname of the Terraform project which you want to deploy. Usually, this is `blue` for `app-` projects and `govuk` for `infra-` projects.
-* Your GitHub username.
-* Your GitHub personal access token (see setup instruction above).
-* The name of your AWS role (`govuk-powerusers` or `govuk-adminstrators`).
-* A local copy of the `govuk-aws` Git repo, which contains the `deploy.rb` script. The example command assumes that your local repo is in the directory `~/govuk/govuk-aws`.
+Manually run the Jenkins job, pasting the token and key from the previous step.
```
issyl0: We shouldn't be recommending this way. The "use the deploy script" way to trigger the Jenkins job should be the default that people do. With the `./tools/deploy.rb` script on the end of the gds-cli command, people don't have to have tokens on their clipboards and accidentally paste them into Slack or somesuch.

source/manual/deploying-terraform.html.md
```diff
@@ -9,77 +9,78 @@ review_in: 3 months
...
-    ```
-    GITHUB_USERNAME=<your GitHub username> \
-      GITHUB_TOKEN=<your GitHub personal access token> \
-      gds aws govuk-integration-admin -- \
-      ~/govuk/govuk-aws/tools/deploy.rb blue app-backend integration plan
-    ```
+### 1. Create a read-only GitHub personal access token
 
-1. If your AWS session has expired, you'll be asked for your MFA (two-factor authentication) code.
+Create a [Github personal access token](https://github.com/settings/tokens) with the `read:org` scope.
```
issyl0:
```suggestion
Create a [GitHub personal access token](https://github.com/settings/tokens) with the `read:org` scope.
```

source/manual/deploying-terraform.html.md
issyl0: IMO copying and pasting three tokens six times (or even three times) is more effort to find the beginning and end, etc. etc. than adding a GitHub token once.

source/manual/deploying-terraform.html.md
benthorner: Fixed 👍

source/manual/deploying-terraform.html.md
benthorner: Having now copied and pasted tokens many times, I agree. However, I also found it more intuitive as a starting point, because now I know what the deploy script is doing and that makes me more confident when using it. I've added a caveat into the section that I'm hoping captures the discussion here. Does that address the issue for you?

source/manual/deploying-terraform.html.md
```diff
@@ -9,77 +9,80 @@ review_in: 3 months
...
-[infra-security project in govuk-aws-data](https://github.com/alphagov/govuk-aws-data/tree/master/data/infra-security).
+You can find which class of user you are [infra-security project in govuk-aws-data](https://github.com/alphagov/govuk-aws-data/tree/master/data/infra-security).
 
-### 2. Install `gds-cli`
+## Deploying on Jenkins
 
-See [instructions for installing `gds-cli`](/manual/access-aws-console.html).
+> **This is not the recommended approach.**
+>
+> It's slightly easier to start with, but requires you to push your changes to GitHub and you'll quickly want to switch to deploying locally (see the next section) to avoid pasting secrets over and again!
```
issyl0:
```suggestion
> It's slightly easier to start with, but you'll quickly want to switch to deploying locally (see the next section) to avoid pasting secrets over and again!
```

Every one of these approaches requires pushing to GitHub.

source/manual/deploying-terraform.html.md
```diff
@@ -9,77 +9,80 @@ review_in: 3 months
...
-* Your GitHub username.
-* Your GitHub personal access token (see setup instruction above).
-* The name of your AWS role (`govuk-powerusers` or `govuk-adminstrators`).
-* A local copy of the `govuk-aws` Git repo, which contains the `deploy.rb` script. The example command assumes that your local repo is in the directory `~/govuk/govuk-aws`.
+👉 [Deploy terraform directly from Jenkins][ci-deploy-jenkins]
 
-**Always `plan` first, check that the output is what you expect, then `apply`.**
 
-1. To start the Terraform plan job, run `deploy.rb` like this.
+## Deploying locally
```
issyl0: This still deploys on Jenkins, it's just less effort for users. I can't think how to word this better though.

source/manual/deploying-terraform.html.md
benthorner: Oops - fixed 👍

source/manual/deploying-terraform.html.md
issyl0: To me, "deploying locally" means "running `terraform apply` from your laptop", which we don't do because audit trails are good.

benthorner: Hmm...I see what you mean. The titles are pretty superficial, but I guess the function of `deploy.rb` is somewhat vague (it doesn't actually deploy anything by itself), so the best title is probably something equally mysterious...updated!

PR #2193, opened 2019-12-27
Update how to work on frontends locally

chao-xian:
Instructions did not include up to date docs when using govuk-docker.

However some frontend devs are no using docker yet, so we are keeping both sets of instructions for now.

### DIFF
--- a/source/manual/local-frontend-development.html.md
+++ b/source/manual/local-frontend-development.html.md
@@ -5,7 +5,7 @@ layout: manual_layout
 section: Frontend
 type: learn
 owner_slack: "#govuk-2ndline"
-last_reviewed_on: 2019-07-08
+last_reviewed_on: 2019-12-27
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manual/local-frontend-development.html.md
```diff
@@ -5,22 +5,62 @@ layout: manual_layout
...
+  $ make frontend
+  $ govuk-docker-up frontend-app
+  ```
+
+  (or you can run the last command from the frontend directory as just `govuk-docker-up`)
+6. changes should be ok for http://frontend.dev.gov.uk/help
+
+## Without docker
+
+A small numver of frontend devs prefer to not use Docker locally, so we are keeping these instructions for a little while longer.
```
ChrisBAshton: Typo:

```suggestion
A small number of frontend devs prefer to not use Docker locally, so we are keeping these instructions for a little while longer.
```

chao-xian: Thanks! Have opened another PR for that https://github.com/alphagov/govuk-developer-docs/pull/2194

PR #2203, opened 2020-01-07
Add page for subscription contents alert

deborahchua:
This is a new alert which replaces one of our `email-alert-api`
heathchecks. See https://github.com/alphagov/email-alert-api/pull/1096
for more information.

Trello card: https://trello.com/c/dtdY0kAK/1606-5-extract-the-subscriptioncontents-check-from-the-healthcheck-endpoint-in-email-alert-api-to-a-separate-icinga-check

### DIFF
--- a/source/manual/alerts/email-alert-api-unprocessed-subscription-contents.html.md
+++ b/source/manual/alerts/email-alert-api-unprocessed-subscription-contents.html.md
@@ -1,0 +1,40 @@
+---
+owner_slack: "#govuk-2ndline"
+title: 'Email Alert API: Unprocessed subscription contents'
+section: Icinga alerts
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2020-01-07
+review_in: 6 months
+---
+
+This means there there are subscription contents being created without emails
+associated with them, and implies that emails aren't being sent out.
+
+### Icinga checks
+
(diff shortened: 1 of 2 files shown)


source/manual/alerts/email-alert-api-unprocessed-subscription-contents.html.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Check which subscription contents this affects
+
+```ruby
+SubscriptionContent.where("subscription_contents.created_at < ?", 10.minutes.ago).where(email: nil).joins(:subscription).merge(Subscription.active)
+```
+
+Check the count, then run the above query again to see if the count has
+decreased. If it's decreasing, then it means that emails are going out and
+there's probably a lot being processed. You can also check the
```
issyl0: If the count is not increasing, what should people do to solve this?

issyl0: I hit the wrong button. 🤦‍♂

deborahchua: Thanks @issyl0 ,I have added a suggestion on what to do if it's not decreasing. It's only a guess but at least a starting point to investigate!

PR #2222, opened 2020-01-10
Add docs on how to add a new column to our Fastly logs and Athena

issyl0:
- We did this recently and it [went wrong](https://trello.com/c/7pAvfM8R/167-only-some-of-our-fastly-cdn-configuration-is-currently-automated), so let's document how to do it _correctly_ for the next people!

### DIFF
--- a/source/manual/logging.html.md
+++ b/source/manual/logging.html.md
@@ -147,8 +147,57 @@ AND "request_received"
 GROUP BY "status";
 ```
 
+### Adding a new field to the CDN logs
+
+Adding a new field to the CDN logs is a half manual, half automated
+process and is [tracked as tech debt](https://trello.com/c/7pAvfM8R/167).
+
+You should do this in the Integration environment first and wait until
+you see the logs coming through to Athena from Fastly. Only then
+should you change Staging and next Production. This helps us catch
+syntax errors and incidents early. Note that the `bouncer` logs are
+only available in Production.
+
+Firstly, make the manual changes via Terraform and a PR. This ensures that you've
+had two pairs of eyes on both the Athena/AWS Glue config changes and the eventual JSON
+you will manually put into the Fastly web UI:
+
+1. Edit the [`infra-fastly-logs` Terraform](https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/infra-fastly-logs/main.tf).
+1. Find the `aws_glue_catalog_table` resource for the Fastly logs you want to add a column to (`govuk_www`, `govuk_assets` or `bouncer`).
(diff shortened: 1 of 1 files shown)


issyl0:
Agreed. There's a load of duplication between these two pages - they both show different example queries, for example. I'll write up a card for consolidating them and put it in Platform Health's backlog. 😂

As for the rest of the changes here, I'll re-arrange the doc to suggest Terraform and non-Prod environments first. Thanks!

issyl0:
I'm almost there on re-writing these docs. What's the AWS Glue Crawler? I can't see any existing references to it.

kevindew:
> I'm almost there on re-writing these docs. What's the AWS Glue Crawler? I can't see any existing references to it.

It's a job that runs regularly (I think every 4 hours) that verifies if the files in s3 meet the schema defined for the table and identifying any new partitions, which it uses to decide whether a particular s3 file matches the schema and whether it needs to either create new tables / delete old ones (or other operations that may cause scary results)

There's a bunch of docs on it: https://docs.aws.amazon.com/glue/latest/dg/add-crawler.html but you can learn a good amount about it by looking at the outcome of a run (https://eu-west-1.console.aws.amazon.com/glue/home?region=eu-west-1#catalog:tab=crawlers while assuming a GOV.UK role)

source/manual/logging.html.md
```diff
@@ -147,8 +147,51 @@ AND "request_received"
...
+1. Click "Clone version xxx (active) to edit".
+1. Click "Logging" in the left hand menu.
+1. Find and click the link for the relevant log, for example "GOV.UK Fastly Logs S3 Bucket".
+1. Paste the new log format (found [in the list of available Fastly logs][fastly-logs-list]) into the "Log Format" text box.
+1. Click "Update".
+1. Click the purple "ACTIVATE" button in the top right corner to make the new configuration live.
+
+Both these sets of steps must be done! Check the S3 bucket and
+query Athena to see the added column and confirm that there's data for
+it. It can take up to half an hour for Athena to recognise that there have
```
kevindew: I think you may have got this half an hour from me from something different I was referring too. It can take up to half an hour for Fastly to send us through a log file with the changed data (we've set it to send them to us every 15 minutes but they seem to often come 15 minutes after the time they stopped logging).

For Athena to recognise the change in table that could potentially take up to 4 hours as that is how often the crawler worker runs. I don't know off the top of my head what happens though at the point of you running the terraform to add the new column. It might be that it is instantaneous and finds that in any data that has it or that it has to wait for the crawler to run which would be within 4 hours.

Sorry for vagueness I've not actually added a new column myself to know what problems it may cause. Simplest thing may be to just suggest that someone runs the crawler worker once they know Fastly is sending the right data.

issyl0: Done! Thanks for the good explanation!

PR #2225, opened 2020-01-13
Clarify release name and update last reviewed date

ChrisBAshton:
I just had to run this to get rid of a "high memory" alert in production. Was momentarily doubting whether the default pre-filled `release` value is the right one to use.

### DIFF
--- a/source/manual/restart-application.html.md
+++ b/source/manual/restart-application.html.md
@@ -4,8 +4,10 @@ title: Restart an application
 parent: "/manual.html"
 layout: manual_layout
 section: Deployment
-last_reviewed_on: 2019-07-03
+last_reviewed_on: 2020-01-13
 review_in: 12 months
 ---
 
 To restart an application go to the Deploy app jenkins job in [Carrenza](https://deploy.publishing.service.gov.uk/job/Deploy_App/build) or [AWS](https://deploy.blue.production.govuk.digital/job/Deploy_App/build), choose your app, the **current release** and select `app:hard_restart`.
+
+For avoidance of doubt, the current release is usually simply called "**release**".


source/manual/restart-application.html.md
```diff
@@ -4,8 +4,10 @@ title: Restart an application
...
 layout: manual_layout
 section: Deployment
-last_reviewed_on: 2019-07-03
+last_reviewed_on: 2020-01-13
 review_in: 12 months
 ---
 
 To restart an application go to the Deploy app jenkins job in [Carrenza](https://deploy.publishing.service.gov.uk/job/Deploy_App/build) or [AWS](https://deploy.blue.production.govuk.digital/job/Deploy_App/build), choose your app, the **current release** and select `app:hard_restart`.
+
+For avoidance of doubt, the current release is usually simply called "**release**".
```
cbaines: This should be OK for Production, but the same can't be said for Staging. Perhaps say "the current release for Production ..."?

ChrisBAshton: You mean because people like to try out branches on Integration and Staging? I didn't feel it needs that level of detail, but I could add a line about "If not on Production, check the name of the current release in the Release app".

Or are you suggesting that the latest 'master' build is _not_ tagged as `release` on staging?

kevindew: We don't always seem to have the release branch either - I've mostly stopped using it since it often doesn't exist for apps. We don't have one for Content Publisher for instance: https://github.com/alphagov/content-publisher/tree/release

cbaines: > You mean because people like to try out branches on Integration and Staging? I didn't feel it needs that level of detail, but I could add a line about "If not on Production, check the name of the current release in the Release app".

Well, I was just thinking if you're trying to restart an application in Staging, you might end up rolling back some code changes if you deploy what's deployed to production.

> Or are you suggesting that the latest 'master' build is _not_ tagged as `release` on staging?

Well, this will only be the case if there are no changes to be deployed, but it won't be the case if there are changes to be deployed.

ChrisBAshton: On reflection, it seems there's little value in this PR! Thanks for explaining the nuances around this.

PR #2226, opened 2020-01-13
Update docs with info for rotating IAM Access Keys

MahmudH:
Co-Authored-By: Alan Gabbianelli <alan.gabbianelli@digital.cabinet-office.gov.uk>
Co-Authored-By: Mahmud Hussain <mahmud.hussain@digital.cabinet-office.gov.uk>

Trello card: https://trello.com/c/cZQQM8Gu/1623-3-find-a-key-that-is-not-in-use-and-update-docs

### DIFF
--- a/source/manual/aws-iam-key-rotation.html.md
+++ b/source/manual/aws-iam-key-rotation.html.md
@@ -5,93 +5,141 @@ layout: manual_layout
 type: learn
 section: AWS
 owner_slack: "#govuk-developers"
-last_reviewed_on: 2019-10-31
+last_reviewed_on: 2020-01-20
 review_in: 6 months
 ---
 
 ## Overview
 
-GOV.UK uses [AWS IAM](https://docs.aws.amazon.com/en_pv/IAM/latest/UserGuide/id_users.html)
-accounts to manage applications access to our AWS infrastructure. The
-[Access Keys](https://docs.aws.amazon.com/en_pv/IAM/latest/UserGuide/id_credentials_access-keys.html)
-associated with these accounts need to be [rotated](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html#Using_RotateAccessKey)
-every 3 months.
+GOV.UK uses [AWS IAM] accounts to manage applications access to our AWS
+infrastructure. The [Access Keys] associated with these accounts need to be
+[rotated] every 3 months.
+
+Rotating AWS IAM keys requires permissions which are available with the `admin`,
+`internal-admin`, `platformhealth-poweruser` or `poweruser` roles. You'll be
+able to rotate keys only if you can assume one of these roles.
 
 ## How to rotate Access Keys
 
-> **Note**
->
-> Rotating AWS IAM keys requires permissions which are available with the
-> `admin`, `internal-admin`, `platformhealth-poweruser` or `poweruser` roles.
-> If you can't assume one of these roles, you won't be able to rotate keys.
-
-Log in to the [IAM Management Console](https://console.aws.amazon.com/iam/home?region=eu-west-1#/users)
-for the environment you want to rotate keys for and you'll see a list of users
-and the age of their access keys.
+Log in to the [IAM Management Console] in the environment you want to rotate
+keys for and you'll see a list of users and the age of their access keys.
 
 ### 1. Find users access key
 
-Click on the user whose key you want to rotate. Click on `Security Credentials`. Make a note of the Access Key ID.
+- Click on the user whose key you want to rotate.
+- Click on the `Security Credentials` tab.
 
-### 2. Find where the key is used
+Users should always only have one access key. The only time a user should have
+two keys is when we are in the process of rotating the key that is in use. For
+this reason, AWS allow users to have a maximum of two Access Keys at the same
+time.
 
-Enter [govuk-secrets/puppet](https://github.com/alphagov/govuk-secrets/tree/master/puppet) in your machine.
-Run one of the [Puppet AWS common actions](https://github.com/alphagov/govuk-secrets/tree/master/puppet_aws#common-actions) and search for the `Access Key ID` in the result.
-
-> **Note**
-> The command you should run is related to the environment you need to target.
-> For example, to edit global staging credentials in the `govuk::apps` namespace, you need to run `bundle exec rake 'eyaml:edit[staging,apps]'`.
-
-Now enter [govuk-secrets/puppet_aws](https://github.com/alphagov/govuk-secrets/tree/master/puppet_aws), run the same command again and search for the Access Key ID in the result.
-
-Keys may also be defined in more then one place, for example the hieradata values
-`govuk::apps::support::aws_access_key_id` and
-`govuk::apps::support_api::aws_access_key_id` contain the same key.
-Keys may also be shared across AWS and Carrenza environments.
-
-If the key is not present in the hieradata, it might be present in the control panel of one of the 3rd party services, such as Fastly or Logit.
-
-### 3. Create and deploy new key
-
-Click the `Create access key` button in `Security credentials` of the Amazon IAM console.
-Replace the old `Access Key ID` and `Secret access key`.
-in [govuk-secrets](https://github.com/alphagov/govuk-secrets/tree/master) or the
-3rd party service with the new ones you just created.
-If the change is in `govuk-secrets`, redeploy [govuk-puppet](https://github.com/alphagov/govuk-puppet) so it picks up the new key.
-This may take up to 30 minutes to take effect. To avoid waiting, you can run `govuk_puppet --verbose` in the machine that uses that key.
-
-Users can have a maximum of 2 access keys. If a user already has 2 keys, one
-key will need to be removed before a new key can be added. If this key is still
-active, you'll need to
-[find where the key is used](#1-find-where-the-key-is-used) and then
-[remove the key](#4-remove-the-old-key).
-
-### 4. Ensure new key is being used
-
-Once the new key has been deployed, wait until the “Last used” values change
-to show that the new key is being used and the old key is not used. Infrequently
-used keys may take a while to update, and updates can take a few minutes to show
-in the control panel.
-
-If the last used time of the old key continues to change, this indicates it
-has not been completely removed so you'll need to
-[find where the key is used](#1-find-where-the-key-is-used) before it can be
-[safely deleted](#4-remove-the-old-key).
-
-### 5. Remove the old key
-
-The old key can either be deleted or made inactive. Making the key inactive will
-prevent the key from being used, but it can quickly be made active again. This
-is useful if you are not confident that all uses of the key have been updated.
-In production systems, this could prevent a minor outage scaling into an
-incident.
+- Make a note of the `Access Key ID`.
+- Proceed to [find out if the key is used](#2-find-out-if-the-key-is-used).
 
 > **Note**
 >
-> Notify [2ndline](/manual/2nd-line.html) before deleting or making any production
-> key inactive.
+> If a user has two keys, the first task is to remove one of them.
+>
+> If one key [is not in use](#2-find-out-if-the-key-is-used), you can
+> [remove it](#6-remove-the-old-key) and rotate the other key (keep reading to
+> see how).
+>
+> If both keys are in use, you need to pick one and
+> [replace all uses of it](#3-find-where-the-key-is-used) with the other key.
+> Then, once it's no longer in use, you can [remove it](#6-remove-the-old-key)
+> and rotate the other key (keep reading to see how).
+
+### 2. Find out if the key is used
+
+- Check the `Last used` value.
+- If the value is a date,
+[find where the key is used](#3-find-where-the-key-is-used).
+- If it's `N/A`, you can safely [remove the key](#6-remove-the-old-key) as it
+isn't used.
+
+### 3. Find where the key is used
+
+When searching for keys, keep in consideration that they may be present in
+[govuk-secrets] or in the control panel of 3rd party services, such as Fastly or
+Logit. Keys may also be shared across AWS and Carrenza environments.
+
+Firstly, search for the key in [govuk-secrets]:
+
+- Enter [govuk-secrets/puppet] in your machine.
+- Run the appropriate [Puppet AWS common actions] considering the following:
+    - The command to run is related to the environment you need to target. For
+    example, to edit global staging credentials in the `govuk::apps` namespace,
+    you need to run `bundle exec rake 'eyaml:edit[staging,apps]'`.
+    - You may need to run more than one command. For example, to edit the global
+    integration credentials both in and out of the `govuk::apps` namespace, you
+    need run both `$ bundle exec rake 'eyaml:edit[integration]'` and
+    `$ bundle exec rake 'eyaml:edit[integration,apps]'`.
+- Search for the `Access Key ID` in the result.
+
+Repeat the previous steps for [govuk-secrets/puppet_aws]:
+
+- Enter [govuk-secrets/puppet_aws].
+- Run the same command(s) again.
+- Search for the `Access Key ID` in the result.
+
+Keys may also be defined in more then one place. For example, the hieradata
+values `govuk::apps::support::aws_access_key_id` and
+`govuk::apps::support_api::aws_access_key_id` contain the same key.
+
+Secondly, search for the key in the control panel of 3rd party services we use,
+for example, Fastly or Logit.
+
+If it isn't present in `govuk-secrets` hieradata, nor in any 3rd party services,
+you can safely [remove the key](#6-remove-the-old-key).
+
+### 4. Create and deploy new key
+
+- Click `Create access key` in `Security credentials`.
+- If the key to change was in a 3rd party service:
+  - Replace the old `Access Key ID` and `Secret access key` with the new ones.
+- If the key to change was in `govuk-secrets`:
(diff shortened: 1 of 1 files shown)


source/manual/aws-iam-key-rotation.html.md
```diff
@@ -51,6 +51,10 @@ Keys may also be shared across AWS and Carrenza environments.
 
 If the key is not present in the hieradata, it might be present in the control panel of one of the 3rd party services, such as Fastly or Logit.
 
+> **Note**
+> If the key is not present in `govuk-secrets` nor in any other 3rd party services,
+> you can [remove the old key](#5-remove-the-old-key)
```
cbaines: If this relates to the key not being in use, then it might be that the user needs removing, not just the key right?

MahmudH: The user may have a new key and the old key is no longer present

AlanGabbianelli:
```suggestion
> you can [remove the old key](#5-remove-the-old-key).
```
😄

MahmudH: 😄 👍

cbaines: Right, so this relates to the limitation of having up to 2 keys per user right? If you've got two keys, and one is unused, you'll need to remove that unused key prior to creating a new key.

If that is the case, then I guess you don't need to check if it's present anywhere right? You can just trust the AWS interface when it tells you it's not in use?

AlanGabbianelli: For what I understand about AWS users and keys, this doesn't only relate to the case where a user already has 2 keys. There could also be users with only one key and that key could be not in use.

In that case, I'm not sure if we can just trust the AWS interface when it tells us it's not in use. Isn't it possible that we have that key in puppet but not have used it?

cbaines: > For what I understand about AWS users and keys, this doesn't only relate to the case where a user already has 2 keys. There could also be users with only one key and that key could be not in use.
> 
> In that case, I'm not sure if we can just trust the AWS interface when it tells us it's not in use. Isn't it possible that we have that key in puppet but not have used it?

In the case where the single acesss key isn't being used, we should be able to trust the AWS interface, but it might be good to check govuk-secrets as well just to see if there's redundant configuration that can be removed.

In this scenario, ideally one would figure out if the user is still relevant, or redundant. It looks like there are users without keys that are being used, so unfortunately it's not clear cut.

I don't know how far we want to go with these docs, but I think it's wise to be quite specific.

AlanGabbianelli: > I don't know how far we want to go with these docs, but I think it's wise to be quite specific.

I agree. Is there a particular change you have in mind?

AlanGabbianelli: @cbaines unless there's something specific you meant by "I think it's wise to be quite specific", are you happy to move this forward?

erino: We can trust the user interface to tell us if a key is in use. If a key is in puppet and not in use then there's a good chance that the key should have been removed as part of a refactor but wasn't.

I think that If we find an unused/irrelevant user then it should be removed. But maybe instead the users could be added to a list of users to investigate and remove?

cbaines: > @cbaines unless there's something specific you meant by "I think it's wise to be quite specific", are you happy to move this forward?

I can't add any clarity as I don't know what the intent is here (apart from "Update the docs...").

I'm OK with merging this change, but at least when I read it, I'm not sure when that additional conditional instruction to remove a key would apply. I can't work out if it's truly a useful side note, or something more important that should be included in the main block of text.

source/manual/aws-iam-key-rotation.html.md
```diff
@@ -5,93 +5,131 @@ layout: manual_layout
...
 ## Overview
 
-GOV.UK uses [AWS IAM](https://docs.aws.amazon.com/en_pv/IAM/latest/UserGuide/id_users.html)
-accounts to manage applications access to our AWS infrastructure. The
-[Access Keys](https://docs.aws.amazon.com/en_pv/IAM/latest/UserGuide/id_credentials_access-keys.html)
-associated with these accounts need to be [rotated](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html#Using_RotateAccessKey)
-every 3 months.
+GOV.UK uses [AWS IAM] accounts to manage applications access to our AWS
+infrastructure. The [Access Keys] associated with these accounts need to be
+[rotated] every 3 months.
```
cbaines: Nice, I like the shorter line length :+1:

source/manual/aws-iam-key-rotation.html.md
```diff
@@ -5,93 +5,131 @@ layout: manual_layout
...
 
 ### 1. Find users access key
 
-Click on the user whose key you want to rotate. Click on `Security Credentials`. Make a note of the Access Key ID.
+- Click on the user whose key you want to rotate.
+- Click on the `Security Credentials` tab.
 
-### 2. Find where the key is used
+Users should always only have one access key. The only time a user should have
+two keys is when we are in the process of rotating the old one. For this reason,
```
cbaines: I'd perhaps say "rotating the key that is in use", rather than "rotating the old one", at least that makes more sense to me.

source/manual/aws-iam-key-rotation.html.md
```diff
@@ -5,93 +5,131 @@ layout: manual_layout
...
+AWS allow users to have a maximum of two Access Keys at the same time.
 
-Enter [govuk-secrets/puppet](https://github.com/alphagov/govuk-secrets/tree/master/puppet) in your machine.
-Run one of the [Puppet AWS common actions](https://github.com/alphagov/govuk-secrets/tree/master/puppet_aws#common-actions) and search for the `Access Key ID` in the result.
+If this user has no keys, consider deleting it if it's not needed.
```
cbaines: It's unlikely that someone reading this documentation will come across this scenario, because they should be looking at a user which has an old key.

Maybe say something like "If the user has a single key, that appears not to be in use, consider removing the user"?

source/manual/aws-iam-key-rotation.html.md
```diff
@@ -5,93 +5,131 @@ layout: manual_layout
...
-> The command you should run is related to the environment you need to target.
-> For example, to edit global staging credentials in the `govuk::apps` namespace, you need to run `bundle exec rake 'eyaml:edit[staging,apps]'`.
+If this user has only one key, make a note of the `Access Key ID` and proceed to
+[find out if the key is used].
+
+If this user has two keys, you'll need to follow this guide twice. The first
+time, you'll need to consider one key as the old one to rotate and the other key
+as if it was a new one just created. The second time, you'll follow the guide as
+normal because the user will then have only one key. Now, choose which key is
+the one you are going to rotate and apply the following steps to that key.
```
cbaines: I think this is pretty much right, but could maybe be slightly clearer.

I'd perhaps stress that the first task is to remove one of the keys. If one isn't in use, then that's easy, as you just remove that key. If they're both in use, then you need to pick one, and then replace uses of it with the other key. Then, once it's no longer in use, you can remove it. I'm not sure that referring generally to other parts of this document will be helpful, as I think it's more likely to cause confusion.

source/manual/aws-iam-key-rotation.html.md
```diff
@@ -5,93 +5,131 @@ layout: manual_layout
...
+
+### 2. Find out if the key is used
+
+- Check the `Last used` value.
+- If the value is a date, [find where the key is used].
```
cbaines: What if it's a really old date? Could it be that the key is no longer in use?

source/manual/aws-iam-key-rotation.html.md
```diff
@@ -5,93 +5,131 @@ layout: manual_layout
...
+- If it's `N/A`, you can safely [remove the key] as it was never used.
```
cbaines: At this point, do you continue following the guide?

source/manual/aws-iam-key-rotation.html.md
```diff
@@ -5,93 +5,131 @@ layout: manual_layout
...
-Now enter [govuk-secrets/puppet_aws](https://github.com/alphagov/govuk-secrets/tree/master/puppet_aws), run the same command again and search for the Access Key ID in the result.
+### 3. Find where the key is used
 
-Keys may also be defined in more then one place, for example the hieradata values
-`govuk::apps::support::aws_access_key_id` and
+- Enter [govuk-secrets/puppet] in your machine.
+- Run the appropriate [Puppet AWS common actions] considering the following:
+  - The command to run is related to the environment you need to target. For
+  example, to edit global staging credentials in the `govuk::apps` namespace,
+  you need to run `bundle exec rake 'eyaml:edit[staging,apps]'`.
```
cbaines: Maybe indent the above two lines a little bit more, as that would make it a little easier to read.

source/manual/aws-iam-key-rotation.html.md
AlanGabbianelli: Yes, I thought about that too, but I would leave it anyway because if someone is in the process of rotating an old key and wats to have a look at the other keys and users at the same time, at least they know what they should do.

source/manual/aws-iam-key-rotation.html.md
```diff
@@ -5,93 +5,131 @@ layout: manual_layout
...
+
+- Enter [govuk-secrets/puppet_aws].
+- Run the same command(s) again.
+- Search for the `Access Key ID` in the result.
+
+Keys may also be defined in more then one place. For example, the hieradata
+values `govuk::apps::support::aws_access_key_id` and
 `govuk::apps::support_api::aws_access_key_id` contain the same key.
+
 Keys may also be shared across AWS and Carrenza environments.
```
cbaines: This line might be redundant as you say above to "Repeat the previous steps for [govuk-secrets/puppet_aws]:".

source/manual/aws-iam-key-rotation.html.md
AlanGabbianelli: Yes, but you jump to `[remove the key]` that I linked.

source/manual/aws-iam-key-rotation.html.md
```diff
@@ -5,93 +5,131 @@ layout: manual_layout
...
 
-If the key is not present in the hieradata, it might be present in the control panel of one of the 3rd party services, such as Fastly or Logit.
+If the key is not present in the hieradata, it might be present in the control
+panel of one of the 3rd party services, such as Fastly or Logit.
 
-### 3. Create and deploy new key
+> **Note**
+> If the key is not present in `govuk-secrets` nor in any other 3rd party
+> services, you can [remove the key].
```
cbaines: It might be good to make this a more general comment on what to do if you haven't managed to find the key. It could be that AWS says the key is in use, or that AWS says the key hasn't been used recently (or maybe the reader wouldn't get to this point if that's the case).

source/manual/aws-iam-key-rotation.html.md
cbaines: I know we've talked about that, but I'd be wary that this might be a bit beyond the scope of the document. I'm also unsure how risky this is, as I haven't tried deleting users in AWS which are in use, so I'm not sure how easy it is to break things.

source/manual/aws-iam-key-rotation.html.md
AlanGabbianelli: It could be. It would mean that we've deleted the key but not removed it from AWS. 🤷‍♂
I'm not sure where to draw the line here though (a month? a year?) as some of our keys are apparently very rarely used.

source/manual/aws-iam-key-rotation.html.md
AlanGabbianelli: Screenshot 2020-01-20 at 10 41 46

In the rendered page it's clear but I can indent them a bit more to make it clearer in "the code" as well.

source/manual/aws-iam-key-rotation.html.md
cbaines: I think the line for "how old a key is that it isn't in use" depends on how often the key should be used, so maybe say something about that?

source/manual/aws-iam-key-rotation.html.md
cbaines: I'd perhaps try and make that more explicit, maybe say "you can skip ahead and [remove the key] ...", otherwise there's no distinction in the text between a link for providing additional information, and one that indicates that you should stop reading the document linearly.

source/manual/aws-iam-key-rotation.html.md
AlanGabbianelli: Ok, let's take it out entirely for the moment. We can then revisit this in a user-specific document or somewhere else if needed.

source/manual/aws-iam-key-rotation.html.md
AlanGabbianelli: That makes sense but I'm not sure how to find out how often a key should be used. Do we need to first find the key and - based on what it's used for - understand how often it should be used? 🤔

source/manual/aws-iam-key-rotation.html.md
AlanGabbianelli: I'd rather make it clear what steps we're talking about.

source/manual/aws-iam-key-rotation.html.md
AlanGabbianelli: I agree, I'm not happy with this part myself and I re-wrote it several times in different ways. In the end, I decided to put down something anyway and see other's inputs.

I find it difficult to convey this message. I know what needs to be done but it isn't easy to explain, especially because to really understand this point, the reader should either be already familiar with the process of rotating keys or read this whole guide.

How about this? 👇

```
If this user has two keys, the first task is to remove one of them. If one isn't
in use, you can remove that key and proceed to rotate the other. If they're both
in use, you need to pick one and replace all uses of it with the other key.
Then, once it's no longer in use, you can remove it and proceed to rotate the
other.
```

source/manual/aws-iam-key-rotation.html.md
AlanGabbianelli: It's possible to have a key in AWS that is not in use anywhere, even if the key has a `Last used` date. Someone might have removed because it was no longer needed and forgotten to remove it from AWS. If that's the case, the dev reading this guide would arrive at this point of the guide and not find the key anywhere. I wanted to make it clear that, in this case, it's safe to simply delete the key. I put it at the end of `### 3. Find where the key is used` because it's the very last possibility (we should first try and find the key) and I made it a note because so it stands out.

source/manual/aws-iam-key-rotation.html.md
cbaines: I think that's better.

I'm uncertain about the way this mentions rotating specific keys though: "proceed to rotate the other". I guess I know that I think this means that you should continue reading and following the documentation, as that's the subject of the document, but I think it could be read as "rotate" is something you need to do now, as part of this step.

But while I'd prefer not saying "rotate", and perhaps just say instead "proceed to the next step.", I think it's OK this way as well.

source/manual/aws-iam-key-rotation.html.md
cbaines: I guess it's hopefully apparent once you understand how the key is used. I'm not sure it's possible (or useful) to be more specific than that.

source/manual/aws-iam-key-rotation.html.md
cbaines: I'm not sure I follow.

But actually, now I'm reading this again, I think I misinterpreted this the first time I read it. If this is trying to say that the key might be present in both the Carrenza and AWS hieradata, it would be probably more useful at the top of this section, as that's something that would be useful to know prior to diving in to the hieradata.

AlanGabbianelli: Ah, sorry, I thought the comment was related to a different line. I'll move this to the start of this step.

AlanGabbianelli: I moved it at the top and added an introduction.

source/manual/aws-iam-key-rotation.html.md
AlanGabbianelli: I made this the last point of this step. No need to be a note.

source/manual/aws-iam-key-rotation.html.md
```diff
@@ -5,93 +5,141 @@ layout: manual_layout
...
-associated with these accounts need to be [rotated](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html#Using_RotateAccessKey)
-every 3 months.
+GOV.UK uses [AWS IAM] accounts to manage applications access to our AWS
+infrastructure. The [Access Keys] associated with these accounts need to be
+[rotated] every 3 months.
 
-## How to rotate Access Keys
+Rotating AWS IAM keys requires permissions which are available with the `admin`,
+`internal-admin`, `platformhealth-poweruser` or `poweruser` roles. If you can't
+assume one of these roles, you won't be able to rotate keys.
```
erino: What about changing this role so it doesn't use so many negatives.

> Only users who can assume these roles will be able to rotate keys.

AlanGabbianelli: Yeah, can do

issyl0:
Shall we 🚢 this and iterate later if @huwd finds anything confusing?

AlanGabbianelli:
Yes @issyl0, do you want to ✅?

PR #2231, opened 2020-01-14
Fix link to legacy Ops manual

ChrisBAshton:


issyl0:
Thanks! Is it worth saying why we've moved it?

issyl0:
There's also a link to the repo in:

```
source/manual/where-to-find-what.html.md
39:<https://github.com/alphagov/govuk-legacy-opsmanual>.
```

ChrisBAshton:
I don't think it's worth explaining where it used to live 🤷‍♂ so long as people can still get the info they need. Seems unnecessary to explain it used to live in a private repository but was moved to reduce maintenance overhead and free up a private repository slot.

Oops - good spot on the other link - I'll push up a change after lunch!

PR #2232, opened 2020-01-15
rebooting-machines: Add instructions for rebooting AWS Cache machines

issyl0:
- These didn't have any special instructions and, bleary-eyed yesterday
  morning, I didn't remember that there was an intricate and tedious
  process across all eight machines.
- So that everyone else can have this knowledge to hand, document the
  process.
- Took the instructions from the "restarting router" long-running card
  we had when there was the router bug.
- Platform Health has a card to make there be the standard number of
  cache machines rather than the current eight, so soon we can change
  the number in step 2 and the reboots become _a little_ less tedious.

https://trello.com/c/QWiMOe9Q/1686-document-how-to-safely-reboot-cache-machines-in-aws

### DIFF
--- a/source/manual/rebooting-machines.html.md
+++ b/source/manual/rebooting-machines.html.md
@@ -45,6 +45,8 @@ and RE will be able to help in these cases. It also means RE can investigate
 any problems so it doesn't happen again and we have confidence in our ability
 to reboot machines in AWS.
 
+There's a [section on rebooting `cache` machines in AWS](#rebooting-cache-machines-in-aws).
+
 ## Unattended upgrades
 
 Machines are configured with [automatic security updates](https://help.ubuntu.com/community/AutomaticSecurityUpdates#Using_the_.22unattended-upgrades.22_package) which install security updates overnight. Sometimes these require a reboot in order to become active.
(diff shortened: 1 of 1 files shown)


source/manual/rebooting-machines.html.md
```diff
@@ -103,6 +105,24 @@ There is a Fabric task to schedule a machine for downtime in Nagios for
 > manual entry on [adding disks](/manual/adding-disks-in-vcloud.html) for
 > more info.
 
+## Rebooting Cache machines in AWS
+
+The `cache` machines run the `router` app which handles live user traffic.
+To safely reboot them without serving too many errors to users, we must
+remove them from the AWS load balancer target group before rebooting:
```
cbaines: `group` should probably be plural, as there are more than one.

source/manual/rebooting-machines.html.md
```diff
@@ -103,6 +105,24 @@ There is a Fabric task to schedule a machine for downtime in Nagios for
...
+
+1. Login to AWS Console. Switch to the prod environment (`gds aws govuk-production-<your-role> -l`)
```
cbaines: I'd probably say switch to the "right" or "relevant" environment, as you'll need to be in Staging for rebooting machines in Staging.

source/manual/rebooting-machines.html.md
```diff
@@ -103,6 +105,24 @@ There is a Fabric task to schedule a machine for downtime in Nagios for
...
+1. Find the [Instance ID](https://eu-west-1.console.aws.amazon.com/ec2/home?region=eu-west-1#Instances:sort=desc:launchTime) of the critical machine(s) (probably all 8 `blue-cache` machines)
+1. `ssh [ip].eu-west-1.compute.internal` (find this from the reboots required alert listing)
+1. Remove the machine from the following [Target Groups](https://eu-west-1.console.aws.amazon.com/ec2/home?region=eu-west-1#TargetGroups:sort=targetGroupName):
+   1. cache-assets-origin
+   1. cache-www
+   1. cache-www-origin
+1. Check traffic again (should only be Smokey healthchecks now).
```
cbaines: Does step (1, well there all 1, the 3rd step) involve checking the traffic? (as I don't think it's mentioned prior to this point)

issyl0: These were amended from the Trello card for the router restarts. No, the "again" should go.

issyl0:
Thanks Chris! Applied all the suggested changes.

PR #2285, opened 2020-02-04
Add playbook entry for check_cdn_log_s3_freshness Icinga alert.

sengi:
See alphagov/govuk-puppet#10113.

### DIFF
--- a/source/manual/alerts/cdn-log-freshness.html.md
+++ b/source/manual/alerts/cdn-log-freshness.html.md
@@ -1,0 +1,41 @@
+---
+owner_slack: "#govuk-2ndline"
+title: CDN logs from Fastly not appearing in S3
+section: Icinga alerts
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2020-02-27
+review_in: 6 months
+---
+
+## Meaning and impact
+
+CDN logs from Fastly might not be getting uploaded to the
+`govuk-{environment}-fastly-logs` S3 bucket.
+
+## Possible causes
+
+ * Configuration change to Fastly may have stopped the log upload from working.
+ * Access controls might have changed on the S3 bucket.
+ * There might be a problem with the alert itself. It was added in Feb 2020.
+
+## Investigation
+
+List the S3 bucket to see what log files are present for today's date.
+Depending on your level of access, you might need to replace `admin` with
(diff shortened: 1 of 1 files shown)


sengi:
Oh yikes I just found https://docs.publishing.service.gov.uk/manual/alerts/logs-are-not-being-received-from-the-cdn.html which is very similar, but it turns out we have two log ingestion pipelines which both take the Fastly logs. Apparently that's for the other one (the non-S3 one).

issyl0:
We should probably ditch the non-S3 one. It pre-dates the S3 and Athena stuff and feels silly to have both - if Amazon goes down we've got bigger problems. 🤔

issyl0:
Also the docs for the manual way still reference the "deployment" repo, so they're very old.

issyl0:
So I looked into this a bit further and I think we _can_ remove the non-AWS code and docs.

The ["Fastly error rate"](https://docs.publishing.service.gov.uk/manual/alerts/fastly-error-rate.html) alert [pulls from the Fastly APIs](https://github.com/alphagov/govuk-puppet/blob/fad3d3b903389a0e145e47d27bc8f7708bfa41f6/modules/collectd/manifests/plugin/cdn_fastly.pp) via the [collectd-cdn](https://github.com/gds-operations/collectd-cdn/blob/master/collectd_cdn/fastly.py) plugin.

sengi:
> So I looked into this a bit further and I think we _can_ remove the non-AWS code and docs.
> 
> The ["Fastly error rate"](https://docs.publishing.service.gov.uk/manual/alerts/fastly-error-rate.html) alert [pulls from the Fastly APIs](https://github.com/alphagov/govuk-puppet/blob/fad3d3b903389a0e145e47d27bc8f7708bfa41f6/modules/collectd/manifests/plugin/cdn_fastly.pp) via the [collectd-cdn](https://github.com/gds-operations/collectd-cdn/blob/master/collectd_cdn/fastly.py) plugin. It doesn't use the files on the monitoring machine disk.

Awesome. Thanks, Issy! I'll follow up with a PR to remove the old stuff then.

issyl0:
The other, confusing docs are gone now.

sengi:
> The other, confusing docs are gone now.

Thanks!

PR #2305, opened 2020-02-11
Update "Our content delivery network (CDN)" page

injms:
- Update links to examples in config VCL
 - Update link to page about contacting Reliability Engineering
 - Update Bouncer CDN link to AWS from Carrenza

### DIFF
--- a/source/manual/cdn.html.md
+++ b/source/manual/cdn.html.md
@@ -5,19 +5,16 @@ section: CDN & Caching
 type: learn
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2019-07-23
+last_reviewed_on: 2020-02-11
 review_in: 6 months
 ---
 
-GOV.UK uses Fastly as a CDN. Citizen users aren't accessing GOV.UK servers
-directly, they connect via the CDN. This is better because:
+GOV.UK uses Fastly as a CDN. Citizen users aren't accessing GOV.UK servers directly, they connect via the CDN. This is better because:
 
-- The CDN "edge nodes" (webservers) are closer to end users. Fastly has
-  servers all around the world but our "origin" servers are only in the UK.
+- The CDN "edge nodes" (webservers) are closer to end users. Fastly has servers all around the world but our "origin" servers are only in the UK.
 - It reduces load on our origin. Fastly uses Varnish to cache responses.
 
-The CDN is responsible for retrying requests against the
-[static mirror](/manual/fall-back-to-mirror.html).
+The CDN is responsible for retrying requests against the [static mirror](/manual/fall-back-to-mirror.html).
 
 ![image](images/cdn-mirror-configuration.png)
 
@@ -28,13 +25,9 @@ Most of the CDN config is versioned and scripted:
 - [CDN configuration](https://github.com/alphagov/govuk-cdn-config/)
 - [CDN config secrets](https://github.com/alphagov/govuk-cdn-config-secrets)
 
-These are deployed to [integration][integration_cdn], [staging][staging_cdn]
-and [production][production_cdn].
+These are deployed to [integration][integration_cdn], [staging][staging_cdn] and [production][production_cdn].
 
-Some configuration isn't scripted, such as logging. The www, bouncer and assets
-services sends logs to S3 and stream them to `monitoring-1`. These logging
-endpoints are configured directly in the Fastly UI. There is
-[documentation](/manual/query-cdn-logs.html) on how to query the CDN logs.
+Some configuration isn't scripted, such as logging. The www, bouncer and assets services send logs to S3 and stream them to `monitoring-1`. These logging endpoints are configured directly in the Fastly UI. There is [documentation on how to query the CDN logs](/manual/query-cdn-logs.html).
 
 [integration_cdn]: https://deploy.integration.publishing.service.gov.uk/job/Deploy_CDN/
 [staging_cdn]: https://deploy.staging.publishing.service.gov.uk/job/Deploy_CDN/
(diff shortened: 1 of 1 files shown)


source/manual/cdn.html.md
```diff
@@ -28,55 +25,37 @@ Most of the CDN config is versioned and scripted:
...
 
-These are deployed to [integration][integration_cdn], [staging][staging_cdn]
-and [production][production_cdn].
+These are deployed to [integration][integration_cdn], [staging][staging_cdn] and [production][production_cdn].
 
-Some configuration isn't scripted, such as logging. The www, bouncer and assets
-services sends logs to S3 and stream them to `monitoring-1`. These logging
-endpoints are configured directly in the Fastly UI. There is
-[documentation](/manual/query-cdn-logs.html) on how to query the CDN logs.
+Some configuration isn't scripted, such as logging. The www, bouncer and assets services sends logs to S3 and stream them to `monitoring-1`. These logging endpoints are configured directly in the Fastly UI. There is [documentation](/manual/query-cdn-logs.html) on how to query the CDN logs.
```
ChrisBAshton:
```suggestion
Some configuration isn't scripted, such as logging. The www, bouncer and assets services send logs to S3 and stream them to `monitoring-1`. These logging endpoints are configured directly in the Fastly UI. There is [documentation on how to query the CDN logs](/manual/query-cdn-logs.html).
```

source/manual/cdn.html.md
```diff
@@ -28,55 +25,37 @@ Most of the CDN config is versioned and scripted:
...
 [integration_cdn]: https://deploy.integration.publishing.service.gov.uk/job/Deploy_CDN/
 [staging_cdn]: https://deploy.staging.publishing.service.gov.uk/job/Deploy_CDN/
 [production_cdn]: https://deploy.publishing.service.gov.uk/job/Deploy_CDN/
 
 ## Fastly Caching
 
-The main www.gov.uk cache is
-[Varnish](https://varnish-cache.org/docs/2.1/index.html), which Fastly run for
-us.
+The main www.gov.uk cache is [Varnish](https://varnish-cache.org/docs/2.1/index.html), which Fastly run for us.
```
ChrisBAshton:
```suggestion
The main www.gov.uk cache is [Varnish](https://varnish-cache.org/docs/2.1/index.html), which Fastly runs for us.
```

source/manual/cdn.html.md
```diff
@@ -28,55 +25,37 @@ Most of the CDN config is versioned and scripted:
...
 
 Varnish lets us configure our caching logic with VCL (Varnish config language).
 
-It also lets us do fancy things, like [only allowing connections to staging from permitted IPs](https://github.com/alphagov/govuk-cdn-config/blob/master/vcl_templates/www.vcl.erb#L193), [forcing SSL](https://github.com/alphagov/govuk-cdn-config/blob/master/vcl_templates/www.vcl.erb#L214) and [blocking IP addresses](https://github.com/alphagov/govuk-cdn-config/blob/master/vcl_templates/www.vcl.erb#L200), among other things.
+It also lets us do fancy things, like [only allowing connections to staging from permitted IPs](https://github.com/alphagov/govuk-cdn-config/blob/master/vcl_templates/www.vcl.erb#L202), [forcing SSL](https://github.com/alphagov/govuk-cdn-config/blob/master/vcl_templates/www.vcl.erb#L222) and [blocking IP addresses](https://github.com/alphagov/govuk-cdn-config/blob/master/vcl_templates/www.vcl.erb#L208), among other things.
```
ChrisBAshton: Is it worth linking to a specific commit hash rather than `master`, as this is likely to change again in future?

source/manual/cdn.html.md
```diff
@@ -95,44 +74,28 @@ This will give you various debugging headers that may be useful:
...
-list of Fastly IPs to match the ones listed by Fastly.
+We have [a Jenkins job "Check CDN IP Ranges"][check-cdn-ip-ranges] which will start to fail if our Fastly IPs don't match the IPs returned from the Fastly API. If you see this alert, you can [let Reliability Engineering know][raise-with-re] and they will update our list of Fastly IPs to match the ones listed by Fastly.
 
-Updating the firewall rules in Carrenza with new Fastly IPs used to be done by
-committing the change to the govuk-provisioning repo and to then deploy the
-firewall through a jenkins job. This process is broken at the moment since the
-code base has diverged from the state of the firewall, while this is remedied
-we have to add the new rules manually, this is how to do it:
+Updating the firewall rules in Carrenza with new Fastly IPs used to be done by committing the change to the govuk-provisioning repo and to then deploy the firewall through a jenkins job. This process is broken at the moment since the code base has diverged from the state of the firewall, while this is remedied we have to add the new rules manually, this is how to do it:
```
ChrisBAshton:
```suggestion
Updating the firewall rules in Carrenza with new Fastly IPs used to be done by committing the change to the govuk-provisioning repo and to then deploy the firewall through a jenkins job. This process is broken at the moment since the code base has diverged from the state of the firewall. While this is being remedied, we have to add the new rules manually; this is how to do it:
```

source/manual/cdn.html.md
```diff
@@ -180,49 +143,36 @@ done
 [firewall rules]: https://github.com/alphagov/govuk-provisioning/blob/master/vcloud-edge_gateway/vars/production_carrenza_vars.yaml
 [vcl_config]: https://github.com/alphagov/govuk-cdn-config/
 [check-cdn-ip-ranges]: https://deploy.publishing.service.gov.uk/job/Check_CDN_IP_Ranges/
-[raise-with-re]: raising-issues-with-reliability-engineering.html
+[raise-with-re]: https://docs.publishing.service.gov.uk/manual/raising-issues-with-reliability-engineering.html
```
ChrisBAshton:
```suggestion
[raise-with-re]: /manual/raising-issues-with-reliability-engineering.html
```

Links to dev docs tend to not have the domain in them. (We should do the same below).

source/manual/cdn.html.md
```diff
@@ -180,49 +143,36 @@ done
...
-  slowing down
+- repeatedly receiving 429 (rate limit) error responses from origin and not slowing down
 - making suspicious requests like attempting SQL injection queries
 
 [robots]: https://www.gov.uk/robots.txt
 
-Banning IPs shouldn't be taken lightly as IP address can be shared my multiple
-user devices and the user behind an IP address can change over time, so there's
-always a chance that we may block a legitimate user when we ban IP addresses.
+Banning IPs shouldn't be taken lightly as IP address can be shared my multiple user devices and the user behind an IP address can change over time, so there's always a chance that we may block a legitimate user when we ban IP addresses.
```
ChrisBAshton:
```suggestion
Banning IPs shouldn't be taken lightly as IP address can be shared by multiple user devices and the user behind an IP address can change over time, so there's always a chance that we may block a legitimate user.
```

source/manual/cdn.html.md
```diff
@@ -180,49 +143,36 @@ done
...
+You can change the list of banned IP addresses by modifying the [YAML config file][ip_ban_config] and [deploying the configuration][ip_ban_deploy].
 
 [ip_ban_config]: https://github.com/alphagov/govuk-cdn-config-secrets/blob/master/fastly/dictionaries/config/ip_address_blacklist.yaml
 [ip_ban_deploy]: https://deploy.publishing.service.gov.uk/job/Update_CDN_Dictionaries/build
 
 ## Bouncer's Fastly service
 
-A Fastly CDN service can normally handle up to 1000 domains (this limit was
-undocumented).
+A Fastly CDN service can normally handle up to 1000 domains (this limit was undocumented).
```
ChrisBAshton: I assume it still is? And if it isn't, we should just remove the parentheses:

```suggestion
A Fastly CDN service can normally handle up to 1000 domains (this limit is undocumented).
```

source/manual/cdn.html.md
injms: Much better link - thanks for the suggestion :thumbs:

source/manual/cdn.html.md
injms: Good idea! I've pointed it at the latest commit, since these are cited as examples and it doesn't matter if they go a bit out of date.

source/manual/cdn.html.md
injms: Done ✅

source/manual/cdn.html.md
injms: I've corrected for both links.

source/manual/cdn.html.md
injms: Hmm, not sure if it still is or not as I don't have access to the ZenDesk ticket about it. I've made the change as it makes more sense.

source/manual/cdn.html.md
injms: Done 👍

source/manual/cdn.html.md
injms: 🔍  😄

injms:
> I notice that you've removed the line-limiting newlines that existed partway between sentences. I don't think a preference is prescribed either way in the GDS Way or anywhere else. In the absence of an agreed rule, we should stick with the convention of the repo - which I think in this case is mostly for newlines rather than against. (It also makes the changes much easier to review - I can't easily see where your 'real' changes are as GitHub isn't picking it up in its diffing algorithm).

I thought the general guidance was that prose should be on a single line for a better diff, as a change that takes a line to over 80 characters could then cause a cascade of changes that needlessly affect subsequent lines. Whilst this diff is harder to see changes, future edits will be clearer.

PR #2314, opened 2020-02-11
Update DNS documentation

sarahseewhy:
* Update dates.
* Tighten language.
* Consolidate duplicate content under a single heading.
* Add backticks for formatting.

I followed the instructions as far as I could without having an actual change to deploy and it all seemed fine.

I don't know if the rest of the content is out of date, that may require more experienced reviews.

I've also tested the docs locally and it looks fine.

### DIFF
--- a/source/manual/dns.html.md
+++ b/source/manual/dns.html.md
@@ -5,7 +5,7 @@ section: DNS
 type: learn
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2019-08-03
+last_reviewed_on: 2020-02-11
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


issyl0:
I'd love to fix this whole process, but instead I guess I need to document some of the caveats that I keep having to explain to people instead (like the GCP Terraform provider being buggy). We can borrow from https://re-team-manual.cloudapps.digital/verify.html#how-to-deploy-gov-uk-dns-changes which I also wrote.

sarahseewhy:
Thank you @issyl0 for adding the caveats about Google provider. I think this doc is ready for a final reviewer.

PR #2345, opened 2020-02-25
Expand DGU deployment docs

vanitabarrett:
Adds the following details:

- Link to DGU staging (staging.data.gov.uk)
- What needs to be provided to create a new release

### DIFF
--- a/source/manual/data-gov-uk-deployment.html.md
+++ b/source/manual/data-gov-uk-deployment.html.md
@@ -60,7 +60,7 @@ You can check the deployment on [staging.data.gov.uk](https://staging.data.gov.u
 The deployment to production is triggered when a [new release] with an appropriate version
 number is created in GitHub.
 
-To create a new release, provide a new version number, release title and description.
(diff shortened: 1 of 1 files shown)


source/manual/data-gov-uk-deployment.html.md
```diff
@@ -50,10 +50,19 @@ deployed.
...
-number is created in GitHub. This behaviour is defined in [Publish's travis.yml]
+can check the [Travis logs of the `master` build](https://travis-ci.org/alphagov/datagovuk_find) to see progress.
+
+You can check the deployment on [staging.data.gov.uk](https://staging.data.gov.uk/)
+
+#### Production
+The deployment to production is triggered when a [new release] with an appropriate version
+number is created in GitHub.
+
+To create a new release, provide a new version number, release title and description.
```
kentsanggds: Worth adding an example version number here, as in the past people have added a version number without the leading `v` and it hasn't deployed

PR #2363, opened 2020-03-02
Update mongodb backups doc

sarahseewhy:
* Tighten language
* Fix broken link to another manual page
* Create "Restoring" sections to focus reader on relevant content
* Add a way for readers to ascertain which machines have enabled S3 backups and thus also contain the `mongodb-restore-s3` script.
* Replace Amazon with AWS for precision

I have tested the page locally and it looks OK.

### DIFF
--- a/source/manual/mongodb.html.md
+++ b/source/manual/mongodb.html.md
@@ -34,9 +34,11 @@ The timings are defined by parameters [set in the manifest](https://github.com/a
 These backups are encrypted using GPG, but the functionality is similar to mongodump.
 
 ### Restoring
-Use the `/usr/local/bin/mongodb-restore-s3` script available on the MongoDB machine cluster. 
+Use the `/usr/local/bin/mongodb-restore-s3` script available on MongoDB machines which have S3 backup enabled. 
 
 This script grabs the latest backup from the S3 bucket, decrypts and unpacks it, and does a `mongo restore`.
+
+Machines which have enabled S3 backups and contain the script will have `mongodb::backup::s3_backups` set to `true` in the yaml configuration (see [`govuk-puppet`](https://github.com/alphagov/govuk-puppet)).
 
 ### mongodumps via `govuk_env_sync` in AWS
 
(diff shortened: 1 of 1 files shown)


source/manual/mongodb.html.md
```diff
@@ -5,37 +5,45 @@ layout: manual_layout
 parent: "/manual.html"
 section: Backups
 type: learn
-last_reviewed_on: 2019-08-06
+last_reviewed_on: 2020-03-02
 review_in: 6 months
 ---
 
-We have 2 ways of taking MongoDB backups.
+There are two ways of taking MongoDB backups.
```
sengi: Aren't there three listed here? 🤔

 * automongodbbackup
 * mongodumps to S3
 * mongodumps via `govuk_env_sync` in AWS?

source/manual/mongodb.html.md
```diff
@@ -5,37 +5,45 @@ layout: manual_layout
...
+These backups are encrypted using GPG, but the functionality is similar to mongodump.
+
+### Restoring
+Use the `/usr/local/bin/mongodb-restore-s3` script available on MongoDB machines which have S3 backup enabled. 
+
+This script grabs the latest backup from the S3 bucket, decrypts and unpacks it, and does a `mongo restore`.
+
+Machines which have enabled S3 backups and contain the script will have `mongodb::backup::s3_backups` set to `true` in their yaml configuration (see [`govuk-puppet`](https://github.com/alphagov/govuk-puppet)).
 
 ### mongodumps via `govuk_env_sync` in AWS
```
sengi: We need restore instructions for this. Doesn't need to happen in this PR though. I think there might already be a card on the RE GOV.UK "Doing" Trello board for it.

source/manual/mongodb.html.md
sarahseewhy: The header formatting makes it a bit ambiguous. "mongodumps via `govuk_env_sync` in AWS" is formatted as a sub-heading of "mongodumps". 

This is likely to distinguish between the two types of mongodumps and the `automongodbbackup` third-party script. 

It boils down to a stylistic choice: yes, there are technically three ways but they fall into two main approaches.

I'm not fussed. 🤷‍♀

source/manual/mongodb.html.md
sarahseewhy: Noted 👍

PR #2372, opened 2020-03-05
Add link to moved info about govuk_env_sync.

sengi:
The sections about govuk_env_sync cron jobs and Icinga checks were moved to a playbook page in #2368. Add a link so that the information is still discoverable from the govuk_env_sync documentation.


sengi:
Verified that the link works, from a local instance.

PR #2389, opened 2020-03-10
Make entire repository pass Rubocop linting

ChrisBAshton:
This is depended on by https://github.com/alphagov/govuk-jenkinslib/pull/66.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,11 +1,8 @@
 require "govuk_tech_docs"
+require "rspec/core/rake_task"
 require_relative "./app/requires"
 
-begin
-  require "rspec/core/rake_task"
-  RSpec::Core::RakeTask.new(:spec)
-rescue LoadError
-end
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -1,11 +1,8 @@
...
+require "govuk_tech_docs"
+require "rspec/core/rake_task"
+require_relative "./app/requires"
 
-begin
-  require 'rspec/core/rake_task'
-  RSpec::Core::RakeTask.new(:spec)
-rescue LoadError
-end
+RSpec::Core::RakeTask.new(:spec)
```
ChrisBAshton: This shouldn't fall into the same [problem as we encountered in the cache-clearing-service](https://github.com/alphagov/cache-clearing-service/pull/260), as RSpec isn't listed as a _developer_ dependency in govuk-developer-docs. It probably should be, but that's out of scope for this PR!

PR #2391, opened 2020-03-10
Get rid of some small sections

barrucadu:
We have a bunch of sections with only one or two docs in, where I'm not convinced the granularity is worth it.  This PR shuffles some things around to reduce that:

- Analytics -> Frontend
- AWS accounts -> AWS
- DNS -> Infrastructure
- Patterns & Style Guides -> Documentation / Dependencies

### DIFF
--- a/source/manual/nagstamon.html.md
+++ b/source/manual/nagstamon.html.md
@@ -1,7 +1,7 @@
 ---
 owner_slack: "@christopherbaines"
 title: Use Nagstamon for monitoring Icinga
-section: AWS accounts
+section: Monitoring
 layout: manual_layout
 parent: "/manual.html"
 last_reviewed_on: 2020-02-14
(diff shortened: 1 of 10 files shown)


source/manual/nagstamon.html.md
```diff
@@ -1,7 +1,7 @@
 ---
 owner_slack: "@christopherbaines"
```
bevanloon: is Chris still the best person for this? Or should we move it to govuk-developers?

barrucadu: Does anyone but Chris use Nagstamon?

bevanloon: I do....very occasionally when I'm on 2ndline and WFH.....but yeah we could leave it for now. I'm sure Chris will updated it if it becomes a burden

barrucadu:
I don't think the search uses the section, eg the results for "DNS" don't have the current DNS page at the top:

Screenshot 2020-03-10 at 15 41 13

PR #2401, opened 2020-03-12
Exclude tmp directory from rubocopping

benjamineskola:
This is causing the build to fail due to linter errors in govuk-content-schemas (which may or may not need to be ignored anyway)


benjamineskola:
Closing in favour of https://github.com/alphagov/rubocop-govuk/pull/23

PR #2404, opened 2020-03-17
Review some 2ndline docs

barrucadu:


issyl0:
> GOV.UK developers and web operations staff

Can we change this to be "reliability engineers"? WebOps isn't a thing we hire for any more.

PR #2406, opened 2020-03-17
Add instructions for connecting to sidekiq in AWS production

leenagupte:
👉 [Rendered version](https://github.com/alphagov/govuk-developer-docs/blob/a0cf82068b226a2518adad3fd92c438cd24dcb4f/source/manual/sidekiq.html.md)

### DIFF
--- a/source/manual/sidekiq.html.md
+++ b/source/manual/sidekiq.html.md
@@ -48,7 +48,8 @@ $ ssh $(ssh integration "govuk_node_list --single-node -c backend").integration 
 or using `govuk-connect`:
 
 ```bash
-$ govuk-connect sidekiq-monitoring -e production aws/backend 
+$ brew install alphagov/gds/govuk-connect
+$ gds govuk connect sidekiq-monitoring -e production aws/backend 
 ```
 
 Then visit [http://127.0.0.1:9000](http://127.0.0.1:9000) to see a list of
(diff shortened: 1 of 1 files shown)


source/manual/sidekiq.html.md
```diff
@@ -45,6 +45,12 @@ Or on AWS:
 $ ssh $(ssh integration "govuk_node_list --single-node -c backend").integration -CNL 9000:127.0.0.1:3211
 ```
 
+or using `govuk-connect`:
+
+```bash
+$ govuk-connect sidekiq-monitoring -e production aws/backend
```
issyl0: I don't think we should do this until govuk-connect is [official](https://github.com/alphagov/govuk-aws/issues/1159). It'll be soon but not yet. And we should do the things mentioned in the comments, like using `gds govuk connect` so there's one tool. @benthorner We should expedite this.

leenagupte: I thought it would useful to have _something_, so we're not stuck like we were last night. If there's some other way that works, I'm happy to document that.

issyl0: You know what yeah let's do this.

```suggestion
$ brew install alphagov/gds/govuk-connect
$ gds govuk connect sidekiq-monitoring -e production aws/backend 
```

leenagupte:
@issyl0 Suggestion accepted!

PR #2407, opened 2020-03-17
Include sidekiq from the console section

bevanloon:
Include a section with some basic commands to query sidekiq from the
console.

Include skip links for easy navigation.


issyl0:
A companion PR is #2406 for @richardTowers' point.

PR #2411, opened 2020-03-18
Change all SSH docs to reference `gds govuk connect` not `govukcli`

issyl0:
https://trello.com/c/7T1NB3ym/135-update-dev-docs-for-govuk-connect

- This is very much WIP and it'll take a while to move things over.
- We'll also do some workshops and I guess training sessions on using
  `govuk-connect`, to get people fully switched over - much like we did
  with the GDS CLI.
- Come and talk to #govuk-dev-tools if you want to get involved or have
  questions!
- Further context can be found in https://github.com/alphagov/govuk-aws/issues/1159.

----

On commit number one, `ssh-access`:

- I feel like this is quite rushed and not very explanatory. Maybe we need a "prerequisite software and internal tooling" section? That said, we never much explained the `govukcli` either.

### DIFF
--- a/source/manual/editing-an-existing-route.html.md
+++ b/source/manual/editing-an-existing-route.html.md
@@ -20,8 +20,7 @@ If there's a need to edit a route in the database:
 1. Connect to a router-backend machine
 
 ```console
-govukcli set-context production-aws
-govukcli ssh router_backend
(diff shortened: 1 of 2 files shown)


issyl0:
Closing this as we can revisit it after #2415.

PR #2414, opened 2020-03-19
Add comment on data scrubbing process

gclssvglx:
Some document editions that are pending publishing are being re-created into the same state each night by the Whitehall scrub-database process.

This change alerts the 2nd line team.

### DIFF
--- a/source/manual/alerts/whitehall-scheduled-publishing.html.md.erb
+++ b/source/manual/alerts/whitehall-scheduled-publishing.html.md.erb
@@ -42,3 +42,14 @@ recently restored, perhaps due to the data sync. In that case, you can try
 running the following Rake task on a `whitehall_backend` machine:
 
 <%= RunRakeTask.links("whitehall", "publishing:scheduled:requeue_all_jobs") %>
+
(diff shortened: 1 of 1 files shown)


source/manual/alerts/whitehall-scheduled-publishing.html.md.erb
```diff
@@ -42,3 +42,14 @@ recently restored, perhaps due to the data sync. In that case, you can try
 running the following Rake task on a `whitehall_backend` machine:
 
 <%= RunRakeTask.links("whitehall", "publishing:scheduled:requeue_all_jobs") %>
+
+Due to the overnight (data anonymisation process)[https://github.com/alphagov/whitehall/blob/7b5c5a086b89cb62ffba62b152a0a8dcfc10c8e6/script/scrub-database] you may notice
```
issyl0:
```suggestion
Due to the overnight [data anonymisation process](https://github.com/alphagov/whitehall/blob/7b5c5a086b89cb62ffba62b152a0a8dcfc10c8e6/script/scrub-database) you may notice
```

gclssvglx: Thanks @issyl0 - I'm always getting those the wrong way.  Fixed.

PR #2417, opened 2020-03-20
Fix broken link and update Re-sync Postgres docs

sarahseewhy:
* Fix broken link to `postgresql-backups` doc in `resync-postgres-standby` doc.
* Tighten wording.
* Remove references to step numbers which risk getting out of sync.
* Update `postgresql-backups` doc to include special case reference to `resync-postgres-standby` instructions.

I followed these docs yesterday when I resynced the disaster recovery postgres database.

Tested the project locally (found a mising `/` and fixed it) and it all looks fine now.

### DIFF
--- a/source/manual/postgresql-backups.html.md
+++ b/source/manual/postgresql-backups.html.md
@@ -79,7 +79,7 @@ Do you really want to do this? (y/N) y
 gpg> save
 ```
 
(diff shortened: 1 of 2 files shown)


sengi:
Aha that makes sense to me now. Thanks so much for the clarification!

PR #2419, opened 2020-03-24
Add coronavirus apps to the dev docs

richardTowers:
Apparently nothing else that's deployed to paas has a deployment script,
but these apps do (well, a concourse pipeline) so I had to patch the
code to work for them.

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -659,6 +659,28 @@
   dashboard_url: false
   description: 'GOV.UK Licensing (formerly ELMS, Licence Application Tool, & Licensify)'
 
+# Coronavirus
+
+- github_repo_name: govuk-coronavirus-vulnerable-people-form
+  private_repo: false
+  type: Specialist apps
+  production_hosted_on: paas
+  team: "#govuk-vulnerable-people-tech"
+  sentry_url: false # TODO - can the reviewer let me know what this should be?
+  deploy_url: 'https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/concourse/pipeline.yml'
+  dashboard_url: 'https://gds.splunkcloud.com/en-GB/app/gds-006-govuk/d006_coronavirus?form.main.earliest=-7d%40h&form.main.latest=now&form.xhost=coronavirus-vulnerable-people.service.gov.uk'
+  description: 'Helps citizens get support during the Covid-19 pandemic'
+
+- github_repo_name: govuk-coronavirus-business-volunteer-form
+  private_repo: false
+  type: Specialist apps
+  production_hosted_on: paas
+  team: "#govuk-corona-forms-tech"
+  sentry_url: false # TODO - can the reviewer let me know what this should be?
+  deploy_url: 'https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/master/concourse/pipeline.yml'
+  dashboard_url: false
+  description: 'Lets businesses volunteer with the response to the Covid-19 pandemic'
+
 # Utility Heroku apps (don't add apps that are only for review here)
 
 - github_repo_name: seal
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -676,7 +676,7 @@
   type: Specialist apps
   production_hosted_on: paas
   team: "#govuk-corona-forms-tech"
-  sentry_url: false https://sentry.io/organizations/govuk/issues/?project=5170680
+  sentry_url: https://sentry.io/organizations/govuk/issues/?project=5172100
   deploy_url: 'https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/master/concourse/pipeline.yml'
   dashboard_url: false
   description: 'Lets businesses volunteer with the response to the Covid-19 pandemic'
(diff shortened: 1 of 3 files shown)


data/applications.yml
```diff
@@ -659,6 +659,28 @@
...
   description: 'GOV.UK Licensing (formerly ELMS, Licence Application Tool, & Licensify)'
 
+# Coronavirus
+
+- github_repo_name: govuk-coronavirus-vulnerable-people-form
+  private_repo: false
+  type: Specialist apps
+  production_hosted_on: paas
+  team: "#govuk-vulnerable-people-tech"
+  sentry_url: false # TODO - can the reviewer let me know what this should be?
```
benthorner: It looks like [this is set in the pipeline config](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/e9e151499ae9a5a5b53dde615172cbf30299b4bd/concourse/tasks/deploy-to-govuk-paas.yml#L46). I don't seem to have access to concourse to see what it's actually configured to though. I expect there's some kind of 'settings' button [here](https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/govuk-corona-vulnerable-people-form) if you do.

richardTowers: Hmmm, the DSN doesn't look like the other URLs in this doc - it's like `https://$user:$password@sentry.io/5170680`

I think this should be a link to the view in sentry.

benthorner: This seems to come up with something, but I don't think it's working properly, as it doesn't have a project name: https://sentry.io/organizations/govuk/issues/?project=5170680

huwd: where do you see the project name ben? Where should it be?
I see coronavirus-vulnerable-people form from the left most of the the three top drop downs?

and coronavirus-business-voulenteer for the one we're doing today?
https://sentry.io/organizations/govuk/issues/?project=5172100

data/applications.yml
```diff
@@ -659,6 +659,28 @@
...
+  deploy_url: 'https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/concourse/pipeline.yml'
+  dashboard_url: 'https://gds.splunkcloud.com/en-GB/app/gds-006-govuk/d006_coronavirus?form.main.earliest=-7d%40h&form.main.latest=now&form.xhost=coronavirus-vulnerable-people.service.gov.uk'
+  description: 'Helps citizens get support during the Covid-19 pandemic'
+
+- github_repo_name: govuk-coronavirus-business-volunteer-form
+  private_repo: false
+  type: Specialist apps
+  production_hosted_on: paas
+  team: "#govuk-corona-forms-tech"
+  sentry_url: false # TODO - can the reviewer let me know what this should be?
```
huwd:
```suggestion
  sentry_url: false https://sentry.io/organizations/govuk/issues/?project=5170680
```

data/applications.yml
benthorner: Yeah, I don't see it (the dropdown is just blank for me). I know there's a strange thing with Sentry where you have to add the project 'manually' sometimes. Perhaps that's it.

I also don't see any coronavirus projects here: https://sentry.io/organizations/govuk/projects/

huwd:
```suggestion
  sentry_url: https://sentry.io/organizations/govuk/issues/?project=5170680
```
If this is just for reference then the above URL should be right,
otherwise if this is supposed to be the DNS might need removing here as that's a secret?

data/applications.yml
huwd: If this is just for reference then the above URL should be right,
otherwise if this is supposed to be the DNS might need removing here as that's a secret?

issyl0: I think that URL is right.

data/applications.yml
```diff
@@ -659,6 +659,28 @@
...
+  deploy_url: 'https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/concourse/pipeline.yml'
+  dashboard_url: 'https://gds.splunkcloud.com/en-GB/app/gds-006-govuk/d006_coronavirus?form.main.earliest=-7d%40h&form.main.latest=now&form.xhost=coronavirus-vulnerable-people.service.gov.uk'
+  description: 'Helps citizens get support during the Covid-19 pandemic'
+
+- github_repo_name: govuk-coronavirus-business-volunteer-form
+  private_repo: false
+  type: Specialist apps
+  production_hosted_on: paas
+  team: "#govuk-corona-forms-tech"
+  sentry_url: false https://sentry.io/organizations/govuk/issues/?project=5170680
```
bilbof:
```suggestion
  sentry_url: https://sentry.io/organizations/govuk/issues/?project=5172100
```

data/dashboard.yml
```diff
@@ -8,6 +8,7 @@
     - name: Frontend apps
     - name: data.gov.uk apps
     - name: Licensing apps
+    - name: Specialist apps
```
bilbof: Perhaps 'Service apps' would be more specific, since this is how we refer to apps by other departments that run on `.service.gov.uk`

data/applications.yml
huwd: my bad! sorry

data/dashboard.yml
richardTowers: There's already a "Services" category - it might be a bit confusing to have "Service apps" and "Services".

I thought about "Coronavirus apps", but maybe that's too on-the-nose?

PR #2431, opened 2020-03-30
Update DGU supporting CKAN docs

kentsanggds:
## What

The DB connections regularly time out every few months, so until it is investigated and resolved these notes should help resolve it.

## Reference 

https://trello.com/c/J9NuRBqs/1883-too-many-database-connections-causing-ckan-to-fail-in-staging

### DIFF
--- a/source/manual/data-gov-uk-supporting-ckan.html.md
+++ b/source/manual/data-gov-uk-supporting-ckan.html.md
@@ -427,7 +427,7 @@ $ fab aws_production class:ckan ckan.restart_harvester
 
 ### CKAN publisher on Staging environment responds with Nginx 504 timeout:
 
-Sometimes CKAN publisher on Staging environemt responds with a 504 from Nginx, this is due to it timing out when connecting to the database as there are too many connections, current limit is 1000.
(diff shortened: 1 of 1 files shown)


source/manual/data-gov-uk-supporting-ckan.html.md
```diff
@@ -424,3 +424,32 @@ running, so is safe to run immediately if you suspect the process has crashed:
 ```bash
 $ fab aws_production class:ckan ckan.restart_harvester
 ```
+
+### CKAN publisher on Staging environment responds with Nginx 504 timeout:
+
+Sometimes CKAN publisher on Staging environemt responds with a 504 from Nginx, this is due to it timing out when connecting to the database as there are too many connections, current limit is 1000.
```
sengi: (trivial) typo
```suggestion
Sometimes CKAN publisher on Staging environment responds with a 504 from Nginx, this is due to it timing out when connecting to the database as there are too many connections, current limit is 1000.
```

sengi: Might be good to mention that it's only supposed to need one connection when it's working normally.

PR #2433, opened 2020-03-31
Fix links leading to 404 pages

beccapearce:
When linking to documentation written in markdown on github we were replacing the `.md` extension with `.html`, however this was linking incorrectly and returning 404 errors. Here we change the behaviour to only modify links pointing internally and not to modify links pointing to external github repos.

### DIFF
--- a/app/external_doc.rb
+++ b/app/external_doc.rb
@@ -92,8 +92,7 @@ class ExternalDoc
 
         href = element["href"].strip
         uri = URI.parse(href)
-
-        if uri.path.end_with?(".md")
+        if uri.path.end_with?(".md") && uri.host == nil
           uri.path.sub!(/.md$/, ".html")
           element["href"] = uri.to_s
         end
--- a/spec/app/external_doc_spec.rb
+++ b/spec/app/external_doc_spec.rb
@@ -22,6 +22,11 @@ RSpec.describe ExternalDoc do
       expect(html).to have_link("turpis ultrices", href: "/ultrices.html")
     end
 
+    it "does not rewrite links to markdown pages with a host" do
+      expect(html).to have_link("Nam eget dui", href: "https://nam.com/eget/dui/uploading.md")
+      expect(html).not_to have_link("Nam eget dui", href: "https://nam.com/eget/dui/uploading.html")
+    end
+
     it "rewrites relative images" do
       expect(html).to have_css('img[src="https://raw.githubusercontent.com/example/lipsum/master/suspendisse_iaculis.png"]')
     end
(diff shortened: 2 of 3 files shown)


thomasleese:
Will this affect the “API docs” pages, such as: https://docs.publishing.service.gov.uk/apis/publishing-api/link-expansion.html ?

issyl0:
Is this mergeable?

beccapearce:
@thomasleese After the update removing the extra slash: https://github.com/alphagov/govuk-developer-docs/pull/2450 , the API docs seem to run fine

beccapearce:
@issyl0 Sorry I didn't see this, there was an error on the publishing API page fixed in https://github.com/alphagov/govuk-developer-docs/pull/2450 which we wanted to fix to test this PR out properly, hopefully it's close now 🤞

thomasleese:
I wonder if this breaks links to other pages in the Publishing API docs? For example in this paragraph https://docs.publishing.service.gov.uk/apis/publishing-api/link-expansion.html the `dependency-resolution` link I think is getting changed from `.md` to `.html`:

> A closely related process to this is **dependency-resolution**. This is something of a reversal in link expansion and has the role of determining which editions will need re-presenting to the content store as a result of an update to a document.

Maybe we could have a system where this `MarkdownLinkFilter` only applies for some documentation?

app/external_doc.rb
```diff
@@ -93,7 +93,7 @@ def call
         href = element["href"].strip
         uri = URI.parse(href)
 
-        if uri.path.end_with?(".md")
+        if uri.path.end_with?(".md") && uri.host == nil
```
ChrisBAshton, line 96: Could you add a test for this to avoid it happening again?

beccapearce:
@ChrisBAshton The two examples I've been testing for 404 links are:
- In the whitehall documentation https://docs.publishing.service.gov.uk/apis/whitehall.html where it says `Please add your application to this list if you’re using the API.`
- In the load testing documentation https://docs.publishing.service.gov.uk/manual/load-testing.html#uploading the link ` how to upload results`

In both of the examples the text comes from separate github repos for whitehall https://github.com/alphagov/whitehall/blob/master/docs/api.md and for load testing https://github.com/alphagov/govuk-load-testing/blob/master/README.md

ChrisBAshton:
Whoops, I misunderstood, and now understand what your PR is doing 👍 
Have raised a PR to fix the broken link I found: https://github.com/alphagov/govuk-developer-docs/pull/2468 (I thought your PR was trying to fix links like these with code).

PR #2438, opened 2020-04-02
Update instructions to use 'gds' cli instead of govukcli

ChrisBAshton:
govukcli is now deprecated.


source/manual/howto-ssh-to-machines-in-aws.html.md
```diff
@@ -19,52 +19,37 @@ hostnames like `ip-10-1-5-53.eu-west-1.compute.internal`. Each Puppeted instance
...
+If you know the class of machine you want, you can SSH straight from the command line:
 
 ```sh
-local$ govukcli list-contexts
+$ gds govuk connect ssh -e staging cache
 ```
 
-To set a persistent context:
+This will automatically SSH into a random `cache` machine on AWS.
+If a class exists in multiple clouds, you will need to choose which one to SSH into:
```
issyl0, line 33: ☁️ ☁️

source/manual/howto-ssh-to-machines-in-aws.html.md
```diff
@@ -19,52 +19,37 @@ hostnames like `ip-10-1-5-53.eu-west-1.compute.internal`. Each Puppeted instance
...
-Now when you SSH into an instance, it will be to one running on the specified environment:
+specify the hosting provider and node type, for example: 
 
-```sh
-# SSH into a 'calculators_frontend' node in current context (Integration)
-local$ govukcli ssh calculators_frontend
-```
-
-Alternatively, you can set a context for the current command only by passing a CLI parameter:
+  govuk connect ssh carrenza/backend
```
issyl0, line 41:
```suggestion
  gds govuk connect ssh carrenza/backend
```

Unless you meant `govuk-connect`, to use the underlying tool. 😭

source/manual/howto-ssh-to-machines-in-aws.html.md
```diff
@@ -19,52 +19,37 @@ hostnames like `ip-10-1-5-53.eu-west-1.compute.internal`. Each Puppeted instance
...
 
-```sh
-# SSH into a 'calculators_frontend' node in Staging, ignoring local context
-local$ govukcli ssh --context staging calculators_frontend
+  govuk connect ssh aws/backend
```
issyl0, line 43:
```suggestion
  gds govuk connect ssh aws/backend
```

source/manual/howto-ssh-to-machines-in-aws.html.md
ChrisBAshton: Hi - this is the error message copied & pasted verbatim 🤔

issyl0: Oh. We should [fix that](https://github.com/alphagov/govuk-connect/blob/master/lib/govuk_connect.rb). We can also test my new "release a new version of the gem" instructions! But for the avoidance of doubt, I don't get that error. What version are you using?

ChrisBAshton: v2.7.0. I'm surprised you don't get that, as the [master version is missing the `gds` prefix](https://github.com/alphagov/govuk-connect/blob/master/lib/govuk_connect.rb#L509). It looks like [the same thing happens for node class and application mame](https://github.com/alphagov/govuk-connect/blob/master/lib/govuk_connect.rb#L462) (sic!)

I'll fix it up in this doc for the time being 👍

issyl0: Oh, I wasn't on the VPN, so the command just timed out rather than erroring on the "ambiguous". 😩

issyl0: Please go ahead and make the changes in govuk-connect, in that case.

ChrisBAshton: Ah - I'd missed a subtlety there. `master` appears to be:

`govuk-connect ssh #{bold(hosting_provider)}/#{node_type}`

Whereas my output is:

`govuk connect ssh carrenza/backend`

There's a discrepancy with the `-` somewhere, I'm not sure where! And also correcting it to:

`gds govuk connect ssh carrenza/backend`

...may not give the desired affect, as copying and pasting that command leads to a `error: you must specify the environment` error. We might want to consider making the error message more useful by:

1) dynamically taking the user's last command as parameters, so we can regurgitate the environment information etc
2) being more generic, e.g. "change backend to aws/backend or carrenza/backend"

...or just live with it as is 🤔

issyl0: 2 sounds like a good option.

PR #2440, opened 2020-04-02
Clarify some commands

barrucadu:
Using $(date...) in every example is confusing, because you need to use the same date in them all.

In secrets, the password isn't called "apt", it's "apt/passphrase".  Also use the edit script to avoid problems with comments in the GPG ID list.

### DIFF
--- a/source/manual/debian-packaging.html.md
+++ b/source/manual/debian-packaging.html.md
@@ -174,10 +174,11 @@ deb file to aptly.
 4. Create a snapshot:
 
     ```
-      $ sudo aptly snapshot create rbenv-ruby-$(date +%Y%m%d) from repo rbenv-ruby
+      $ snapshot="rbenv-ruby-$(date +%Y%m%d)"
+      $ sudo aptly snapshot create $snapshot from repo rbenv-ruby
         Snapshot rbenv-ruby-20190212 successfully created.
 
-      $ sudo aptly snapshot show rbenv-ruby-20190212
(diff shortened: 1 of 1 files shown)


source/manual/debian-packaging.html.md
```diff
@@ -187,16 +188,16 @@ deb file to aptly.
...
     ```
-      $ sudo aptly snapshot diff rbenv-ruby-20181023 rbenv-ruby-20190212
+      $ sudo aptly snapshot diff rbenv-ruby-20181023 $snapshot
         Arch   | Package          | Version in A  | Version in B
       + amd64  | rbenv-ruby-2.6.1 | -             | 1
 
     ```
 6. Publish the new snapshot, you will be prompted to enter the passphrase for our
-APT account which is in [govuk-secrets](https://github.com/alphagov/govuk-secrets/tree/master/pass) `PASSWORD_STORE_DIR=~/govuk/govuk-secrets/pass/2ndline pass apt`
+APT account which is in [govuk-secrets](https://github.com/alphagov/govuk-secrets/tree/master/pass) `./edit.sh 2ndline apt/passphrase`
```
thomasleese, line 197: Hmm... I don't get problems with the `.gpg-id` files when using `pass` to get the secret, only if I use `pass` directly for editing or deleting.

barrucadu: Oh, it's been a while since I tried that, maybe it is fine

PR #2451, opened 2020-04-08
Switch dependencies to govuk-developers

benthorner:
This is because the Publishing Workflow team is pausing, so it's
unlikely anyone will be monitoring the -dev channel and actioning
the open PRs.


benthorner:
Thanks for checking @ChrisBAshton - I've updated. Could you take another look?

nicholsj:
Probably superseded by https://github.com/alphagov/govuk-developer-docs/pull/2455 - which makes more changes

PR #2454, opened 2020-04-08
covid-19-services: Add docs about the new find-support form

issyl0:
- Where relevant I changed "both" to "all" or got rid of the word
  entirely as there are now >2 forms.
- Fix some casing of products, like "PaaS" and fix some accidental,
  unnecessary double spacing between sections.
- Move the blurb about the IVR to the vulnerable people section, as
  that's the only thing it's for.

Trello card: https://trello.com/c/3yABBo1i

### DIFF
--- a/source/manual/covid-19-services.html.md
+++ b/source/manual/covid-19-services.html.md
@@ -165,14 +165,15 @@ settings and credentials for the backing service automatically.
 We have two AWS accounts (staging and production) for CDN and data
 storage. They contain DynamoDB, IAM and CloudFront resources,
 which were provisioned by Terraform in the [tech-ops-private
-repository](techops-repo).
+repository](techops-repo). Members of the GOV.UK Coronavirus Services
+team have access.
 
 To log in to AWS:
 
 1. [Install and set up the gds-cli](/manual/get-started.html#8-use-your-aws-access), then log in to the AWS Console with either:
 
-   - `gds aws govuk-corona-data-staging -l` for staging
-   - `gds aws govuk-corona-data-prod -l` for production
+   - `gds aws govuk-corona-data-staging-poweruser -l` for staging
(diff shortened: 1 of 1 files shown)


source/manual/covid-19-services.html.md
```diff
@@ -38,17 +32,32 @@ message from the NHS, or been advised by their GP to fill in the form.
...
+- [Service](https://find-coronavirus-support.service.gov.uk/urgent-medical-help)
+- [GitHub](https://github.com/alphagov/govuk-coronavirus-find-support)
 
 ## Architecture
 
 Both applications have a similar architecture of:
 
 - **Framework:** Ruby on Rails
-- **Database:** AWS RDS (business volunteering) and DynamoDB (vulnerable people)
+- **Database:** AWS RDS (business volunteering, find support) and DynamoDB (vulnerable people)
```
JonathanHallam, line 60: Isn't the find support page using postgres?

source/manual/covid-19-services.html.md
```diff
@@ -91,8 +99,9 @@ This is also stored for 2 hours.
 
 ### Data storage
 
-For both applications, when the user submits the form on the "Check
-your answers" page, the application writes user data to the database.
+For all but the find support form, when the user submits the form on
+the "Check your answers" page, the application writes user data to the
+database.
```
JonathanHallam, line 104: Might be worth adding when the find support page does this, ie. on submission of the final question

source/manual/covid-19-services.html.md
brucebolt: RDS is AWS's Postgres/MySQL/etc solution

JonathanHallam: 👍

PR #2456, opened 2020-04-09
Explain what to do if search analytics job fails

barrucadu:


source/manual/alerts/fetch-analytics-data-for-search-failed.html.md
```diff
@@ -14,10 +14,22 @@ which runs every night and updates all documents in the search index with pagevi
...
 
-The popularity data is used to promote more popular pages in search results.
+This job has two parts:
+
+1. Page view data is fetched from Google Analytics.  This process seems to sometimes hang, resulting in stuck jobs.
+2. The search index is locked and popularity score of every page is updated.
+
+If the job fails or gets stuck in part 1, you can cancel it and leave it to run the next day.
+
+if the job fails or gets stuck in part 2, the search indices may still be locked.  If that's the case, documents will be missing from search.  First unlock the indices:
```
AlanGabbianelli, line 26:
```suggestion
If the job fails or gets stuck in part 2, the search indices may still be locked.  If that's the case, documents will be missing from search.  First unlock the indices:
```

PR #2459, opened 2020-04-11
Improve hosting section

ChrisBAshton:
This PR exposes which apps are hosted with which hosters, which otherwise requires examining each application's dev doc page individually. We also rely on a Google Spreadsheet showing the current status of migration from Carrenza to AWS. We should have this information more readily available and more automated so that we have a reliable single source of truth.

Before:

Screenshot 2020-04-11 at 15 24 28

After:

Screenshot 2020-04-11 at 16 49 54


app/app_docs.rb
```diff
@@ -14,6 +14,19 @@ def self.pages
     end
   end
 
+  def self.hosters_descending
+    ordered_keys = HOSTERS.keys.sort do |a, b|
+      [apps_on_host(b).count, a] <=> [apps_on_host(a).count, b]
+    end
+    ordered_keys.map { |key| [key, HOSTERS[key]] }.to_h
```
thomasleese, line 21:
```suggestion
    ordered_keys.index_with { |key| HOSTERS[key] }
```

I think this is equivalent: https://api.rubyonrails.org/classes/Enumerable.html#method-i-index_with

ChrisBAshton: Looks like that's only available in Rails 6, but govuk-developer-docs isn't a Rails app 😢

thomasleese: Ah... looks like we could do with upgrading the version of ActiveSupport we use! https://github.com/alphagov/govuk-developer-docs/blob/bb4f5690f42358f688e771b9082410e781160c91/Gemfile.lock#L4

PR #2462, opened 2020-04-14
Replace all occurrences of `govukcli ssh` with `gds govuk connect ssh`

issyl0:
- Now the rest of the docs match the Get Started guide, we will no
  longer confuse new people (or old people trying to decipher which tool
  they should use).
- Some of these could be replaced with `gds govuk connect -e production
  app_console <app>`, but that's an exercise for later.
- Relates to https://github.com/alphagov/govuk-aws/issues/1159. Come and talk to us in #govuk-dev-tools!

### DIFF
--- a/source/manual/pagerduty.html.md
+++ b/source/manual/pagerduty.html.md
@@ -23,7 +23,7 @@ and that they're correctly scheduled in.
 When an alert that triggers PagerDuty goes off, someone on the escalation schedule must acknowledge
 them, otherwise they will be escalated further. NB, 2nd line shadowers are not required to be on PagerDuty.
 
-## PagerDuty drill
+## PagerDuty drill
 
 Every week we test PagerDuty to make sure it can phone to alert us to
 any issues. This happens every Wednesday morning at 10am UTC.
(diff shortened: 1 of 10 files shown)


source/manual/pagerduty.html.md
```diff
@@ -23,7 +23,7 @@ and that they're correctly scheduled in.
 When an alert that triggers PagerDuty goes off, someone on the escalation schedule must acknowledge
 them, otherwise they will be escalated further. NB, 2nd line shadowers are not required to be on PagerDuty.
 
-## PagerDuty drill
+## PagerDuty drill
```
ChrisBAshton, line 26: Was this change deliberate? I'm slightly concerned about merging as the heading is already rendering correctly: https://docs.publishing.service.gov.uk/manual/pagerduty.html#pagerduty-drill

issyl0: It was [automatic](https://github.com/issyl0/dotfiles/blob/master/.vimrc#L33), which is odd because the header is currently rendering correctly. :shrug:

ChrisBAshton: Maybe our header rendering is less brittle than GitHub's?

issyl0:
Thanks Chris!

PR #2464, opened 2020-04-15
Fix get started

ChrisBAshton:
A missing ` caused this issue on the Get Started page:

> Screenshot 2020-04-15 at 09 08 39

### DIFF
--- a/app/external_doc.rb
+++ b/app/external_doc.rb
@@ -81,7 +81,7 @@ class ExternalDoc
   # contain links to other pages of documentation also formatted with Markdown.
   # When the documentation is rendered as part of GOV.UK Developer Docs we
   # render it as HTML so we need to rewrite the links so that they have a .html
-  # extension to match out routing.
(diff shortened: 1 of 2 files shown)


issyl0:
THANK YOU! I've been trying to figure this out for days, going through all the Markdown links docs, trying different combinations, comparing with other files to see that the link syntax was indeed exactly the same... And it's all down to a missing backtick!

ChrisBAshton:
Ha, you're welcome! It took a while, I resorted to deleting chunks from the document and refreshing the page until I found the offending chunk. My syntax highlighter gave me a clue, as the first heading following the missing backtick wasn't coloured in.

I wonder if we could enforce better validation on our markdown, or do some post-processing check for `[`/`]` characters (which almost always indicate an unprocessed bit of markdown!). I'll add it to my list - hoping to start tackling these things as part of an epic in govuk-dev-tools (will bring it up in tomorrow's meeting).

PR #2469, opened 2020-04-16
nagstamon: Link to the canonical source for installing our tooling

issyl0:
- Otherwise different places in this manual will diverge from the Get
  Started messaging and that'd be bad.
- Also `brew install` is easier to understand than strange, cryptic
  errors people were getting about Ruby versions when they used `gem
  install`.


issyl0:
Turns out I missed multiple places where we're still referring to raw SSH. Maybe I should split this PR up. 🤔

PR #2473, opened 2020-04-16
Link to architectural deep-dive from 'Getting started'

ChrisBAshton:
Creates a new "You're all done" section to round off the guide,
just to confirm the developer has finished setting everything up.
This seems a natural point to introduce them to the architecture.

I considered linking them to the [architectural overview](https://docs.publishing.service.gov.uk/manual/architecture.html)
instead, but in my experience the diagram doesn't mean much until
it has been explained, so the deep-dive is a better, if more
daunting piece of documentation to link to.

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -307,6 +307,13 @@ You can also chain commands, like this one to list S3 buckets in integration:
 gds aws govuk-integration-poweruser aws s3 ls
 ```
 
+## You're all done!
+
+You're set up and ready to go. It might be worth reading and bookmarking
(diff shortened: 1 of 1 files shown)


source/manual/get-started.html.md
```diff
@@ -307,6 +307,13 @@ You can also chain commands, like this one to list S3 buckets in integration:
 gds aws govuk-integration-poweruser aws s3 ls
 ```
 
+## You're all done!
+
+You're set up and ready to go. It might be worth reading and bookmarking
```
issyl0, line 312: If I had to absorb all this on my first day I think I'd leave. 😂

PR #2481, opened 2020-04-21
Review COVID-19 documentation

brucebolt:
This PR:
- adds information about managing GOV.UK Notify for the COVID-19 services
- reviews all COVID-19 documentation which was out of date in a few places

Trello card: https://trello.com/c/ecivql30

### DIFF
--- a/source/manual/covid-19-services.html.md
+++ b/source/manual/covid-19-services.html.md
@@ -4,14 +4,15 @@ title: COVID-19 Services
 section: Services
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2020-03-25
+last_reviewed_on: 2020-04-21
 review_in: 1 month
 ---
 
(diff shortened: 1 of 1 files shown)


source/manual/covid-19-services.html.md
```diff
@@ -63,13 +64,14 @@ Both applications have a similar architecture of:
 - **CI:** GitHub Actions (branches and PRs)
 - **CD:** Concourse (master builds and production deploys)
 - **Logging:** Sentry and Logit
+- **Email**: GOV.UK Notify
```
bilbof, line 68: Do you think it would be worth adding Sidekiq to this list? We're running a mailers queue for the confirmation emails.

brucebolt: I've added it here anyway.  Useful to have a summary of the full tech stack.

PR #2482, opened 2020-04-21
Review docs about our use of GitHub

issyl0:
See commits for more details.


issyl0:
Do you think Jenkins knows its days are numbered? 😆

issyl0:
I wonder if Jenkins can't handle `#` characters in branches? 😱

issyl0:
Let's try this again in #2484 with a normal branch name...

PR #2487, opened 2020-04-22
Review page on receiving emails from staging

AlanGabbianelli:
This edits this doc page to reflect the fact that it relates only to
public facing emails and not to publisher emails. In order to do so, I
changed the title and added a note at the top of the page.

I also:
 - changed the file name to reflect the new tile
 - created a new placeholder file that will (in a later PR) document the
   process to receive publisher emails from Email Alert API in
   integration and staging
 - added a link to the new file in the aforementioned note

All this should hopefully clarify the distinction between public facing
emails and publishers emails and the different processes to get them in
our non-production environments.

The part on what to change in puppet was still showing outdated
instructions that are not relevant anymore (since we migrated
email_alert_api to AWS), so I edited that part too.

### DIFF
--- a/source/manual/receive-publisher-emails-from-email-alert-api-in-integration-and-staging.html.md
+++ b/source/manual/receive-publisher-emails-from-email-alert-api-in-integration-and-staging.html.md
@@ -1,0 +1,15 @@
+---
+owner_slack: "#govuk-developers"
+title: Receive publisher emails from Email Alert API in integration and staging
+section: Emails
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2020-04-21
+review_in: 6 months
+---
+
+> **Note**
+>
+> This only relates to publisher emails, not to [public facing emails].
+
+[public facing emails]: /receive-public-facing-emails-from-email-alert-api-in-integration-and-staging.html
--- a/source/manual/receive-public-facing-emails-from-email-alert-api-in-integration-and-staging.html.md
+++ b/source/manual/receive-public-facing-emails-from-email-alert-api-in-integration-and-staging.html.md
@@ -1,0 +1,88 @@
+---
+owner_slack: "#govuk-developers"
+title: Receive public facing emails from Email Alert API in integration and staging
+section: Emails
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2020-04-21
+review_in: 6 months
+---
+
+> **Note**
+>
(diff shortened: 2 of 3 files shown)


cbaines:
Previously the page title was "Receive emails from Email Alert API in integration and staging", does the Email Alert API send emails to publishers?

source/manual/receive-publisher-emails-from-email-alert-api-in-integration-and-staging.html.md
```diff
@@ -0,0 +1,15 @@
+---
+owner_slack: "#govuk-developers"
+title: Receive publisher emails from Email Alert API in integration and staging
```
thomasleese, line 3: I'm not sure if this makes sense because the publishing apps don't go through the Email Alert API to send emails, they use Notify directly.

AlanGabbianelli:
@cbaines @thomasleese thanks for your comments. How about changing the titles to:
- Receive public-facing emails in integration and staging

cbaines:
> @cbaines @thomasleese thanks for your comments. How about changing the titles to:
> 
>     * Receive public-facing emails in integration and staging
>     * Receive publishers emails in integration and staging

I haven't thought too much about it, but structuring the documentation around the apps makes the most sense to me. So one section or page about Email Alert API, one section/page about Publisher... That way when you want to test an email from Email Alert API say, you just find the documentation related to that app.

AlanGabbianelli:
> I haven't thought too much about it, but structuring the documentation around the apps makes the most sense to me. So one section or page about Email Alert API, one section/page about Publisher... That way when you want to test an email from Email Alert API say, you just find the documentation related to that app.

Wouldn't that involve a lot of duplication for apps that send emails similarly?

thomasleese:
> Wouldn't that involve a lot of duplication for apps that send emails similarly?

Possibly, although I'm not sure that we have many apps that send emails in a similar way. The only commonality between Email Alert API, Publisher, Support, etc is that they use Notify. Apart from that, the way of getting them to actually send out an email in integration/staging is all quite different.

For Email Alert API, we've already got docs that cover this. For Support it involves making a request for some feedback and pressing the "Export CSV" button and for Publisher it involves sending a fact-check email.

benthorner:
@AlanGabbianelli I think this is superseded by @kevindew's iteration: https://github.com/alphagov/govuk-developer-docs/pull/2886.

PR #2489, opened 2020-04-22
github-trello-poster: Emma has left, note the use of a `govuk-ci` token

issyl0:
https://trello.com/c/ia2MkYpQ/1916-2-get-github-trello-poster-using-an-api-key-we-own


ChrisBAshton:
(Might be worth linking to https://trello.com/c/ia2MkYpQ/1916-2-get-github-trello-poster-using-an-api-key-we-own in the commit message)

PR #2492, opened 2020-04-22
Review more out of date docs for 2nd line

thomasleese:
These pages look up to date to me.


issyl0:
> -   This check is there to flag up potential problems occurring upstream
    -- a service not returning fast enough so we have the set write
    connecting building up. This was a pattern spotted in a number of
    recent outages as at April 2013.

Maybe remove the word "recent"? 🧓Also I am astonished that nothing has changed since 2013 and that we still have this at all, but...

thomasleese:
> > * This check is there to flag up potential problems occurring upstream
> >   -- a service not returning fast enough so we have the set write
> >   connecting building up. This was a pattern spotted in a number of
> >   recent outages as at April 2013.
> 
> Maybe remove the word "recent"? 🧓Also I am astonished that nothing has changed since 2013 and that we still have this at all, but...

Yes, I was half tempted to remove that alert and document because I don't remember it happening at any point, but I didn't feel confident enough!

issyl0:
I'd remove it, if we need it we can put it back. Looking at Icinga it hasn't fired in... 154 days?

thomasleese:
> I'd remove it, if we need it we can put it back. Looking at Icinga it hasn't fired in... 154 days (and probably way beyond that, I don't know how far back Icinga's data goes...)?

Cool, I've opened a PR to remove it!

PR #2496, opened 2020-04-26
Condense Sidekiq Monitoring setup docs into main Sidekiq doc

issyl0:
- It felt like a waste to have two docs for this, when the second doc about setting up a new app to have Sidekiq Monitoring wasn't actually that long.
- Here's one doc, that we only have to review once every six months.


source/manual/sidekiq.html.md
```diff
@@ -38,14 +38,45 @@ $ gds govuk connect sidekiq-monitoring -e production aws/backend
 
 Go to the `127.0.0.1:port` URL in the command output to see the UI.
 
+#### Set up Sidekiq Monitoring for your application
+
+- Choose a port that isn't already taken for the Sidekiq Monitoring
+  app to be served from.
+- Add it to the Procfile in the [sidekiq-monitoring repository](https://github.com/alphagov/sidekiq-monitoring)
```
alex-ju, line 45:
```suggestion
- Add it to the [Procfile in the sidekiq-monitoring repository](https://github.com/alphagov/sidekiq-monitoring/blob/master/Procfile)
```

issyl0: Thanks, Alex!

PR #2497, opened 2020-04-26
Use GitHub Actions, not Jenkins, for CI

issyl0:
- This runs Rake and RuboCop checks with the Ruby version defined in
  `.ruby-version`, for ease of updating later.
- The deployment of this app is done on the deploy Jenkins, once an hour
  on a timer, and the job `bin/deploy-to-s3` builds the site and pushes
  the files to S3. The CI Jenkins is not involved in the deployment
  process at all, and we don't seem to use the release tags generated in
  the master build process for anything.
- This reduces the amount of load on our Jenkins CI servers, and gets us
  further towards RFC-123.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,6 +9,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+
+      - name: Clone govuk-content-schemas
+        uses: actions/checkout@v2
+        with:
+          repository: alphagov/govuk-content-schemas
+          ref: deployed-to-production
+          path: ./govuk-content-schemas
       - uses: ruby/setup-ruby@v1
       - uses: actions/cache@v1
         with:
@@ -22,4 +29,6 @@ jobs:
           bundle install --jobs 4 --retry 3
 
       - run: bundle exec rake
+        env:
+          GOVUK_CONTENT_SCHEMAS_PATH: ./govuk-content-schemas
       - run: bundle exec rubocop
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,16 @@
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+
+      - run: bundle install --jobs 4 --retry 3
+      - run: bundle exec rake
+      - run: bundle exec rubocop
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,20 +15,20 @@ jobs:
         with:
           repository: alphagov/govuk-content-schemas
           ref: deployed-to-production
-          path: ./govuk-content-schemas
+          path: tmp/govuk-content-schemas
       - uses: ruby/setup-ruby@v1
       - uses: actions/cache@v1
         with:
-          path: actions-gem-cache/bundle
+          path: tmp/bundle
           key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
           restore-keys: |
             ${{ runner.os }}-gems-
 
       - run: |
-          bundle config path actions-gem-cache/bundle
+          bundle config path tmp/bundle
           bundle install --jobs 4 --retry 3
 
       - run: bundle exec rake
         env:
-          GOVUK_CONTENT_SCHEMAS_PATH: ./govuk-content-schemas
+          GOVUK_CONTENT_SCHEMAS_PATH: tmp/govuk-content-schemas
       - run: bundle exec rubocop
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,16 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: ruby/setup-ruby@v1
+      - uses: actions/cache@v1
+        with:
+          path: actions-gem-cache/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-
 
-      - run: bundle install --jobs 4 --retry 3
+      - run: |
+          bundle config path actions-gem-cache/bundle
+          bundle install --jobs 4 --retry 3
+
       - run: bundle exec rake
       - run: bundle exec rubocop
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,10 +6,5 @@ inherit_mode:
   merge:
     - Exclude
 
-AllCops:
-  Exclude:
-    - "actions-gem-cache/**/*"
-    - "govuk-content-schemas/**/*"
-
 Style/ClassVars:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -9,6 +9,7 @@ inherit_mode:
 AllCops:
   Exclude:
     - "actions-gem-cache/**/*"
+    - "govuk-content-schemas/**/*"
 
 Style/ClassVars:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,5 +2,13 @@ inherit_gem:
   rubocop-govuk:
     - config/default.yml
 
+inherit_mode:
+  merge:
+    - Exclude
+
+AllCops:
+  Exclude:
+    - "actions-gem-cache/**/*"
+
 Style/ClassVars:
   Enabled: false
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,0 @@
-#!/usr/bin/env groovy
-
-library("govuk")
-
-node {
-  govuk.buildProject(skipDeployToIntegration: true)
-}


issyl0:
RuboCop is trying to [lint the entire world](https://github.com/alphagov/govuk-developer-docs/pull/2497/checks?check_run_id=620018718#step:8:6) instead of just our app files, but this is good to go once that is fixed. As a result, I'm marking it as not a draft and will come back to the RuboCop failure tomorrow.

This whole PR turned out to be more complicated than I expected, because I entirely forgot that this app does things with `govuk-content-schemas` - because it's abstracted away so much into `govuk-jenkinslib`! But now it works, and we have an example for CI that uses the `govuk-content-schemas`.

Related improvements could include changing the directory that `govuk-schemas` searches for the govuk-content-schemas repo, so that we don't have to [override](https://github.com/alphagov/govuk-jenkinslib/blob/master/vars/govuk.groovy#L614) it with an envvar in [all the places](https://github.com/alphagov/govuk-developer-docs/pull/2497/files#diff-353995fb5bebaef307ec25fa5be39042R28).

.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
+        with:
+          repository: alphagov/govuk-content-schemas
+          ref: deployed-to-production
+          path: govuk-content-schemas
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: .ruby-version
```
kevindew, line 15: As this is the default we could drop this line?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
+on: [push, pull_request]
```
kevindew, line 1: I guess this means that once we have a PR every push will be 2 identical builds to wait for?

I saw you justifying we want this for external contributors - it'd be good to work out if we can avoid the double build situation before propagating across our apps.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
+        with:
+          repository: alphagov/govuk-content-schemas
+          ref: deployed-to-production
+          path: govuk-content-schemas
```
kevindew, line 12: As I said on Slack I think we're going to find this tries to be linted as well. We may need to clone the govuk-developer-docs repo into a directory of it's own to isolate it. I guess either that or some env var or config manipulation, though path separation seems cleaner.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/cache@v1
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-
+
+      - run: |
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
```
kevindew, line 25: It seems a bit of a shame for us to propagate these same magic numbers around every build. I can see they come from the action examples, wonder if they could be persuaded to set up env vars to specify the default. 

By the way, not relevant to this repo but for repos with gems in groups we should use the --with=test to only install test gems.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
...
+      - run: bundle exec rake
```
kevindew, line 26: I feel like we're going to hit a consistency issue soon with some apps that use rake default to only run tests and others that use it to run all build tasks including linting. 

As an example Content Publisher does everything in Rake default intentionally: https://github.com/alphagov/content-publisher/blob/c6e81e4a71b9b4a9710638988a6ce0a0366743f9/Rakefile#L10

We're going to have to decide whether we want a simple action file that just calls rake but fails fast or if we want each step called out in steps so that we can see whether lint passed but tests failed (for instance).

.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
...
+        env:
+          GOVUK_CONTENT_SCHEMAS_PATH: ./govuk-content-schemas
+      - run: bundle exec rubocop
```
kevindew, line 29: I wondered whether we were missing sass linting too - but looks like it was never set-up on this repo 😞

.github/workflows/tests.yml
benthorner: Indeed - I think we can just use `push` here.

ChrisBAshton:
@kevindew thanks for the detailed review - I always learn something from your comments!

I think the `release` tag is in fact used by the data sync job, e.g. https://deploy.integration.publishing.service.gov.uk/job/Deploy_App/27880/

.github/workflows/tests.yml
issyl0: I'd be in favour of all the steps being separate, for fast failures. I did a bunch of work last year to remove the `lint` Rake tasks as Jenkins itself ran `RuboCop` for us - but I'd be in favour of running each step individually for fewer abstractions. I'll think about this.

benthorner: +1 for `rake`, as then it's consistent with local development. Also less work than having to manually specify all the steps.

.github/workflows/tests.yml
issyl0: NIIIICE! I failed at reading the docs. 😭

.github/workflows/tests.yml
benthorner: @issyl0 would it actually lead to faster failures? If `rake` is running all the same steps, then it should fail at the same point; arguably, it's faster because there's less startup overhead. Having been annoyed by Jenkins doing something 'different' to running `rake` locally, I'm very keen on us standardising on just `rake`.

issyl0:
Thanks for the comments, @kevindew! I should have spent yesterday working on `alphagov/actions` rather than the individual apps. So far, I'm imagining that we have one that's "set up Ruby, bundle gems" as one step, then the app does whatever else it wants on its own.

> I think the `release` tag is in fact used by the data sync job, e.g. https://deploy.integration.publishing.service.gov.uk/job/Deploy_App/27880/

@ChrisBAshton Oh wow. That means I have more questions than answers. If we don't deploy this app to integration, and it's not a production "app" in the sense that we deploy it to S3, nothing to do with GOV.UK's infrastructure, we can probably stop syncing the data across environments as there's no point.

issyl0:
> Also, I saw you've put a few PR's up like this yesterday. What's the line between which of these are trials to inform the RFC and what are actually implementing the RFC? Just asking as it seems unusual to be implementing the RFC prior to approval

Alright, I apologise, you got me. I was very bored on yesterday's Sunday shift and took the relay and ran with it. I am definitely, I suppose, implementing the RFC now. This one is a good case study of "an app that needs content schemas", and the others... well, Jenkins annoys me. 😳 Maybe an RFC isn't the best format for my brain - I just want to do all the things to make people's developer experience better, and I had time enough to do that yesterday, so I did. 😣 But anyhow, the RFC is overall positive and has had some good comments. We've still got some things to iron out (notably secrets - but none of these jobs require secrets), and there _is_ a week left to do that in, but based on current comments, thoughts and the enthusiasm of other people - rather than just me - for trying Actions out and saying that they're good - I don't imagine at this point that it'll be shouted down to a "no not at all" prior to the deadline.

ChrisBAshton:
@issyl0 ah I meant that we probably can't do away with the `release` tag universally as there is still code that uses it. But in the case of govuk-developer-docs it probably isn't used 🤔 Apologies for the confusion.

issyl0:
@ChrisBAshton Yeah, I wasn't suggesting doing away with it for "normal" apps - I just don't think it's useful for "tooling" apps like this one and other dev tools that aren't deployed to the normal GOV.UK infra. For creating release tags for normal apps, I'll write up some things in the RFC. :-)

.github/workflows/tests.yml
issyl0: If, across all the apps, we can standardise on `rake` running the tests AND RuboCop, that's be good on the whole. But I would like to take into account [Michael's RFC comment](https://github.com/alphagov/govuk-rfcs/pull/123#issuecomment-617175480) about individualised steps being easier to see what failed - but maybe that's for the future.

.github/workflows/tests.yml
issyl0: I'll switch this to use `pull_request` and the `master` branch, as discussed in the RFC. Then we can see how we go.

.github/workflows/tests.yml
issyl0: I've attempted to clone it into `../govuk-content-schemas`, because then we don't need to set the envvar either as that's the default directory that the `govuk_schemas` gem (⁉️) looks in.

issyl0: Looks like it doesn't work like that:

```
 Clone govuk-content-schemas0s
##[error]Repository path '/home/runner/work/govuk-developer-docs/govuk-content-schemas' is not under '/home/runner/work/govuk-developer-docs/govuk-developer-docs'
Run actions/checkout@v2
  with:
    repository: alphagov/govuk-content-schemas
    ref: deployed-to-production
    path: ../govuk-content-schemas
    token: ***
    ssh-strict: true
    persist-credentials: true
    clean: true
    fetch-depth: 1
    lfs: false
    submodules: false
##[error]Repository path '/home/runner/work/govuk-developer-docs/govuk-content-schemas' is not under '/home/runner/work/govuk-developer-docs/govuk-developer-docs'
```

I suppose if Actions cleans up after itself, I guess it only cleans up the current runner directory...

kevindew: Heh - yeah I think that was quite optimistic. Would feel like a bit of a hack. Worth a try though.

We'll have to do similar with vendor/bundle but that'd feel a bit less odd to have the linting ignore directory.

issyl0:
I've updated the gem caching step. The Actions gem cache now writes to `actions-gem-cache`, which is excluded in `.rubocop.yml`. We could have excluded `vendor`, but I don't know if other apps use this directory for important things, and I didn't want to make a pattern that wasn't a pattern. If people like this approach, I'll upstream it to be a default exclude in `rubocop-govuk`.

issyl0:
I excluded the `govuk-content-schemas` directory from RuboCop for this app, as well as `actions-gem-cache`. There's rationale in both the commit messages. And now CI is green. It felt really weird to be linting things that we don't have any direct control over in this app - govuk-content-schemas is a totally different app, where linting errors should be fixed at source - and the gem cache directory.

This seems ready to go now. It was a long ol' slog, but thanks very much for your comments and sticking with it!

.rubocop.yml
```diff
@@ -2,5 +2,14 @@ inherit_gem:
   rubocop-govuk:
     - config/default.yml
 
+inherit_mode:
+  merge:
+    - Exclude
+
+AllCops:
+  Exclude:
+    - "actions-gem-cache/**/*"
```
benthorner, line 11: Would it be possible to put these in a 'tmp' directory? This would mean they automatically get excluding during linting.

issyl0: That's an entirely better idea! Thanks, Ben!

issyl0: That worked! 🎉

.github/workflows/tests.yml
```diff
@@ -10,7 +10,16 @@ jobs:
...
       - uses: ruby/setup-ruby@v1
+      - uses: actions/cache@v1
+        with:
+          path: actions-gem-cache/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-
+
+      - run: |
+          bundle config path actions-gem-cache/bundle
```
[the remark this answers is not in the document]
issyl0: So am I. I'm sure that was intentional to remind me to go and look something up, but I don't remember. 😳It's OK, that's out of date now after Ben's comment. I've put them all in /tmp instead.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          repository: alphagov/govuk-content-schemas
+          ref: deployed-to-production
+          path: tmp/govuk-content-schemas
+      - uses: ruby/setup-ruby@v1
+      - uses: actions/cache@v1
+        with:
+          path: tmp/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: |
```
benthorner, line 24: Could you help me understand what this is for? It seems to be a fallback, but I'm not sure where it came from.

issyl0: The short answer is, I don't know. I copied all the caching stuff from the [actions/cache Bundler example](https://github.com/actions/cache/blob/master/examples.md#ruby---bundler), which has this included. The docs say that it's "an ordered list of keys to use for restoring the cache if no cache hit occurred for key", but that's not terribly enlightening as to _why_.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,34 @@
...
+  push:
+    branches:
+      - master
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
```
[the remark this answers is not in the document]
issyl0: Yeah, we probably should!

PR #2500, opened 2020-04-27
Improve "Deploying" section

ChrisBAshton:
- Correct the reference to 'govuk-jenkinslib', which is a library rather than a job
- Explain how the Jenkins jobs that run app tests are created in the first place
- Remove the confusing 'release tag' terminology, as we currently _do_ create a `release` tag but it is unused. We actually use the build number release tag (e.g. `release_123`) in the deployment pipeline, not the `release` tag.
- Suggest that the CI and Deploy Jenkins live in separate environments.
- Link to some of the code where the tag-pushing and send-message magic happens, to demystify it and to make the docs easier to review.

### DIFF
--- a/source/manual/architecture-deep-dive.html.md
+++ b/source/manual/architecture-deep-dive.html.md
@@ -393,16 +393,21 @@ although apps not hosted on AWS are configured in [env-sync-and-backup].
 
 ### Deploying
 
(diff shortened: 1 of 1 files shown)


ChrisBAshton:
🙄 

Branch name problems. Moving to https://github.com/alphagov/govuk-developer-docs/pull/2501...

PR #2502, opened 2020-04-28
Update ssh instructions

ChrisBAshton:
See commits for details


source/manual/howto-ssh-to-machines-in-aws.html.md
```diff
@@ -74,17 +74,23 @@ And if you can't remember the names of the node classes, there's a built-in help
...
 
 ```sh
-local$ ssh ip-10-1-5-22.eu-west-1.compute.internal.integration
+$ gds govuk connect ssh -e integration ip-10-1-5-22.eu-west-1.compute.internal
+```
+
+To avoid this cumbersome workflow, you could pass an additional argument to the CLI:
+
+```sh
+gds govuk connect ssh -e integration jumpbox "govuk_node_list -c backend"
```
benthorner, line 93:
```suggestion
gds govuk connect ssh -e integration backend
```

We should try to advise the shorter approach 💪

ChrisBAshton: Your suggestion SSHs into a random machine of the `backend` class. I'm trying to document how to list the IPs of the machines of the `backend` class - so that you can then SSH into a specific one if needed. (I needed to recently and got a bit muddled!)

benthorner: Ah, I see. Still a bit confused: how does the IP tell you which one you want to go in to?

ChrisBAshton: In my case, I was trying to find an IP for a specific machine which I hoped had some local logs on it. I ran the equivalent of `gds govuk connect ssh -e integration jumpbox "govuk_node_list -c backend"`, saw a couple of different IP addresses, one of which was the one that matched the log, and I SSH'd.
(Looking back I probably could have SSH'd without ever needing the jumpbox, but I wasn't sure of the `[xx].eu-west-1.compute.internal` syntax, or whether the AWS region might be different, so it was nice to have something concrete to refer to).

A better example is something I'm looking at today - I need to go into every Mapit machine on Integration and do something. I need to find out which IPs exist, and then make sure I SSH into them individually and do the thing, not just rely on whichever one the shorthand command will give me.

issyl0: You can do `gds govuk connect ssh -e integration mapit:1` and `:2` etc. for this.

ChrisBAshton: Ah, nice. And I guess just keep incrementing the number until I hit an error?

benthorner: @ChrisBAshton ah, I see what you mean. Reading the whole thing: this document is pretty superfluous/misleading now - it's all in the `--help` for `govuk-connect`.

I'm tempted to suggest: "delete the whole thing", as it's encouraging people to have an inefficient workflow, with a pointless 'jumpbox' step halfway through.

As a less drastic change, how about removing the "Jumpbox" section, and bringing your last command up to just after "Alternatively, you can use the jumpbox."?

I think that achieves the purpose of the manual, without making it too longwinded. What do you think?

ChrisBAshton: I see this has been done in #2555 - thanks Ben!

PR #2507, opened 2020-04-29
Add some docs on how we use GOV.UK Notify

thomasleese:
We're starting to use GOV.UK Notify in lots of places (not just Email Alert API), so this updates the docs to make this clearer and start to introduce separation between GOV.UK Notify and Email Alert API in peoples minds.

### DIFF
--- a/source/manual/govuk-notify.html.md
+++ b/source/manual/govuk-notify.html.md
@@ -1,0 +1,71 @@
+---
+owner_slack: "#govuk-developers"
+title: How we use GOV.UK Notify
+section: Emails
+type: learn
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2020-04-29
+review_in: 6 months
+---
+
+[GOV.UK Notify][notify] is a Government-as-a-Platform service that allows
+clients of their API to send emails, text messages and letters. We use GOV.UK
+Notify to send emails to users - both citizens and publishers. Historically
+we've also used AWS SES to send emails, but that's being phased out in favour
+of GOV.UK Notify.
+
+[notify]: https://www.notifications.service.gov.uk/
+
+We have two main services on GOV.UK Notify (six in total as we have a service
+for each environment):
+
+- **GOV.UK Emails**
+
+  This service is used by Email Alert API only. It's used to send citizen-facing
+  email updates about pieces of content on GOV.UK. It's our biggest sender of
+  emails and regularly exceeds one million emails per day.
+
+- **GOV.UK Publishing**
+
+  This service is used by our publishing applications. It's used to send
+  publisher-facing emails. The type of email can vary a lot, but a typical
+  example might be to receive an email when a scheduled document on GOV.UK has
+  been automatically published.
+
+_In the future we may set up a new 'GOV.UK Public' service which is used to
+send citizen-facing emails which haven't gone through Email Alert API._
+
+## Accessing the dashboard
+
+**[👉 Sign in to the GOV.UK Notify dashboard](https://www.notifications.service.gov.uk/sign-in)**
+
+You can either use your own credentials (if you have them) or you can use the
+credentials in [govuk-secrets] (found in the `govuk-notify/2nd-line-support`
+entry).
+
+[govuk-secrets]: https://github.com/alphagov/govuk-secrets
+
+## Receiving emails from GOV.UK Notify
+
+GOV.UK Notify services have two modes: live and [trial][trial-mode]. Only our
+production services are in live mode, the others are in trial mode. In this
+mode emails will only be sent to members of the service or email address in the
+allow-list, any request to send an email to another email address will fail.
+
+[trial-mode]: https://www.notifications.service.gov.uk/using-notify/trial-mode
+
+To receive emails in integration and staging through GOV.UK Notify, you should
+add yourself to the service:
+
+1. Choose the service in the appropriate environment and navigate to
+   "Team members".
+
+2. The members with the permission `Manage settings, team and usage` will be
+   able to add you to this team.
+
+**Note:** some of our applications, notably Email Alert API, has an extra level
+of protection and [there is an extra step][email-alert-api-receive-emails]
+before you can receive emails through Email Alert API.
+
+[email-alert-api-receive-emails]: /manual/receiving-emails-from-email-alert-api-in-integration-and-staging.html
--- a/source/manual/receiving-emails-from-email-alert-api-in-integration-and-staging.html.md
+++ b/source/manual/receiving-emails-from-email-alert-api-in-integration-and-staging.html.md
@@ -4,7 +4,7 @@ title: Receive emails from Email Alert API in integration and staging
 section: Emails
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2020-02-28
+last_reviewed_on: 2020-04-29
 review_in: 6 months
 ---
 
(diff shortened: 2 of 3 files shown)


thomasleese:
I think this might help to structure #2487.

source/manual/govuk-notify.html.md
```diff
@@ -0,0 +1,73 @@
...
+
+[GOV.UK Notify][notify] is as Government-as-a-Platform service that allows
+clients of their API to send emails, text messages and letters. We use GOV.UK
+Notify to send emails to users, both citizens and publishers. Historically
+we've also used AWS SES to send emails, but that's being phased out in favour
+of GOV.UK Notify.
+
+[notify]: https://www.notifications.service.gov.uk/
+
+We have two main services on GOV.UK Notify (six in total as we have an service
```
issyl0, line 20:
```suggestion
We have two main services on GOV.UK Notify (six in total as we have a service
```

source/manual/govuk-notify.html.md
```diff
@@ -0,0 +1,73 @@
...
+for each environment):
+
+- **GOV.UK Emails**
+
+  This service is used by Email Alert API only. It's used to send citizen-facing
+  email updates about pieces of content on GOV.UK. It's our biggest sender of
+  emails and can regularly exceed one million emails per day.
```
issyl0, line 27:
```suggestion
  emails and regularly exceeds one million emails per day.
```

source/manual/govuk-notify.html.md
thomasleese: Ah... I did a find a replace of `account` to `service` to be consistent!

thomasleese:
Thanks @issyl0, those should be resolved now!

source/manual/govuk-notify.html.md
```diff
@@ -0,0 +1,73 @@
...
+title: How we use GOV.UK Notify
+section: Emails
+type: learn
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2020-04-29
+review_in: 6 months
+---
+
+[GOV.UK Notify][notify] is as Government-as-a-Platform service that allows
```
ChrisBAshton, line 12:
```suggestion
[GOV.UK Notify][notify] is a Government-as-a-Platform service that allows
```

source/manual/govuk-notify.html.md
```diff
@@ -0,0 +1,73 @@
...
+clients of their API to send emails, text messages and letters. We use GOV.UK
+Notify to send emails to users, both citizens and publishers. Historically
```
ChrisBAshton, line 14:
```suggestion
Notify to send emails to users - both citizens and publishers. Historically
```

source/manual/govuk-notify.html.md
```diff
@@ -0,0 +1,73 @@
...
+
+- **GOV.UK Publishing**
+
+  This service is used by our publishing applications. It's used to send
+  publisher-facing emails. The type of email can vary a lot, but a typical
+  example might be to receive an email when a scheduled document on GOV.UK has
+  been automatically published.
+
+_In the future we may set up a new 'GOV.UK Public' service which is used to
+send citizen-facing emails which haven't gone through Email Alert API._
```
ChrisBAshton, line 37: I'd remove this line - it's missing the 'why', and is an unnecessary complication until it's implemented.

source/manual/govuk-notify.html.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Accessing the dashboard
+
+**[👉 Sign in to the GOV.UK Notify dashboard](https://www.notifications.service.gov.uk/sign-in)**
```
ChrisBAshton, line 41: This should be an aliased markdown link like the others, for consistency.

source/manual/govuk-notify.html.md
```diff
@@ -0,0 +1,73 @@
...
+
+You can either use your own credentials (if you have them) or you can use the
+credentials in [govuk-secrets].
+
+[govuk-secrets]: https://github.com/alphagov/govuk-secrets
```
ChrisBAshton, line 46: Minor: Alias links are usually grouped together 'reference style', at the end of the page or section. I don't see much of an advantage to making them aliased if they're just dropped in below the sentence they're used.

source/manual/govuk-notify.html.md
```diff
@@ -0,0 +1,73 @@
...
+
+_In the future we may set up a new 'GOV.UK Public' service which is used to
+send citizen-facing emails which haven't gone through Email Alert API._
+
+## Accessing the dashboard
+
+**[👉 Sign in to the GOV.UK Notify dashboard](https://www.notifications.service.gov.uk/sign-in)**
+
+You can either use your own credentials (if you have them) or you can use the
+credentials in [govuk-secrets].
```
ChrisBAshton, line 44:
```suggestion
credentials in [govuk-secrets] (in the `govuk-notify/2nd-line-support` directory).
```

Then we could remove lines 48-49.

source/manual/govuk-notify.html.md
```diff
@@ -0,0 +1,73 @@
...
+
+[govuk-secrets]: https://github.com/alphagov/govuk-secrets
+
+The credentials in `govuk-notify/2nd-line-support` has access to all of our
+services.
+
+## Receiving emails from GOV.UK Notify
+
+GOV.UK Notify services have two modes: live and trial. Only our production
```
ChrisBAshton, line 53: I'd move the `https://www.notifications.service.gov.uk/using-notify/trial-mode` link to here:

```suggestion
GOV.UK Notify services have two modes: live and [trial][trial-mode]. Only our production
```

...otherwise the link on "the others are in trial mode" suggests we're linking to our configuration, not an explainer doc.

source/manual/receiving-emails-from-email-alert-api-in-integration-and-staging.html.md
```diff
@@ -13,20 +13,11 @@ to a single test address: `success@simulator.amazonses.com`. This is used to
 simulate a successful email sending.
 
 However, you can override this for specific email addresses for testing
-purposes. To do this, you will need to be added as a team member to
-the GOV.UK Email Integration or Staging service in Notify and make
-changes to govuk-puppet.
+purposes. To do this, [you will need to be added as a team member to
+the GOV.UK Email Integration or Staging service in Notify][add-in-notify] and
+make changes to govuk-puppet.
```
ChrisBAshton, line 18:
```suggestion
make the following changes to [govuk-puppet]:
```

Then remove the `In [govuk-puppet][]` sentence below.

source/manual/receiving-emails-from-email-alert-api-in-integration-and-staging.html.md
```diff
@@ -13,20 +13,11 @@ to a single test address: `success@simulator.amazonses.com`. This is used to
...
-   [govuk-secrets][] under `govuk-notify/2nd-line-support`. You should
-   then receive an email via 2nd Line Support containing a link to
-   sign in to Notify.
-2. Choose the service (GOV.UK Email) in the appropriate environment
-   and navigate to "Team members".  The members with the permission
-   `Manage settings, team and usage` will be able to add you to this
-   team.
+[add-in-notify]: /manual/govuk-notify.html#receiving-emails-from-govuk-notify
 
 In [govuk-puppet][]:
```
ChrisBAshton, line 22:
```suggestion
```

source/manual/govuk-notify.html.md
thomasleese: I added this because it's come up a few times (and more recently with the Coronavirus services) about which GOV.UK Notify account to use when sending public facing emails and not using Email Alert API. It's likely we're going to need to create an account for this very soon (@kevindew knows more at the moment) but it's not there yet, which is why Ieft the line this like.

I could remove it, but I wanted it there as a kind of reminder to the next person who reviews it...

source/manual/govuk-notify.html.md
thomasleese: I tend to only use the aliased links if the link appears in a large paragraph (and therefore having the entire URL in there makes it harder to read the Markdown in a text editor) of if I link to the place multiple times (to avoid the duplication). I would say it's fine to leave this link with the full URL since it's kind of a special "call-to-action", but I'm also fine with making this an aliased link.

source/manual/govuk-notify.html.md
thomasleese: Kind of similar to above, the advantage I see of the aliased links is mostly to improve the readability of the paragraph while reading it as Markdown, and I prefer to keep the aliased links relatively close to where they are used so you don't need to scroll too much to find the links.

source/manual/govuk-notify.html.md
ChrisBAshton: Oh ok, I was missing that context. On the basis that it becomes reality soon, I'm happy for this to be in the docs.

source/manual/govuk-notify.html.md
ChrisBAshton: I guess we don't have a markdown style guide, so we're all going to develop our own preferences 😅

PR #2509, opened 2020-05-01
Add Publishing API docs, App docs, DocumentTypes/Supertypes to search

ChrisBAshton:
Add Publishing API docs, App docs, DocumentTypes/Supertypes to search
  
- Moves this config into a new class 'ProxyPages' so that it can be tested
- Adds a 'title' frontmatter property so that it is indexed by middleman-search

This makes the search far more comprehensive and useful, although it does
slow the initial indexing down a bit, so the 'Loading search index'
message will stick around slightly longer. I think this is worth it for
the improved relevancy of search results, and if it becomes a problem we
can investigate a server-side search/index solution, or make better use
of caching. It should also only be a short term decrease in search speed;
when the latest version of middleman-search-gds is published, we'll
benefit by only indexing the titles of these external pages, not their
contents: https://github.com/alphagov/middleman-search/pull/5

Unfortunately, until that gem is released, I can't add the
govuk_schema_names pages as they break the indexing altogether.
It will be possible to add them in once the gem is updated.

Trello: https://trello.com/c/lVTGSjCA/142-imported-dev-docs-dont-appear-in-search

## Before
Screenshot 2020-05-01 at 16 40 02

## After
Screenshot 2020-05-01 at 16 39 28

### DIFF
--- a/app/proxy_pages.rb
+++ b/app/proxy_pages.rb
@@ -1,0 +1,110 @@
+class ProxyPages
+  def self.resources
+    publishing_api_docs +
+      govuk_schema_names +
+      app_docs +
+      app_docs_json +
+      document_types +
+      supertypes
+  end
+
+  def self.publishing_api_docs
+    PublishingApiDocs.pages.map do |page|
+      {
+        path: "/apis/publishing-api/#{page.filename}.html",
+        template: "templates/publishing_api_template.html",
+        frontmatter: {
+          title: "Publishing API: #{page.title}",
+          locals: {
+            title: "Publishing API: #{page.title}",
+            page: page,
+          },
+        },
+      }
+    end
+  end
+
+  def self.govuk_schema_names
+    GovukSchemas::Schema.schema_names.map do |schema_name|
+      schema = ContentSchema.new(schema_name)
+
+      {
+        path: "/content-schemas/#{schema_name}.html",
+        template: "templates/schema_template.html",
+        frontmatter: {
+          # title: "Schema: #{schema.schema_name}", # FIXME - this slows down the search index too much
+          locals: {
+            title: "Schema: #{schema.schema_name}",
+            description: "Everything about the '#{schema.schema_name}' schema",
+            schema: schema,
+          },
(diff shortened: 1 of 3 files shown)


app/proxy_pages.rb
```diff
@@ -0,0 +1,110 @@
...
+      app_docs +
+      app_docs_json +
+      document_types +
+      supertypes
+  end
+
+  def self.publishing_api_docs
+    PublishingApiDocs.pages.map do |page|
+      {
+        url: "/apis/publishing-api/#{page.filename}.html",
```
thomasleese, line 14:
```suggestion
        path: "/apis/publishing-api/#{page.filename}.html",
```

I think this is better named `path` as it's not a full URL.

app/proxy_pages.rb
```diff
@@ -0,0 +1,110 @@
...
+        template: "templates/publishing_api_template.html",
+        frontmatter: {
```
thomasleese, line 16: I'm unsure about the naming `frontmatter` here as I was a bit unsure what this was, perhaps `options` is a better name here...

ChrisBAshton: Agreed, it's not very intuitive, except that it is Googlable ([it's what Middleman uses](https://middlemanapp.com/basics/frontmatter/) to refer to the page-level attributes, so seems reasonable to use the same terminology). I think it makes it clearer that it's a Middleman-specific thing, rather than some app-specific thing we've come up with ourselves. Happy to rename if you think we should though.

app/proxy_pages.rb
ChrisBAshton: Thanks - have now updated 👍

app/proxy_pages.rb
thomasleese: Hmm, I was under the assumption that `locals` was the frontmatter (i.e. they get used when rendering the page), and the third argument to the function was just an `options`. I don't think it matters too much though... I'm happy to leave it as it is.

PR #2510, opened 2020-05-04
Fix link to entry for GOV.UK Notify

chao-xian:


source/manual/govuk-notify.html.md
```diff
@@ -41,8 +41,7 @@ send citizen-facing emails which haven't gone through Email Alert API._
 **[👉 Sign in to the GOV.UK Notify dashboard](https://www.notifications.service.gov.uk/sign-in)**
 
 You can either use your own credentials (if you have them) or you can use the
-credentials in [govuk-secrets] (found in the `govuk-notify/2nd-line-support`
-entry).
+credentials in [govuk-secrets] (https://github.com/alphagov/govuk-secrets/blob/master/pass/2ndline/govuk-notify/2nd-line-support.gpg).
```
thomasleese, line 44:
```suggestion
credentials in [govuk-secrets][] (found in the `govuk-notify/2nd-line-support`
entry).
```

Ah, I think I know why it wasn't working. It was interpreting the part in brackets as the URL for `govuk-secrets`. I think instead if we just change it to `[govuk-secrets][]` that should work now.

PR #2511, opened 2020-05-04
Use more inclusive terminology (public vs citizen)

AlanGabbianelli:
"Citizen" is a very specific term and should only be used when we
specifically mean someone with a particular citizenship. Users of GOV.UK
services are generally not only British citizens but the broader public.
Using more general terms ("people", "public", "members of the public")
instead of "citizen" makes this clear and our docs more inclusive.

### DIFF
--- a/source/manual/covid-19-services.html.md
+++ b/source/manual/covid-19-services.html.md
@@ -26,7 +26,7 @@ services such as medical equipment, hotel rooms or childcare.
 
 ## Get coronavirus support as an extremely vulnerable person
 
-This service allows citizens identified as vulnerable by the NHS to
+This service allows people identified as vulnerable by the NHS to
 tell us if they need help accessing essential supplies and support.
 Users will have received a link to the service in a letter or a text
 message from the NHS, or been advised by their GP to fill in the form.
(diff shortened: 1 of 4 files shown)


source/manual/covid-19-services.html.md
```diff
@@ -26,7 +26,7 @@ services such as medical equipment, hotel rooms or childcare.
 
 ## Get coronavirus support as an extremely vulnerable person
 
-This service allows citizens identified as vulnerable by the NHS to
+This service allows the public identified as vulnerable by the NHS to
```
joelanman, line 29: maybe 'people' is better here?

AlanGabbianelli: Or maybe "individuals"? 🤔

richardTowers: I like "people" too. If not, maybe "members of the public" reads more naturally here.

source/manual/govuk-notify.html.md
```diff
@@ -11,7 +11,7 @@ review_in: 6 months
 
 [GOV.UK Notify][notify] is a Government-as-a-Platform service that allows
 clients of their API to send emails, text messages and letters. We use GOV.UK
-Notify to send emails to users - both citizens and publishers. Historically
+Notify to send emails to users - both public and publishers. Historically
```
richardTowers, line 14:
```suggestion
Notify to send emails to users - both members of the public and publishers. Historically
```

0atman:
_Denizen_ is a 1:1 replacement word, if that's useful in the future :ok_hand:

PR #2515, opened 2020-05-04
Review "Re-create an AWS mongo instance"

camdesgov:
Information is correct and up to date.


issyl0:
The first line:

> Deprecation note

The linked PR was merged on 12th September 2019. Do we still need this doc at all? Has anyone seen a recurrence of this problem since?

camdesgov:
> The first line:
> 
> > Deprecation note
> 
> The linked PR was merged on 12th September 2019. Do we still need this doc at all? Has anyone seen a recurrence of this problem since?

Hmmm I don't believe they have....Also our self hosted mongo instances will be deleted after the migration finishes as everything is being moved to DocumentDB....But I kind of feel that this might be worth keeping around until then if some poor confused soul needs to terraform a mongo instance at 2am...

PR #2516, opened 2020-05-05
Add link for docs on configuring secrets to COVID-19 services page

brucebolt:
The process for managing deployment secrets in the COVID-19 services differs from other GOV.UK applications, therefore adding a link to the relevant documentation so anyone searching for "secrets" will be able to find it.

### DIFF
--- a/source/manual/covid-19-services.html.md
+++ b/source/manual/covid-19-services.html.md
@@ -184,9 +184,9 @@ To log in to AWS:
 
 Deployment is managed via [TechOps shared Concourse][big-concourse].
 CI/CD pipelines are configured under the `govuk-tools` team. You can
-read about what Concourse is, its access controls, and how to
-administer it, on the [Concourse](/manual/concourse.html) developer
-docs page.
+read about what Concourse is, its access controls, [configuring secrets][re-secrets],
+and how to administer it, on the [Concourse](/manual/concourse.html)
+developer docs page.
 
 In terms of these applications, any changes to application code on the
 master branch are continuously deployed to staging.  Smoke tests are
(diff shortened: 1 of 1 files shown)


source/manual/covid-19-services.html.md
```diff
@@ -155,6 +155,12 @@ All applications are Rails applications, and use the
 about [managing PaaS applications][paas-managing] and [monitoring PaaS
 applications][paas-monitoring].
 
+#### Manage deployment secrets
+
+Secrets are configured using Reliability Engineering's secrets management
+system.  [Detailed documentation][re-secrets] is available in the
```
ChrisBAshton, line 161: [One space please](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style) 😄 

```suggestion
system. [Detailed documentation][re-secrets] is available in the
```

issyl0:
This duplicates what's in the [main page about Concourse](https://docs.publishing.service.gov.uk/manual/concourse.html). Might be worth trimming those words down instead?

source/manual/covid-19-services.html.md
brucebolt: This goes anything everything I was ever taught .... but it's in the style guide, so I've made the change 😁

brucebolt:
> This duplicates what's in the [main page about Concourse](https://docs.publishing.service.gov.uk/manual/concourse.html). Might be worth trimming those words down instead?

Thanks Issy.  I have updated the section where we talk about Concourse instead so the word "secrets" is explicitly included.  I previously had no idea how these were configured for these services and was searching for this word without any success.

PR #2519, opened 2020-05-05
Consume application READMEs into the Dev Docs

ChrisBAshton:
This adds to the App page template so that the corresponding README is rendered on the page. This is also indexed by search, so gives us a richer searching experience. Note that the size of the search.json file rises by about 200 KB with this change. This is a reasonable tradeoff for the increased volume of content and more powerful search.

This change allows us to be more consistent about the way we consume external docs, and resulted in the removal of various 'stub pages' where there is no real content other than the pulled-in README. It is also something of a precursor to auto-ingesting `docs/` from repos, which I hope to put on the roadmap before too long.

As well as the README ingestion, this PR makes several small improvements to the layout of the app page, consolidating related sections and fixing the semantics of the headings.

# API limit

We now make 73 additional API requests to GitHub (one for every app page). The requests are made at startup time and the responses cached, so the user shouldn't see a difference, but it is quite easy to hit the API limit in local development, as the [limit for unauthenticated requests is 60 per hour](https://developer.github.com/v3/#rate-limiting), versus 5000 per hour for if you provide an environment `GITHUB_TOKEN`.

For this reason, I've wrapped the README inclusion in a conditional, so that the API calls aren't made unless the GITHUB_TOKEN is defined. This should protect developers who run `./startup.sh` without defining a token, which shouldn't be necessary for them if they're working on something unrelated.

# Visual difference

In my haste I've screenshotted two different pages, but hopefully you get the idea.

App page before:

before

App page after:

screencapture-localhost-4567-apps-content-store-html-2020-05-21-10_53_24

Local development if no GITHUB_TOKEN ENV variable set:

> Screenshot 2020-05-05 at 17 36 05

Trello: https://trello.com/c/rzunnimb/143-automatically-consume-all-govuk-repo-readmes


source/templates/application_template.html.md.erb
```diff
@@ -16,37 +16,63 @@ This application is retired.
...
-### Links
+<% if application.can_run_rake_tasks_in_jenkins? %>
+## Run a rake task
 
-<ul>
-  <li><%= link_to "GitHub repo", application.repo_url %></li>
+<%= RunRakeTask.links(application) %>
+<% end %>
 
+<h2>Links</h2><!--FIXME - markdown not working-->
```
benthorner, line 74: I had a play with this, and found that we can fix it by [putting the `<ul>`s on newlines](https://github.com/alphagov/govuk-developer-docs/blob/master/app/run_rake_task.rb#L19). You'll need to restart the app to reload the code.

source/templates/application_template.html.md.erb
```diff
@@ -136,24 +139,44 @@ gds govuk connect ssh -e production aws/<%= application.aws_puppet_class %>
...
-<h3>Example pages rendered by <%= application.app_name %></h3>
+## GitHub
 
-<ul>
-<% application.example_rendered_pages.each do |example| %>
-  <li><%= link_to example['title'], "https://www.gov.uk#{example['link']}" %></li>
+<%= link_to "github.com/alphagov/#{application.github_repo_name}", application.repo_url %>
+
+<% unless application.topics.nil? %>
+### Topics
```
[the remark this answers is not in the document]
ChrisBAshton: Agreed - I only kept it in case I was missing some historical context and others find it useful. I can't imagine it being particularly useful! Will remove.

app/external_doc.rb
```diff
@@ -42,6 +42,19 @@ def self.fetch(repository:, path:)
       .to_html(contents, context)
   end
 
+  def self.parse(markdown)
+    filters = [
+      HTML::Pipeline::MarkdownFilter,
+      HTML::Pipeline::AbsoluteSourceFilter,
+      PrimaryHeadingFilter,
+      HeadingFilter,
+      # AbsoluteLinkFilter, # FIXME
```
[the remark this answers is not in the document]
ChrisBAshton: Fair enough, I'll remove the line 👍

source/templates/application_template.html.md.erb
```diff
@@ -16,37 +16,63 @@ This application is retired.
...
+
+<% if application.team == application.dependencies_team %>
+**<%= application.team %>** owns the application and is responsible for updating its dependencies.
+<% elsif application.team && application.dependencies_team %>
+**<%= application.team %>** owns the application, but **<%= application.dependencies_team %>** is responsible for updating its dependencies.
+<% elsif application.team %>
+**<%= application.team %>** owns the application.
+<% elsif application.dependencies_team %>
+**<%= application.dependencies_team %>** is responsible for updating its dependencies.
+<% end %>
```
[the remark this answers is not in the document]
ChrisBAshton: I could certainly put an `else` in 😂

source/templates/application_template.html.md.erb
```diff
@@ -16,37 +16,63 @@ This application is retired.
...
 
-Owned by team **<%= application.team %>**.
+<% if application.hosting_name == "None" %>
+There is no production version of this application yet.
+<% else %>
+The production version of this application is hosted on **<%= application.hosting_name %>**.
 <% end %>
 
-<% if application.dependencies_team %>
+<% case application.production_hosted_on %>
```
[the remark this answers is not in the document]
ChrisBAshton: I already have something like this on the epic backlog 👍

source/templates/application_template.html.md.erb
ChrisBAshton: Thanks for investigating, it was driving me crazy! I'll see if I can incorporate your suggestion into the commit.

ChrisBAshton:
@benthorner I've just had a play at incorporating the tabs component into the page, as I agree it would be nice to have, say, two tabs for `Overview` and `Readme`. Unfortunately, govuk_publishing_components is not included in govuk-developer-docs at the moment, and an attempt to do so fails because [govuk_publishing_components is locked to Sprockets <4](https://github.com/alphagov/govuk_publishing_components/commit/e33e32c91bab8c2126318dc849b471c94bf785b8) and [tech_docs_gem requires v4 or above](https://github.com/alphagov/tech-docs-gem/blob/master/govuk_tech_docs.gemspec#L49).

I managed to hack around this locally, but then struggled to include the necessary CSS and JS, as govuk-developer-docs is not a Rails app. This feels like quite a big leap to make in an already sizeable PR, for a somewhat subjective change (and don't fancy opening this can of worms with the front-end devs just yet!), so I'm inclined to leave as-is for now and look at iterating it later.

app/github_repo_fetcher.rb
```diff
@@ -17,6 +17,14 @@ def repo(app_name)
     end
   end
 
+  # Fetch a README for an alphagov application and cache it.
+  # Note that it is cached as pure markdown and requires further processing.
+  def readme(app_name)
+    CACHE.fetch("alphagov/#{app_name} README", expires_in: 1.hour) do
+      Base64.decode64(client.readme("alphagov/#{app_name}").content)
```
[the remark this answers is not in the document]
ChrisBAshton: I figured all of our apps have READMEs 🤷‍♂️ 

But you're right, we could one day find ourselves pulling in a repo that has no README and we shouldn't bork if that happens. This should return `nil` in that case. I'll tweak the code and add a test for it.

source/templates/application_template.html.md.erb
ChrisBAshton: Hi Ben, I may be misunderstanding your suggestion, but every I attempt I make is _not_ converting the markdown 😢 

```md


##Links


<ul>


  <% if application.sentry_url %>
```

Screenshot 2020-05-21 at 10 42 24

benthorner: So I'm saying you need to reformat [this specific line](https://github.com/alphagov/govuk-developer-docs/blob/master/app/run_rake_task.rb#L19) with new lines. That worked for me.

spec/app/github_repo_fetcher_spec.rb
```diff
@@ -18,4 +18,34 @@
...
+        .to_return(status: 200, body: '{ "content": {} }', headers: { content_type: "application/json" })
+
+      GitHubRepoFetcher.new.readme(repo_name)
+      GitHubRepoFetcher.new.readme(repo_name)
+      expect(stubbed_request).to have_been_requested.once
+    end
+
+    it "retrieves the README content from the GitHub API response" do
+      repo_name = SecureRandom.uuid
+      response = { "content": "IyB0ZW1wb3JhcnktdGVzdA==\n" }
```
benthorner, line 36: Can we explain where this comes from?

source/templates/application_template.html.md.erb
ChrisBAshton: Thanks Ben - I've made you a co-author!

spec/app/github_repo_fetcher_spec.rb
ChrisBAshton: I've rewritten this to be a bit more self documenting. (But it came from making an API call to https://github.com/ChrisBAshton/temporary-test, which I've since deleted).

PR #2524, opened 2020-05-06
Import Email Alert API docs

ChrisBAshton:
This uses the same approach as the Publishing API. It's not
sustainable in the long term, but we plan on abstracting
the way we import docs in the future.

I've imported all the existing markdown files in the dir,
apart from those in the 'arch' subfolder, which don't seem
hugely important to import at this stage.

Screenshot:

Screenshot 2020-05-06 at 10 03 30

Trello: https://trello.com/c/KNesH0LZ/1939-3-improve-documentation-for-email-alert-api


app/proxy_pages.rb
```diff
@@ -24,6 +25,22 @@ def self.publishing_api_docs
     end
   end
 
+  def self.email_alert_api_docs
+    EmailAlertApiDocs.pages.map do |page|
+      {
+        path: "/apis/email-alert-api/#{page.filename}.html",
+        template: "templates/publishing_api_template.html",
```
thomasleese, line 32: I wonder if we could rename this template to `external_doc_template` perhaps.

app/email_alert_api_docs.rb
```diff
@@ -0,0 +1,32 @@
...
+    "api" => "Available endpoints",
+    "fields" => "Field Reference",
+    "matching-content-to-subscriber-lists" => "Matching content to subscriber lists",
+    "tasks" => "Tasks",
+    "transition-from-govdelivery-to-notify" => "Transition from Govdelivery to Notify",
+    "troubleshooting" => "Troubleshooting",
+  }.freeze
+
+  def self.pages
+    PAGES.map { |k, v| Page.new(k, v) }
```
AlanGabbianelli, line 13:
```suggestion
    PAGES.map { |filename, title| Page.new(filename, title) }
```
Not a blocker but it makes this clearer and reflects the `Page` attributes.

PR #2529, opened 2020-05-08
Review whitelist-site-in-transition.html.md

AlanGabbianelli:
We are slowly transitioning from the use of "whitelist" and "blacklist"
to more inclusive terms like "allowlist" and "blocklist". These changes
reflect this transition.


config/tech-docs.yml
```diff
@@ -69,6 +69,7 @@ redirects:
   /manual/use-terraboard-to-monitor-terraform-state.html: /manual.html
   /manual/user-management-in-aws.html: /manual/get-started.html
   /manual/virus-scanning.html: /manual/alerts/virus-scanning.html
+  /manual/whitelist-site-in-transition.html: /manual/allowlist-site-in-transition.html
```
brucebolt, line 72: 👍

PR #2533, opened 2020-05-11
GitHub actions conventions

kevindew:
I thought since we've now got [RFC 123](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-123-github-actions-ci.md) approved 🎉  and a variety of implementations made that it'd be good to try get some documentation in that defines some conventions. This is to try avoid every implementation being done based on different preferences.

I've opened this as a draft as I've pushed in suggestions that aren't agreed and I'm hoping we can work towards agreement. Hopefully there is broad agreement that we want conventions and consistency, I'm quite happy for there to be disagreements with some of the proposed suggestions.

Key thing to read: 👉 https://github.com/alphagov/govuk-developer-docs/blob/github-actions-conventions/source/manual/test-and-build-a-project-with-github-actions.html.md

Highlighting areas that might be points of contention which may want discussion:

- suggesting that we name our workflow files ci.yml, this seems to be a [common](https://medium.com/@michaelekpang/creating-a-ci-cd-pipeline-using-github-actions-b65bb248edfe) [convention](https://github.com/rust-lang/rust/blob/master/.github/workflows/ci.yml) and to me feels a bit more specific than the names we've used already such as [default.yml](https://github.com/alphagov/rubocop-govuk/blob/b8411dfbcc401298e341178460d0b761d72086e7/.github/workflows/default.yml) or [tests.yml](https://github.com/alphagov/govuk-coronavirus-content/blob/866437e706dcf2dfad1e57044246637e17f076f3/.github/workflows/tests.yml). I'm not set on ci.yml though and happy for suggestions
- suggest we give the workflow a name of "Continuous Integration", it's slightly nicer to give a workflow a name than tell people not to set a name - I don't really mind what our preference is here
- using a cache for Gems. We already seem to have [some projects with this](https://github.com/alphagov/govuk-developer-docs/blob/3baf8267dd2a9fa4220800e4405bcd2e38379652/.github/workflows/tests.yml#L20-L25) and [some without](https://github.com/alphagov/govuk-ask-export/blob/5333650927c2f146a5a9332305fc50a534f63a1c/.github/workflows/default.yml), and then [discussions about the value](https://github.com/alphagov/govuk-coronavirus-content/pull/205#discussion_r420345837). Life is probably simpler if all projects just have it and then it's not a decision point.
- usage of vendor for third party code. This seemed a relatively uncontroversial place to put it since it's a pretty standard directory name and already conventionally used by Bundler, this saves having to choose where to put this for any other repos.
- including a Content Publisher configuration that isn't used. I wanted to include a full Rails app to illustrate a conventions for a variety of techniques needed for a conventional, modern GOV.UK application, however these aren't used in anger anywhere so may mean they age badly.

I did consider putting in a small section on security and trust of actions, where we suggest only using actions from trusted providers. However, I felt the advice for this applied the same for Dockerfiles, Ruby gems and all sorts of other usages of open source code so it felt a bit superfluous to point it out.

When reviewing try not to worry too much about phrasing at this point and more on the conventions and what content is required. We can save the phrasing improvements once any opinionated aspects are resolved.


source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -0,0 +1,266 @@
...
+  example, Node.js versions past 8)
+- As GitHub Actions are not executed within GDS internal network they are a safer
+  choice for testing contributions from external partners (such as pull
+  requests from members of the public)
+
+## GOV.UK Conventions for GitHub Actions
+
+### Name of CI workflow file
+
+You should name the file used for configuring the CI workflow `ci.yml` and name
```
issyl0, line 46: Having a consistent name across all of the CI builds Actions is a good thing. It'll make [looking for Actions in repos to enable branch protection for](https://github.com/alphagov/govuk-saas-config/issues/42) easier.

source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -0,0 +1,266 @@
...
+- Since the same actions are used in different contexts the `name` option has
+  been used in steps to distinguish the different steps.
+- The `bundle exec rails assets:precompile` task is used to build a warm cache
+  of assets before starting tests to reduce risk of timeouts, this is due to
+  an [underlying performance issue][govuk-frontend-issue] with GOV.UK asset
+  compilation.
+- In the underlying project (Content Publisher) `bundle exec rake` performs:
+  Ruby & JS unit testing; Ruby, SCSS, JavaScript and FactoryBot linting; and
+  brakeman security audit.
+- This example does not include tagging the [master branch with a particular
```
issyl0, line 247: Should it? Or does this have bearing on our "no secrets" decision? Though it should use the `GITHUB_TOKEN` from the Action run, not one of our own...

source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -0,0 +1,266 @@
...
+
+- As Jenkins is hosted within GDS' network it can be used to trigger other
+  internal systems to perform jobs during or following a build (for example,
+  triggering [publishing end-to-end tests][e2e-tests] or triggering a
+  deployment)
+- Jenkins can be used for any services that require authentication to external
+  systems (for example, [releasing a gem
+  automatically][gem-release] during a master build), since secrets are not
+  permitted within GitHub Actions
+- As GitHub Actions are not constrained by the [GOV.UK Puppet
```
benthorner, line 35: Do we need to say this? I've assumed that, since we have an accepted RFC, we only need to cover reasons to postpone using GitHub Actions?

source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Base your workflow on one of our documented examples
+
+GitHub Actions currently [lack templating functionality][action-templates]
+to enable re-use of workflow configuration across projects. Therefore, in order
+to maintain consistency across GOV.UK projects, it is beneficial that we
+document and re-use common practices.
+
+This documentation contains examples of a number of common build needs for
+GOV.UK projects. You are encouraged to base your projects configuration on
```
benthorner, line 87:
```suggestion
GOV.UK projects. You are encouraged to base your project's configuration on
```

source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -0,0 +1,266 @@
...
+        ports:
+          - 6379:6379
+        options: --health-cmd "redis-cli ping" --health-interval 10s --health-timeout 5s --health-retries 5
+    env:
+      RAILS_ENV: test
+      REDIS_URL: redis://localhost:6379/0
+      TEST_DATABASE_URL: postgresql://postgres@localhost/content-publisher
+      GOVUK_CONTENT_SCHEMAS_PATH: vendor/govuk-content-schemas
+    steps:
+      - name: Clone project
```
benthorner, line 199:
```suggestion

```

This would be more consistent with the other examples. Perhaps we could just have names for the 'odd' action in a pair?

source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -0,0 +1,266 @@
...
+          restore-keys: yarn
+      - run: yarn install --frozen-lockfile
+      - run: bundle exec rails db:setup
+      - run: bundle exec rails assets:precompile
+      - run: bundle exec rake
+```
+
+Notes:
+
+- Similar configuration to PostgreSQL and Redis can be applied to other software
```
benthorner, line 232: Minor: how would this help someone reading it?

source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -0,0 +1,266 @@
...
+this repository to propose a change here. This is preferential to taking a
+project in a different direction and losing consistency.
+
+## GitHub Action Examples
+
+### Simple Ruby Application
+
+```yml
+# .github/workflows/ci.yml
+name: Continuous Integration
```
benthorner, line 99: Could you help me understand the advantage here? I suspect this will override the current GitHub UI, which prints the path to the file, which is useful if you're new.

issyl0: It will indeed. I was wondering about this too.

source/manual/test-and-build-a-project-with-github-actions.html.md
kevindew: Yeah the GITHUB_TOKEN will have access. It'd probably be useful to have an example but that might not be applicable until we have one to use in anger.

I'm not too sure at this point whether we'd continue with incrementing integer or use an approach based on sha hash?

source/manual/test-and-build-a-project-with-github-actions.html.md
kevindew: Yeah, could probably turn this into a list of restrictions rather than a more comparison style of list. 

My concern though is that might open a different can of worms which leads to a doc on choosing your CI platform.

source/manual/test-and-build-a-project-with-github-actions.html.md
kevindew: I felt that things got a bit more complex if you only name some actions with the same name. In this example we've got the schemas as the odd example but then you could have a second or third. I don't mind too much but just would want to go with the route that involves the least decisions for an implementor.

source/manual/test-and-build-a-project-with-github-actions.html.md
kevindew: Mostly this is to discourage adding more examples for every variation of a db system to this doc. Ideally it'd be good to have links to db configurations that could be used but it wasn't as easy to find them as I thought.

source/manual/test-and-build-a-project-with-github-actions.html.md
kevindew: I put some text about this in the PR description. Myself, I don't mind if we give workflows a name or not but would prefer it to be consistent - given we have some workflows with names and some without.

The UI does show the name, the path is the fallback, though clicking on any action shows you the workflow file so I don't think it obscures much.

My logic was that it was a more positive approach to suggest a consistent name than to say "do not set a name" and that it meant that any potential future actions also didn't need to be no name for consistency too. 

Where do you two stand on it? What would be your preference for guidance?

source/manual/test-and-build-a-project-with-github-actions.html.md
benthorner: So my understanding is that: all CI should be GitHub actions, and we're working on a couple of issues with the approach, and any affected repos should be excluded _for now_. 

Hopefully that avoids any implication of choice. Does that help?

source/manual/test-and-build-a-project-with-github-actions.html.md
benthorner: I suppose you end up making a decision either way, as you need to be aware that you're writing a flow with repeated actions. Since you can always expand each action to see what it's doing (e.g. which repo it's fetching), do the names really add much? I appreciate it's surprising to see something twice; my hunch is that in practice, people will just forget to add the names, since it requires more effort than not.

source/manual/test-and-build-a-project-with-github-actions.html.md
benthorner: Ah, I see 👍.

source/manual/test-and-build-a-project-with-github-actions.html.md
benthorner: For me, the less text we have, the more consistent we're likely to be (plus less typing). In this case, I don't find the name particularly helpful. As long as we don't need to set a name on any future actions, then it seems OK to not have one ever - this might encourage people to use a descriptive name for the file, and avoids it getting out-of-sync with any name.

issyl0: Thinking more about it, like Ben I'd be in favour of `.github/workflows/ci.yml` being the filename and each action having no defined `name`.

kevindew: Sure - that's cool. We can go with no name.

source/manual/test-and-build-a-project-with-github-actions.html.md
kevindew: Sure, I guess we'll see what people actually do in practice. I think it's stuff worth suggesting through conventions that identical steps should be named.

One thing to ponder is whether we should actually highlight a convention for naming steps. Thinking about it we might want to specify an overall convention for names since we have them at workflow level (we have a preference for no-name), job level (preference for no name as job key should be sufficient) and step (preference for no-name unless necessary to resolve ambiguity), which I think wrapped together sounds like everything has a preference for no-name unless resolving some ambiguity.

benthorner: An overall note on convention sounds good to me 👍. Still not convinced resolving ambiguities meets a need: if a step fails, then the name is pretty irrelevant, since the person debugging it needs to look into the detailed log anyway; if a step passes, then I wouldn't expect anyone to pay much attention to its name.

However, I'm not too fussed one way or the other.

kevindew:
Wow so many references 💯thanks for the review Issy. I'll try spend my last 30 mins today cleaning this up so it can be (hopefully) ready for merging

kevindew:
Ok this is out of draft now, I think I've resolved all the comments thus far.

PR #2543, opened 2020-05-12
Clarify installation instructions for Mac users

ChrisBAshton:


ChrisBAshton:
I've heard from @beccapearce that that has the same security issues. Probably the same remedy though!

ChrisBAshton:
Agreed it feels gross, but I guess if the Security aspect of it involves some GUI manipulation, doesn't hurt to get straight into the GUI headspace with the install steps 😅

PR #2544, opened 2020-05-12
Point to the new rake task for viewing subscriptions

ChrisBAshton:


source/manual/subject-access-requests.html.md
```diff
@@ -24,10 +24,11 @@ So far we've received requests that have asked for information on:
 ### Email subscriptions
 
 We've had requests to get information on what is held about an email address
-in terms of subscriptions. We have [some documented queries that can help get
-this][email-alert-api-analytics].
+in terms of subscriptions. [There is a rake task for that][rake-task].
```
thomasleese, line 27: It might be nice to use the `RunRakeTasks` helper here to automatically get some links to Jenkins: https://github.com/alphagov/govuk-developer-docs/blob/e051e0bf72dfb862bc25e5ce95561f2a6c85cbc2/source/manual/alerts/whitehall-scheduled-publishing.html.md.erb#L25

ChrisBAshton: Ooh, didn't know about that helper! I see that being useful in other docs but not particularly this one (we always want to run against Prod for this kind of FOI request, not much value in rendering a list. Plus it would break up the flow of the document)

thomasleese: I agree, it's only really useful in production, but I do think the ability to click on a link and be taken straight to Jenkins is useful in these kinds of docs. I don't mind too much though, happy to leave it as it is.

PR #2557, opened 2020-05-18
Speed up search and start indexing Schemas

ChrisBAshton:
- Speeds up the search by utilising the ['content' override in middleman-search-gds](https://github.com/alphagov/middleman-search/pull/5)
- Adds Schemas to the search (these were not previously indexed)

## Before
5.6MB search.json

Screenshot 2020-05-18 at 10 22 45

## After
4.8MB search.json, and schema now shows up in search results.

Screenshot 2020-05-18 at 11 05 33

This is being tackled as part of https://trello.com/c/lVTGSjCA/142-external-docs-arent-indexed-in-search.


app/proxy_pages.rb
```diff
@@ -98,6 +98,7 @@ def self.document_types
         template: "templates/document_type_template.html",
         frontmatter: {
           title: "Document type: #{document_type.name}",
+          content: "",
```
benthorner, line 102: I don't know much about the way this repo does indexing. My assumption from reading this is that `content: ""` stops the indexing. However, it's also present in the next commit, where we're wanting to _start_ indexing. 

Can you help me understand what the change is doing?

ChrisBAshton: Ah, apologies. A lot of that knowledge is carried over from https://github.com/alphagov/govuk-developer-docs/pull/2509.

Proxied pages aren't indexed by default, unless we pass a `title` parameter to middleman (see the ProxyPages snippet in the [Usage section of middleman-search](https://github.com/alphagov/middleman-search#usage)). So prior to #2509, these weren't being indexed at all. In #2509, I added `title` for DocumentType and Supertype, but not for Schema, as the schema pages are so large and complicated that they broke the search functionality. DocumentType and Supertype slowed down the search a bit, but within an acceptable (subjective) limit.

This PR overrides the `content` property to stop the indexing of the page's contents, meaning the only thing that is indexed is the `title`, speeding up the search again, and allowing us to index Schemas for the first time. I hope that makes sense!

benthorner: Aha, I understand now - I appreciate the commit title better now! Thanks for explaining ⭐.

PR #2558, opened 2020-05-18
Rewrite errors manual to give the big picture

benthorner:
https://trello.com/c/Uzpw1bDA/243-document-how-error-pages-work-on-govuk

Previously I found this manual hard to read, because the policy at
the top only applied to two of the subsequent sections: most errors
are *not* reported to Sentry. Most of the linked URLs also did not
work for me, and seem largely unnecessary/verbose.

This rewrites the manual to surface the two approaches we have for
errors more concisely, while maintaining the original policy of the
associated RFC. Shortening this general manual about 'errors' will
also allow for a planned addition that describes what happens when
an application errors on GOV.UK (this will be in a future PR).

### DIFF
--- a/source/manual/errors.html.md
+++ b/source/manual/errors.html.md
@@ -1,109 +1,50 @@
 ---
 owner_slack: "#govuk-developers"
-title: Triage and handle errors
+title: How we handle errors
 section: Monitoring
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2019-11-11
+last_reviewed_on: 2020-05-18
 review_in: 6 months
 ---
 
+## How we categorise errors
+
 Sometimes applications will encounter exceptions. This policy describes what we should do for different types of errors. It was first proposed in [RFC 87](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-087-dealing-with-errors.md).
 
-## Principles
+### High priority errors
 
-### 1. When something goes wrong, we should be notified
+1. **When something goes wrong, we should be notified**. Applications should report exceptions to Sentry. Applications must not swallow errors.
 
-Applications should report exceptions to Sentry. Applications must not swallow errors.
+1. **Notifications should be actionable**. Sentry notifications should be something that requires a developer of the app to do something about it. Not just a piece of information.
 
-### 2. Notifications should be actionable
+1. **Applications should not continue to have these errors**. The goal of GOV.UK is that applications should not error. When something goes wrong it should be fixed.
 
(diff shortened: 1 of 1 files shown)


source/manual/errors.html.md
```diff
@@ -1,109 +1,50 @@
...
-
-Desired behaviour: our environment is set up such that these errors do not occur.
-
-### Bad request errors
-
-User makes a request the application can't handle ([example][bad-request]).
+Sometimes applications will encounter exceptions. This policy describes what we should do for different types of errors. It was first proposed in [RFC 87](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-087-dealing-with-errors.md).
 
-Often happens in [security checks](https://sentry.io/govuk/app-frontend/issues/400074979).
+### High priority errors
```
vanitabarrett, line 15: How have we categorised errors into high and low here?

source/manual/errors.html.md
```diff
@@ -1,109 +1,50 @@
...
+1. **When something goes wrong, we should be notified**. Applications should report exceptions to Sentry. Applications must not swallow errors.
 
-Desired behaviour: user gets feedback, error is not reported to Sentry
+1. **Notifications should be actionable**. Sentry notifications should be something that requires a developer of the app to do something about it. Not just a piece of information.
 
-[bad-request]: https://sentry.io/govuk/app-service-manual-frontend/issues/400074003
+1. **Applications should not continue to have these errors**. The goal of GOV.UK is that applications should not error. When something goes wrong it should be fixed.
 
-### Incorrect bubbling up of errors
+Errors that should be reported:
```
vanitabarrett, line 23: This suggests to me that there are high-priority errors that shouldn't be reported? Which goes a bit against "Applications should report exceptions to Sentry"? 

Is this just a list of what we consider to be a high priority error?

benthorner: Oops, you're right, it's a strange implication left over from a previous draft. Updated 👍.

source/manual/errors.html.md
benthorner: The two kinds of error we've got here is based on the previous version of this manual. I've introduced the terminology 'high' and 'low' as a way to refer to them.

- If we're just logging the errors and not actively doing anything about them, it seems fair to say these are 'low priority'.

- If we alert on the errors and are expected to act on that alerts, it seems fair to say these are 'high priority'.

In reality, I don't think people pay as much attention to Sentry as this manual suggests they should, but that's a different problem.

Does that help? I'm open to other terminology, if you have any suggestions.

PR #2564, opened 2020-05-20
Update removing content from search text

edwardkerry:
Clarify that not only unpublished content may be removed from search-admin

### DIFF
--- a/source/manual/incorrect-content-in-search-or-navigation.html.md
+++ b/source/manual/incorrect-content-in-search-or-navigation.html.md
@@ -63,6 +63,7 @@ reindexing via a rake task in the publishing app.
 
 Try republishing the content.
 
-#### Unpublished content is still showing up in search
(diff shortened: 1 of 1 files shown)


source/manual/incorrect-content-in-search-or-navigation.html.md
```diff
@@ -63,6 +63,7 @@ reindexing via a rake task in the publishing app.
 
 Try republishing the content.
 
-#### Unpublished content is still showing up in search
+#### Unpublished or outdated content is still showing up in search
```
barrucadu, line 66: I'm not sure about "outdated", as content can be outdated but still supposed to be in search (it should just be up-to-date).

Maybe "Unpublished, deleted, or redirected content is still showing up in search"

edwardkerry: very good point, thank you

PR #2565, opened 2020-05-20
Document how we serve error pages on GOV.UK

benthorner:
https://trello.com/c/Uzpw1bDA/243-document-how-error-pages-work-on-govuk

### DIFF
--- a/source/manual/errors.html.md
+++ b/source/manual/errors.html.md
@@ -8,9 +8,48 @@ last_reviewed_on: 2020-05-18
 review_in: 6 months
 ---
 
-## How we categorise errors
+## How we serve errors on GOV.UK
 
-Sometimes applications will encounter exceptions. This policy describes what we should do for different types of errors. It was first proposed in [RFC 87](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-087-dealing-with-errors.md).
+When a request to GOV.UK fails, we need to handle the error in some way, so that GOV.UK does not look completely broken. Rather than embedding this into every app, we have multiple layers of error handling.
+
+> Note that publishing apps do not have these same layers of error handling; they are not behind a CDN. A publishing app is expected to handle all errors itself, according to the policy in the next section.
+
+1. **CDN**. There are a few scenarios here:
+
+  - When the origin servers are offline, we:
+
+      - [Skip normal request handling and begin error handling](https://varnish-cache.org/docs/trunk/users-guide/vcl-built-in-subs.html#vcl-backend-error)
+      - Try to [serve the page from a sequence of static mirrors](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L274)
+      - Serve [a basic inline error page](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L531) if [all the mirrors fail](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L363) (uh oh!)
+
+  - When the origin servers are online but return a 5xx response:
+
+    - For [GET or HEAD requests](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L339), we try the mirrors, or else serve an error (as above)
+
+    - For [other requests](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L367), we [serve the response as-is from the origin](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L374) (see below)
+
+  - When the origin servers return a 4xx response, we:
+      - Serve the response as-is from the origin (no special handling)
+
+2. **Origin**. When a request fails in an upstream app, we:
+
+  - [Intercept the response in Nginx](https://github.com/alphagov/govuk-puppet/blob/7dafec7cccd8308ec90c28835de70243d79b323b/modules/router/templates/router_include.conf.erb#L81)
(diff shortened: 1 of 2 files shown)


source/manual/errors.html.md
```diff
@@ -8,9 +8,33 @@ last_reviewed_on: 2020-05-18
...
+
+1. **Origin**. When a request fails in an upstream app, we:
+
+  - [Intercept the response in Nginx](https://github.com/alphagov/govuk-puppet/blob/7dafec7cccd8308ec90c28835de70243d79b323b/modules/router/templates/router_include.conf.erb#L81)
+  - Replace the response with [a pre-rendered one from Static](https://github.com/alphagov/govuk-puppet/blob/7dafec7cccd8308ec90c28835de70243d79b323b/modules/router/manifests/errorpage.pp#L14)
+
+1. **App**. When an app raises an exception, we:
+
+  - Should [catch the error and return an appropriate error code](https://github.com/alphagov/email-alert-frontend/blob/a2bd35b5b17b7da40cd43df9c2756b564597b66e/app/controllers/application_controller.rb#L10)
+  - Return a "500 Internal Server Error" response (by default)
```
ChrisBAshton, line 29: This reads as contradictory to me. E.g. if an app can't find the resource a user has requested, should it return a 404 or a 500?
(I think line 29 can be omitted, unless I'm missing a nuance).

source/manual/errors.html.md
```diff
@@ -8,9 +8,33 @@ last_reviewed_on: 2020-05-18
...
+
```
ChrisBAshton, line 30: I would also want to signpost to the "How we handle errors in our apps" section here, as otherwise it reads as though our work is done already.

We could probably rewrite the above couple of points as:

```md
1. **App**. When an app raises an exception, we:

  - Should [catch the error and return an appropriate error code](https://github.com/alphagov/email-alert-frontend/blob/a2bd35b5b17b7da40cd43df9c2756b564597b66e/app/controllers/application_controller.rb#L10)
  - We should also log the error so that we can debug it later. See ["How we handle errors in our apps"](#how-we-handle-errors-in-our-apps) below.
```

source/manual/errors.html.md
benthorner: I see what you mean. I've updated this to imply that the behaviour is inconsistent across our apps.

source/manual/errors.html.md
benthorner: Hopefully the changes in https://github.com/alphagov/govuk-developer-docs/pull/2565#discussion_r428599344 will clarify this better. I'd like to separate the mechanism from the policy; whether we log or not doesn't affect what the user sees.

Does that help?

benthorner:
Thanks for taking a look @ChrisBAshton. I've added a link to this in the deep dive manual, as well as making a few tweaks to the CDN section, which was slightly wrong.

Could you take another look?

benthorner:
@ChrisBAshton sorry about this: I've been doing some more investigation into our CDN config, which is more complicated than I originally realised. While I'm still trying to resolve one question, I've rewritten that first point after talking with @kevindew on Slack.

It's worth another look.

source/manual/errors.html.md
```diff
@@ -8,9 +8,50 @@ last_reviewed_on: 2020-05-18
...
+    - For [other requests](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L367), we:
+      - [Serve the response as-is from the origin](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L374) (see below)
+
+  - When the origin servers are offline, we:
+      - [Skip normal request handling and begin error handling](https://varnish-cache.org/docs/trunk/users-guide/vcl-built-in-subs.html#vcl-backend-error)
+
+      - Try to [serve the page from a sequence of static mirrors](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L274)
+      - Serve [a basic inline error page](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L531) if [all the mirrors fail](https://github.com/alphagov/govuk-cdn-config/blob/d77313abdb5098e2b350de7a0992375e50ff03a3/vcl_templates/www.vcl.erb#L363) (uh oh!)
```
ChrisBAshton, line 30: The two points above link to the same code as the identical points in the section above it. I think it could be simplified / easier to parse:

|Before | After |
|-------|------|
|Screenshot 2020-05-28 at 14 17 48|Screenshot 2020-05-28 at 14 17 04|

source/manual/errors.html.md
```diff
@@ -8,9 +8,50 @@ last_reviewed_on: 2020-05-18
...
+
+3. **App**. When an app raises an exception:
+
+  - If we have chosen to handle it, we:
+    - [Catch it and return a more helpful error response page](https://github.com/alphagov/email-alert-frontend/blob/a2bd35b5b17b7da40cd43df9c2756b564597b66e/app/controllers/application_controller.rb#L10)
+
+  - Otherwise, we:
+    - Let Rails return a "500 Internal Server Error" (by default)
+
+> Due to the way errors are handled by the origin servers, it's not currently possible for a frontend app to render a contextual error page for an error response. One option to get around this is to deviate from normal web semantics, and [return a 200 response](https://github.com/alphagov/email-alert-frontend/blob/a2bd35b5b17b7da40cd43df9c2756b564597b66e/app/controllers/email_alert_signups_controller.rb#L10). This should be avoided because such responses are cacheable, indexable (by search engines), and surprising for anyone trying to look for errors in logs.
```
ChrisBAshton, line 48: Is there a reason why these two paragraphs are indented with `>`? I understand it causes them to 'pop out' and draw attention to them - in which case, perhaps this should only apply to the second paragraph?

source/manual/errors.html.md
benthorner: Sure, that makes sense, updated. My original thinking was that they could change independently.

source/manual/errors.html.md
benthorner: I've made a few layout tweaks to clarify. They are both notes:

- The first is a note about app handling (last bullet point). I've replaced this with a possibly more convention "*" caveat.

- The second is a note about the whole section. I've moved this to the top, as it seems a bit pointless to find out after reading the section.

PR #2567, opened 2020-05-21
Remove embedded dashboard

nicholsj:
The dashboard is only available to VPN users - and doesn't add value by being embedded here. So let's remove it. 

[Trello](https://trello.com/c/O7Ktz17S/1998-remove-embedded-dashboard-from-developer-docs)


thomasleese:
> It only works reliably in Carrenza anyway.

It should be work in AWS as well since https://github.com/alphagov/govuk-puppet/pull/10117.

PR #2568, opened 2020-05-22
Fix build issues introduced in README-ingestion

ChrisBAshton:
https://github.com/alphagov/govuk-developer-docs/pull/2519 tests passed but the subsequent [building of the docs in Jenkins](https://deploy.publishing.service.gov.uk/job/deploy-developer-docs) broke.

I've fixed each issue one-by-one, and then added a build instruction to the GitHub Action config to prevent this sort of thing from happening again. This will slow down future PRs slightly, as now we're testing _and_ building, but we can be confident we haven't broken anything (and this also paves the way for post-processing, e.g. checking that all markdown has successfully been parsed into HTML).

I was a little worried we might hit our GitHub API limit with this but [GitHub Actions create a new token on every run](https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token) (at least that's how I understand it), so we're good to go.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -9,8 +9,6 @@ task :lint, :environment do
   sh "bundle exec rubocop --format clang"
 end
 
-task default: %i[lint spec]
-
 namespace :cache do
   desc "Clear the cache of external content"
   task :clear do
@@ -63,3 +61,5 @@ task :check_puppet_names do
     HTTP.get(app.puppet_url)
   end
 end
+
+task default: %i[lint spec cache:clear assets:precompile]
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -20,3 +20,6 @@ jobs:
       - run: bundle exec rake
         env:
           GOVUK_CONTENT_SCHEMAS_PATH: tmp/govuk-content-schemas
+      - run: bundle exec middleman build --verbose
```
benthorner: Instead of doing this, can we change the `default` rake task to also do `cache:clear` and `assets:precompile`? This avoids duplicating the knowledge.

ChrisBAshton: Good idea!

PR #2575, opened 2020-05-28
Improve ProxyPages tests and make robust to non-UTF-8 content

ChrisBAshton:
This PR was originally to do with fixing search indexing of README content (hence the branch name), but on further investigation the search indexing was already working.

However, by that point I'd already made some good improvements to the code, so have now refactored this PR to contain only the improvements:

1) Improved ProxyPages tests (helps document the 'why' behind the `content: ""` decision)
2) Force convert README content to UTF-8

Trello: https://trello.com/c/rzunnimb/143-automatically-consume-all-govuk-repo-readmes

### DIFF
--- a/spec/app/github_repo_fetcher_spec.rb
+++ b/spec/app/github_repo_fetcher_spec.rb
@@ -24,7 +24,7 @@ RSpec.describe GitHubRepoFetcher do
       repo_name = SecureRandom.uuid
       api_endpoint = "https://api.github.com/repos/alphagov/#{repo_name}/readme"
       stubbed_request = stub_request(:get, api_endpoint)
-        .to_return(status: 200, body: '{ "content": {} }', headers: { content_type: "application/json" })
+        .to_return(status: 200, body: '{ "content": "" }', headers: { content_type: "application/json" })
 
       GitHubRepoFetcher.new.readme(repo_name)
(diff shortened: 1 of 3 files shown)


spec/app/github_repo_fetcher_spec.rb
```diff
@@ -42,6 +42,18 @@
       expect(GitHubRepoFetcher.new.readme(repo_name)).to eq(readme_contents)
     end
 
+    it "forces encoding to UTF-8" do
+      repo_name = SecureRandom.uuid
+      stub_request(:get, "https://api.github.com/repos/alphagov/#{repo_name}/readme")
+        .to_return(status: 200, body: '{ "content": "" }', headers: { content_type: "application/json" })
+      encoded_input = "abc½½½".force_encoding("iso-8859-1")
+      allow(Base64).to receive(:decode64).and_return(encoded_input.dup)
```
benthorner, line 50: Can you help me understand why we need this stub?

ChrisBAshton:
Thanks for the review @benthorner . I had a think about your comment and then tried a fresh build of the docs to see if I could dig into the error message again... but then the error didn't appear!

It's not clear this was causing any issues other than some logs on my machine, so I'm closing in favour of #2581 which just concentrates on the test improvement.

PR #2603, opened 2020-06-10
Consolidate and rewrite deployment manuals

benthorner:
https://trello.com/c/4YxwfpDs/163-consolidate-and-update-dev-docs-for-deployment-to-mention-cd

Previously we had several manuals about developing and deploying a
change to an app, with the 'development-pipeline' manual acting as
a kind of starting point. This consolidates all the manuals into a
single guide for how to raise and deploy a change, so that we have
less to review. Putting all the knowledge in one place will also
make it easier to add caveats for Continuous Deployment.

👉 [Rendered version](https://github.com/alphagov/govuk-developer-docs/blob/rewrite-deployment-manuals/source/manual/development-pipeline.html.md)


bevanloon:
Is it worth setting up a redirect for the files you are removing/renaming? 

I think there's also currently a redirect pointing to one of the pages you are removing:
e.g. https://github.com/alphagov/govuk-developer-docs/blob/master/config/tech-docs.yml#L62

benthorner:
@bevanloon I didn't know we could do that! I've replaced the faulty one, and added another for the larger "deploying" manual, which seems like something people could easily link to externally.

PR #2605, opened 2020-06-10
Clarify which team needed for prod jenkins

ChrisBAshton:


source/manual/access-jenkins.html.md
```diff
@@ -10,4 +10,7 @@ review_in: 12 months
 
 Our Jenkins installations ([deploy](https://deploy.integration.publishing.service.gov.uk/) and [CI](https://ci.integration.publishing.service.gov.uk/)) use GitHub for authentication and authorisation. You need to be added to the [GOV.UK team][govuk-team] by a GitHub administrator (usually a senior tech), otherwise you won't see any projects, or will see a message that you're unauthorised.
 
+For access to the staging/production Jenkins instances, you also need to be added to the [GOV.UK Production team][govuk-production].
```
cbaines, line 13: Doesn't this contradict the paragraph above: "You need to be added to the [GOV.UK team][govuk-team] by a GitHub administrator"

cbaines: They're possibly both right, maybe on Integration the gov-uk team is sufficient, and for Staging/Production you need to be in the gov-uk-production team.

ChrisBAshton: I don't think so - you need to be added by a GitHub admin in both cases.

ChrisBAshton: I've now rewritten it 👍 let me know what you think.

PR #2612, opened 2020-06-12
Add warning about Continuous Deployment

benthorner:
https://trello.com/c/4YxwfpDs/163-consolidate-and-update-dev-docs-for-deployment-to-mention-cd

I debated adding another warning inside the "Deployment" section,
but that would undermine the need to know about it *before* merging.

### DIFF
--- a/source/manual/development-pipeline.html.md
+++ b/source/manual/development-pipeline.html.md
@@ -16,15 +16,10 @@ Our development and deployment pipeline looks like this:
 1. Open a Pull Request (PR)
 1. [Wait for Continuous Integration to pass](#wait-for-continuous-integration-to-pass)
 1. [Get someone to review your Pull Request](#get-someone-to-review-your-pull-request)
+1. [Check if there is a deploy freeze](#check-if-there-is-a-deploy-freeze)
 1. [Merge your own Pull Request](#merge-your-own-pull-request)
 1. [Wait for the release to deploy to Integration](#wait-for-the-release-to-deploy-to-integration)
 1. [Manually deploy to Staging, then Production](#manually-deploy-to-staging-then-production)
-
-In exceptional circumstances, we may wish to block or _freeze_ deployments for a short period of time. In this case, add a note to the application in the [Release app][release]. This should explain:
-
-- Why the app is not deployable at the moment
-- Who to contact about deploying the app
-- When you expect the app to be deployable again
 
 ## Wait for Continuous Integration to pass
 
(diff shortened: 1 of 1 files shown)


benthorner:
Added a "DO NOT MERGE" instruction until we have a date for activating the CD pipeline.

source/manual/development-pipeline.html.md
```diff
@@ -55,6 +55,12 @@ We've got [guidelines on merging of Pull Requests](/manual/merge-pr.html).
 
 Code that is merged to `master` is tested again on CI. This is because the `master` branch may have changed since the tests last ran on the PR. If the tests on `master` pass, Jenkins pushes a `release_123` git tag to GitHub.
 
+> **WARNING**: some applications have Continuous Deployment enabled, which means the deployment process is fully automated. You should do any manual testing with [a temporary, branch deployment](#branch-deploy-review), before you merge.
+>
+> - Check the notes in the [Release app][release] to see if Continuous Deployment is enabled.
+> - You should still monitor the progress of your deployment in the [Release app][release].
```
theseanything, line 61: Just wondering why someone would need to monitor the progress of the deployment? Is it because this is still an "alpha" feature and it might break? What are people looking for?

benthorner: > "alpha"

"beta" - we skipped alpha 'cause we know we want it ;-)

Yeah, that's basically it. The pipeline also has no notification mechanism, so if some code does actually cause a critical issue, the pipeline will fail without anyone knowing, unless they actually check their release made it to production.

I've been thinking about ways to address this. I don't think notifications are a good idea, as that's just a load of noise whatever medium it's on. As we start to switch everything over, I think the deploy badger will fulfil the role of notifying us about undeployed PRs, and I also some people will naturally want to double-check anyway.

theseanything: Yup - might be good to add to that point in the docs, so people understand that they are looking for i.e. deployments actually succeed. But yeah +1 of deploy badger - make sense, "only notify us if something is wrong".

benthorner:
@theseanything I've also added a commit about freezing deployments to cover the work you're doing. Could you take another look?

source/manual/development-pipeline.html.md
```diff
@@ -49,12 +44,28 @@ Sometimes you may need to deploy your change in Integration in order to test it
...
 
+## Check if there is a deploy freeze
+
+In exceptional circumstances, we may wish to block or _freeze_ deployments for a short period of time. This should be done by checking "Freeze deployments?" and adding an explanatory note to the application in the [Release app][release].
+
+> Checking "Freeze deployments?" will turn off all automatic deployments for the application. You can still deploy urgent changes manually if necessary.
+
+When a deploy freeze is in effect, you should avoid merging any PRs. This is because your changes may block other, urgent changes related to the deploy freeze. Your changes will also remain undeployed for a long time.
+
+> People don't always check the Release app before merging their PRs. If you need to freeze deployments for an application, you should also make people aware using other channels.
```
theseanything, line 55: Yup - this all makes sense! 

Side thought and (probably out of scope) of this PR:
On line 55, the docs are telling people that they should check deploy freezes, then acknowledges that people don't always check. Which probably leaves the reader thinking - "do I actually need to do this?". I don't think the answer to this is with how the documentation is written, and probably highlights the ambiguity in the "deploy freeze" process.

theseanything:
> @theseanything I've also added a commit about freezing deployments to cover the work you're doing. Could you take another look?

Looks good!

source/manual/development-pipeline.html.md
benthorner: I know what you mean! We focus a lot on freezing the deployment, but not freezing merging; people do usually mention this in comms, though. This should a least be enough to encourage someone to send out comms as part of a freeze.

Issue #1821, opened 2019-06-24
Find & fix root cause of 'html-pipeline' issue


ChrisBAshton:
First spotted by Tim Blair in April (https://github.com/alphagov/govuk-developer-docs/commit/e9541eb4d01d1fb4b883041ca211613086618641), html-pipeline 2.11.0 causes issues with ASCII/UTF-8 processing:

```
Error processing resource for index: apis/search/faceted-search.html
"\xE2" from ASCII-8BIT to UTF-8
 govuk-developer-docs/.bundle/gems/commonmarker-0.19.0/lib/commonmarker.rb:26:in `encode'
```

We pinned html-pipeline to 2.5.0 in https://github.com/alphagov/govuk-developer-docs/pull/1818/commits/6f81229297ab2a3107b317313de8101c1898a7a7, but should investigate the root cause and fix it.

ChrisBAshton:
We saw a very similar error here: https://github.com/alphagov/govuk-developer-docs/pull/2575/commits/d95bcbd509b92b0da14af36d3bf49a962f3b5c10 - but with a different stack trace.

PR #2626, opened 2020-06-23
Upgrade dependencies, remove unused 'rack' dependency, force UTF-8

ChrisBAshton:
- Can't upgrade activesupport to 6+ as Middleman requires < 6. We've upgraded to 5.2.4.3, which is the [highest pre-v6 version of activesupport](https://rubygems.org/gems/activesupport/versions).
  - Can't upgrade Middleman beyond 4.3.7 as [that's the latest version](https://rubygems.org/gems/middleman/versions), although there's a 5.0.0.rc.1 from May 2019.

Trello: https://trello.com/c/Rk03inLB/169-upgrade-dependencies-for-govuk-developer-docs


Gemfile.lock
```diff
@@ -63,7 +63,7 @@ GEM
       faraday (~> 1.0)
     fast_blank (1.0.0)
     fastimage (2.1.7)
-    ffi (1.13.1)
+    ffi (1.12.2)
```
benthorner: Can we apply this lock in the Gemfile, so someone in the future can avoid it?

Gemfile
```diff
@@ -15,7 +15,7 @@ gem "middleman", "~> 4.3.7"
 gem "middleman-search_engine_sitemap", "~> 1.4"
 
 gem "github-markdown"
-gem "html-pipeline", "~> 2.5.0"
+gem "html-pipeline"
```
benthorner, line 18: Minor: could we link to the fix in the commit?

Gemfile.lock
ChrisBAshton: Good idea.

Gemfile
ChrisBAshton: I did try to track down the exact version/fix but couldn't find it.

benthorner: I might be missing something here: how do we know it's fixed?

ChrisBAshton: Ha, your comment made me go back and run the build - everything passes. So then I switched it back to 2.11.0 (the [buggy version referenced in this issue](https://github.com/alphagov/govuk-developer-docs/issues/1821)) to confirm it was broken in that version, and it was.

So then I proceeded to increment the version and run the build, to see which version fixed the issue.
I worked my way through all of the versions and found that _none_ of them work.
Then I realised the `Gemfile.lock` was still pointing to the pinned 2.5.0 version when the build last passed 🤦 

So it does appear to still be an issue. Tomorrow I'll have a go at delving in and fixing this specific issue.

benthorner: Ah, I feel validated for my awkward question 😸.

ChrisBAshton: I've now found the subdependency that was causing the error, and thought we might have to [expand the context options passed to HTML::Pipeline](https://github.com/jch/html-pipeline/pull/333), but on second thoughts it's easier and more consistent to just do the UTF-8 conversion here.

PR #2631, opened 2020-06-24
WIP: standardise pipeline

ChrisBAshton:
It makes no sense to have different pipelines depending on
whether code is fetched or passed directly as markdown.
Fixing this properly also means fixing links on content
passed to the 'parse' method.


ChrisBAshton:
Superseded by #2870.

PR #2634, opened 2020-06-25
Document steps you can take to monitor CD deployment

ChrisBAshton:
It's quite a black box experience otherwise! And devs shouldn't need to dig around for
these links / pipeline information.

### DIFF
--- a/source/manual/development-pipeline.html.md
+++ b/source/manual/development-pipeline.html.md
@@ -65,8 +65,12 @@ Code that is merged to `master` is tested again on CI. This is because the `mast
 > **WARNING**: some applications have Continuous Deployment enabled, which means the deployment process is fully automated. You should do any manual testing with [a temporary, branch deployment](#branch-deploy-review), before you merge.
 >
 > - Check the notes in the [Release app][release] to see if Continuous Deployment is enabled.
-> - You should still monitor the progress of your deployment in the [Release app][release] to check it succeeds.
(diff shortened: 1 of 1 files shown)


ChrisBAshton:
Agreed it should all be handled for you and that sitting and monitoring a release isn't ideal. It's a shame we don't have something like the GOV.UK PaaS UI, which shows you how the release is proceeding and where it fails.

It's quite unnerving to not see any change to the Release page (smart-answers build took about 5 minutes before the release tag was registered). And if it were to stall halfway through because of a Smokey error, we have no real visibility of that - I guess I'd just have to do some digging if I notice after 15 minutes or so that the release hasn't gone beyond Integration.

In that case it's useful to be able to point to the various bits of the pipeline to help guide the developer to where it went wrong. That slideshow looks very useful, but I do think that info should live in the dev docs, where it can be reviewed and kept up to date, and have its various subsections linked to.

Could I have a go at writing up that slideshow as a 'how to' doc?

benthorner:
I'd like to hold off on that for now. The "Learning GOV.UK" decks should be replayed at roughly the same frequency as we review the dev docs, so I regard them as some form of documentation. It would be a shame to repeat the information if we don't have to. After I present it, I'll populate the speaker notes to actually explain what's going on with links and stuff. Can we try linking to that (next week) as a lower effort/maintenance first pass?

ChrisBAshton:
Superseded by #2649.

PR #2637, opened 2020-06-30
Explain Grafana's data sources and usage tips

ChrisBAshton:
Thanks @bilbof for the presentation yesterday! ⭐ 
Learned a few things and thought, rather than take notes for myself, may as well share this info more widely.


source/manual/grafana.html.md
```diff
@@ -5,11 +5,20 @@ section: Monitoring
...
-last_reviewed_on: 2020-02-26
+last_reviewed_on: 2020-06-30
 review_in: 6 months
 ---
 
-Grafana lets us create nice dashboards using Graphite data.
+[Grafana](https://grafana.com/) is an open-source visualisation tool.
+It does not store data, but consumes data sources to create real-time
+graphs displayed on custom dashboards. Data sources include Prometheus,
+[Graphite], [Logit] and CloudWatch. The query language of the data
```
bilbof, line 15: I think we need extra square brackets here, such as [Graphite][].

source/manual/grafana.html.md
```diff
@@ -28,6 +37,30 @@ Useful Grafana dashboards:
...
+
+We often show multiple metrics on the same graph. The position of the
+key shows which Y-axis each metric corresponds to:
+
+![screenshot of multiple metrics selected](/manual/images/grafana-y-axis.png)
+
+You can click on a metric in a graph to show only that metric, or you
+can `CMD + click` to select multiple:
+
+![screenshot of multiple metrics selected](/manual/images/grafana-select-multiple.png)
```
bilbof, line 60: I wonder if these will render when deployed? They're not rendering for me, but that might just be Github! https://github.com/alphagov/govuk-developer-docs/blob/2966abfb749f525761647b188163a847e0327eb8/source/manual/grafana.html.md

source/manual/grafana.html.md
ChrisBAshton: [Shouldn't need to](https://github.github.com/gfm/#example-565) (we use this style in a few other places). (I only discovered that myself a couple of months ago - it's not intuitive at first!)

source/manual/grafana.html.md
ChrisBAshton: I think it's just GitHub - see [the source](https://github.com/alphagov/govuk-developer-docs/blame/c54ad9c3515c3bde25cea1dada2740b99b2ff9bc/source/manual/logging.html.md#L12) and [the output here](https://docs.publishing.service.gov.uk/manual/logging.html).

bilbof: Aha! Thanks :D

PR #2643, opened 2020-07-01
Remove install yarn from GitHub Action example

theseanything:
This removes the step to install yarn, as it appears that's already installed with `actions/setup-node@v1`. See the section about ]publishing with yarn in the README](https://github.com/actions/setup-node#usage).


kevindew:
Oh cool - I hadn't realised that. That's handy.

PR #2652, opened 2020-07-08
[DO NOT MERGE] Remove priorities from Smokey alert manual

benthorner:
Related to: https://github.com/alphagov/govuk-puppet/pull/10499

Previously we deprecated priorities in Smokey [1]. This manual was never
specific to high priority tests in the first place. I've also done a
general review of this manual, which was confusing because it conflated
alerts for the manual Smokey job, and the Smokey loop.

Please see the commits for more details.

[1]: https://github.com/alphagov/smokey/pull/693

### DIFF
--- a/source/manual/alerts/high-priority-tests.html.md
+++ b/source/manual/alerts/high-priority-tests.html.md
@@ -1,6 +1,6 @@
 ---
 owner_slack: "#govuk-2ndline"
-title: Run high priority tests
+title: Smokey loop tests
 parent: "/manual.html"
 layout: manual_layout
 section: Icinga alerts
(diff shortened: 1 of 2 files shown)


source/manual/alerts/high-priority-tests.html.md
```diff
@@ -1,15 +1,19 @@
...
 layout: manual_layout
 section: Icinga alerts
 last_reviewed_on: 2020-06-15
 review_in: 6 months
 ---
 
-The high priority tests [come from Smokey][smokey] and [the Icinga check is defined in Puppet][icinga].
-Smokey can fail with an error message "Run high priority tests", or simply with "Smokey failed".
+[Smokey][smokey] runs in a continuous loop in each environment.
+We have [Icinga checks][] for [most Smokey features][], so that
```
ChrisBAshton, line 13: We don't need the extra `[]` characters here - it defaults to using the link text as the link alias value.

benthorner: Ahhh, I was sticking to the known-to-work approach. Updated to drop them 👍

PR #2656, opened 2020-07-09
Create 'deploy' GitHub Action

ChrisBAshton:
Currently, a Jenkins job builds and deploys the developer docs to
an S3 bucket [using the 'deploy-to-s3' script][deploy].
It runs when a PR is merged, or between 9am and 7pm on weekdays
on a [timed cron job][job].

Historically we've used the S3 bucket in order to be proxied
through us for the TLS certificate, but we can now point the
CNAME to GitHub pages for hosting, which is one less moving part
to think about (two if you count both Jenkins _and_ S3), and
will cut some costs.

Step 1 is to get gh-pages building as regularly as the current
Jenkins builds. The next step will be to point
https://docs.publishing.service.gov.uk/ to GitHub Pages. The final
step will be to remove the legacy code and infrastructure.

Trello: https://trello.com/c/nEP3URQF/171-host-the-dev-docs-on-github-pages

[deploy]: https://github.com/alphagov/govuk-puppet/blob/7f8de3c9ea11697b054aa644d1dec8d8071bb8b1/modules/govuk_jenkins/templates/jobs/deploy_docs.yaml.erb#L29
[job]: https://github.com/alphagov/govuk-puppet/blob/7f8de3c9ea11697b054aa644d1dec8d8071bb8b1/modules/govuk_jenkins/templates/jobs/deploy_docs.yaml.erb#L22


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,50 @@
+name: "Deploy to gh-pages when PR is merged"
+on:
+  pull_request:
+    # Deploy after merging PRs to master
+    types: closed
+    branches:
+      - master
+  schedule:
+    # Deploy hourly between 9am and 7pm on weekdays
+    - cron: "0 9-19 * * 1-5"
```
benthorner, line 10: While the existing job may do this, I'm confused why a periodic deploy is necessary. Could you help me understand why we need it for GitHub Pages?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,50 @@
...
+        id: commit_message_writer
+        run: | # `github.event.number` will be blank if this is a cron
+          if [ "${{ github.event.number }}" == "" ]; then
+            echo "::set-output name=commit_message::Hourly scheduled redeploy"
+          else
+            echo "::set-output name=commit_message::Deploy via merge"
+          fi
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
```
benthorner, line 26: This repo doesn't have submodules, so I don't think we need this.

.github/workflows/deploy.yml
benthorner: Looks like removing the cron trigger would also simplify the steps vvv.

ChrisBAshton: Because the dev docs pull in external content (`docs/` folders for various repos, some hardcoded `ExternalDoc.fetch` references, govuk-content-schemas to name a few). Having a periodic rebuild ensures the docs are always no more than an hour out of date.

.github/workflows/deploy.yml
ChrisBAshton: Agreed. I'll do some experimenting to see if I can remove or change this.
(I think it was needed at one point because it was missing a local `gh-pages` branch)

ChrisBAshton: It wasn't needed 👍 now removed.

ChrisBAshton:
Thanks Ben - I like the idea of enforcing a passing test suite prior to deploying. I've added another commit to take the [same approach as rubocop-govuk](https://github.com/alphagov/rubocop-govuk/blob/master/.github/workflows/default.yml).

It's a little complicated by the `schedule` event condition, but I think that's worth keeping in.

PR #2662, opened 2020-07-16
Update clamav documentation

benjamineskola:
Removing the reference to fabric as it's far easier to do so with ssh

### DIFF
--- a/source/manual/alerts/clamav-definitions-out-of-date.html.md
+++ b/source/manual/alerts/clamav-definitions-out-of-date.html.md
@@ -4,15 +4,16 @@ title: ClamAV definitions out of date
 parent: "/manual.html"
 layout: manual_layout
 section: Icinga alerts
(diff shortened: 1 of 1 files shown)


source/manual/alerts/clamav-definitions-out-of-date.html.md
```diff
@@ -4,15 +4,16 @@ title: ClamAV definitions out of date
...
 layout: manual_layout
 section: Icinga alerts
-last_reviewed_on: 2019-10-01
+last_reviewed_on: 2020-07-16
 review_in: 6 months
 ---
 
 This could be because of a number of reasons. Check that the database is up to
-date by calling the freshclam command directly with verbose:
+date by [logging into a backend machine][] and calling the freshclam command directly with verbose:
```
ChrisBAshton, line 12: I think you could get away without linking this, as you've copied the relevant command into the bash snippet below.

PR #2663, opened 2020-07-17
Update and correct process file handle count docs

benjamineskola:
Still valid, but title/filename mispelled


benjamineskola:
> Good to fix spelling. But can you put [a redirect in](https://github.com/alphagov/govuk-developer-docs/blob/master/config/tech-docs.yml#L22) from old to new to catch any places we've linked to this that we've missed?

Ah, I wondered if that was possible! I've checked for links to it, but there could be others of course.

PR #2687, opened 2020-07-29
Move Google spreadsheet reference earlier in the documentation

ChrisBAshton:
We now have a number of keys documented in the spreadsheet, and the IAM
roles have been around long enough that we're starting to re-rotate some keys
we've already rotated in the past. It therefore makes sense to point the dev
to the spreadsheet in the first instance, to use it as a guide.

I've updated the last_reviewed_on, but have also tightened up the 'review_in' to
one month from now, as I think we should go into more detail describing where
to find and rotate keys in third-party services. There is a [Trello card][card] for
that, which I hope 2nd line get to in the next few weeks.

[card]: https://trello.com/c/q4gOMIP3/953-rotate-aws-logginglogit-read-iam-key

### DIFF
--- a/source/manual/aws-iam-key-rotation.html.md
+++ b/source/manual/aws-iam-key-rotation.html.md
@@ -5,7 +5,7 @@ parent: "/manual.html"
 layout: manual_layout
 type: learn
 section: AWS
-last_reviewed_on: 2020-01-20
+last_reviewed_on: 2020-07-29
 review_in: 6 months
 ---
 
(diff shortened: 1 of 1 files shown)


source/manual/aws-iam-key-rotation.html.md
```diff
@@ -5,8 +5,8 @@ parent: "/manual.html"
 layout: manual_layout
 type: learn
 section: AWS
-last_reviewed_on: 2020-01-20
-review_in: 6 months
+last_reviewed_on: 2020-07-29
+review_in: 1 month
```
kevindew, line 9: I don't think this is necessary as the information in here won't be invalid if this 1 month time period passes. As the Trello card is to add extra information rather than contradicting information it should just enhance the documentation, should the trello card not be done someone might be rather confused as to why this is on such a short review cycle.

PR #2697, opened 2020-07-31
[INCIDENT ACTION] Update CKAN support doc

1pretz1:
Updates the doc to include some support information which includes what
to do if you find a bug or security vulnerability. I haven't reviewed
the full doc as I don't believe I'm the best person to do this.

Trello:
https://trello.com/c/FABYywyd/433-update-2nd-line-documentation-supporting-ckan-to-include-correct-ckan-email-address-if-theres-a-security-incident

### DIFF
--- a/source/manual/data-gov-uk-supporting-ckan.html.md
+++ b/source/manual/data-gov-uk-supporting-ckan.html.md
@@ -12,8 +12,23 @@ review_in: 6 months
 [ckanext-datagovuk]: /apps/ckanext-datagovuk.html
 [ckandocs]: http://docs.ckan.org/en/ckan-2.7.4/maintaining/paster.html
 [ckan-api]: https://docs.ckan.org/en/2.8/api/
+[security_policy]: https://github.com/ckan/ckan/security/policy
+[issues]: https://github.com/ckan/ckan/issues
+[ckan_repo]: https://github.com/ckan/ckan
 
 [CKAN][dgu-ckan] is the publishing application for data.gov.uk.  [ckanext-datagovuk] is the primary [CKAN] extension for data.gov.uk.
+
+## Support
+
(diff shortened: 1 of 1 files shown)


source/manual/data-gov-uk-supporting-ckan.html.md
```diff
@@ -12,9 +12,18 @@ review_in: 6 months
...
+[ckan_repo]: https://github.com/ckan/ckan
 
 [CKAN][dgu-ckan] is the publishing application for data.gov.uk.  [ckanext-datagovuk] is the primary [CKAN] extension for data.gov.uk.
 
+## Support
+
+If you've found a bug in CKAN either try and fix it yourself and open a PR against their [main repo][ckan_repo] or open a new issue on
+CKAN's [GitHub Issues][issues] (try searching first to see if there's already an issue for your bug).
+
+If you find a potential security vulnerability please email security@ckan.org.
```
kevindew, line 25: I think you should change the ordering of this section around as it arguably ends with the most important information and should you find a security issue CKAN advise you not to create a [GitHub issue](https://github.com/ckan/ckan/security/policy)

I'd go with something like:

If you find an issue with CKAN that you believe may be a potential security vulnerability please contact security@ckan.org rather disclose the information publicly (as per their [security policy](https://github.com/ckan/ckan/security/policy)).

If you find an issue that does not appear to be a security vulnerability you should first search the [issues](https://github.com/ckan/ckan/issues) of the [upstream CKAN repository](https://github.com/ckan/ckan) to see if it has already been reported and if there is a workaround. If there is not an issue, considering opening a new one or - if you know how to fix it - open a PR to resolve the problem.

1pretz1: Good point on the ordering. Your explanation also reads better than mine so I'm stealing it! 😄

PR #2704, opened 2020-08-04
Remove Vale language linting

ChrisBAshton:
This commit removes Vale from govuk-developer-docs. Hopefully
we can reintroduce it one day in a more useful way.

Vale was added in November 2017 with little explanation, but with
good intentions:
https://github.com/alphagov/govuk-developer-docs/commit/b3dcb10783b238de3fbc0acc808da2647ee851f7

It was made part of the CI build in February 2018:
https://github.com/alphagov/govuk-developer-docs/commit/fd67db213786389798fc73d58f695f6c17f0ac42#diff-58231b16fdee45a03a4ee3cf94a9f2c3R42

But looking at the configs, it isn't possible to raise an error;
all this does is raise warnings, which no developer is likely to
see on CI.
For example, the 'Words' config has a level of 'error' (added in
the 2017 commit), which we've explicitly overridden in the .vale
file to only raise a warning.

It was subsequently removed in November 2019:
https://github.com/alphagov/govuk-developer-docs/commit/e39eab147116ef810c232d5fd274112261b46341#diff-58231b16fdee45a03a4ee3cf94a9f2c3
Note the commit message:

> The only thing that the default library Jenkinsfile doesn't do is
  `vale` linting. We're using a [really
  old](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/manifests/packages/vale.pp#L11)
  version of Vale anyway (the current is 1.7.1), so I'm not sure of its
  value any more. If we want it, we can migrate it to be run on
  Concourse, or make it a pre-commit hook, or somesuch.

The pre-commit linting feature is an opt-in-only feature, requiring
devs to read the README (!) and install some additional dependencies.
It's not clear how many devs do this, if any; moreover, it isn't
compatible with the common workflow of editing docs directly in
the GitHub UI. There's little to no value to keeping this as an
opt-in, local-only thing.

I experimented with what it might take to run Vale on CI:
https://github.com/alphagov/govuk-developer-docs/pull/2702

Unfortunately there's no Ruby gem for running Vale, so we couldn't
build it into the standard Rake default task. The best solution
I could see was to use the dedicated GitHub Action for it, which
worked after a bit of experimentation. But the same point around
it only giving warnings, not errors, limits its usefulness.
(Moreover, I could not see any warnings in the CI output, even
when I tried to deliberately trigger them. Only raising errors
appeared to do anything).

Tangentially related Trello card: https://trello.com/c/XSgGyudd/181-fix-url-markdown-parsing
Tangentially related PR (linting markdown): https://github.com/alphagov/govuk-developer-docs/pull/2703


issyl0:
In Homebrew we implemented it [like this](https://github.com/Homebrew/brew/pull/6826). But the difference there is that we do actually define a whole bunch of rules and have a prose style guide.

We don't here, and our rules here have languished somewhat. It was a good intention, but your conclusion that this can be removed is sensible, given it's currently providing no real value.

ChrisBAshton:
I've just discovered a [style guide in the dev docs](https://docs.publishing.service.gov.uk/manual/docs-style-guide.html, though it hasn't had a meaningful content change since October 2017, and I don't know how rigidly we want to stick to it. We could probably run Vale via a GitHub Action and have it tackle this much smaller subset of rules, which hopefully wouldn't involve too much doc fixing.

I might have a play with that today. 🤔

ChrisBAshton:
I didn't find time to get around to it...!

@issyl0 shall we approve this and look at adding Vale linting later?

issyl0 (approved):
We can bring this back later if we want to.

PR #2708, opened 2020-08-05
Expose GITHUB_TOKEN as ENV variable when deploying

ChrisBAshton:
We have some [build-time logic](https://github.com/alphagov/govuk-developer-docs/blob/130c6ac6f87f1fe5e6402a534a73ed0128b16a52/source/templates/application_template.html.md.erb#L168-L176) which checks for the presence of a GITHUB_TOKEN ENV variable
before attempting to call `application.readme`. This is so that
devs working on their local machine can compile the dev docs
without having to specify a token, and without hitting the GitHub
API limit.

This commit should fix issues like we're seeing on https://docs.publishing.service.gov.uk/apps/imminence.html:

> Screenshot showing error message 'Error: cannot fetch README without `GITHUB_TOKEN`.'

Running `GITHUB_TOKEN=$(cat ~/github_token.txt) ./startup.sh`
locally builds the pages as expected.

It also explicitly explains that private repos shouldn't have README content pulled in:

> Screenshot 2020-09-04 at 09 31 26

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,8 @@ jobs:
           GOVUK_CONTENT_SCHEMAS_PATH: tmp/govuk-content-schemas
       - name: Deploy
         uses: peaceiris/actions-gh-pages@v3
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./build
--- a/source/templates/application_template.html.md.erb
+++ b/source/templates/application_template.html.md.erb
@@ -156,15 +156,6 @@ gds govuk connect ssh -e production aws/<%= application.aws_puppet_class %>
 
 ## README
 
-<div class="govuk-warning-text">
-  <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
-  <strong class="govuk-warning-text__text">
-    <span class="govuk-warning-text__assistive">Warning</span>
-    The content below is pulled in directly from the repository.<br />
-    Links might not function properly.
-  </strong>
-</div>
-
 <% if application.private_repo? %>
   <span class="govuk-error-message">
     Cannot fetch README of private repo.
@@ -174,6 +165,15 @@ gds govuk connect ssh -e production aws/<%= application.aws_puppet_class %>
     Cannot fetch README without `GITHUB_TOKEN`.
   </span>
 <% else %>
+  <div class="govuk-warning-text">
+    <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+    <strong class="govuk-warning-text__text">
+      <span class="govuk-warning-text__assistive">Warning</span>
+      The content below is pulled in directly from the repository.<br />
+      Links might not function properly.
+    </strong>
+  </div>
+
   <div class="embedded-readme">
     <%= ExternalDoc.parse(application.readme) %>
   </div>
--- a/source/templates/application_template.html.md.erb
+++ b/source/templates/application_template.html.md.erb
@@ -165,14 +165,18 @@ gds govuk connect ssh -e production aws/<%= application.aws_puppet_class %>
   </strong>
 </div>
 
-<% if ENV["GITHUB_TOKEN"] %>
-<div class="embedded-readme">
-  <%= ExternalDoc.parse(application.readme) %>
-</div>
+<% if application.private_repo? %>
+  <span class="govuk-error-message">
+    Cannot fetch README of private repo.
+  </span>
+<% elsif ENV["GITHUB_TOKEN"].nil? %>
+  <span class="govuk-error-message">
+    Cannot fetch README without `GITHUB_TOKEN`.
+  </span>
 <% else %>
-<span class="govuk-error-message">
-  Error: cannot fetch README without `GITHUB_TOKEN`.
-</span>
+  <div class="embedded-readme">
+    <%= ExternalDoc.parse(application.readme) %>
+  </div>
 <% end %>
 
 <% end %>


issyl0:
Interesting that this works, because that goes against [the docs](https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#naming-conventions-for-environment-variables).

ChrisBAshton:
> Interesting that this works, because that goes against [the docs](https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#naming-conventions-for-environment-variables).

@issyl0 ah, that puts a spanner in the works. It's hard to know if this commit _does_ work, I just know that providing that ENV var locally compiles the pages properly, and that's somehow not compiling on production.

I guess the implication here is we're not allowed to set `GITHUB_ENV`, and it doesn't get set for us automatically.

I can try renaming the ENV var to not begin with the `GITHUB_` prefix?

issyl0:
@ChrisBAshton Turns out that envvars for `GITHUB_TOKEN` [work fine](https://github.com/Homebrew/brew/blob/master/.github/workflows/spdx.yml#L24), I don't know why the docs say they don't.

ChrisBAshton:
> @ChrisBAshton Turns out that envvars for `GITHUB_TOKEN` [work fine](https://github.com/Homebrew/brew/blob/master/.github/workflows/spdx.yml#L24), I don't know why the docs say they don't.

Thanks for the update, @issyl0. Are you happy for this to be merged then?

kevindew:
Hmm I'm not sure I fully understand what's going on here but I feel like this might not be the right path - as I'd guess you have the access of GITHUB_TOKEN without needing the env var when running on Actions.

Do we need the token set in order to not hit rate limits or is it that some of the repos are private and we need a token with access to the private repos?

ChrisBAshton:
The token is needed to avoid hitting rate limits (unauthenticated requests are limited to something tiny like 50 per hour).
I think it may also be needed for private repos as you say.

When compiling the website, Middleman checks for the presence of `ENV["GITHUB_TOKEN"]` and avoids pulling in README content if it is missing. We could change the logic around that, e.g. remove the conditional and make all developers provide a token.

We could also avoid hitting the API altogether by just CURLing the README endpoints, which is on my todo list, but it would only be a temporary fix because if/when we import _all_ `docs/`, we'll need to hit the API quite a lot.

As far as I know, GitHub Actions provide a fresh token every time, so it can't hit the limit - but they won't automatically authenticate any API requests you make, I think that still needs to be done manually as per this PR.

kevindew:
Thanks Chris. The private repos thing seems a key point to me, as I guess the token we get from GitHub doesn't have access to the private repos nor would the curling suggestion work. If we don't have private repos it seems a bit overkill to force someone to have a token set-up and perhaps more something that should be suggested were GitHub to raise a 429 response.

ChrisBAshton:
I've had another look, @kevindew, and we only have 2 private repos we reference in the dev docs. (We had 3, but one was a mistake - I've raised #2717). I think we're in agreement that a dev shouldn't be forced to create a token when they clone this repo.

I think as a first pass, we should merge this PR, which will expose the auto-generated GITHUB_TOKEN to the Middleman build script and allow it to pull in README content, so that the [README section of app pages](https://docs.publishing.service.gov.uk/apps/asset-manager.html#readme) start working. It's likely that we'll encounter an error for the 2 private repos we have, since the generated token likely won't have access, but we can think about those in a future PR. What do you think?

kevindew:
> I've had another look, @kevindew, and we only have 2 private repos we reference in the dev docs. (We had 3, but one was a mistake - I've raised #2717). I think we're in agreement that a dev shouldn't be forced to create a token when they clone this repo.
> 
> I think as a first pass, we should merge this PR, which will expose the auto-generated GITHUB_TOKEN to the Middleman build script and allow it to pull in README content, so that the [README section of app pages](https://docs.publishing.service.gov.uk/apps/asset-manager.html#readme) start working. It's likely that we'll encounter an error for the 2 private repos we have, since the generated token likely won't have access, but we can think about those in a future PR. What do you think?

I see, hmm the big smell about this is that we have this silent, yet broken build situation. It seems like we should have the build blow up rather than leave error messages in the resulting output. But that can wait until after this PR.

I'm a bit uncomfortable with the half fix that merging this brings. It seems like we'll still have this weird error but less incentive to fix it as it'd only affect a couple of pages. Sounds like the thing we need to is a personal access token that can read the private repos and store that in secrets - might be worth catching up with @richardTowers on that front to see if he's made any progress with IA and if, in his interpretation, a GitHub personal access token is ok to be added as a secret (I believe we essentially do the same with Heroku). Otherwise it'd probably be good to add a conditional to not try pulling in the files if a repo is marked as private so at least this fixes all the error states.

ChrisBAshton:
Thinking about it more, this isn't a half-fix - we _shouldn't_ pull in README content from private repos, as they're private and the dev docs are public! I have pushed up another commit which documents this in code. Hopefully that is sufficient.

I agree the build should error rather than silently fail if the ENV is missing, and agree that we could look at that after this PR.

richardTowers:
I agree with @ChrisBAshton's view that we shouldn't be automatically pulling in READMEs from private repos and publishing it publicly. I can (tenuously) imagine a situation where someone puts something they don't want to be public in a private repo's README and gets surprised by this behaviour.

Maybe we could get the dev docs to do something clever-ish for the private repos to say "This is a private repo, readme is here - link" or whatever. That'd be fine by me.

I chased IA about github actions last week - they're still looking into it.

kevindew:
Sure, I agree that sounds the best way to avoid a nasty surprise, though it does mark a change in this projects direction. Are we confident we only talk to private repos to pull out documentation? I can see there is this line https://github.com/alphagov/govuk-developer-docs/blob/9223a94fe85cfd477b6b353c61e8833f9a3e15ea/app/github_repo_fetcher.rb#L54-L58 but I don't know what code relies on this but it seems like we'll end up with a reduced list - perhaps we'll be lucky and find we can delete this code.

Given not showing a readme will become expected behaviour for private repos would we be able to render something other than an error message? It feels like we'd want to provide a link to the readme instead, perhaps in the links section? maybe it'd also be useful to have `(Private repo)` suffixing the link to GitHub repo in links?

karlbaker02:
For private repos, is there actually a case to even _show them_ in the dev docs? If a repo is private I'd expect it not to be referenced publicly. A use case of this would be if we're building something new that we don't want to announce to the world until it is announced by the government - by making it available in the dev docs, even if only by name, could lead to speculation about what might be going on behind the scenes.

ChrisBAshton:
@karlbaker02 I agree it's strange to reference private repos publicly, but I suppose it allows for those occasional use cases where we've had to make public repos temporarily private, e.g. #1934 and #1970, without having to add/remove lots of config.

For repos we want to remain truly private, I'm guessing we don't reference them in the docs full stop, which seems like the right approach.

ChrisBAshton:
RE adding a link to the repo (@richardTowers , @kevindew) the link already appears above and below the README area.

Here's a mock-up of how this will look when merged:

> Screenshot 2020-09-04 at 09 31 26

Are we good to approve this today? ✅

richardTowers (approved):
Yeah, I think we should merge this.

If people have strong feelings about showing a more prominent link instead of an error, feel free to raise a follow up PR for it.

PR #2710, opened 2020-08-05
Add some info for delivery_transactional latency

1pretz1:
We decided to start monitoring the `delivery_transactional` queue in
[this PR][monitor] and so we need some documenation on what to do if we
see this alert. These are being processed super quick on average (under
a second).

There has been a lack of documentation for the existing queue checks but
as we're planning to remove them all (apart from `delivery_transactional`)
in future work we don't need to document these here too.

Trello:
https://trello.com/c/HwvZx0V9/456-update-the-queue-latency-check

[monitor]: https://github.com/alphagov/govuk-puppet/pull/10574

### DIFF
--- a/source/manual/alerts/email-alert-api-high-queue-latency.html.md
+++ b/source/manual/alerts/email-alert-api-high-queue-latency.html.md
@@ -5,19 +5,40 @@ section: Icinga alerts
 subsection: Email alerts
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2020-07-21
+last_reviewed_on: 2020-08-05
 review_in: 6 months
 ---
 
 ### High queue latency (queue_latency)
 
-This means the time it takes for a job to be processed is unusually high. This
-may indicate a problem down the line which is preventing workers from being
-processed. It may also imply the threshold is too low if a large number of
-emails have being sent out due to a content change.
+This means the time it takes for a job to be processed is unusually high. See
+the Sidekiq ['Queue Latency'][Sidekiq dash] graph for more detail.
 
-See the [sidekiq][sidekiq] section for more information about the Sidekiq queues,
-or read [email troubleshooting].
+For more information on the individual queues and what they do, see
+[here][email queues].
 
-[email troubleshooting]: /manual/email-troubleshooting.html
-[sidekiq]: /manual/sidekiq.html
+Some diagnostic steps you could take are below:
+
+* Check the Sidekiq ['Queue Length'][Sidekiq dash] graph to see if we have a
+  high number of emails queued up.
+* If there is a high amount of emails in the queue, are we processing them at a
+  normal speed? Under high load we regularly send requests to Notify at a rate
+  of ~300 email requests per second. To check this, look at the ['Email Send
+  Attempts'][technical dash] graph.
+* If we're sending email requests slower than usual, refer to the [Email Alert
+  API machine metrics dashboard][machine metrics] or the [AWS RDS postgres
+  dashboard][postgres dash] to see if we're experiencing resourcing issues.
+* Are the workers reporting any problems or any issues being raised in
+  [Sentry]?
+* Check [Kibana] to see our Sidekiq logs for Email Alert API.
+
+See [here][Sidekiq] for more information about Sidekiq.
+
+[Sidekiq]: https://docs.publishing.service.gov.uk/manual/sidekiq.html
+[Sentry]: https://sentry.io/organizations/govuk/issues/?project=202220&statsPeriod=12h
+[Sidekiq dash]: https://grafana.blue.production.govuk.digital/dashboard/file/sidekiq.json?refresh=1m&orgId=1&var-Application=email-alert-api&var-Queues=All&from=now-3h&to=now
+[technical dash]: https://grafana.blue.production.govuk.digital/dashboard/file/email_alert_api_technical.json
+[Kibana]: https://kibana.logit.io/s/2dd89c13-a0ed-4743-9440-825e2e52329e/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-1h,mode:quick,to:now))&_a=(columns:!('@message',host),index:'*-*',interval:auto,query:(query_string:(query:'@type:%20sidekiq%20AND%20application:%20email-alert-api')),sort:!('@timestamp',desc))
+[email queues]: https://docs.publishing.service.gov.uk/apis/email-alert-api/queues.html
+[machine metrics]: https://grafana.blue.production.govuk.digital/dashboard/file/machine.json?refresh=1m&orgId=1&var-hostname=email_alert_api*&var-cpmetrics=cpu-system&var-cpmetrics=cpu-user&var-filesystem=All&var-disk=All&var-tcpconnslocal=All&var-tcpconnsremote=All
+[postgres dash]: https://grafana.production.govuk.digital/dashboard/file/aws-rds.json?orgId=1&var-region=eu-west-1&var-dbinstanceidentifier=blue-postgresql-primary&from=now-3h&to=now


source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -14,10 +14,44 @@ review_in: 6 months
 This means the time it takes for a job to be processed is unusually high. This
 may indicate a problem down the line which is preventing workers from being
 processed. It may also imply the threshold is too low if a large number of
```
benthorner, line 16:
```suggestion

```

benthorner: We can remove these last two sentences: the steps should be specific to the queue.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -14,10 +14,44 @@ review_in: 6 months
...
-emails have being sent out due to a content change.
+emails are currently being processed.
+
+#### `delivery_transactional`
+
+The `delivery_transactional` queue is used for when we need to send one time
```
benthorner, line 21: This is a bit redundant, since we have [this doc](https://github.com/alphagov/email-alert-api/blob/master/docs/queues.md). Perhaps we could move this text into that?

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -14,10 +14,44 @@ review_in: 6 months
...
+emails. Some examples of one time emails could be a confirmation of a
+subscription or when validating a new subscribers email address.
+
+The [Sidekiq priority][priority] for this queue is weighted highly and there
```
benthorner, line 25: I don't think we need this paragraph - it doesn't affect the next steps.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -14,10 +14,44 @@ review_in: 6 months
...
+should only be a very small number of jobs on the queue at any one time which
+means that these emails should be processed pretty promptly. If this is not the
+case and this alert is firing then we should investigate.
+
+We should initially look for degradation of the service as high latency
```
benthorner, line 30: Minor: this is redundant - let's get to the steps ASAP.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -14,10 +14,44 @@ review_in: 6 months
...
+normally indicates degraded performance or high load, or both.
+
+Some diagnostic steps you could take are below:
+
+* Check to see if we have a high number of emails queued up. If these is a
```
benthorner, line 35:
```suggestion
* Check to see if we have a high number of emails queued up. If there is a
```

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -14,10 +14,44 @@ review_in: 6 months
...
+  large amount of `delivery_immediate_high` emails backed up (same Sidekiq
```
benthorner, line 36: This could confuse people: let's just focus on the queue, and assume it's high priority.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
benthorner, line 35: This is a bit of a non-step: it's the definition of the alert. One useful thing we could do here, or in the intro, is point people to the specific Grafana dashboard/panel they can see this on.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -14,10 +14,44 @@ review_in: 6 months
...
+  priority) then these transactional emails could be stuck waiting to be
+  processed.
+* If there is a high amount of emails in the queue, are we processing them at
```
benthorner, line 39: How can I find this out?

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -14,10 +14,44 @@ review_in: 6 months
...
+  a normal speed? Under high load we regularly send requests to Notify at a
+  rate of ~300 email requests per second.
+* Are the workers reporting any problems or any issues being raised in
+  [sentry][sentry]?
+* If everything looks okay from the above you could also check the Sidekiq
```
benthorner, line 44: Could we give a Kibana query instead? Otherwise, it can be pot luck as to whether the logs will show something useful on a particular machine, especially if there are lots of logs.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -14,10 +14,44 @@ review_in: 6 months
...
+  logs which are located on each email-alert-api machine here ->
+  `/var/apps/email-alert-api/sidekiq.json.log`
+* You could also look at the jobs in the queue directly to see if anything
```
benthorner, line 47: Not sure we'd gain anything here. We should have been able to diagnose the problem from the previous steps. The fix may involve doing something in the console, but it's hard to say what in advance. At least we can help people get to that point.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
1pretz1: Sure, I think the explanation is fine in the queues doc so have left it as is.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
1pretz1: Good spot!

source/manual/alerts/email-alert-api-high-queue-latency.html.md
1pretz1: Sure

source/manual/alerts/email-alert-api-high-queue-latency.html.md
1pretz1: Sure, I've linked to the relevant graphs.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
1pretz1: By our swanky email send requests graph, have updated accordingly.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
1pretz1: Good point Ben! I've added a Kibana query.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
1pretz1: Sure, I've removed this step.

1pretz1:
Thanks for the review @benthorner!

I've updated the doc to include your suggestions. Since we decided to make this doc concise I've remove the specific `delivery_transactional` section as these debugging steps could be applied more generically and there wasn't much point having it without the additional fluff.

I have however linked to the queues doc at the end of the page which is a nice way for someone to learn what each queue does without having to repeat the information.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -5,19 +5,38 @@ section: Icinga alerts
...
+[email troubleshooting].
+
+For more information on the individual queues and what they do, see
+[here][email_queues].
 
 [email troubleshooting]: /manual/email-troubleshooting.html
-[sidekiq]: /manual/sidekiq.html
+[Sidekiq]: /manual/sidekiq.html
+[Sentry]: https://sentry.io/organizations/govuk/issues/?project=202220&statsPeriod=12h
+[Sidekiq queue size]: https://grafana.blue.production.govuk.digital/dashboard/file/sidekiq.json?refresh=1m&orgId=1&var-Application=email-alert-api&var-Queues=All&from=now-3h&to=now&panelId=1&fullscreen
```
benthorner, line 38: Minor: we could just link to the dashboard, so it's clear both graphs are in the same place.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -5,19 +5,38 @@ section: Icinga alerts
...
+  queued up.
+* If there is a high amount of emails in the queue, are we processing them at a
+  normal speed? Under high load we regularly send requests to Notify at a rate
+  of ~300 email requests per second. To check this, look at our [email send
+  attempts] graph.
+* Are the workers reporting any problems or any issues being raised in
+  [Sentry]?
+* Check [Kibana] to see our Sidekiq logs for Email Alert API.
+
+See the [Sidekiq] section for more information about Sidekiq or read
```
benthorner, line 29: We shouldn't link to generic steps unless we know they are helpful. In this case, the first two sections of the email troubleshooting are irrelevant.

Also, both these links are broken.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -5,19 +5,38 @@ section: Icinga alerts
...
+[email troubleshooting].
+
+For more information on the individual queues and what they do, see
```
benthorner, line 32: Minor: we could move this to the beginning, as it does provide some useful context about the type of email we're sending.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
```diff
@@ -5,19 +5,38 @@ section: Icinga alerts
...
+This means the time it takes for a job to be processed is unusually high. See
+our [Sidekiq queue latency] graph for more detail.
 
-See the [sidekiq][sidekiq] section for more information about the Sidekiq queues,
-or read [email troubleshooting].
+Some diagnostic steps you could take are below:
+
+* Check the [Sidekiq queue size] graph to see if we have a high number of emails
+  queued up.
+* If there is a high amount of emails in the queue, are we processing them at a
```
benthorner, line 21: What do I do if we're not processing at normal speed?

I think it's worth having this step last, as abnormal speed could be diagnosed directly from errors/logs.

If there's no obvious cause for the speed, we could suggest people look at [the machine dashboard](https://grafana.blue.production.govuk.digital/dashboard/file/machine.json?refresh=1m&orgId=1&var-hostname=email_alert_api*&var-cpmetrics=cpu-system&var-cpmetrics=cpu-user&var-filesystem=All&var-disk=All&var-tcpconnslocal=All&var-tcpconnsremote=All), in case we're running out of resources.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
1pretz1: Sure

source/manual/alerts/email-alert-api-high-queue-latency.html.md
1pretz1: I've removed the troubleshooting link. The links are made up of relative paths so are only broken when rendered by Github. I've made them absolute so you can click em here 😄

source/manual/alerts/email-alert-api-high-queue-latency.html.md
1pretz1: Good point!

source/manual/alerts/email-alert-api-high-queue-latency.html.md
1pretz1: I've kept the ordering the same as I think digging through logs can be / is laborious whereas just looking at some graphs to see an obvious problem could help diagnose the problem quicker. 

I have added your recommendation of adding a link to the machines dashboard as I think that's a good addition.

1pretz1:
Thanks for the 2nd review @benthorner! 

I did think about re-naming the alert but with upcoming work to removing some of the `delivery_*` queue checks then this alert will probably be only acting for the `delivery_transactional` queue and so can be named / changed when this happens to be more specific to just that use case.

source/manual/alerts/email-alert-api-high-queue-latency.html.md
benthorner: Fair enough. I'm not sure that resource issues are obvious to diagnose either. Someone can always iterate the doc when they have an actual incident to work from, though.

Minor: we've had resource issues with DB IOPS in the past, so we could link to [this](https://grafana.production.govuk.digital/dashboard/file/aws-rds.json?orgId=1&var-region=eu-west-1&var-dbinstanceidentifier=blue-postgresql-primary) as well.

1pretz1: Oh cool, didn't know that existed. Have added a link.

PR #2718, opened 2020-08-12
Improve documentation on fact check alert

thomasleese:
This includes some more explicit steps on what 2nd line can do when they see this alert so hopefully it will be easier for them to handle. We're doing this as an experiment as we don't believe the alert will fire enough for us to need to put in place some automated systems, but if it does end up being a problem we will revisit this decision.

[Trello Card](https://trello.com/c/ti6D5GBW/2076-5-manage-unprocessed-fact-check-emails-for-publisher)

### DIFF
--- a/source/manual/alerts/publisher-unprocessed-fact-check-emails.html.md
+++ b/source/manual/alerts/publisher-unprocessed-fact-check-emails.html.md
@@ -1,63 +1,84 @@
 ---
-owner_slack: "#govuk-2ndline"
+owner_slack: "#govuk-platform-health"
 title: "Publisher: Unprocessed fact-check emails"
 section: Icinga alerts
 subsection: Email alerts
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2020-07-02
+last_reviewed_on: 2020-08-12
 review_in: 6 months
 ---
 
-### Unprocessed fact-check emails
+As part of the [Publisher fact checking process], this alert appears if emails
+have arrived in the inbox but weren't able to be processed. This is usually
+because they're missing the identification for the edition they relate to
+(which is currently stored in the subject line).
 
-This means there are unprocessed emails in the [fact-check] inbox (currently
-`factcheck@alphagov.co.uk` but soon to change to
-`govuk-fact-check@digital.cabinet-office.gov.uk`).
-Publisher has a [script (`mail_fetcher`)](script) that checks the inbox
-[every 5 minutes](schedule) and [processes all emails present](process).
+[Publisher fact checking process]: https://github.com/alphagov/publisher/blob/master/doc/fact-checking.md
 
-If emails are not processed or deleted, this alert is triggered. The level of
-the [alert] depends on the number of emails present - which was chosen
-arbitrarily and is currently:
+## Dealing with the alert
 
-- [warning with 1 email](warning)
-- [critical with 2 or more emails](critical)
+### Log in to the inbox
 
-### Log into the fact-check inbox
+To investigate the emails we first need to [log in to the inbox][login].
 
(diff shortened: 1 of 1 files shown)


source/manual/alerts/publisher-unprocessed-fact-check-emails.html.md
```diff
@@ -1,63 +1,83 @@
...
+
+#### Missing edition identification
+
+If the subject line is missing the edition identification but is clearly a
+valid fact check response, the first thing to check is if a follow up
+email was sent with the correct subject.
+
+This can be checked by [logging in to Publisher][publisher] and finding the
+relevant edition by using the title. Usually you'll need to filter on editions
+with the state of "Publisher" or "Out for fact check".
```
AlanGabbianelli, line 72:
```suggestion
with the state of "Published" or "Out for fact check".
```

thomasleese: Oops, I've pushed up an amended commit to fix this!

source/manual/alerts/publisher-unprocessed-fact-check-emails.html.md
```diff
@@ -1,63 +1,83 @@
...
-[warning]: https://github.com/alphagov/govuk-puppet/blob/68507dba280dd6e2fd4e5663f915c0fd7be06bef/modules/govuk/manifests/apps/publisher/unprocessed_emails_count_check.pp#L19
-[critical]: https://github.com/alphagov/govuk-puppet/blob/68507dba280dd6e2fd4e5663f915c0fd7be06bef/modules/govuk/manifests/apps/publisher/unprocessed_emails_count_check.pp#L20
-[login]: https://support.google.com/accounts/answer/1721977?co=GENIE.Platform%3DDesktop&hl=en
-[alert]: https://github.com/alphagov/govuk-puppet/blob/68507dba280dd6e2fd4e5663f915c0fd7be06bef/modules/govuk/manifests/apps/publisher/unprocessed_emails_count_check.pp
+There are a number of different ways of proceeding here based on why the email
+wasn't able to be processed:
+
+#### Spam/out of office
+
+If the email is clearly spam and not relevant, **it can be deleted**.
```
AlanGabbianelli, line 62:
```suggestion
If the email is clearly spam or not relevant, **it can be deleted**.
```
I wouldn't consider out of office emails as spam, but they are indeed not relevant.

source/manual/alerts/publisher-unprocessed-fact-check-emails.html.md
```diff
@@ -1,63 +1,83 @@
...
+
+If the subject line is missing the edition identification but is clearly a
+valid fact check response, the first thing to check is if a follow up
+email was sent with the correct subject.
+
+This can be checked by [logging in to Publisher][publisher] and finding the
+relevant edition by using the title. Usually you'll need to filter on editions
+with the state of "Published" or "Out for fact check".
+
+Once you've found the edition, you can check in the "History" tab to see if
```
AlanGabbianelli, line 74:
```suggestion
Once you've found the edition, you can check in the "History and notes" tab to see if
```

source/manual/alerts/publisher-unprocessed-fact-check-emails.html.md
thomasleese: Ah yep, that's wrong. I've reworded it to make it clearer!

source/manual/alerts/publisher-unprocessed-fact-check-emails.html.md
thomasleese: Good point, I was writing this from memory!

PR #2725, opened 2020-08-13
Review MySQL backups

camdesgov:
Add deprecation note and remove reference to "whitehall-mysql-backup-1.backend"
as this refers to a machine in Carrenza, and Whitehall has now been
migrated to AWS.

I have not yet removed this entire documentation page as we need to investigate if mysql is
relevant to Performance Platform (the last app left on Carrenza).

### DIFF
--- a/source/manual/mysql.html.md
+++ b/source/manual/mysql.html.md
@@ -5,9 +5,13 @@ section: Backups
 type: learn
 layout: manual_layout
 parent: "/manual.html"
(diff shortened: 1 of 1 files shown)


issyl0:
CI was complaining about trailing whitespace, so I fixed it.

camdesgov:
> CI was complaining about trailing whitespace, so I fixed it.

Thanks Issy!

PR #2740, opened 2020-08-25
Add new CNAME to consider AKA domains step

huwd:
Relates to commit b4e5e77b, forgot to change this CNAME record for the AKA step too, also added a line of context in case a reader is confused to find a different CNAME compared to previously migrated domains, or if there's any potential for a clash with older docs (Google Docs from 2017 are still doing the rounds for example)

### DIFF
--- a/source/manual/transition-a-site.html.md
+++ b/source/manual/transition-a-site.html.md
@@ -61,7 +61,10 @@ preview according to the following pattern:
 There are lots of examples of these in [hosts currently configured in
 Transition][transition-hosts].
 
-These AKA domains should be CNAMEd to: `redirector-cdn.production.govuk.service.gov.uk`
+These AKA domains should be CNAMEd to: `bouncer-cdn.production.govuk.service.gov.uk`
+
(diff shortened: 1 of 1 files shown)


source/manual/transition-a-site.html.md
```diff
@@ -61,7 +61,10 @@ preview according to the following pattern:
 There are lots of examples of these in [hosts currently configured in
 Transition][transition-hosts].
 
-These AKA domains should be CNAMEd to: `redirector-cdn.production.govuk.service.gov.uk`
+These AKA domains should be CNAMEd to: `bouncer-cdn.production.govuk.service.gov.uk`
+
+Previously we used a `redirector-cdn` address which will now fail if added as the
```
seb-szy, line 66:
```suggestion
Previously we used a `redirector-cdn` address which reached its capacity and will now fail if added as the
```

huwd:
Much clearer, thanks Seb!

seb-szy (approved):
LGTM

PR #2741, opened 2020-08-25
move deploy_terraform link to AWS integration jenkins

fredericfran-gds:


issyl0 (approved):
I suspect this is on your radar as you wrote it, but you'll have to fixup the `gds govuk terraform` command to point at this new Jenkins too.

PR #2743, opened 2020-08-25
Convert GitHub API calls to HTTP requests

ChrisBAshton:
We save around 70 API calls by curling the URL of each
app README directly. This helps prevent hitting API resource
limits locally, as well as making the code easier to read
since there is no base64 conversion required any more.

It also ensures that we don't accidentally expose any
private repo information. If we use the API and the relevant
access token, we could potentially download private repo content
which we wouldn't want to expose on the Dev Docs. By fetching
the public GitHub URL directly, if it's private we'll run
into a 404.

Finally, this PR removes deprecated logic around organisation names
on GitHub, which is no longer needed now that everything is under
@alphagov. Other parts of the code (README and docs/ ingestion)
make this assumption, so `repo` should too.

Trello: https://trello.com/c/of45ytdG/182-reduce-unnecessary-github-api-requests

### DIFF
--- a/app/github_repo_fetcher.rb
+++ b/app/github_repo_fetcher.rb
@@ -23,7 +23,7 @@ class GitHubRepoFetcher
   # Note that it is cached as pure markdown and requires further processing.
   def readme(app_name)
     CACHE.fetch("alphagov/#{app_name} README", expires_in: 1.hour) do
-      Base64.decode64(client.readme("alphagov/#{app_name}").content)
+      HTTP.get("https://raw.githubusercontent.com/alphagov/#{app_name}/master/README.md")
     rescue Octokit::NotFound
       nil
     end
--- a/app/github_repo_fetcher.rb
+++ b/app/github_repo_fetcher.rb
@@ -3,20 +3,9 @@ require "octokit"
 class GitHubRepoFetcher
   include Singleton
 
-  # Fetch a repo from GitHub.
-  #
-  # If the app_name is just a repo name we'll assume that it's in the alphagov
-  # org. Because 99% of the repos will be from that org, we can download and
-  # cache all alphagov-repos and save on API calls.
+  # Fetch a repo from GitHub
   def repo(app_name)
-    if app_name =~ %r{/}
-      # Not on alphagov, make a separate call to the API. Cache it for
-      # development speed.
-      @cache ||= {}
-      @cache[app_name] ||= client.repo(app_name)
-    else
-      all_alphagov_repos.find { |repo| repo.name == app_name } || raise("alphagov/#{app_name} not found")
-    end
+    all_alphagov_repos.find { |repo| repo.name == app_name } || raise("alphagov/#{app_name} not found")
   end
 
   # Fetch a README for an alphagov application and cache it.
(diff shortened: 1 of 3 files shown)


app/github_repo_fetcher.rb
```diff
@@ -3,27 +3,16 @@
...
-    end
+    all_alphagov_repos.find { |repo| repo.name == app_name } || raise("alphagov/#{app_name} not found")
   end
 
   # Fetch a README for an alphagov application and cache it.
   # Note that it is cached as pure markdown and requires further processing.
   def readme(app_name)
     CACHE.fetch("alphagov/#{app_name} README", expires_in: 1.hour) do
-      Base64.decode64(client.readme("alphagov/#{app_name}").content)
+      HTTP.get("https://raw.githubusercontent.com/alphagov/#{app_name}/master/README.md")
```
issyl0, line 15: If we ever start changing some default branch names to non-`master` but not others, this will break. But we can cross that bridge when we come to it...

spec/app/github_repo_fetcher_spec.rb
```diff
@@ -1,18 +1,29 @@
 RSpec.describe GitHubRepoFetcher do
   describe "#repo" do
-    it "returns a repo if the user is specified" do
-      stub_request(:get, "https://api.github.com/repos/some-user/some-repo")
-        .to_return(body: "{}", headers: { content_type: "application/json" })
+    before do
+      stub_request(:get, "https://api.github.com/users/alphagov/repos?per_page=100")
```
issyl0, line 4:
```suggestion
      stub_request(:get, "https://api.github.com/orgs/alphagov/repos?per_page=100")
```

I know that users and orgs are functionally equivalent - the API returns the same thing - but pulling `alphagov` org data makes more sense to my brain.

issyl0 (commented):
In general, this looks great!

app/github_repo_fetcher.rb
ChrisBAshton: Argh, that's a very good point! I'd be tempted to hold off merging this for that alone 😕 unless we plan on changing branch defaults en masse rather than in dribs and drabs.

issyl0: The GitHub API does return `default_branch` for a repo, so we could reintroduce some code and grab it programatically to mean we have to do less manual work when things change, and we can make this improvement to README grabbing now.

ChrisBAshton: Thanks - looks like we already have access to that information, as we cache all the repo info up front, so it's a fairly trivial code change. Unfortunately the tests are another matter entirely 😓 

This is what I have so far but I'm hoping to refactor tomorrow:
https://github.com/alphagov/govuk-developer-docs/pull/2743/commits/3f12972ab4ce0152322b7e073fd3ed3db61c73d7

spec/app/github_repo_fetcher_spec.rb
ChrisBAshton: Don't think we can do that - this before block is just stubbing the endpoint that gets called by the Octokit SDK.

app/github_repo_fetcher.rb
ChrisBAshton: Have had another look at this today and whilst it's a bit of a smell in places, I think it's mergeable. Think it would take quite a rewrite to make it cleaner, but with the repo branch name changes happening over the next few weeks I'd sooner just remove my last commit and wait for everything to be on `main` than attempt a rewrite now.

Will request a review from you and see what you think 😄

spec/app/github_repo_fetcher_spec.rb
issyl0: 👍🏻

issyl0 (approved):
Thanks for doing this, Chris!

PR #2749, opened 2020-09-04
Updates rabbitmq-no-consumers-listening doc

1pretz1:
Updates the doc to be more actionable and useful.

The original discussion around the decisions made in this doc can be
found [here][original_pr].

Trello:
https://trello.com/c/0OHZT3cq/458-investigate-rabbitmq-and-improve-the-documentation

[original_pr]: https://github.com/alphagov/govuk-developer-docs/pull/2733

### DIFF
--- a/source/manual/alerts/rabbitmq-no-consumers-listening.html.md
+++ b/source/manual/alerts/rabbitmq-no-consumers-listening.html.md
@@ -4,75 +4,92 @@ title: 'RabbitMQ: No consumers listening to queue'
 parent: "/manual.html"
 layout: manual_layout
 section: Icinga alerts
-last_reviewed_on: 2019-12-05
+last_reviewed_on: 2020-09-07
 review_in: 6 months
 ---
 
-[Read more about how we use RabbitMQ][rabbitmq]
+## Check that there is at least one non-idle consumer for rabbitmq queue {queue_name}
 
-We run a check that there is at least one non-idle consumer for a named
-RabbitMQ queue. (For example, for the `email_alert_service` queue.) The queue
-name should indicate the app responsible for consuming the queue.
+Icinga connects to RabbitMQ's admin API to check on the activity of the
+consumers and that at least one consumer is running for a given RabbitMQ
+message queue. See [here][check_rabbitmq_plugin] for the plugin that implements
+the alert.
 
-The Icinga check is performed by connecting to RabbitMQ's admin API, so the
-information given is from Rabbit's point of view.
+For information about how we use RabbitMQ, see [here][rabbitmq_doc].
 
-Since we send heartbeats every minute (from each instance of the Publishing
-API) at least one of the consumers for each queue should be active every
-minute. To be conservative, the check checks that at least one consumer of the
-queue is running and has been active within the last 5 minutes.
+### `No consumers listening to queue`
 
-If the check succeeds, it will return the most recent time at which activity
-happened on the queue.
+This check reports a critical error when **no** consumers are listening to the
+queue, meaning messages entering the queue will never be processed.
 
-If the check fails due to no consumers having been active recently, it will
-report a critical failure, giving the most recent idle time, and the idle
-times of each consumer connected to RabbitMQ:
+### `No activity for X seconds: idle times are [X, Y, Z]`
 
-```
-CRITICAL: No activity for X seconds: idle times are [X, Y, Z]
-```
+This checks whether the RabbitMQ consumers for a queue have been active in the
+last 5 minutes. Consumers in an `idle` state are listening to the queue but are
+unable to process the messages on it.
 
-If the check fails due to there being no consumers connected to the queue, it
-will report a critical failure:
+Publishing API [sends a heartbeat message][heartbeat_messages] every minute to
+the following queues - `email-alert-service`, `cache_clearing_service-high` and
+`content_data_api`. This is configured via the queues bindings (e.g
+[email-alert-service's binding][email_alert_binding]) matching the routing key
+used in the [heartbeat][heartbeat]. This should ensure that the consumers are
+never idle for these queues.
 
-```
-CRITICAL: "No consumers listening to queue"
-```
+> **Note**
+>
+> You may see the [high unprocessed messages][high_unprocessed_messages] alert too,
+> as issues with consumers processing messages could then lead to a high
+> backlog of messages.
 
-## Consequences of idle consumers
+## Troubleshooting
 
-If consumers are idle, messages sent to the queue will not be being processed.
-This means that the actions which are meant to happen in response to the
-messages will not happen. For example, email alerts about updated content will
-not be sent if the `email_alert_sevice` queue isn't running.
+### Check the RabbitMQ logs
 
-Unless there is a wider RabbitMQ failure, messages will not be lost - they will
-be processed once the problem is resolved.
+Check the logs for the `rabbitmq` application which can be achieved logging
+into Kibana and searching for `application: rabbitmq`. Is there evidence of any
+errors?
 
-## Investigation
+### Check the RabbitmQ Grafana dashboard
 
-RabbitMQ has an admin interface, which allows details of recent activity on the
-queues to be seen. To log in, [follow the instructions in the RabbitMQ
-docs][rabbitmq_control_panel].
+This [Grafana dashboard][rabbitmq_grafana_dashboard] shows activity across
+multiple exchanges and queues. The main exchange we expect to be monitoring is
+`published_documents` which handles broadcasting to services such as search and
+email-alert-service when content changes across GOV.UK.
 
-The admin interface allows you to see details (and graphs!) of the messages
-sent to each queue, and the number of messages held on the queues. Look at the
-queue for which the alert happened. If no messages have been sent to the queue
-over the last several minutes, the most likely failure is that the heartbeat
-messages are no longer being sent correctly. Look for recent changes to the
-[Publishing API][publishing_api].
+Looking at the queue graphs we should look out for the following:
 
-If messages have been received, and the queue has messages held on it and not
-being processed, the consumers have either got stuck or stopped running for
-some reason. Restarting the consumers for the relevant app will probably clear
-this up (but please make sure the team owning the app are also alerted, since
-this shouldn't normally happen). For example:
+* **Check for high 'ready' messages for the queue** - indicates these messages are
+  waiting to be processed in RabbitMQ by the consumer (e.g
+  email-alert-service).
 
-```sh
+* **Check for high 'unacknowledged' messages for the queue** - implies that
+  messages have been read by the consumer but the consumer has never sent back
+  an ACK to the RabbitMQ broker to say that it has finished processing.
+
+* **Check for high 'redeliver' rate for the queue** - in the event of network
+  failure (or a node failure), messages can be redelivered. An example is if
+  the consumer dies (its channel is closed, connection is closed, or TCP
+  connection is lost).
+
+If we're seeing high 'redeliver' rates, high 'ready' or 'unacknowledged'
+messages then this could indicate an issue with the consumer.
+
+### Troubleshooting steps
+
+1. You could try restarting the consumer applications. After restarting, check
+   to see if the problem is solved. E.g for email-alert-service, run:
+
+```bash
 $ fab $environment class:email_alert_api app.restart:email-alert-service
 ```
 
-[publishing_api]: https://github.com/alphagov/publishing-api
-[rabbitmq]: /manual/rabbitmq.html
-[rabbitmq_control_panel]: /manual/rabbitmq.html#connecting-to-the-rabbitmq-web-control-panel
+2. If the issue has not resolved, we should check in the consumers application
+   logs to see if any errors are being thrown.
+
+[high_unprocessed_messages]: https://docs.publishing.service.gov.uk/manual/alerts/rabbitmq-high-number-of-unprocessed-messages.html
+[email_alert_binding]: https://github.com/alphagov/email-alert-service/blob/4412a1b3b0f281733801e1631416ab02fac90e25/lib/tasks/message_queues.rake#L17
+[rabbitmq_doc]: https://docs.publishing.service.gov.uk/manual/rabbitmq.html
+[check_rabbitmq_plugin]: https://github.com/alphagov/govuk-puppet/blob/eb8a04a7883d4772fa7266c909c7f40563f8f7a0/modules/icinga/files/usr/lib/nagios/plugins/check_rabbitmq_consumers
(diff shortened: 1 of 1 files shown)


source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,106 @@ title: 'RabbitMQ: No consumers listening to queue'
...
+## Check that there is at least one non-idle consumer for rabbitmq queue <queue>
 
-We run a check that there is at least one non-idle consumer for a named
-RabbitMQ queue. (For example, for the `email_alert_service` queue.) The queue
-name should indicate the app responsible for consuming the queue.
+For information about how we use RabbitMQ, see [here][rabbitmq_doc].
 
-The Icinga check is performed by connecting to RabbitMQ's admin API, so the
-information given is from Rabbit's point of view.
+## Criticals
```
benthorner, line 15: We can lose this heading, as these are really just one section.

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,106 @@ title: 'RabbitMQ: No consumers listening to queue'
...
-happened on the queue.
+This check reports a critical error when no consumers are listening to the
+queue.
 
-If the check fails due to no consumers having been active recently, it will
-report a critical failure, giving the most recent idle time, and the idle
-times of each consumer connected to RabbitMQ:
+Icinga connects to RabbitMQ's admin API to check that at least one consumer is
+running for a given RabbitMQ message queue. See
+[here][no_consumers_listening_config] for the alert config.
```
benthorner, line 24: This isn't the config, although it's helpful to point to the implementation 👍

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,106 @@ title: 'RabbitMQ: No consumers listening to queue'
...
-If the check fails due to there being no consumers connected to the queue, it
-will report a critical failure:
+This checks whether the RabbitMQ consumers for a queue have been active in the
+last 5 minutes.
 
-```
-CRITICAL: "No consumers listening to queue"
-```
+Icinga connects to RabbitMQ's admin API to check on the activity of the
+consumers of a given RabbitMQ message queue. See [here][idle_config] for the
```
benthorner, line 32: Similarly: the link doesn't point to config.

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,106 @@ title: 'RabbitMQ: No consumers listening to queue'
...
+[here][no_consumers_listening_config] for the alert config.
 
-```
-CRITICAL: No activity for X seconds: idle times are [X, Y, Z]
-```
+### `No activity for X seconds: idle times are [X, Y, Z]`
 
-If the check fails due to there being no consumers connected to the queue, it
-will report a critical failure:
+This checks whether the RabbitMQ consumers for a queue have been active in the
```
benthorner, line 28: This sentence is repetitive compared to the one above, but the nature of the alert is very different. Can we incorporate the note (below) more prominently here, to emphasise these are difference between the checks? Why do we have two checks?

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,106 @@ title: 'RabbitMQ: No consumers listening to queue'
...
-```
+Icinga connects to RabbitMQ's admin API to check on the activity of the
+consumers of a given RabbitMQ message queue. See [here][idle_config] for the
+alert config.
+
+**Note**
+> Publishing API [sends a heartbeat message][heartbeat_messages] every minute to
+> the following queues - `email-alert-service`, `cache_clearing_service-high` and
+> `content_data_api`. These queues receive the heartbeat due to being classed as
+> particularly important. This is configured via the queues bindings matching the
```
benthorner, line 39: Minor: we could give an example link to a binding, for one of these queues.

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,106 @@ title: 'RabbitMQ: No consumers listening to queue'
...
+> routing key used in the [heartbeat][heartbeat]. This should ensure that the
+> consumers are never idle for these queues.
+
+## Troubleshooting
+
+### Check the RabbitMQ logs
+
+* Check the logs for the `rabbitmq` application which can be achieved logging
```
benthorner, line 47: Why is this a bullet point?

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,106 @@ title: 'RabbitMQ: No consumers listening to queue'
...
+This [Grafana dashboard][rabbitmq_grafana_dashboard] shows activity across
+multiple exchanges and queues. The main exchange we expect to be monitoring is
+`published_documents` which handles broadcasting to services such as search and
+email-alert-service when content changes across GOV.UK.
+
+Looking at the queue graphs we should look out for the following:
+
+* Check for high 'ready' messages for the queue
+
+This indicates these messages are waiting to be processed in RabbitMQ by the
```
benthorner, line 62: Minor: the formatting here means it's ambiguous whether "this" refers to the previous bullet point, or the next one. Can we just tack is on to the end?

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,106 @@ title: 'RabbitMQ: No consumers listening to queue'
...
 
-The admin interface allows you to see details (and graphs!) of the messages
-sent to each queue, and the number of messages held on the queues. Look at the
-queue for which the alert happened. If no messages have been sent to the queue
-over the last several minutes, the most likely failure is that the heartbeat
-messages are no longer being sent correctly. Look for recent changes to the
-[Publishing API][publishing_api].
+3. If messages are being received by the consumer check they are actually being
+   actioned (e.g email-alert-api is receiving the messages email-alert-service
+   has received from RabbitMQ). To do this we could compare the content changes
```
benthorner, line 89: This is too specific, given the alert applies to several apps. I think we should see errors in all cases, somewhere, so we can just stick with the previous step.

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,106 @@ title: 'RabbitMQ: No consumers listening to queue'
...
+   in [email-alert-api][email-alert-api] against whats recorded in the
+   email-alert-service logs.
 
-If messages have been received, and the queue has messages held on it and not
-being processed, the consumers have either got stuck or stopped running for
-some reason. Restarting the consumers for the relevant app will probably clear
-this up (but please make sure the team owning the app are also alerted, since
-this shouldn't normally happen). For example:
+4. If messages aren't being processed then the consumers could be stuck or
```
benthorner, line 93: This could replace the first bullet point: it's worth trying a blind restart ([esp. for RabbitMQ](https://docs.publishing.service.gov.uk/manual/alerts/rebooting-machines.html#rebooting-rabbitmq-machines)) to see if that fixes it, and then seeing if there are any errors.

benthorner (commented):
Thanks for having another go at this 🏅. It's much clearer now that there are really two potential alerts for this documentation. And I like that we have more steps than "restart and hope"! Just a few comments to clarify a couple of things, and then this should be ready to go!

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
1pretz1: Good point - I've moved the link to near the top of the doc and referenced it as the alerts plugin instead of mentioning it twice and pointing to different lines.

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
1pretz1: See the comment above ^

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
1pretz1: Sure, I've made the note into a paragraph just below the alert description. I've also emphasised what each alert actually means in each description which should make it clear they're distinct. Let me know what you think?

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
1pretz1: Not sure, it has been removed! 😆

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
1pretz1: Sure, I've replaced step 1.

1pretz1:
Thanks for the review @benthorner, I've updated accordingly.

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,94 @@ title: 'RabbitMQ: No consumers listening to queue'
...
-report a critical failure, giving the most recent idle time, and the idle
-times of each consumer connected to RabbitMQ:
+### `No activity for X seconds: idle times are [X, Y, Z]`
 
-```
-CRITICAL: No activity for X seconds: idle times are [X, Y, Z]
-```
+This checks whether the RabbitMQ consumers for a queue have been active in the
+last 5 minutes. Consumers in an `idle` state are listening to the queue but are
+unable to process the messages on it. Icinga connects to RabbitMQ's admin API
```
benthorner, line 28: Minor: you could DRY this up by saying it after the link to the plugin (above).

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,94 @@ title: 'RabbitMQ: No consumers listening to queue'
...
+to check on the activity of the consumers of a given RabbitMQ message queue.
 
-If the check fails due to there being no consumers connected to the queue, it
-will report a critical failure:
+Publishing API [sends a heartbeat message][heartbeat_messages] every minute to
+the following queues - `email-alert-service`, `cache_clearing_service-high` and
+`content_data_api`. These queues receive the heartbeat due to being classed as
```
benthorner, line 33: Minor: this is a bit convoluted - my original suggestion wasn't great. We could just say "... because they listen for `*.major` messages".

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,94 @@ title: 'RabbitMQ: No consumers listening to queue'
...
+If we're seeing high 'redeliver' rates, high 'ready' or 'unacknowledged'
+messages then this could indicate an issue with the consumer.
 
-## Investigation
+### Troubleshooting steps
 
-RabbitMQ has an admin interface, which allows details of recent activity on the
-queues to be seen. To log in, [follow the instructions in the RabbitMQ
-docs][rabbitmq_control_panel].
+1. Reboot all RabbitMQ nodes, you can do this safetly by following the steps
```
benthorner, line 74: Noooooo! This would probably make it worse. I was actually referring to restarting the consumers processes, which can sometimes get stuck _when RabbitMQ reboots_.

benthorner (commented):
Almost there! I think something got lost in translation at the bottom 😬.

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
1pretz1: I've just got rid of the sentence as I don't think it adds much value as the sentence after it describes how it's configured.

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
1pretz1: Mentioning restarting and linking to a rebooting guide made me assume you meant rebooting RabbitMQ. 🙈 

Oh well, I've added what you were actually thinking. 😄

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,89 @@ title: 'RabbitMQ: No consumers listening to queue'
...
+  connection is lost).
 
-```sh
+If we're seeing high 'redeliver' rates, high 'ready' or 'unacknowledged'
+messages then this could indicate an issue with the consumer.
+
+### Troubleshooting steps
+
+1. You could try restarting the consumer applications. After restarting, check
+   to see if the problem is solved, if not, check to see if any errors are
```
benthorner, line 74:
```suggestion
```

(we say this in the next step, although it can be irrespective of whether the service is running)

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -4,75 +4,89 @@ title: 'RabbitMQ: No consumers listening to queue'
...
+```bash
 $ fab $environment class:email_alert_api app.restart:email-alert-service
 ```
 
-[publishing_api]: https://github.com/alphagov/publishing-api
-[rabbitmq]: /manual/rabbitmq.html
-[rabbitmq_control_panel]: /manual/rabbitmq.html#connecting-to-the-rabbitmq-web-control-panel
+2. If the services are running we should look in the consumers application logs.
+
+3. If messages aren't being processed then the consumers could be stuck or
```
benthorner, line 83:
```suggestion
```

We've already said this in (1).

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
benthorner: Ah, sorry about that. The bit I was referring to was:

> Publishing API jobs became stuck - This required restarting Publishing API workers to resolve.

But I see now how it's confusing, as the following looks more relevant:

> No non-idle RabbitMQ consumers - This required killing RabbitMQ processes to resolve.

Looking in the incident report, I think this was something of a one-off due to a planned reboot going wrong. While the alert was triggered, the actual problem was with RabbitMQ itself, and this alert is only tangentially related.

benthorner (approved):
Thanks for pursuing all the comments 🚲. I think the actions at the end still need a few changes, but all the content is there, so I'll approve this now.

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
1pretz1: Updated the text on both steps! 👍

source/manual/alerts/rabbitmq-no-consumers-listening.html.md
1pretz1: Ah yes 🤦

PR #2753, opened 2020-09-08
Update process for unprocessed fact check emails

brucebolt:
By forwarding the message rather than copying it into Publisher, it retains the status of a fact check response (rather than an internal note)  and updates the edition's status in Publisher to "Fact check received" which alerts the content designer to the response having been received.


thomasleese (approved):

cbaines (approved):
Sounds reasonable to me, you might want to check with @thomasleese as he's more familiar with broken fact check emails.

PR #2759, opened 2020-09-10
Remove unused .puml diagram files

ChrisBAshton:
These don't appear to be referenced anywhere, and haven't had a
programatic update for 5 months (there was a manual update
4 months ago, but I think this was just a generic find-and-remove)
5bdff643f5a569a01b5a1d6bf14d500534ca702d

It was never ideal that these used to update with a 'bundle
exec rake' anyway - auto-generated files should stay out of
commit history
4bf03398221a3cf79b882a2a9ba93d54ae019cb2

If we need to add them back in in future, we can do so, hopefully
with more automation.


erino:
@VitalieMogoreanu Can these files be deleted?

thomasleese (approved):
This makes sense to me, and we can always revert this PR if we need them back.

PR #2760, opened 2020-09-10
Copy over Licensing documentation from Legacy Ops Manual

### LINKED ISSUE #1777: Add support info for licensing apps
Debugging alerts for licensing is hard because it differs from the standard GOV.UK architecture in some ways, but not in others. It's not something we're actively working on, so there's little shared knowledge about where it exists on GOV.UK and what the different machines are for. 

It was also surprising that it exists in Signon, given that it's commonly referred to as being 'special'. Appreciate all that is quite vague, but some kind of overview doc that explains the support status and touches on the infrastructure and user journeys would be helpful :-).

ChrisBAshton:
Copies everything with the exception of the Contact section, which
is likely to be outdated now anyway. Nothing else looks particularly
sensitive so can exist in the public domain.

Fixes #1777.

### DIFF
--- a/source/manual/licensing.html.md
+++ b/source/manual/licensing.html.md
@@ -17,11 +17,11 @@ There is a [PDF user manual](https://licensify-admin.publishing.service.gov.uk/a
 
 ### Applications
 
-The Licensing service consists of three separately deployed applications:
+The Licensing service consists of [three separately deployed applications](https://github.com/alphagov/govuk-puppet/blob/f45e303e4e2555ac8cdc003898f17e4dbe86add5/hieradata_aws/common.yaml#L132-L140):
 
-- Frontend: the interface used by citizens and businesses who want to apply for a license. Allows them to download the relevant forms, submit them once completed, and also handles redirection to the relevant payment service if a fee is required.
-- Backend: administrative interface for LAs to allow them to view and download applications from users. Also provides the ability for LA admin users to set up customisations (the specifics of the licence types) for their LA, and for GDS admins to control which licence types are available for each LA, and configuration like which payment provider the LA uses.
-- Feeds: backend (asynchronous) processing of application uploads and processing (via Apto), email notifications (via GOV.UK Notify), and expiry and purging of old applications. Also provides an interface for some debugging-type reporting.
+- `licensify` (on the `licensing_frontend` machine): the interface used by citizens and businesses who want to apply for a license. Allows them to download the relevant forms, submit them once completed, and also handles redirection to the relevant payment service if a fee is required.
+- `licensify-admin` (on the `licensing_backend` machine): administrative interface for LAs to allow them to view and download applications from users. Also provides the ability for LA admin users to set up customisations (the specifics of the licence types) for their LA, and for GDS admins to control which licence types are available for each LA, and configuration like which payment provider the LA uses.
+- `licensify-feed` (on the `licensing_backend` machine): backend (asynchronous) processing of application uploads and processing (via Apto), email notifications (via GOV.UK Notify), and expiry and purging of old applications. Also provides an interface for some debugging-type reporting.
 
 ### Finding a licence
 
--- a/source/manual/licensing.html.md
+++ b/source/manual/licensing.html.md
@@ -60,7 +60,13 @@ You can give the application reference to Kibana in the relevant environment to 
 The Feeds application has a UI that runs on port `9400`, but that isn't exposed publicly, so you need to tunnel to it. For example, the following will tunnel to the production instance of the Feeds admin and make it available via `localhost:9400`:
 
 ```bash
-ssh licensing-backend-1.licensify.production -L 9400:localhost:9400 -N
+gds govuk connect -e production ssh licensing_backend
+
+# Look at the output and grab the IP address, e.g.
+# Running command: ssh -J username@jumpbox.production.govuk.digital username@ip-10-13-5-62.eu-west-1.compute.internal
+
+# now exit the machine and run this command locally, replacing the IP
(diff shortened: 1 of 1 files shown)


source/manual/licensing.html.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Overview
+
+There is a [PDF user manual](https://licensify-admin.publishing.service.gov.uk/assets/manual.pdf) available which includes a lot of detail about the backend (administration) facilities available through the GOV.UK Licensing service.
+
+### Applications
+
+The Licensing service consists of three separately deployed applications:
+
+- Frontend: the interface used by citizens and businesses who want to apply for a license. Allows them to download the relevant forms, submit them once completed, and also handles redirection to the relevant payment service if a fee is required.
```
benthorner, line 22: We should avoid using "Frontend" here and subsequently, since we have an app called that already!

source/manual/licensing.html.md
```diff
@@ -0,0 +1,111 @@
...
+- Backend: administrative interface for LAs to allow them to view and download applications from users. Also provides the ability for LA admin users to set up customisations (the specifics of the licence types) for their LA, and for GDS admins to control which licence types are available for each LA, and configuration like which payment provider the LA uses.
+- Feeds: backend (asynchronous) processing of application uploads and processing (via Apto), email notifications (via GOV.UK Notify), and expiry and purging of old applications. Also provides an interface for some debugging-type reporting.
+
+### Finding a licence
+
+Users use [Licence Finder](https://www.gov.uk/licence-finder) to locate licences on the GOV.UK frontend, but from a technical perspective this is separate from Licensify.
+
+Licence Finder sends users to a "Licence" page served by [Frontend](/apps/frontend.html). Users enter their postcode and Frontend uses [Mapit](/apps/mapit.html) to find their local authority. Frontend makes a request to a Licensify API to find out whether the licence is available on GOV.UK.
```
benthorner, line 30: A code reference for the use of Mapit would be handy here.

source/manual/licensing.html.md
```diff
@@ -0,0 +1,111 @@
...
+1. Login to Licensify's admin interface to verify that your application is visible
+
+You can give the application reference to Kibana in the relevant environment to see the logs and any error messages. The application should speak to both the frontend and backend app servers at least once during the transaction. It may take up to five minutes to process the licence application, so it might be worth waiting for a few minutes before checking Kibana.
+
+## Debugging
+
+The Feeds application has a UI that runs on port `9400`, but that isn't exposed publicly, so you need to tunnel to it. For example, the following will tunnel to the production instance of the Feeds admin and make it available via `localhost:9400`:
+
+```bash
+ssh licensing-backend-1.licensify.production -L 9400:localhost:9400 -N
```
benthorner, line 63: This doesn't work anymore, and involves the convoluted steps of:

```
gds govuk connect -e production ssh licensing_backend

Running command: ssh -J benthorner@jumpbox.production.govuk.digital benthorner@ip-10-13-5-62.eu-west-1.compute.internal

...

ssh benthorner@ip-10-13-5-62.eu-west-1.compute.internal -J benthorner@jumpbox.production.govuk.digital -L 9400:localhost:9400 -N
```

source/manual/licensing.html.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+You can then copy this file onto an integration machine (possibly via your local machine) and restore the data:
+
+```bash
+tar xvzf licensifydump.tar.gz
+mongorestore licensifydump/
+```
+
+Confusingly, some collections exist in both the `licensify` and the `licensify-refdata` database. If `db.collectionName.count() == 0` in one database, and `db.collectionName.count() != 0` in the other, then you should assume that the one with no objects is not the source of truth.
```
benthorner, line 103:
```suggestion
Confusingly, some collections exist in both the `licensify` and the `licensify-refdata` database. If `db.collectionName.count() == 0` in one database, and `db.collectionName.count() != 0` in the other, then you should assume that the one with objects is the source of truth.
```

source/manual/licensing.html.md
benthorner: https://github.com/alphagov/frontend/blob/8dbe0f0f6ca0f5a777a0d6ca77b858fe5adc2494/app/controllers/licence_controller.rb#L83

benthorner (approved):
Thanks for actioning my issue 🏅 - that was lazy of me to just leave it there.

source/manual/licensing.html.md
ChrisBAshton: The Legacy Ops manual actually links to the Frontend application, so I think they're referring to the same thing 🤔 
It is confusing, however, as we don't have a "Backend" application...

source/manual/licensing.html.md
ChrisBAshton: Thanks for finding the reference.

source/manual/licensing.html.md
ChrisBAshton: That works - thanks!

source/manual/licensing.html.md
benthorner: Ah, so I think that's (confusingly!) true [below](https://github.com/alphagov/govuk-developer-docs/pull/2760/files#diff-b5c81e34e93e8b8b37fae6de48881dc0R30), but here, we're referring to [these ones](https://github.com/alphagov/govuk-puppet/blob/master/hieradata_aws/common.yaml#L132-L140).

ChrisBAshton: Thanks - I've rewritten this bit, hopefully it's clearer now.

benthorner (approved):

PR #2761, opened 2020-09-10
remove-user-from-puppet: Remove mention of the training environment

issyl0:
- The training environment was a spike that no longer exists, and its configuration was removed in https://github.com/alphagov/govuk-puppet/pull/10509.


kevindew (approved):
🎉

PR #2768, opened 2020-09-14
Add accessibility statement

thomasleese:
This was written for us by the technical writers team ahead of the 23rd September deadline.

There are numerous accessibility issues with the developer docs and we plan to fix those before September next year.

### DIFF
--- a/source/accessibility.html.md
+++ b/source/accessibility.html.md
@@ -1,0 +1,86 @@
+---
+layout: core
+title: Accessibility statement
+review_in: 6 months
+hide_in_navigation: true
+last_reviewed_on: 2020-09-02
+---
+
+# Accessibility statement
+
+This accessibility statement applies to the GOV.UK developer documentation at [https://docs.publishing.service.gov.uk/](https://docs.publishing.service.gov.uk/).
+
+This website is run by the GOV.UK developer team at the Government Digital Service (GDS). We want as many people as possible to be able to use this website. For example, that means you should be able to:
+
+- change colours, contrast levels and fonts
+- zoom in up to 300% without the text spilling off the screen
+- navigate most of the website using just a keyboard
+- navigate most of the website using speech recognition software
+- listen to most of the website using a screen reader (including the most recent versions of JAWS, NVDA and VoiceOver)
+
+We’ve also made the website text as simple as possible to understand.
+
+[AbilityNet](https://mcmw.abilitynet.org.uk/) has advice on making your device easier to use if you have a disability.
+
+## How accessible this website is
+
+We know some parts of this website are not fully accessible.
+
+- Some links are broken.
+- Some tables are structured in an inaccessible way.
+- Some headings do not fit with the content that follows.
+- Some links do not clearly explain where they lead, or that they lead to external sites.
+- Some pages skip heading levels.
+- Some images do not have good alternative text, or have no alternative text.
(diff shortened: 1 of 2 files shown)


source/accessibility.html.md
```diff
@@ -0,0 +1,87 @@
...
+We’ll consider your request and get back to you within 5 working days.
+
+## Reporting accessibility problems with this website
+
+We’re always looking to improve the accessibility of this website. If you find any problems not listed on this page or think we’re not meeting accessibility requirements, [contact us using the GOV.UK contact form](https://www.gov.uk/contact/govuk).
+
+### Enforcement procedure
+
+The Equality and Human Rights Commission (EHRC) is responsible for enforcing the Public Sector Bodies (Websites and Mobile Applications) (No. 2) Accessibility Regulations 2018
+(the ‘accessibility regulations’). If you’re not happy with how we respond to your complaint, [contact the Equality Advisory and Support Service (EASS)](https://www.equalityadvisoryservice.com/).
```
ChrisBAshton, line 51: This looks like an unexpected newline (all other paragraphs are encapsulated on their own line).

source/accessibility.html.md
```diff
@@ -0,0 +1,87 @@
...
+
+#### Non-compliance with the accessibility regulations
+
+- Some links are broken, which fails WCAG 2.1 success criterion 2.1.1 Keyboard.
+- Some tables are structured in a way that fails WCAG 2.1 success criteria 1.3.1 Info and Relationships and 1.3.2 Meaningful Sequence.
+- Some content uses emojis without alternative text, which fails WCAG 2.1 success criteria 1.1.1 Non-text Content.
+- Some headings do not fit with the content that follows, which fails WCAG 2.1 success criterion 2.4.6 Headings and Labels.
+- Some links do not clearly explain where they lead, or that they lead to external sites. This fails WCAG 2.1 success criteria 2.4.4 Link Purpose (In Context) and 2.4.4 Link Purpose (In Context).
+- Some pages skip heading levels, which fails WCAG 2.1 success criteria 1.3.1 Info and Relationships, 2.4.1 Bypass Blocks and 2.4.6 Headings and Labels.
+- Some images do not have good alternative text or are missing alternative text, which fails WCAG 2.1 success criterion 1.1.1 Non-text Content.
```
ChrisBAshton, line 73: This list duplicates the one further above - I wonder if it's worth considering keeping to one list to ensure that we don't have to update in multiple places?

ChrisBAshton (approved):
LGTM 👍 couple of minor points below.

source/accessibility.html.md
thomasleese: Yep, good point, I've fixed that!

source/accessibility.html.md
thomasleese: I'll ask the tech writers and see what they think.

m-green: Hi Chris - good call and we'd definitely normally avoid this kind of duplication! However we need to follow the structure of the [accessibility statement template](https://www.gov.uk/government/publications/sample-accessibility-statement/), which means we cover the list twice - once without WCAG criteria and once with. Hope that's ok.

ChrisBAshton: No problem, just thought I'd check - thanks for the response!

PR #2775, opened 2020-09-16
Add documentation for getting CKAN dataset list

oscarwyatt:
Useful for Zendesk queries where an organisation
wants a list of all their datasets, whether
public, drafted or deleted


deborahchua (approved):
🎉

PR #2777, opened 2020-09-17
Review longstanding Alerts documentation

Tetrino:
Spent some time testing each of these alert docs such that I know the
content within them are valid. Testing methodology involved running
any safe associated commands to look for errors that don't actually
exist (right now) because we're not seeing the alerts, checking links
and processes, etc.

We can't wait for these alerts to actually trigger to verify this
documentation so I did the closest next-best-thing I could think of and
tested them as if there was a valid alert I was trying to solve - within
reason.

### DIFF
--- a/source/manual/alerts/aws-lb-healthyhosts.html.md
+++ b/source/manual/alerts/aws-lb-healthyhosts.html.md
@@ -4,7 +4,7 @@ title: AWS LB Healthy Hosts
 parent: "/manual.html"
 layout: manual_layout
 section: Icinga alerts
-last_reviewed_on: 2020-03-05
+last_reviewed_on: 2020-09-17
 review_in: 6 months
 ---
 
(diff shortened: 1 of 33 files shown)


beccapearce (approved):
Looks good to me! :)

1pretz1:
Nice! One comment is we should probably stagger the dates of when the docs were reviewed (last_reviewed_on) otherwise we'll have a big dump of them all on one day. Same goes for https://github.com/alphagov/govuk-developer-docs/pull/2778. 😄

Tetrino:
> Nice! One comment is we should probably stagger the dates of when the docs were reviewed (last_reviewed_on) otherwise we'll have a big dump of them all on one day in 6 months time. Same goes for #2778. 😄

Agreed! I'll do that when I have some time in a bit.

PR #2785, opened 2020-09-21
Switch to using Kramdown for parsing markdown

ChrisBAshton:
Redcarpet has an [issue from 2015][issue] whereby shortcut
reference links get confused. For example:

```md
See [foo] (example: bla)
[foo]: /bar.html
```

This was being parsed as `See <a href="(example: bla)">foo<a>`,
instead of: `See <a href="/bar.html">foo</a> (example: bla)`

Changing `[foo]` to `[foo][]` explicitly marks the link as a
shortcut reference, fixing the issue, so I originally proposed
that we enforce such a style using Markdownlint. Unfortunately,
Markdownlint has no such rule, so I [raised a proposal][proposal]
with them.

Talking it through with @kevindew, we realised we could swap out
the markdown parsing library instead. Kramdown seems the obvious
choice, given that is [the library we use for GovSpeak][govspeak],
and it can be swapped in relatively easily.

The only modification I had to make was to explicitly state we're
using GitHub flavoured markdown ("GFM"), otherwise newlines inside
code blocks were not being correctly rendered. Credit to our very
own @bevanloon for that [tip][bevan].

Trello: https://trello.com/c/XSgGyudd/181-fix-url-markdown-parsing

[issue]: https://github.com/vmg/redcarpet/issues/471
[proposal]: https://github.com/markdownlint/markdownlint/issues/351
[govspeak]: https://github.com/alphagov/govspeak/blob/e19f80810c717c04cccb6e42d7f6df074a142330/govspeak.gemspec#L36
[bevan]: https://www.intenseagile.com/how-to-configure-middleman-to-use-backticks-for-fenced-code-blocks/

### DIFF
--- a/app/string_to_id.rb
+++ b/app/string_to_id.rb
@@ -1,9 +1,4 @@
 class StringToId
-  # Redcarpet uses a different algo to create fragment ids than github
-  # which has caused a TOC bug // ref https://trello.com/c/Re6fSBKj/24-change-internal-links-to-work-or-remove-internal-links
-  # this implementation modified for our purposes from version at jch/html-pipeline
-  # https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/toc_filter.rb
-  # We now use Kramdown instead of Redcarpet, but if it ain't broke, don't fix it.
   def self.convert(string)
     string
       .downcase # lower case
--- a/app/string_to_id.rb
+++ b/app/string_to_id.rb
@@ -3,6 +3,7 @@ class StringToId
   # which has caused a TOC bug // ref https://trello.com/c/Re6fSBKj/24-change-internal-links-to-work-or-remove-internal-links
   # this implementation modified for our purposes from version at jch/html-pipeline
   # https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/toc_filter.rb
+  # We now use Kramdown instead of Redcarpet, but if it ain't broke, don't fix it.
   def self.convert(string)
     string
       .downcase # lower case
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,5 +19,5 @@ updates:
         dependency-type: direct
       - dependency-name: octokit
         dependency-type: direct
-      - dependency-name: redcarpet
+      - dependency-name: kramdown
         dependency-type: direct
--- a/Gemfile
+++ b/Gemfile
@@ -17,8 +17,8 @@ gem "middleman-search_engine_sitemap", "~> 1.4"
 
 gem "github-markdown"
 gem "html-pipeline"
+gem "kramdown", "~> 2.3.0"
 gem "mdl", "~> 0.9.0"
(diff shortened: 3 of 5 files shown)


app/string_to_id.rb
```diff
@@ -3,6 +3,7 @@ class StringToId
   # which has caused a TOC bug // ref https://trello.com/c/Re6fSBKj/24-change-internal-links-to-work-or-remove-internal-links
   # this implementation modified for our purposes from version at jch/html-pipeline
   # https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/toc_filter.rb
+  # We now use Kramdown instead of Redcarpet, but if it ain't broke, don't fix it.
```
benthorner, line 6: But...can we fix it?

benthorner (commented):
Looks like a surprisingly simple change. Have you tried it out as a temporary deployment?

app/string_to_id.rb
ChrisBAshton: My comment was a bit facetious, I should have explained the value in leaving this alone 😅 Have now removed the comment in a separate commit which explains why things have moved on since this comment.

kevindew:
Oh exciting, glad to see that idea seems to have worked out.

Should we (or did we once have) Heroku deployments set-up for PRs?

ChrisBAshton:
@benthorner I haven't, but all looks good locally. I'm confident enough that there should be nothing dodgy about the deployment process, and we can revert the PR if it came to that. Would appreciate someone else having a look at the built output though, in case there's anything different about the rendering that I've missed 😀

ChrisBAshton:
@kevindew I believe the docs used to be hosted on Heroku - I'm not sure we ever had Heroku builds per PR though.
We also removed rack (needed for Heroku) in 1f10050e2fe822e261b30bb1d3e055fea3a7af20 in June.

I have a "Add Heroku integration" proposal on the [improve dev docs Epic](https://trello.com/c/IFR1BTCD/138-improve-dev-doc-usability), I do think it would be a good idea one day 👍

bevanloon:
💁 🙇

app/string_to_id.rb
benthorner: Thanks! I think it's out-of-scope, but worth noting this class isn't tested, so it may be even less clear now why it's necessary. As a follow-up, it may help future devs to have a couple of tests that showcase the specific use cases for this method that we don't get from the library itself.

ChrisBAshton: It does have [some tests](https://github.com/alphagov/govuk-developer-docs/commit/77f176ff84f4059af86d75ae3b09ade622f46cdf#diff-1b36fb860ebf4c903d116690329a1a8eR41-R44), but I agree it's tested indirectly and needs a bit of digging to find!

benthorner (approved):
Sure, that's fine with me. I've tried it out locally and all the pages I visited seem to work OK. No console errors either. I'm confident it's good to go.

PR #2791, opened 2020-09-24
Don't hard-wrap newlines when rendering markdown

ChrisBAshton:
In #2785 we changed the parsing engine from Redcarpet to
Kramdown, but this slipped through. Newlines were now being
treated as explicit newlines (`<br>`).

Many of our markdown files split sentences across multiple lines
for clarity in diffs, so we don't want to actually render them
across multiple lines.

Before:

> Screenshot 2020-09-24 at 18 36 48

After:

> Screenshot 2020-09-24 at 18 36 40

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -14,7 +14,8 @@ set :markdown,
     fenced_code_blocks: true,
     input: "GFM",
     tables: true,
-    no_intra_emphasis: true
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Nice one 🎉

PR #2796, opened 2020-09-29
Update local frontend development docs

injms:
- Bump review date
- Example changed from `frontend` to `government-frontend` to be less confusing about "frontend apps" vs "the frontend app"
- Docker guidance - the `bundle install` command should be run inside the Docker container
- Added comment showing the `govuk-docker-up` shortcut in example
- Removed reference to `vim` because different people use different tools

### DIFF
--- a/source/manual/local-frontend-development.html.md
+++ b/source/manual/local-frontend-development.html.md
@@ -5,16 +5,16 @@ parent: "/manual.html"
 layout: manual_layout
 section: Frontend
 type: learn
-last_reviewed_on: 2020-01-09
+last_reviewed_on: 2020-09-29
 review_in: 6 months
 ---
 
 ## Using startup scripts
 
-If you are making changes to a frontend app and nothing else, you can view these changes by running the application `./startup.sh` script:
+If you are making changes to a frontend app and nothing else, you can view these changes by running the application's `./startup.sh` script. This example is for [government-frontend], but these instructions apply to any frontend app.
 
 ```shell
-cd /govuk/frontend
+cd /govuk/government-frontend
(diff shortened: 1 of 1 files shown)


source/manual/local-frontend-development.html.md
```diff
@@ -5,16 +5,16 @@ parent: "/manual.html"
...
 type: learn
-last_reviewed_on: 2020-01-09
+last_reviewed_on: 2020-09-29
 review_in: 6 months
 ---
 
 ## Using startup scripts
 
-If you are making changes to a frontend app and nothing else, you can view these changes by running the application `./startup.sh` script:
+If you are making changes to a frontend app and nothing else, you can view these changes by running the application's `./startup.sh` script. This example is for [government-frontend], but these instructions apply to any frontend app.
```
andysellick, line 14: Is `[government-frontend]` supposed to be a link? (or does that just work magically somehow and I don't know about that?)

andysellick (approved):
Looks sensible, one question.

source/manual/local-frontend-development.html.md
injms: It magically works somehow :)

There's a list of links at the bottom of the page that `[government-frontend]` references kind of like footnotes - for example:

```markdown
[government-frontend]: https://github.com/alphagov/government-frontend
```

You have to look at the [raw unrendered file](https://raw.githubusercontent.com/alphagov/govuk-developer-docs/master/source/manual/local-frontend-development.html.md) to see it.

You can also go 

```markdown
[Government Frontend][government-frontend]
```

to set the text in the link.

PR #2802, opened 2020-10-06
Combine and rewrite unprocessed work docs

benthorner:
https://trello.com/c/seBH6Dhz/470-update-warning-critical-digest-runs-check

Please see the commits for more details.

- TODO: Update alerts in Puppet to point to the new manual


source/manual/alerts/email-alert-api-unprocessed-work.html.md
```diff
@@ -0,0 +1,57 @@
...
+title: 'Email Alert API: Unprocessed work'
+section: Icinga alerts
+subsection: Email alerts
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2020-10-05
+review_in: 6 months
+---
+
+> Sometimes we lose work due to [a flaw with the Sidekiq queueing system](https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting#my-sidekiq-process-is-crashing-what-do-i-do). At 30 minute intervals a job, [RecoverLostJobsWorker], will try to requeue work that has not been processed [within an hour](https://github.com/alphagov/email-alert-api/blob/2f3931ac1ca25fe8c79b2405af98d1de55e1d47b/app/workers/recover_lost_jobs_worker/unprocessed_check.rb#L13).
```
kevindew, line 12: This feels a bit strange to be a blockquote?

source/manual/alerts/email-alert-api-unprocessed-work.html.md
```diff
@@ -0,0 +1,57 @@
...
+
+Things to check:
+
+* Check [Sentry] for errors.
+
+* Check the [Sidekiq dashboard] for worker failures.
+
+* Check [Kibana] for errors - use ```@fields.worker: <worker class>``` for the query.
+
+* Manually retry any pending jobs from a console: `Sidekiq::RetrySet.new.retry_all`.
```
kevindew, line 34: This feels inconsistent with the list being an action rather than something to check. It'd feel more appropriate to suggest they check the [Sidekiq monitoring](https://docs.publishing.service.gov.uk/manual/sidekiq.html#sidekiq-web-aka-sidekiq-monitoring) to see if there are large queues or the item is in the retry set.

It also feels a potentially rather big action doing the whole retry set.

source/manual/alerts/email-alert-api-unprocessed-work.html.md
```diff
@@ -0,0 +1,57 @@
...
+
+> A digest run may be "complete" - all work items generated, all work items processed - but not marked as such. In this case, you will need to use slightly different commands to investigate the incomplete run:
+>
+> ```ruby
+> # find which digests are "incomplete"
+> DigestRun.where("created_at < ?", 1.hour.ago).where(completed_at: nil)
+>
+> # try manually marking it as complete
+> DigestRunCompletionMarkerWorker.new.perform
+> ```
```
kevindew, line 48: I'm not too sure how helpful this aside is - as this is just following the same step as given above to try `new.perform`. 

I'm figuring that you might instead be wanting to reference that `DigestRunCompletionMarkerWorker.new.perform` can successfully perform but not clear the alert as there could be DigestRunSubscribers that aren't processed? 

In general I don't think it's actually needed as all of the workers have nuances to learn if you're trying to debug the alert and we're not covering all of them.

kevindew (commented):
Looks good, just a couple of minor comments.

source/manual/alerts/email-alert-api-unprocessed-work.html.md
benthorner: After re-reading, I'm inclined to agree. I've moved and rewritten this paragraph to go after the list of alerts.

source/manual/alerts/email-alert-api-unprocessed-work.html.md
benthorner: This is something I found useful when attempting to fix an issue with Content Data API, where the backoff had grown large, and I wanted more immediate feedback. On reflection, I think it was pretty niche, and the main scenarios are:

- Broken code that we fix pretty quickly
- A transient error that goes away quickly

In both scenarios, the retry schedule should mean the alert goes away by itself, and we don't need to manually flush the retry set.

I've removed this suggestion (the one above covers the monitoring part).

source/manual/alerts/email-alert-api-unprocessed-work.html.md
benthorner: The re-running part is indeed the same. However, the previous paragraph was specific to manually trying jobs that are covered as part of automatic recovery, which doesn't include this particular worker.

The intro paragraph explains that the only scenario where this worker is important is when the marking itself is broken for some reason (but the work has all been done). I thought it helpful to indicate the code to investigate further.

> I'm figuring that you might instead be wanting to reference that DigestRunCompletionMarkerWorker.new.perform can successfully perform but not clear the alert as there could be DigestRunSubscribers that aren't processed?

That should not be the case, as the preceding guidance is inviting the reader to investigate this, since it's covered by the recovery worker.

Perhaps the distinction is too subtle?

benthorner (commented):
Thanks @kevindew! I found your comments thought provoking, and have made some structural and textual changes to try and make the manual a bit clearer.

Could you take another look?

source/manual/alerts/email-alert-api-unprocessed-work.html.md
```diff
@@ -0,0 +1,67 @@
...
+owner_slack: "#govuk-2ndline"
+title: 'Email Alert API: Unprocessed work'
+section: Icinga alerts
+subsection: Email alerts
+layout: manual_layout
+parent: "/manual.html"
+last_reviewed_on: 2020-10-05
+review_in: 6 months
+---
+
```
kevindew, line 11: nit, surplus new line

source/manual/alerts/email-alert-api-unprocessed-work.html.md
kevindew: Ah sure, that makes sense. Hopefully we're getting rid of the thing soon anyway and can then remove the info.

kevindew (approved):

PR #2807, opened 2020-10-12
Bump review date on browser support

alex-ju:
Nothing changed; content is still accurate.

### DIFF
--- a/source/manual/browser-support.html.md
+++ b/source/manual/browser-support.html.md
@@ -5,7 +5,7 @@ parent: "/manual.html"
 layout: manual_layout
 section: Frontend
 type: learn
-last_reviewed_on: 2019-10-08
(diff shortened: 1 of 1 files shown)


andysellick (approved):

source/manual/browser-support.html.md
```diff
@@ -5,7 +5,7 @@ parent: "/manual.html"
 layout: manual_layout
 section: Frontend
 type: learn
-last_reviewed_on: 2019-10-08
+last_reviewed_on: 2019-10-12
```
alex-ju, line 8: Only if I would've updated the year too 🤦

andysellick: Aw wow I definitely should have spotted that too 🤦

PR #2821, opened 2020-10-15
Bump date of the mongodb backups pages

fredericfran-gds:
Nothing has changed yet since Performance Platform is still in
Carrenza Production and uses some of the backup methods mentioned on
that page.

We expect performance platformed to be moved in November 2020 so
therefore, we set the next review date to be in 3 months rather than the
usual 6.

Future Work:
Once Performance Platform is moved from Carrenza, we can delete the
first 2 backup methods on that page and keep only the `mongodumps via
govuk_env_sync in AWS`

Refs:
1. [trello
card](https://trello.com/c/XpeVz301/3891-bump-date-of-mongobackups-in-developer-docs)

### DIFF
--- a/source/manual/mongodb.html.md
+++ b/source/manual/mongodb.html.md
@@ -6,7 +6,7 @@ parent: "/manual.html"
 section: Backups
 type: learn
 last_reviewed_on: 2020-10-15
-review_in: 6 months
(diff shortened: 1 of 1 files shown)


fredericfran-gds:
no longer needed....

PR #2830, opened 2020-10-22
Document two more Rails app conventions

kevindew:
This records two conventions for GOV.UK Rails applications that don't seem to be previously documented.

The first of these is to configure Rails applications to use London as the time zone. This seems to be a long standing convention that isn't consistently applied to apps (we hit some problems with [co-ordinating the local Coronavirus restrictions](https://github.com/alphagov/collections/pull/1985/commits/35e3cb4850ea72b43f64dd5b207d4f3c4369ad74) due to the lack of this in Collections). I'm taking this opportunity to define what our collective convention is to resolve ambiguities in future.

The second is the approach to organise Gemfiles with a distinct preference to not pin Gems to versions unless there is a clear documented reason to do so. This convention is quite new, it was recently trialled in [Email Alert API](https://github.com/alphagov/email-alert-api/commit/4630e8eeaf618d0111f8244ef2d618240264f5af#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55f) and, thanks to Unboxed, has now been conventionally applied across most GOV.UK apps. So I'm taking this opportunity to document this and the why.

There are further details in the individual commits.

### DIFF
--- a/source/manual/conventions-for-rails-applications.html.md
+++ b/source/manual/conventions-for-rails-applications.html.md
@@ -5,7 +5,7 @@ section: Applications
 type: learn
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2020-05-26
+last_reviewed_on: 2020-10-22
 review_in: 12 months
 ---
 
@@ -386,7 +386,7 @@ different abstractions (unit, integration and functional)][cp-testing-strategy].
 
 When testing Rails applications from the perspective of an end user it is
 conventional to use RSpec Rails' [feature specs][] (new applications should
-use the more modern equivalent [system specs][]) via [govuk_test][]'s
+use the more modern equivalent: [system specs][]) via [govuk_test][]'s
 [Capybara](https://teamcapybara.github.io/capybara/) configuration.
 
 GOV.UK have adopted the [Future Learn readable feature test][future-learn]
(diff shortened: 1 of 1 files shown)


source/manual/conventions-for-rails-applications.html.md
```diff
@@ -93,6 +93,44 @@ introduced by these gems.
...
+easier read - [example][email-alert-api-gemfile].
+
+Should you have a need to specify a particular version of the gem (for example,
+to indicate lack of compatibility with a newer release) leave a comment to
+explain the particular version. This documents why we need to care
+about the gem version. For example:
+
+```
+# We need a 6.x release to interface with Elasticsearch 6
+gem "elasticsearch", "~> 6"
```
thomasleese, line 128: A very minor/picky comment, that I don't feel super strongly on but anyway...

```suggestion
gem "elasticsearch", "~> 6" # We need a 6.x release to interface with Elasticsearch 6
```


I've experienced times in the past where it's easy to miss a comment like this if it's not on the line itself. When Dependabot opens a PR for example, the diff highlights the line itself and the comment is very obvious, whereas if it's on the line above or below it's much easier to miss [(example diff)](https://github.com/alphagov/datagovuk_publish/pull/1013/files).

kevindew: Sure thanks, does seems better on the line if we can phrase it concisely enough but can't be a constant rule as we'd not want a single line for something. I'll see if I can amend the text to not be specific on placement.

bilbof (approved):

barrucadu (approved):
LGTM

PR #2832, opened 2020-10-23
Change ownership to Notifications team

nicholsj:
Makes the Notifications team the owners of some email-related pages - they have the most current working knowledge of these things, so are the best team to ensure the docs are accurate.

[Trello card](https://trello.com/c/ardPpkNW/2191-transfer-ownership-of-some-dev-docs-to-notifications-team)

### DIFF
--- a/source/manual/bulk-email.html.md
+++ b/source/manual/bulk-email.html.md
@@ -1,5 +1,5 @@
 ---
-owner_slack: "#govuk-developers"
+owner_slack: "#govuk-notifications"
 title: Send a bulk email
 section: Emails
 layout: manual_layout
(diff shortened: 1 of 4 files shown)


benthorner (approved):
Definitely good to reduce the extensive list we get in `#govuk-developers` every day. If we're moving these for the purpose of review, then I'm not sure the Notifications team is the best place for some of them:

- General manuals make sense - the one about how they work 👍.
- Support manuals could be better reviewed by 2ndline (who use them).

I guess there's a counter-argument that the notifications team has more time to review the manuals thoroughly. So let's try it and see if it makes sense in practice!

PR #2835, opened 2020-10-28
Revert "Add 'last updated' banner to internal and externally version controlled docs"

ChrisBAshton:
Reverts alphagov/govuk-developer-docs#2784

See https://github.com/alphagov/govuk-developer-docs/pull/2784#issuecomment-717786517


benthorner (approved):
😢

Thanks for all your work on this so far! I'm sure we'll find a solution to the bug.

PR #2836, opened 2020-10-28
Add 'last updated' banner to internal and externally version controlled docs (NEW)

ChrisBAshton:
This PR is identical to #2784 other than one final commit which fixes a [deployment issue we were seeing](https://github.com/alphagov/govuk-developer-docs/pull/2784#issuecomment-717786517), and caused us to have to [revert](https://github.com/alphagov/govuk-developer-docs/pull/2835).

---

This PR removes the old "last reviewed on" banner from the bottom of the internally hosted docs. In their place - and as a new feature for the externally hosted docs - there's now a "Last updated" banner at the top of each doc. The date is clickable and links to the commit SHA on GitHub so that the changes can be viewed easily.

"Last updated" on internal docs:

Screenshot 2020-10-12 at 18 42 30

"Last updated" on external docs:

Screenshot 2020-10-12 at 18 42 39

This is phase 1 of removing Daniel the Manual Spaniel. Once the 'last reviewed' date is no longer surfaced visually, there's no point in keeping it - we'll remove the 'last_reviewed' and 'review_in' properties from files and discontinue the spaniel.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Clone govuk-content-schemas
         uses: actions/checkout@v2
         with:
@@ -41,6 +43,8 @@ jobs:
             echo "::set-output name=commit_message::Deploy via merge"
           fi
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Clone govuk-content-schemas
         uses: actions/checkout@v2
         with:
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,7 @@ gem "govuk_tech_docs"
 gem "middleman"
 gem "middleman-search_engine_sitemap"
 
+gem "git"
 gem "github-markdown"
 gem "html-pipeline"
 gem "kramdown"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -65,6 +65,8 @@ GEM
     fast_blank (1.0.0)
     fastimage (2.2.0)
     ffi (1.12.0)
+    git (1.7.0)
+      rchardet (~> 1.8)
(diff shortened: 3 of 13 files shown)


ChrisBAshton:
I've pushed up a branch to test this, by logging out the timestamps vs paths:

https://github.com/alphagov/govuk-developer-docs/runs/1320989727

Sample:

```
TIMESTAMP FOR source/manual/query-cdn-logs.html.md IS 2020-10-06 13:48:51 +0000
TIMESTAMP FOR source/manual/add-authentication-to-an-application.html.md IS 2020-06-10 10:09:53 +0000
```

This appears to be working 🎉

.github/workflows/ci.yml
```diff
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
benthorner, line 13: Do we need it when running tests? Similarly for the token, now I come to think of it.

ChrisBAshton: We do need the token, as we run a full build as part of the test suite (to catch things like unparseable markdown, which should error the build).

We don't strictly need the `fetch-depth: 0` for the tests but I'm wary of the test and build environments diverging (and causing exactly the kind of bug this commit is trying to fix!).

benthorner: Ah, true. I think it's OK to digress here: we do benefit from speeding up the build, and it's highly unlikely that we'll be writing a test that depends on previous commits (since a future commit may always supersede them!).

benthorner (approved):
Nice one (again) tracking down the bug 💯.

(I'm assuming nothing else has changed.)

.github/workflows/ci.yml
ChrisBAshton: Agreed the risk is low 👍 I may raise a follow-up PR at some point, concentrating on bringing the build time down more generally.

PR #2837, opened 2020-10-29
Remove old review mechanism (and therefore Daniel the Manual Spaniel)

ChrisBAshton:
## What
This PR removes the `last_reviewed_on` and `review_in` fields from the documentation, as they are no longer displayed in the frontend as of #2836. Instead, our 'freshness' indicator is a "Last updated" banner, displayed prominently at the top of each document. 

As a result of this change, **Daniel the Manual Spaniel will no longer post messages to Slack** for GOV.UK developers. The spaniel will continue to work for other departments in GDS who may use the tool more effectively. For GOV.UK, the 'spaniel' approach to keeping documentation up to date, while well intentioned, was not working well.

I made the changes using a script, to preserve the previous commit timestamp of each file, as I didn't want everything to suddenly have a "Last updated" value of 29th October. The script is embedded below, for posterity. 

<details>
<summary>Script for committing changes whilst keeping the old commit timestamp</summary>

```ruby
require "git"
require "date"

hook = Git.open(".")

commit_message = <<~EOS
Remove last_reviewed_on/review_in fields

Whilst this change happened on 29th October 2020, the commit date has been
backdated so that the 'last_updated' behaviour is unaffected. We don't want
to give a misleading impression that this document is more up to date than
it is because of the side-effect of removing some of its frontmatter.
EOS

# `changed_files` was a hardcoded `%w[source/...etc]` array, omitted for brevity
changed_files.each_with_index do |file, index|
  date_of_prev_commit = hook.log.path(file).first.date.strftime("%a %b %d %H:%M %Y %z")
  command = "git add #{file} && git commit -m \"#{commit_message}\" --date=\"#{date_of_prev_commit}\""
  puts command
  `#{command}`
end
```
</details>

I've tested this locally to make sure the 'last updated' value still reflects the most recent 'real' change:

> Screenshot 2020-10-29 at 10 26 29

## Why
Daniel the Manual Spaniel's purpose is to make sure we regularly review and keep our documentation up to date. In practice:

- Many pages go unreviewed long past their "review by" date. This is down to a number of reasons:
  - lack of ownership (in the case of docs attributed to the generic #govuk-developers)
  - lack of knowledge on the subject matter (a lot of our docs are very niche and therefore difficult to review)
  - frequency of review on certain pages. Our 'review_in' values are picked quite arbitrarily, so many docs are probably reviewed too often; others, not enough.
- When pages do eventually get re-reviewed, the quality of the reviewing ranges enormously. Many developers simply bump the review date; it's hard to believe they've thoroughly examined the documentation and ensured it's all fully working and up to date.

People are 'reviewing' docs because there's an annoying Slack message every day, not because the doc needs updating.

We had an open meeting on Thursday 17th September (attended by Chris A, Kevin, Ben, Deborah, and with notes from Bruce) to discuss the effectiveness of our documentation reviewing workflow, and concluded that a more 'self-organising' approach made sense. A "Last updated" banner was decided on as an improvement to the current tool.

With the new banner, it is up to the developer to determine how much they trust the documentation they are reading, based on the subject matter and the last updated value, much like people are cautious when they see a 'This article is more than 3 months old' notice on various news sites. Developers will still be empowered to update the 'last updated' value by editing the reviewing and editing the document.

The new banner works for both docs hosted inside govuk-developer-docs _and_ docs pulled in from other repositories. This allows us to localise our documentation to their respective repositories (i.e. closer to the source code), which is hugely beneficial particularly in cases such as documenting rake tasks. We anticipate auto-importing all `docs/` folders into govuk-developer-docs in the near future.

### DIFF
--- a/source/manual/review-page.html.md
+++ b/source/manual/review-page.html.md
@@ -4,8 +4,6 @@ title: Review a page in this manual
 section: Documentation
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2020-01-23
-review_in: 12 months
 ---
 
 We have [a system to make sure we regularly review pages](https://github.com/alphagov/tech-docs-monitor) in the manual.
--- a/source/manual/review-page.html.md
+++ b/source/manual/review-page.html.md
@@ -6,11 +6,9 @@ layout: manual_layout
 parent: "/manual.html"
 ---
 
-We have [a system to make sure we regularly review pages](https://github.com/alphagov/tech-docs-monitor) in the manual.
+Every page in the manual has a "Last updated" value, based on the timestamp of the last commit associated with the file. The more recent the date, the more confidence the reader can have that the content is accurate.
 
-Main pages in the manual are reviewed every 3-12 months.
-
-We also document Icinga alerts. These explain what to do when a certain alert triggers ([example](/manual/alerts/fastly-error-rate.html)). Because these can be very rare and hard to review, we should review these around the same time every six months. This is best done by people with a lot of 2nd line context. To make sure we've documented all alerts, you can [run `rake lint_alert_docs` in govuk-puppet](https://github.com/alphagov/govuk-puppet/blob/master/lib/tasks/lint_alert_docs.rake), which will output alerts that can be removed/need to be added.
+It is helpful to review a document at the point of using it, as you likely have the most context for the problem at that moment in time.
 
 ## How to review
 
@@ -23,23 +21,14 @@ We also document Icinga alerts. These explain what to do when a certain alert tr
 
 ### When you've reviewed the page, either:
 
-- Update the page
+- Update the page by committing your edits
+  - You can do this by scrolling to the bottom of the page and clicking the "View source" link to take you to the relevant page in GitHub
 - Remove it (don’t forget to [set up a redirect][redirects])
-- Assign it for someone else to review
-- Confirm the page is OK and set a new review date
-
-These tasks can be achieved by editing the metadata at the top of the source for that page. See the [govuk-developer-docs github repo](https://github.com/alphagov/govuk-developer-docs).
-
-### Choose a review by date
-
-- How likely is it that the information will change?
-- Is work happening in this area?
-- What would the impact be if the information were wrong?
 
 ### Relevant blog posts to link to
 
 Is there a relevant blog post to the page you're reviewing? Would linking to it help the reader complete their task?
 
-You can find [GOV.UK blog posts on the  GDS Tech blog](https://gdstechnology.blog.gov.uk/category/gov-uk/).
+You can find [GOV.UK blog posts on the GDS Tech blog](https://gdstechnology.blog.gov.uk/category/gov-uk/).
 
 [redirects]: https://github.com/alphagov/govuk-developer-docs/blob/master/config/tech-docs.yml
--- a/helpers/commit_helpers.rb
+++ b/helpers/commit_helpers.rb
@@ -13,7 +13,7 @@ module CommitHelpers
     timestamp = if current_page.data.latest_commit
                   current_page.data.latest_commit[:timestamp].to_s
                 else
-                  Git.open(".").log.path(source_file(current_page)).first.date.to_s
+                  Git.open(".").log.path(source_file(current_page)).first.author_date.to_s
                 end
 
     Time.parse(timestamp).strftime("%e %b %Y").strip # e.g. "3 Sep 2020"
--- a/source/manual/provision-machines-for-data-science-research.html.md
+++ b/source/manual/provision-machines-for-data-science-research.html.md
@@ -4,8 +4,6 @@ title: Provision machines for data science research
 section: Infrastructure
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2020-03-09
-review_in: 6 months
 ---
 
 Data science research frequently requires high-powered machines for
--- a/source/manual/publish-special-routes.html.md
+++ b/source/manual/publish-special-routes.html.md
@@ -4,8 +4,6 @@ title: Publish special routes
 section: Deployment
 layout: manual_layout
 parent: "/manual.html"
-last_reviewed_on: 2020-09-23
-review_in: 6 months
 old_paths:
  - /manual/publish_special_routes.html
 ---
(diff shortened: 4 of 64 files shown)


source/manual/review-page.html.md
```diff
@@ -26,20 +24,14 @@ We also document Icinga alerts. These explain what to do when a certain alert tr
 - Update the page
 - Remove it (don’t forget to [set up a redirect][redirects])
 - Assign it for someone else to review
```
benthorner, line 26: I don't think this makes sense any more, since there's no concept of "assignment". We could say: "Ask someone who has more context to review it".

source/manual/review-page.html.md
```diff
@@ -26,20 +24,14 @@ We also document Icinga alerts. These explain what to do when a certain alert tr
...
-- Confirm the page is OK and set a new review date
-
-These tasks can be achieved by editing the metadata at the top of the source for that page. See the [govuk-developer-docs github repo](https://github.com/alphagov/govuk-developer-docs).
-
-### Choose a review by date
+- Confirm the page is OK and set a new 'last updated' date
```
benthorner, line 27: Could we be more specific here? I think we're saying either:

- Update the displayed timestamp by committing your edits
- Update the displayed timestamp with a no-op commit

source/manual/review-page.html.md
```diff
@@ -26,20 +24,14 @@ We also document Icinga alerts. These explain what to do when a certain alert tr
 - Update the page
```
benthorner, line 24: Could we add some explanation for how to do this for internal and external docs? I remember it being difficult the first time (having to find where the manual text was stored).

benthorner (commented):
Thanks for the separate discussions we've had about this @ChrisBAshton! I'm keen we make this change, although there are a few things I think we should iron out first:

- I think we should omit the new guidance around "no-op" commits, which is mentioned in the PR description and the updated review manual. This seems preemptive to me, and I'm worried the suggested approach leading inconsistency and confusion. Can we wait and see if this is a common problem people encounter? I suspect it won't be.

- The review manual is quite specific to internal docs, but this PR seems to be treating internal and external docs together. I've added a comment around that.

- The subtle "Last updated" banner is quite disconnected from the review guidance, so it may not be clear to people what the date means, and how it can change. Could we add something near to the banner that links to the review guidance?

- The new "Last updated" banner is somewhat different from the "This page hasn't been reviewed ..." comparison we're making in the PR description. Have we considered highlighting manuals that are particularly out-of-date? Although it sounds minor, having this highlighted _before_ reading the manual could really help the reader.

Overall, I'm worried we're jumping to a relatively extreme opposite of our current approach, and I'm hoping we can make a few simple adjustments so there's a bit more support in place before we proceed. Happy to chat about this if you prefer, too!

benthorner:
Oops, I forgot one other thing: since this is quite a big and subjective change to our processes, it really ought to be reviewed by more people than just me, so we're representing more than two points of view. On the other hand, we don't want a committee! How about a relatively light touch approach:

- Ask one of the other people you mention in the PR description to review it.
- Ask someone who's not been involved at all to review it.

ChrisBAshton:
Thanks for the review, @benthorner. I've updated the PR and code to not refer to no-op commits at this stage, and agree we should wait to see if the absence becomes a problem. I also am keen to get approval from several more reviewers, who I'll request a review from in the GitHub UI. 👍 

The previous "Last reviewed on" banner used to appear right at the bottom of the page, so I don't think that the new "Last updated" banner makes it any less obvious if a given document is old (if anything, it's clearer now, as the date is rendered in all circumstances).

Whilst I think there are a few low-effort things we can do to improve how the banner looks under certain conditions, I think this should be tackled in a separate PR. I'm keen to avoid making changes to the display of the "Last updated" banner here, as it could very easily descend into design-by-committee and scope creep. This PR is really about deprecating Daniel the Manual Spaniel.

source/manual/review-page.html.md
```diff
@@ -25,23 +21,14 @@ We also document Icinga alerts. These explain what to do when a certain alert tr
 
 ### When you've reviewed the page, either:
 
-- Update the page
+- Update the page by committing your edits
+  - You can do this in the GitHub UI by scrolling to the bottom of the page and clicking the "View source" link to take you to the relevant page in GitHub
```
benthorner, line 25:
```suggestion
  - You can do this by scrolling to the bottom of the page and clicking the "View source" link to take you to the relevant page in GitHub
```

benthorner (approved):
> This PR is really about deprecating Daniel the Manual Spaniel.

I think we need to be careful here. I agree that is the function of the PR, but the "why" is because we're supposed to have "an adequate replacement process".

I'm not sure that's true, since we're not making any effort to support people reviewing old docs (the banner makes no mention of this, for example). However, I can also see how I'm overreacting: we can't really get much worse than we are now 😆. 

I can empathise with wanting to get this in quickly, since it's difficult to keep up-to-date. And I know you'll continue improving it, so I'm not worried about that. We've communicated about this change several times already, and there's no clear objection.

Let's just get this in. If we're surprised by strong objections later, we can always revert.

One parting thought: we should probably have a test to _prevent_ people adding review metadata in, otherwise it may creep back in e.g. out of habit, or notions of "importance".

ChrisBAshton:
Thanks for the approval, @benthorner.

FYI, I'm working on a "improve Last Updated banner" PR here: #2842. These two PRs can be reviewed/merged independently. I'll hold on for another approval or two before I click 'merge' on this one - probably next week at this rate! 😄

deborahchua (approved):
This all looks good to me, thanks for doing this!

Maybe it's also worth mentioning this at our next Tech Fortnightly so that everyone is aware? :)

brucebolt (approved):
Thank you for raising this PR Chris.

> Many pages go unreviewed long past their "review by" date. This is down to a number of reasons:
> lack of ownership (in the case of docs attributed to the generic #govuk-developers)
> lack of knowledge on the subject matter (a lot of our docs are very niche and therefore difficult to review)
> frequency of review on certain pages. Our 'review_in' values are picked quite arbitrarily, so many docs are probably reviewed too often; others, not enough.
> When pages do eventually get re-reviewed, the quality of the reviewing ranges enormously. Many developers simply bump the review date; it's hard to believe they've thoroughly examined the documentation and ensured it's all fully working and up to date.

Reviews of documentation have definitely been a problem for some time and I agree fully with all of these issues you've listed here.

This looks like a nice solution and leaves the reader to decide whether they trust the documentation or not.

One thing that could be a later improvement is some type of "I've just used this documentation and can confirm it is still current" process that bumps this date, without the need for the regular review process.

ChrisBAshton:
Force-pushed to resolve conflicts (another reason to get this in early, given how many files it touches!)

PR #2838, opened 2020-10-29
Fix link to email-alert-api support tasks

barrucadu:


benthorner (approved):
Thanks - looks like we missed this one in our earlier doc updates 👍

PR #2839, opened 2020-10-29
Remove 'review_in' field (and therefore Daniel the Manual Spaniel)

ChrisBAshton:
## What
This PR removes the `review_in` field from the documentation, so that **Daniel the Manual Spaniel will no longer post messages to Slack** for GOV.UK developers. The spaniel will continue to work for other departments in GDS who may use the tool more effectively. For GOV.UK, the 'spaniel' approach to keeping documentation up to date, while well intentioned, was not working well.

## Why
Daniel the Manual Spaniel's purpose is to make sure we regularly review and keep our documentation up to date. In practice:

- Many pages go unreviewed long past their "review by" date. This is down to a number of reasons:
  - lack of ownership (in the case of docs attributed to the generic #govuk-developers)
  - lack of knowledge on the subject matter (a lot of our docs are very niche and therefore difficult to review)
  - frequency of review on certain pages. Our 'review_in' values are picked quite arbitrarily, so many docs are probably reviewed too often; others, not enough.
- When pages do eventually get re-reviewed, the quality of the reviewing ranges enormously. Many developers simply bump the review date; it's hard to believe they've thoroughly examined the documentation and ensured it's all fully working and up to date.

People are 'reviewing' docs because there's an annoying Slack message every day, not because the doc needs updating.

We had an open meeting on Thursday 17th September (attended by Chris A, Kevin, Ben, Deborah, and with notes from Bruce) to discuss the effectiveness of our documentation reviewing workflow, and concluded that a more 'self-organising' approach made sense. A "Last updated" banner was decided on as an improvement to the current tool.

With the new banner, it is up to the developer to determine how much they trust the documentation they are reading, based on the subject matter and the last updated value, much like people are cautious when they see a 'This article is more than 3 months old' notice on various news sites. Developers will still be able to edit the `last_reviewed_on` field to mark a document as up-to-date.

The new banner works for both docs hosted inside govuk-developer-docs _and_ docs pulled in from other repositories. This allows us to localise our documentation to their respective repositories (i.e. closer to the source code), which is hugely beneficial particularly in cases such as documenting rake tasks. We anticipate auto-importing all `docs/` folders into govuk-developer-docs in the near future.

### DIFF
--- a/helpers/commit_helpers.rb
+++ b/helpers/commit_helpers.rb
@@ -9,9 +9,10 @@ module CommitHelpers
   end
 
   def last_updated(current_page)
-    # e.g. "2020-09-03 09:53:56 UTC"
     timestamp = if current_page.data.latest_commit
                   current_page.data.latest_commit[:timestamp].to_s
+                elsif current_page.data.last_reviewed_on
+                  current_page.data.last_reviewed_on.strftime("%Y-%m-%d")
                 else
                   Git.open(".").log.path(source_file(current_page)).first.date.to_s
                 end
--- a/source/accessibility.html.md
+++ b/source/accessibility.html.md
@@ -1,7 +1,6 @@
 ---
 layout: core
 title: Accessibility statement
-review_in: 6 months
 hide_in_navigation: true
 last_reviewed_on: 2020-09-02
 ---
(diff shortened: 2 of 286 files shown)


ChrisBAshton:
Just an experiment. @benthorner and I have decided to proceed with 2837 instead.

PR #2843, opened 2020-11-02
Add documentation on how to import Mapit data

thomasleese:
The automatic process is currently not working and when we recently scaled up these machines we had to do this manual process instead.

### DIFF
--- a/source/manual/mapit-database-not-available.html.md
+++ b/source/manual/mapit-database-not-available.html.md
@@ -1,0 +1,77 @@
+---
+owner_slack: "#re-govuk"
+title: Fix Mapit database not available
+layout: manual_layout
+parent: "/manual.html"
+section: Databases
+last_reviewed_on: 2020-11-02
+review_in: 6 months
+---
+
+Each Mapit machine has its own PostgreSQL database which powers the Mapit
+application. When the machine is first created by AWS, the latest data is
+[supposed to be imported into the database][autoimport].
+
+[autoimport]: https://github.com/alphagov/govuk-puppet/pull/9657
+
+This feature is currently not working which means that new Mapit machines are
+not added to the load balancer until the data has been imported manually. There
+will also be numerous errors in Icinga related to 5xx errors and events in
+Sentry:
+
+```
+OperationalError: could not connect to server: Connection refused
+        Is the server running on host "127.0.0.1" and accepting
+        TCP/IP connections on port 5432?
+```
+
+> RE are aware that the automatic import process doesn't work.
+
+## Manual process
+
+SSH onto the machine which is experiencing the problem and run the following
+comments:
+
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
👍

deborahchua (approved):

source/manual/mapit-database-not-available.html.md
```diff
@@ -0,0 +1,83 @@
...
+will also be numerous errors in Icinga related to 5xx errors and events in
+Sentry:
+
+```
+OperationalError: could not connect to server: Connection refused
+        Is the server running on host "127.0.0.1" and accepting
+        TCP/IP connections on port 5432?
+```
+
+> RE are aware that the automatic import process doesn't work.
```
ChrisBAshton, line 28: Could you add something about whether they intend to fix it, and to what sort of timescale? Is there a Trello card we can link to?

source/manual/mapit-database-not-available.html.md
```diff
@@ -0,0 +1,83 @@
...
+   $ sudo kill -9 <pid>
+   ```
+
+1. Recreate the PostgreSQL cluster
+
+   ```sh
+   $ sudo pg_dropcluster --stop 9.3 main; sudo pg_createcluster 9.3 main
+   ```
+
+1. Run Puppet
```
ChrisBAshton, line 48: What's the advantage of running Puppet twice? (We do so at the end of this doc too).

ChrisBAshton (commented):
Just tested on integration and seems to work 👍  thanks. Just a couple of questions as I went through it:

source/manual/mapit-database-not-available.html.md
thomasleese: @rtrinque Do you have any thoughts on this above? ^

It came from the work we did to scale up the Mapit machines a couple of weeks ago.

source/manual/mapit-database-not-available.html.md
thomasleese: If I remember correctly, this Puppet run will make sure the Mapit service is up and running correctly after the import script runs. I can remove it if we don't think we need it.

ChrisBAshton: I don't think it hurts, though we may want to explain why we're running it, e.g.

```suggestion
1. Run Puppet (to ensure the Mapit service is up and running)
```

Presumably the second Puppet run isn't testing that the import worked correctly, as otherwise this first one would fail, so I assume it's only there to ensure the import hasn't interrupted the Mapit service? Sounds like the second Puppet run could be removed.

thomasleese: I think you're right that we can remove it, I've pushed up an amend which removed it.

PR #2845, opened 2020-11-02
Update SAR manual with specific steps to follow

benthorner:
https://trello.com/c/GZqSmo0H/1030-update-sar-requests-manual

Previously this was quite open-ended.

### DIFF
--- a/source/manual/subject-access-requests.html.md
+++ b/source/manual/subject-access-requests.html.md
@@ -13,24 +13,26 @@ Subject Access Requests (often referred to as a 'SAR') allow members of the
 public to get information being held about them by an organisation.
 
 2nd line might be needed to answer some of the questions in the request, and
-this will usually come in through the form of a Zendesk ticket. Once 2nd line
-has found the necessary information, the request might need to be passed to
-another team to fully complete the request, for example, Licensing.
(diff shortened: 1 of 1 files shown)


benthorner:
Note: we're still trying to work out if there are more specifics steps to follow with regard to licensing.

source/manual/subject-access-requests.html.md
```diff
@@ -13,24 +13,26 @@ Subject Access Requests (often referred to as a 'SAR') allow members of the
...
-### Others
+### Signon
 
-If any other information is requested, we should update this documentation so
-they become easier to deal with in the future.
+Check if the user [has an account](https://signon.publishing.service.gov.uk/users).
+
+### Licensing
+
+Ask in #govuk-licensing.
```
barrucadu, line 38: When we hear back about the query it would be good to put it here

barrucadu (approved):
There will be some accounts stuff to add to this page too, but it's on our list

source/manual/subject-access-requests.html.md
benthorner: Indeed. I'm going to merge this is now, as plan to raise a PR when we know more about licencing.

PR #2848, opened 2020-11-03
Clarify overall process for SAR tickets

benthorner:
https://trello.com/c/GZqSmo0H/1030-update-sar-requests-manual

### DIFF
--- a/source/manual/subject-access-requests.html.md
+++ b/source/manual/subject-access-requests.html.md
@@ -12,8 +12,14 @@ related_applications: [email-alert-api]
 Subject Access Requests (often referred to as a 'SAR') allow members of the
 public to get information being held about them by an organisation.
 
-2nd line might be needed to answer some of the questions in the request, and
-this will usually come in through the form of a Zendesk ticket.
+Requests should come through in the form of a Zendesk ticket from our policy
+team, on behalf of the GDS Privacy Team. The process goes like this:
+
+1. User raises a request with the privacy team, who then turn it into a 2ndline ticket.
+
+2. 2ndline add notes after performing relevant checks (see below), and then reassign to the blogs/campaigns team with a "low" priority (since 2ndline are no longer working on it).
+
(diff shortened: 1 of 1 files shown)


source/manual/subject-access-requests.html.md
```diff
@@ -12,8 +12,14 @@ related_applications: [email-alert-api]
...
 public to get information being held about them by an organisation.
 
-2nd line might be needed to answer some of the questions in the request, and
-this will usually come in through the form of a Zendesk ticket.
+Requests should come through in the form of a Zendesk ticket from our policy
+team, on behalf of the GDS Privacy Team. The process goes like this:
+
+1. User raises a request with the privacy team, who then turn it into a 2ndline ticket.
+
+2. 2ndline add notes after performing relevant checks (see below), and then reassign to the blogs/campaigns team (James Whitmarsh) with a "low" priority (since 2ndline are no longer working on it).
```
deborahchua, line 20: Do we want to include full names here as the docs are publicly accessible?

benthorner: Oh yeah, good point. I've updated the commit to remove the name, and instead documented it on a new "useful info" Trello card: https://trello.com/c/Vzg3Fze8/1035-subject-access-requests.

I wonder if it makes more sense to have the whole manual on the Trello card, so there's one place to look. What do you think?

deborahchua: Yeh I think that's a good idea! Or possibly our internal Wiki (I'm not sure if we still use this) 🤔

deborahchua (approved):

source/manual/subject-access-requests.html.md
benthorner: OK. I'll merge this in for now, and then look at shifting it over onto the Trello card: I don't think we have anything related to 2ndline in the Wiki, so I reckon it's better to use Trello, as long as it's practical to do so.

PR #2861, opened 2020-11-11
Various improvements to Sidekiq documentation

ChrisBAshton:
- Made clearer the distinction between Sidekiq and govuk_sidekiq
- Describe how there is one shared Redis instance between machines
- Clarify Sidekiq Web port instructions
- Add an example Sidekiq console command for finding named workers

### DIFF
--- a/source/manual/sidekiq.html.md
+++ b/source/manual/sidekiq.html.md
@@ -8,33 +8,65 @@ type: learn
 ---
 
 Many of our applications use
-[Sidekiq](https://github.com/mperham/sidekiq) for background job processing.
+[Sidekiq](https://sidekiq.org) ([see repo](https://github.com/mperham/sidekiq))
+for background job processing.
 
-There's a [GOV.UK wrapper](https://github.com/alphagov/govuk_sidekiq) that will
-help you set it up.
+## Sidekiq on GOV.UK
+
+For redundancy, we have publishing apps running on multiple machines.
+We would have all sorts of race conditions and difficulties querying
+Sidekiq if each app on each machine had its own instance of Sidekiq.
+
+Therefore, we've build a GOV.UK wrapper for Sidekiq, called
+[govuk_sidekiq](https://github.com/alphagov/govuk_sidekiq). This
+allows all Sidekiq processes to talk to a single Redis instance. It
+also enables [request tracing](/manual/setting-up-request-tracing.html)
+and sends activity stats to Statsd, which forwards data to Graphite
+to be stored.
+
+## Retry logic
+
+Sidekiq has in built retry logic (turned on by default, but configurable).
+Successes, failures, job timings and retry counts are sent to `statsd`.
+
+Jobs do fail, this is not inherently bad and can happen for a number of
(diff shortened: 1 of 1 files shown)


source/manual/sidekiq.html.md
```diff
@@ -8,33 +8,65 @@ type: learn
...
 
 ## Monitoring
 
-There are three approaches for monitoring Sidekiq, via the [Sidekiq Web interface](#sidekiq-web),
-or the [Grafana dashboard](#sidekiq-grafana-dashboard) or the [Console](#sidekiq-from-the-console).
+There are three approaches for monitoring Sidekiq:
+
+- via the [Sidekiq Web interface](#sidekiq-web-aka-sidekiq-monitoring)
+- or the [Grafana dashboard](#sidekiq-grafana-dashboard)
+- or the [Console](#sidekiq-from-the-console)
```
AlanGabbianelli, line 43:
```suggestion
- via the [Sidekiq Web interface](#sidekiq-web-aka-sidekiq-monitoring)
- via the [Grafana dashboard](#sidekiq-grafana-dashboard)
- via the [Console](#sidekiq-from-the-console)
```

source/manual/sidekiq.html.md
```diff
@@ -8,33 +8,65 @@ type: learn
...
-multiple [Sidekiq] configurations used throughout [GOV.UK].
+that can display the current state of a Sidekiq installation. We have
+configured this as the [sidekiq-monitoring web app](https://github.com/alphagov/sidekiq-monitoring)
+to monitor multiple Sidekiq configurations used throughout GOV.UK.
 
-We have restricted public access as the Web UI allows modifying the state of
-[Sidekiq] queues. To access the Sidekiq monitoring UI for Production AWS,
-run:
+We have restricted public access because the Web UI allows modifying the
+state of [Sidekiq] queues. To access the Sidekiq monitoring UI for
```
AlanGabbianelli, line 54:
```suggestion
state of Sidekiq queues. To access the Sidekiq monitoring UI for
```

ChrisBAshton: Missed this one, thanks!

source/manual/sidekiq.html.md
ChrisBAshton: Personally I prefer what we have already, as it's clearer you're only expected to choose one of the options.

AlanGabbianelli: If we say `There are three approaches for monitoring Sidekiq:`, it's clear that the list that follows is of those three approaches and that we're expected to choose one. I think the `or` works in a sentence (like it was before the change) but is unnecessary in a bullet point list.
Not a blocker though.

AlanGabbianelli (approved):

source/manual/sidekiq.html.md
ChrisBAshton: Actually on second read, I agree! I read your comment from the 'landing page' of the PR, which only shows the suggestion, and not the sentences that precedes it. Now I prefer the `or`.

Have overwritten my commit 👍

AlanGabbianelli (approved):

PR #2863, opened 2020-11-12
Update note about this alert on Staging and Integration

ChrisBAshton:
Changes following https://github.com/alphagov/govuk-puppet/pull/10842.

I had raised https://github.com/alphagov/whitehall/pull/5882 to remove the rake task from Whitehall, as it is no longer called by govuk-puppet. However, this alert still runs on Production, and if the queue length become mismatched with the number of scheduled editions, this rake task could still be useful. So for now I've updated the documentation and will close the Whitehall PR.

Trello: https://trello.com/c/RethzUSW/2088-5-fix-assetmanagerservicehelperassetnotfound-errors-on-staging-integration


AlanGabbianelli (approved):
Looks clear 👍
Non-blocking suggestion: it’d be good for the commit message to have the same context as the PR description.

PR #2865, opened 2020-11-16
Add details how to fix the Licensify VPN

fredericfran-gds:
One fix for the Licensify VPN is to switch it off and on.

In this PR, we describe the process so that people on support
know this potential fix.


edwardkerry (approved):
Thank you for this!

PR #2868, opened 2020-11-19
Remove superfluous HTML element

kevindew:
This closing element doesn't have a respective opening element so this
displays as plain text. It looks like this is a mistake where someone
forgot this file was markdown.

Before:

Screenshot 2020-11-19 at 11 12 29

After:

Screenshot 2020-11-19 at 11 13 07


ChrisBAshton (approved):
Nice catch.

PR #2869, opened 2020-11-19
Add documentation on how to redirect assets

thomasleese:
This is to match the new Rake tasks that I've added in https://github.com/alphagov/asset-manager/pull/822. I've also taken the opportunity to consolidate the various bits of asset management documentation into one page.

[Trello Card](https://trello.com/c/14zssUpz/2221-8-investigate-and-fix-whitehall-file-attachments-not-redirecting-%F0%9F%8D%90)


ChrisBAshton (approved):
Nice job ⭐

PR #2870, opened 2020-11-20
Fix relative links & images in external docs

### LINKED ISSUE #2851: Relative links on 'external documentation' don't resolve correctly on docs.publishing.service.gov.uk
Problem with '' (https://docs.publishing.service.gov.uk/apis/search-api/how-search-works.html)

Links e.g <a href="learning-to-rank.md">learning to rank documentation</a> 404

e.g https://docs.publishing.service.gov.uk/apis/search-api/learning-to-rank.md -> 404

but https://docs.publishing.service.gov.uk/apis/search-api/learning-to-rank.html -> this works!

ChrisBAshton:
Fixes #2851. Now relative links and images in imported READMEs and documents will be rewritten to either:

1. An absolute GitHub URL to the resource, or
2. If appropriate, a relative link to the Developer Docs hosted version of the resource.

The latter can be seen for example on the "How Search Works" page under Search API, where the "learning to rank documentation" link is rewritten from `./learning-to-rank.md` to `/apis/search-api/learning-to-rank.html`.

Trello: https://trello.com/c/PLV6EPXh/199-fix-relative-links-images-on-imported-docs

### DIFF
--- a/app/external_doc.rb
+++ b/app/external_doc.rb
@@ -5,7 +5,7 @@ require_relative "./string_to_id"
 class ExternalDoc
   def self.fetch(repository:, path:)
     contents = HTTP.get(
-      "https://raw.githubusercontent.com/#{repository}/master/#{path}",
+      "https://raw.githubusercontent.com/alphagov/#{repository}/master/#{path}",
     )
 
     context = {
@@ -13,11 +13,11 @@ class ExternalDoc
       gfm: false,
       base_url: URI.join(
         "https://github.com",
-        "#{repository}/blob/master/",
+        "alphagov/#{repository}/blob/master/",
       ),
       image_base_url: URI.join(
         "https://raw.githubusercontent.com",
-        "#{repository}/master/",
+        "alphagov/#{repository}/master/",
       ),
     }
 
--- a/app/external_doc.rb
+++ b/app/external_doc.rb
@@ -41,16 +41,38 @@ class ExternalDoc
       .to_html(contents.force_encoding("UTF-8"), context)
   end
 
-  def self.parse(markdown)
+  def self.parse(markdown, repository: "", path: "")
+    context = {
+      # Turn off hardbreaks as they behave different to github rendering
+      gfm: false,
+      base_url: URI.join(
+        "https://github.com",
+        "alphagov/#{repository}/blob/master/",
+      ),
+      image_base_url: URI.join(
+        "https://raw.githubusercontent.com",
+        "alphagov/#{repository}/master/",
+      ),
+    }
+
+    context[:subpage_url] =
+      URI.join(context[:base_url], File.join(".", File.dirname(path), "/"))
+
+    context[:image_subpage_url] =
+      URI.join(context[:image_base_url], File.join(".", File.dirname(path), "/"))
+
     filters = [
       HTML::Pipeline::MarkdownFilter,
+      HTML::Pipeline::AbsoluteSourceFilter,
       PrimaryHeadingFilter,
       HeadingFilter,
+      AbsoluteLinkFilter,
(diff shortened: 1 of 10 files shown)


app/external_doc.rb
```diff
@@ -5,6 +5,7 @@
 class ExternalDoc
   def self.parse(markdown, repository: "", path: "")
     context = {
+      repository: repository,
```
ChrisBAshton, line 8: Note to self: rewrite this commit message.

app/external_doc.rb
```diff
@@ -60,7 +61,7 @@ def call
         uri = URI.parse(href)
         path = uri.path
 
-        next if uri.scheme || href.start_with?("#") || path.end_with?(".md")
+        next if uri.scheme || href.start_with?("#")
```
ChrisBAshton, line 64: Note to self: rewrite this commit message. It should make reference to this change ^

spec/app/proxy_pages_spec.rb
```diff
@@ -1,7 +1,7 @@
 RSpec.describe ProxyPages do
   before do
     allow(AppDocs).to receive(:apps_with_docs)
-      .and_return([double("App with docs", app_name: "", page_title: "", description: "")])
+      .and_return([double("App with docs", app_name: "", page_title: "", description: "", github_repo_name: "")])
```
ChrisBAshton, line 4: Should consider if I can use `AppDocs.apps_with_docs` and filtering based on presence. Probably better than duplicating business logic, and means I can avoid adding this line.

spec/app/external_doc_spec.rb
```diff
@@ -33,15 +44,31 @@
...
+      end
+
+      context "the repository is set to `consume_docs_folder: true`" do
+        let(:consume_docs_folder) { true }
+
+        it "rewrites GitHub relative links to Developer Docs HTML links if link is in the `docs` folder" do
+          expect(html).to have_link("Read the docs", href: "/apis/lipsum/foo.html")
+          expect(html).to have_link("Read another doc", href: "/apis/lipsum/bar.html")
+        end
+      end
```
ChrisBAshton, line 70: These tests could be written more cleanly, with clearer contexts. Also missing explicit tests for when we're parsing something as a `docs` (rather than as root).

source/templates/external_doc_template.html.erb
```diff
@@ -4,4 +4,4 @@ parent: /apis.html
 hide_from_sitemap: true
 ---
 
-<%= ExternalDoc.parse(markdown) %>
+<%= ExternalDoc.parse(markdown, repository: repository, path: "docs/placeholder") %>
```
ChrisBAshton, line 7: Rather than this code smell, could perhaps just pass `page[:path]`.

app/external_doc.rb
```diff
@@ -3,44 +3,6 @@
 require_relative "./string_to_id"
 
 class ExternalDoc
-  def self.fetch(repository:, path:)
```
thomasleese, line 6: I think it might still be useful to keep this method around because it is much more convenient than doing `HTTP.get` in all places, but I don't feel too strongly about it.

thomasleese (approved):

app/external_doc.rb
ChrisBAshton: Thanks for the review, Tom. I don't feel hugely strongly about it either, but it's only being used [in one place at the moment](https://github.com/alphagov/govuk-developer-docs/blob/master/source%2Fmanual%2Fupdating-brexit-transition-checker.html.md.erb), and that will be removed shortly when we enable pulling in all repos' `docs/` folders anyway. So I suggest we leave it as removed, and add it back in later if there is a need.

PR #2874, opened 2020-11-24
Add instructions for reusing a redirect's route

injms:
Added documentation to explain how to remove a redirect and how to repurpose the route so a document can be published at that URL.

### DIFF
--- a/source/manual/redirect-routes.html.md.erb
+++ b/source/manual/redirect-routes.html.md.erb
@@ -58,6 +58,86 @@ and using non-default values for the `segments_mode`.
 [short-url-manager]: https://short-url-manager.publishing.service.gov.uk
 [short-url-manager-permissions]: https://github.com/alphagov/short-url-manager/#permissions
 
+### Removing a route created in the Short URL Manager
+
+First find the `content_id` of the URL that needs to be removed. This can be found by using the Content Store API: `www.gov.uk/api/content/<path goes here>`.
+
+For example, finding the `content_id` for the page about [keeping a pet pig or micropig](https://www.gov.uk/guidance/keeping-a-pet-pig-or-micropig) is found by going to [`www.gov.uk/api/content/guidance/keeping-a-pet-pig-or-micropig`](https://www.gov.uk/api/content/guidance/keeping-a-pet-pig-or-micropig). The `content_id` should be a UUID; for example: `602144f8-7631-11e4-a3cb-005056011aef`.
+
+The content_id is with a rake task that removes the shortlink. The [Short URL Manager documentation](https://docs.publishing.service.gov.uk/apps/short-url-manager.html#run-a-rake-task) has handy links for running the rake tasks in the different environments.
+
+The rake task for removing a redirect is:
+
+```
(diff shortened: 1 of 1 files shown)


source/manual/redirect-routes.html.md
```diff
@@ -58,6 +58,82 @@ and using non-default values for the `segments_mode`.
...
+First find the `content_id` of the URL that needs to be removed. This can be found by using the Content Store API: `www.gov.uk/api/content/<path goes here>`.
+
+For example, finding the `content_id` for the page about [keeping a pet pig or micropig](https://www.gov.uk/guidance/keeping-a-pet-pig-or-micropig) is found by going to [`www.gov.uk/api/content/guidance/keeping-a-pet-pig-or-micropig`](https://www.gov.uk/api/content/guidance/keeping-a-pet-pig-or-micropig). The `content_id` should be a UUID; for example: `602144f8-7631-11e4-a3cb-005056011aef`.
+
+The content_id is with a rake task that removes the shortlink. The [Short URL Manager documentation](https://docs.publishing.service.gov.uk/apps/short-url-manager.html#run-a-rake-task) has handy links for running the rake tasks in the different environments.
+
+The rake task for removing a redirect is:
+
+```
+redirect:destroy["<uuid goes here>"]
```
brucebolt, line 72: The dev docs now have a helper that will produce links to the 'Run rake task' job in Jenkins for all the environments.  It might be easier for those reading the documentation to have a link they can click?
```suggestion
<%= RunRakeTask.links("asset-manager", "redirect:destroy[<uuid goes here>]") %>```

File: source/manual/redirect-routes.html.md [hunk 2]
```diff
@@ -58,6 +58,82 @@ and using non-default values for the `segments_mode`.
...
+PathReservation.find_by(base_path: "/example-path").delete
+```
+
+In content-store _and_ the content-store-draft, remove the content item that used in inhabit the URL:
+
+```ruby
+ContentItem.find_by(base_path: "/example-path").delete
+```
+
+In `publisher`, update and republish the content that should now sit at `/example-path`.
```
brucebolt, line 123: This would only be for documents published in Mainstream Publisher, so it's probably worth noting here that the remaining instructions will vary depending on the publishing app that the new page is using.

brucebolt (commented):
Looks great 👍 

I've just added a couple of comments: there's a helper we can use to create links to the rake tasks and a clarification on the final steps.

source/manual/redirect-routes.html.md
injms: That is a nifty helper - thank you for sharing. I've added it in (and also discovered that it really doesn't like `<uuid goes here>` as that then closes one of the HTML tags...)

source/manual/redirect-routes.html.md
injms: Good point - I've hopefully clarified that this is for Mainstream Publisher. I don't know how to do this for other apps, so rather than just guess I've left it blank.

brucebolt (approved):

PR #2876, opened 2020-11-26
Reference potential cause for medical safety alert

benthorner:
https://trello.com/c/q3rh4ngJ/543-document-surprise-all-behaviour-for-finder-email-signups

While we are pursuing a fix for the underlying issue as tech debt,
this ensures we're aware of it in case it happens again.


kevindew (approved):
Nice one.

PR #2878, opened 2020-11-27
Improve the "Redirect a route" documentation

ChrisBAshton:
It is crucial that we know how to manually redirect routes if we run into issues. Prior to this commit, this knowledge wasn't
particularly well documented anywhere.

Rather than list all the different scenarios in which we want to redirect and making the reader scan the page for the option
relevant to them, I've attempted to provide some brief context at the top of the document, followed by a flow chart of sorts.
The first heading, "Choosing a redirect strategy", describes the thought processes we should follow before applying any redirects.

I've documented the preferred approach for manual redirects, as well as warn the reader that it won't work for Short URL Manager routes (something I discovered by accident while trying the code out!). And, in a recent incident, we were forced to apply redirects at the Router API level, which is also a useful thing to document.

Additionally, there was recently a case where a redirect had been removed, and was showing as such in the Content Store, but the redirect was still happening. It turns out the route registration had failed because of reprovisioning of the Router machines.

This same situation could easily happen in reverse: a redirect being ADDED, but failing to make its way into Router API. Rather
than semi-polluting the content item in Publishing API to force through the redirect, it's better in cases such as these to just
give Content Store a bit of a kick, as it's already in the state we want it to be in. Router API just needs to catch up.

### DIFF
--- a/source/manual/redirect-routes.html.md.erb
+++ b/source/manual/redirect-routes.html.md.erb
@@ -7,31 +7,39 @@ parent: "/manual.html"
 related_applications: [short-url-manager]
 ---
 
-There is a regular requirement for creating short URLs on behalf of
-departments, usually for use with for campaign materials. Content
-designers do this using the [Short URL Manager][short-url-manager] app.
-Read [more about Short URL Manager][short-url-manager-section] below,
-including how to remove routes originating from it.
+There are a number of different ways of applying redirects on GOV.UK.
+Here is a guide for choosing the right one:
 
-There are also processes for [redirecting HRMC manuals][hmrc-manuals-section]
-and [redirecting from campaign sites][campaign-sites-section].
+- Is the redirect a short URL, such as `gov.uk/dfe`?
+  [Use Short URL Manager][short-url-manager-section].
+- Are you redirecting a HMRC manual?
+  Read the [HMRC section][hmrc-manuals-section].
+- Are you redirecting from a campaign site?
+  Read the [campaign sites section][campaign-sites-section].
 
-Finally, sometimes there is a need to manually redirect existing URLs to
-another internal location, either due to an incident, content being
-archived or because the slug has been changed.
+That covers the service-specific redirects. We then have several options
+for applying redirects to most GOV.UK content:
 
-## Choosing a redirect strategy
+- Are you having an incident and need to apply a redirect _urgently_?
+  You can [apply the redirect directly in Router API][apply-redirect-in-router-api-section]
+  (though you'll need to do some housekeeping afterwards).
+- Is the redirect correctly showing in Content Store, but not in Router API?
+  You may just need to [nudge Content Store into updating Router API][register-redirect-from-content-store-section].
+- Is the redirect correctly showing in Publishing API, but not in Content Store?
+  Try [re-presenting the content item downstream][represent-content-item-section].
+- Does the redirect not exist anywhere yet?
+  You'll need to [apply a redirect from within the relevant publishing app][apply-redirect-from-publishing-app-section].
 
-Does something exist at the incoming URL already? There's a
-[simple way of checking][check-route-exists-section].
-
-If no route exists, and if you have the time to wait for a content
-designer to review the proposal, you should just use Short URL Manager.
-If the route already exists, or if you're in a hurry, you'll need to take
-a different approach.
+## Manually apply a redirect from any publishing app
+[apply-redirect-from-publishing-app-section]: #manually-apply-a-redirect-from-any-publishing-app
 
 The preferred method of creating redirects is to use the publishing app that
-originally published the content to set up the redirect. This usually
+originally published the content to set up the redirect. You can find out
+[what exists at the incoming URL already][check-route-exists-section] (if
+anything), and [which app owns the route][find-owning-app].
+
+Once you've established which app owns the route, you should see if it is
+possible to set up the redirect from within the app UI itself. This usually
 involves unpublishing the content, whereupon the app gives the option to
 redirect the old route to another route. This approach is preferred as it
 keeps the workflow and publishing history within the app, as well as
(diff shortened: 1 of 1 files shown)


source/manual/redirect-routes.html.md.erb
```diff
@@ -7,17 +7,131 @@ parent: "/manual.html"
...
 
-To find if there is already a redirect for a particular path, open a router-api
-console:
+## Choosing a redirect strategy
+
+Does something exist at the incoming URL already? There's a
+[simple way of checking][check-route-exists-section].
+
+If no route exists, and if you have the time to wait for a content
+designer to review the proposal, you should just use Short URL Manager.
```
thomasleese, line 29: I think we should discourage this if the route isn't what we could consider to be a "short URL", for example `gov.uk/dfe`.

source/manual/redirect-routes.html.md.erb
```diff
@@ -7,17 +7,131 @@ parent: "/manual.html"
...
+involves unpublishing the content, whereupon the app gives the option to
+redirect the old route to another route. This approach is preferred as it
+keeps the workflow and publishing history within the app, as well as
+correctly propagating the changes through Publishing API, Content Store and
+Router API.
+
+Not all applications support unpublishing / setting redirects. In these
+cases, you should change the content item in Publishing API manually.
+
+### Manually apply a redirect in Publishing API
```
thomasleese, line 44: I think this section is useful to have as it explains how things work inside the Publishing API, but I suspect the simpler option in most cases is to unpublish from the publishing app itself [but using the API directly](https://github.com/alphagov/publishing-api/blob/master/docs/api.md#post-v2contentcontent_idunpublish). For example, doing something like this:

```rb
GdsApi.publishing_api.unpublish(
  content_id,
  type: "redirect",
  explanation: "manually redirected by YOUR_NAME",
  alternative_path: "/new-path",
  discard_drafts: true
)
```

source/manual/redirect-routes.html.md.erb
```diff
@@ -7,17 +7,131 @@ parent: "/manual.html"
...
+[Content Store talks directly to Router API](https://github.com/alphagov/content-store/blob/dd79a03d74f130650bc97d1c84aae557ccea58d3/app/models/content_item.rb#L33)
+to update the route. Now when you visit the route in your
+browser, you should be redirected.
+
+### Manually register a route in Content Store
+[register-redirect-from-content-store-section]: #manually-register-a-route-in-content-store
+
+Sometimes Content Store is unsuccessful in updating a route in
+Router API, e.g. if the call from Content Store happens while
+the Router machines are being deployed.
```
thomasleese, line 102: When Content Store can't set the routes [it will actually remove the content item from its own database](https://github.com/alphagov/content-store/blob/8651b9245046f0a0ae125dde018d39f8507226c5/app/models/content_item.rb#L35), so this might not be the right solution. Instead you'll likely need to [represent it from the Publishing API](https://github.com/alphagov/publishing-api/blob/master/lib/tasks/represent_downstream.rake).

thomasleese: It might also be worth mentioning that if Content Store notices the routes on a Content Item haven't changed, [it won't send them to the Router API](https://github.com/alphagov/content-store/blob/8651b9245046f0a0ae125dde018d39f8507226c5/app/models/content_item.rb#L224-L232). This means representing something which has already been published and is in the Content Store but the routes aren't there, won't actually do anything, and the routes _will_ need to registered manually using the command you've given below.

source/manual/redirect-routes.html.md.erb
```diff
@@ -7,17 +7,131 @@ parent: "/manual.html"
...
+
+### Manually apply a redirect in Router API
+
+You might need to urgently apply a redirect, and Publishing API
+might have a large backlog of publishing in its downstream_high
+queue, making the Publishing API approach too slow.
+
+First, connect to the Router API console:
+
+`$ gds govuk connect -e production app-console draft_cache/router-api`
```
thomasleese, line 119: This will only update the route in the draft stack, which is fine for the docs, but might be easy for someone who wants to do this in the live stack to miss.

thomasleese (commented):
Thanks for putting this together, I think it'll be really helpful documentation to have. I've left a few comments where I think we could be a little bit clearer.

source/manual/redirect-routes.html.md.erb
ChrisBAshton: Good spot. I actually can't seem to connect to the app console on the live stack 🤔 

```
$gds govuk connect app-console -e integration cache/router-api

/usr/local/bin/govuk_app_console: 17: cd: can't cd to /var/apps/router-api
```

thomasleese: It's actually on a `router_backend` machine class! But I think you should be able to do `gds govuk connect app-console -e integration router-api` and it will find the appropriate backend for you.

source/manual/redirect-routes.html.md.erb
ChrisBAshton: Thanks @thomasleese - I've pushed up another commit with a few improvements based on your feedback.

source/manual/redirect-routes.html.md.erb
```diff
@@ -7,17 +7,169 @@ parent: "/manual.html"
...
+First, you'll need the content ID of the content item. Here we have
+a Publishing API console and we've retrieved the Unpublishing record
+(which has the redirect information) for the content item. We find
+its associated content ID and can then use this in one of the linked
+rake tasks.
+
+```
+unpublishing = Unpublishing.find_by(alternative_path: "/foo")
+
+Edition.find(unpublishing.edition_id).content_id
```
thomasleese, line 124:
```suggestion
unpublishing.edition.content_id
```

This might be slightly neater.

thomasleese (approved):
Looks good! I've just left one small comment.

source/manual/redirect-routes.html.md.erb
ChrisBAshton: Much better!

PR #2882, opened 2020-12-02
Fix docs on domain name requests

chao-xian:
The Zendesk queue has been renamed. Also fixed the heading to be more helpful.


richardTowers (approved):
Thanks for this @chao-xian !

PR #2885, opened 2020-12-03
Adjust Email Alert API queue latency for alert changes

kevindew:
Trello: https://trello.com/c/il1TsI3e/658-tweak-alert-latency-thresholds-for-email-alert-api-sidekiq-queues

We're changing the queue latency alerts in https://github.com/alphagov/govuk-puppet/pull/10868 this commit updates the documentation to match those better.

### DIFF
--- a/source/manual/alerts/email-alert-api-high-queue-latency.html.md
+++ b/source/manual/alerts/email-alert-api-high-queue-latency.html.md
@@ -7,36 +7,53 @@ layout: manual_layout
 parent: "/manual.html"
 ---
 
-### High queue latency (queue_latency)
+This alert triggers when there is a significant delay in the time from the
+Email Alert API system creating an email until it is sent.
 
-This means the time it takes for a job to be processed is unusually high. See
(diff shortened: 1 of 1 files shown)


benthorner (commented):
Nice one documenting the nuance of this alert, so it's clear it may not require action. 

I think it's still worth keeping a bit of the directness of the original manual, to help someone trying to debug things in a rush. How about breaking it with a couple of sections:

- One-liner intro sentence
- Interpreting the alert
- User impact
- Steps to investigate

kevindew:
> Nice one documenting the nuance of this alert, so it's clear it may not require action.
> 
> I think it's still worth keeping a bit of the directness of the original manual, to help someone trying to debug things in a rush. How about breaking it with a couple of sections:
> 
> * One-liner intro sentence
> * Interpreting the alert
> * User impact
> * Steps to investigate

Thanks I had concerns it was a bit of drag to read. I've used some headers to break it up.

PR #2889, opened 2020-12-08
Remove manual about stopping Email Alert API

benthorner:
This applies equally to any GOV.UK app or process, yet we have no
other manuals to e.g. "stop all publishing".

I don't believe this has ever been used, or that we would have cause
to stop all email. The original PR doesn't give a use case [1].

[1]: https://github.com/alphagov/govuk-developer-docs/pull/790


ChrisBAshton (approved):
I wonder if @kevindew has access to the linked Trello card? https://trello.com/c/JCuEPDye/716-add-documentation-on-how-to-stop-email-alert-api-workers-in-an-emergency. Would be good to know for sure this isn't needed.

Otherwise this PR seems reasonable to me.

benthorner:
Thanks @ChrisBAshton! I'll wait to see if @kevindew has any more insight before merging. I have access to the card...

```
# What
Add documentation to the dev docs on how to disable puppet and stop the email-alert-api workers if email sending needs to be stopped.

# Why
So that on-call people without context of the email system know what to do.
```

kevindew (approved):
Yup, I vaguely remember this. I recall it not being a reaction to a problem but instead an action taken as a pre-caution. Time wise it'll have been a couple of weeks after the service was launched so presumably correlates with the team being unexpectedly cancelled. 

I'm also confident it's never been used and is good to remove.

PR #2891, opened 2020-12-08
Deprecate startup.sh

benthorner:
**Deadline for discussion: 2020-12-16**

These are superseded by GOV.UK Docker, which is the official way we
develop GOV.UK apps [1]. Having alternative scripts in our repos is
confusing for new developers, and is just another thing we have to
maintain and help people troubleshoot [2].

[1]: https://github.com/alphagov/govuk-rfcs/blob/master/rfc-106-docker-for-local-development.md
[2]: alphagov/email-alert-frontend#914

This is as much a proposal as a PR, so feel free to leave
comments about your own thoughts on this. In particular, I'd
like to know if there are objective reasons why we still need
these scripts for some apps.

If this PR does go in, them I'm also proposing to modify the
startup.sh scripts to print a deprecation notice advising people
to use GOV.UK Docker instead, and warning the scripts may be
removed without notice.

### DIFF
--- a/source/manual/local-frontend-development.html.md
+++ b/source/manual/local-frontend-development.html.md
@@ -66,7 +66,9 @@ govuk-docker-up app-live
 # You can now view the app on government-frontend.dev.gov.uk
 ```
 
-## Using startup scripts
+## Using startup scripts (DEPRECATED)
+
+**DEPRECATED:** these scripts are superseded by GOV.UK Docker, and may no longer be present in some frontend apps. **Try GOV.UK Docker for local frontend development before using these scripts.**
 
 If you are making changes to a frontend app and nothing else, you can view these changes by running the application's `./startup.sh` script. This example is for [government-frontend], but these instructions apply to any frontend app.
 
--- a/source/manual/local-frontend-development.html.md
+++ b/source/manual/local-frontend-development.html.md
@@ -7,28 +7,6 @@ section: Frontend
 type: learn
 ---
 
-## Using startup scripts
-
-If you are making changes to a frontend app and nothing else, you can view these changes by running the application's `./startup.sh` script. This example is for [government-frontend], but these instructions apply to any frontend app.
-
-```shell
-cd /govuk/government-frontend
-./startup.sh --live
-# Check the output to see what port the app is running on, e.g: localhost:3005
-```
-
-If you want to test changes in [govuk_publishing_components] against a frontend app, you need to edit your frontend app Gemfile and then run the startup script:
-
-```ruby
-gem 'govuk_publishing_components', path: '../govuk_publishing_components'
-```
-
-```shell
-bundle install
-./startup.sh --live
-# Check the output to see what port the app is running on, e.g: localhost:3005
-```
-
 ## Using govuk-docker
 
 This assumes that you have already installed and setup [govuk-docker]. We will use [government-frontend] as an example here, but these instructions apply to any frontend app.
@@ -86,6 +64,28 @@ make government-frontend
 cd /govuk/government-frontend
 govuk-docker-up app-live
 # You can now view the app on government-frontend.dev.gov.uk
+```
+
+## Using startup scripts
+
+If you are making changes to a frontend app and nothing else, you can view these changes by running the application's `./startup.sh` script. This example is for [government-frontend], but these instructions apply to any frontend app.
+
+```shell
+cd /govuk/government-frontend
+./startup.sh --live
+# Check the output to see what port the app is running on, e.g: localhost:3005
+```
+
+If you want to test changes in [govuk_publishing_components] against a frontend app, you need to edit your frontend app Gemfile and then run the startup script:
+
+```ruby
+gem 'govuk_publishing_components', path: '../govuk_publishing_components'
+```
+
+```shell
+bundle install
+./startup.sh --live
+# Check the output to see what port the app is running on, e.g: localhost:3005
 ```
 
 ## Components pulled in by Static
--- a/source/manual/local-frontend-development.html.md
+++ b/source/manual/local-frontend-development.html.md
@@ -40,7 +40,7 @@ cd /govuk/govuk-docker
 make government-frontend
 
 cd /govuk/government-frontend
-govuk-docker up government-frontend-app-live # or govuk-docker-up app-live
+govuk-docker-up app-live
 # You can now view the app on government-frontend.dev.gov.uk
 ```
 
@@ -53,7 +53,7 @@ gem 'govuk_publishing_components', path: '../govuk_publishing_components'
 ```shell
 cd /govuk/government-frontend
 govuk-docker-run bundle install
-govuk-docker up government-frontend-app-live
+govuk-docker-up app-live
 # You can now view the app on government-frontend.dev.gov.uk
 ```
 
@@ -84,7 +84,7 @@ cd /govuk/govuk-docker
 make government-frontend
 
 cd /govuk/government-frontend
-govuk-docker up government-frontend-app-live
+govuk-docker-up app-live
 # You can now view the app on government-frontend.dev.gov.uk
 ```
 


ChrisBAshton (approved):
Seems reasonable to me 👍

alex-ju (approved):
While I share the view about converging to a single way of running apps locally, as it makes sense especially from a maintenance point of view, the startup script allows, however, to run an application without any prior setup, so we might need to keep the warning for a while until everyone has Docker setup and is confident using it.

richardTowers (changes requested):
I don't agree. I think having a simple way of running the applications on the host is valuable.

The startup scripts are nice because they remind you that, at the end of the day, it's just `rails s` with some environment variables. If we were to push people towards govuk-docker as the only way of running apps, I worry that people would forget that there's no magic needed to run them locally.

On the other hand, `startup.sh` is itself a bit of abstraction over `rails s`, so you could argue that suffers from the same issue as govuk-docker - it's teaching people that you need some "magic shell" before you can start an app. If you don't read the script, you might assume it's doing more magic than it is.

I wonder if a middle ground would be to remove the startup scripts, but document how to start the apps on the host with `rails s`?

```
GOVUK_APP_DOMAIN=www.gov.uk \
GOVUK_WEBSITE_ROOT=https://www.gov.uk \
PLEK_SERVICE_CONTENT_STORE_URI=https://www.gov.uk/api \
PLEK_SERVICE_RUMMAGER_URI=https://www.gov.uk/api \
PLEK_SERVICE_SEARCH_URI=https://www.gov.uk/api \
PLEK_SERVICE_STATIC_URI=${assets.publishing.service.gov.uk} \
bundle exec rails s
```

... that way we're removing all the magic. If you need magic, then govuk-docker is there for you. But otherwise, it's valuable to understand what's actually happening when you run an app.

karlbaker02:
I echo Alex in that converging to a single way of running apps locally makes sense from a maintenance point of view, as well as removing potential points of confusion for new developers.

That being said, I'm currently against this proposal. The start-up scripts are very lightweight in what they do, either pointing to local instances of services that they need or allowing you to point to `live` (which is what I tend to use them for).

It feels somewhat overkill to need to build the container images for any dependent apps (say if you want to run `frontend`) when you don't actually require these locally to be able to run the app assuming you point to live. I think that keeping the start-up scripts reduces the barrier to entry to get the apps which contain these scripts running, which is no bad thing.

benthorner:
Thanks @richardTowers and @karlbaker02 for your comments. I need to respond to both of them separately, but there's a common theme that I'd like to give a different perspective on: whether the scripts are lightweight.

I'd actually argue GOV.UK Docker is equally lightweight. Compare:

- Steps with GOV.UK Docker:

   - Install Docker / GOV.UK Docker (works for all development)
   - Run `make` for the app (works for all development)
   - Run `govuk-docker-up app-live`

- Steps with startup.sh:

   - Install Rbenv and Ruby manually (works for all development)
   - Install other system packages manually (specific to the repo)
   - Manually do `bundle install`
   - Run `./startup.sh`

So while I agree running `rails s` appears to be lightweight, it's also what's _not_ in those scripts that's the maintenance effort, where people pop up on Slack wondering why something doesn't work on their local machine.

karlbaker02:
> Thanks @richardTowers and @karlbaker02 for your comments. I need to respond to both of them separately, but there's a common theme that I'd like to give a different perspective on: whether the scripts are lightweight.
> 
> I'd actually argue GOV.UK Docker is equally lightweight. Compare:
> 
> * Steps with GOV.UK Docker:
>   
>   * Install Docker / GOV.UK Docker (works for all development)
>   * Run `make` for the app (works for all development)
>   * Run `govuk-docker-up app-live`
> * Steps with startup.sh:
>   
>   * Install Rbenv and Ruby manually (works for all development)
>   * Install other system packages manually (specific to the repo)
>   * Manually do `bundle install`
>   * Run `./startup.sh`
> 
> So while I agree running `rails s` appears to be lightweight, it's also what's _not_ in those scripts that's the maintenance effort, where people pop up on Slack wondering why something doesn't work on their local machine.

I think what I'd say to this is that even though GOV.UK Docker might be fairly lightweight in terms of what it does as you've described above, on balance it's easier, quicker and feels more transparent running said apps with the start-up scripts. The effort to create a new gemset and install gems is minimal, the effort to debug problems with GOV.UK Docker when things go wrong can be more involved. I see that this could be a particular issue for frontend devs, but it's something that would be experienced by all.

richardTowers:
For me it's not whether the scripts are lightweight, it's how much abstraction they bring (more abstraction == bad, in this argument, although that's obviously not universal).

In your example:

* Run make for the app
   * Okay, what will this do? Read the Makefile in govuk-docker. Hmmm, my make's not up to scratch... how does the $app-name target appear? There's an APPS variable that does some things with `ls` and `xargs` - I guess there's another Makefile in `projects/$app/Makefile` then. Guess I'll read that one too. `government-frontend: bundle-$app content-store router static` 🤔  what does this do I wonder? I guess the magic is in the `bundle-$app` target, so back to the big Makefile. Right, so it's going to clone the app, then run `govuk-docker build $app-lite` (what does that do? Guess I better read exe/govuk-docker too - okay, it runs docker-compose build $app-lite), then run rbenv install, gem install, and bundle. Gotcha.
* Run govuk-docker-up app-live
  * Same thing here, but less bad. Once you've understood govuk-docker's abstractions, it's obvious that this is going to call `docker-compose up`.

If we document this as _the way_, then people will naturally assume "there's obviously a reason for this complexity, so it can't be as simple as just running `rails s`", when actually it's exactly that simple.

I'm not trying to 💩  on govuk-docker here - there's no question that it solves the problem of running lots of apps in the nicest way we've been able to come up with, but it's _a lot_ more abstraction than `startup.sh`.

maxgds:
Are we confident that everything can be run in docker? I thought there were some outliers, still. I think possibly email-alert-frontend was easier to get running locally for some frontend changes? Haven't encountered any issues myself on the apps I'm familiar with.

If removing this code might be worth checking whether you need to do a parallel update to the docs on govuk_publishing_components as I think they still refer to the startup scripts.

injms:
As long as the app runs, I don't mind what command I run to get it to do so - as long as the app runs. 

But I do think that deprecating the startup scripts at this particular time is not a good idea. Developers that aren't using `govuk-docker` don't need any extra stress and shouldn't need to worry about a deprecation notice that states that `startup.sh` could disappear at any time without warning

chao-xian:
Also, the docs were written for the user: Frontend Developers working on GOV.UK (I think I added this doc). Most frontend devs have told me that they find running the startup script more productive (as per Alex's comment). I deliberately included the `govuk-docker` instructions. Of course there'd be situations where you'd need more apps running locally, but for frontend development that isn't often the case.

DilwoarH:
I'm in favour of keeping the startup scripts as they make it easy for anyone to quickly get the environment up. Though it is easy to run the docker image - I still prefer this method and use it every day. It also helps onboarding new starting really easy.

I find also docker can take up more memory on my system than the start up scripts (that may just be my set up).

benthorner:
Thanks for all your comments so far everyone. I've now had some time to read through them properly. I'm going to try and summarise what I've read so far, and provide some answers / suggestions.

## Summary of where we've got to

- Several people have commented the `startup.sh` scripts are easier to use. @richardTowers also commented they can be easier to understand because they're closer to the actual command being run: `rails s`. @richardTowers also suggested putting the content of the script in the README. This would make the approach harder (due to copy/paste), so I suspect some people here would be against that, if they rely on these scripts currently.

- @karlbaker02 and @maxgds asked about the level of support in GOV.UK Docker. All apps are [supported](https://github.com/alphagov/govuk-docker/blob/master/docs/compatibility.md), and all frontend apps fully support the [app-live stack](https://github.com/alphagov/govuk-docker#the-app--stacks) (the equivalent of `startup.sh --live`).

- Most of our apps still have a `startup.sh` script (inc. all the backend ones). However, I think we only use them for certain frontend apps with the `--live` flag. Even this doesn't work for stateful frontend apps e.g. Email Alert Frontend.

- @maxgds pointed out some READMEs still refer to these scripts (only). We should fix.

- @injms pointed out we should avoid making any breaking changes right now. I agree.

## Suggested actions to make progress

Taking all that together, let's see if we can agree on some (less extreme) actions:

- Update various documentation:

   - Frontend app READMEs should Point to this documentation.
   - Other apps should point to the GOV.UK Docker README.
   - GOV.UK Docker compatibility incorrect for Frontend ([is supported](https://github.com/alphagov/govuk-docker/blob/master/projects/frontend/docker-compose.yml))
   - GOV.UK Docker compatibility out-of-date for Whitehall (is supported)
   - GOV.UK Docker compatibility incorrect for Calculators (is supported)
   - Update this manual to say `startup.sh` only (actually) works for some frontend apps.

- Remove `startup.sh` scripts for all non-frontend apps. These are [no different](https://github.com/alphagov/asset-manager/blob/master/startup.sh) to running `rails s` manually, so we're actually making things clearer / easier by removing them for non-frontend apps.

- Change `startup.sh` scripts for frontend apps to always run with `--live`, since this is the only way they currently work. Keep / remove `startup.sh` scripts for certain frontend apps:

   - Remove for Email Alert Frontend (doesn't work / stateful, requires local Email Alert API).
   - Remove for Feedback (doesn't work / stateful, requires local Support API).
   - Remove for Whitehall (doesn't work / stateful, requires local database and Email Alert API).
   - Keep for Collections
   - Keep for Frontend
   - Keep for Finder Frontend
   - Keep for Service Manual Frontend
   - Keep for Calculators
   - Keep for govuk_publishing_components
   - Keep for GOV.UK Developer Docs
   - Keep for Government Frontend
   - Keep for Info Frontend
   - Keep for Manuals Frontend
   - Keep for Smart Answers
  
## What you can do

If you agree with these actions, you could give this comment a 👍. If you think we need to do something different, please comment which actions you agree with, and which you think we need to change / add / remove.

maxgds:
Just want to question the Whitehall side of things, I don't use the startup script myself and when I tried just now I got into a rabbithole of dependencies needing installing locally so stopped, BUT, for pages such as https://www.gov.uk/government/how-government-works would the startup script be a good and fast option? Does Whitehall's frontend side support the --live option in the script? Whitehall is currently doing a bunch of frontend delivery (long term plan is to change that but we're not there yet) and it doesn't have a use live data option in docker.

benthorner:
@maxgds that's a good point: Whitehall doesn't have a "live" option, either in [it's `startup.sh` script](https://github.com/alphagov/whitehall/blob/master/startup.sh) or in GOV.UK Docker. This is because it's not really a frontend app - it's tightly coupled to its database / APIs. It's a monolith.

While some pages may work with a simple `rails s`, I think we should only have a `startup.sh` script if it works for all pages. Otherwise, it will only disappoint / confuse people. Most frontend apps are fine - the ones I suggest keeping it for, where all pages should render with `--live` - but not apps like Whitehall, which require a database and/or privileged access to protected APIs. These need the GOV.UK Docker ecosystem, and often some local data, in order to work with them.

maxgds:
> While some pages may work with a simple `rails s`, I think we should only have a `startup.sh` script if it works for all >pages. Otherwise, it will only disappoint / confuse people. 

I'm not going to make a big fuss about this because I don't use it and we want to rip out the frontend bits of Whitehall anyway, but I strongly disagree with this. If I need to update the CSS for the specific page I linked above, it's one hell of a requirement to install docker and download a database full of content in order to view what amounts to a static template full of hard coded html. At that point I'm going to not bother to run it at all, do live coding in the browser, paste it into the codebase and hope for the best when it reaches integration or heroku.

benthorner:
Ah, I think I need to clarify what I meant by "disappoint / confuse". Whitehall renders _a lot_ of pages. Some of them work against live APIs, but many don't. All of them work in GOV.UK Docker, because it has all the dependencies. The problem for someone using Whitehall's `startup.sh` is that we don't know what will work, and what won't. So when someone tries to develop a page, they may encounter an error and be confused because we said "it should work".

Hopefully a developer will already have GOV.UK Docker for developing other GOV.UK apps.

If you feel strongly about having partial frontend support for Whitehall, perhaps we could make the experience better by updating the README to explain `startup.sh` will only work for some things, and to try GOV.UK Docker if not.

How does that sound?

maxgds:
> How does that sound?

That makes sense. I was in a position for a long while of running lots of things in Docker but rarely needing to use Whitehall so I hadn't downloaded the required data. That eventually became necessary but it was no small undertaking - an alternative to that, depending on what it is you need to do, is a good thing to provide for where it works.

chris-gds:
> As long as the app runs, I don't mind what command I run to get it to do so - as long as the app runs. 

I think this is my general feeling, to navigate through all app pages and functionality easily, locally.  

(From a FE POV, part of me also wants one command to spin up most of site locally and move through it like I'm on live, I'm not sure if microservices architecture but with a monolith local "feel" conceptually has a name..) 

> While I share the view about converging to a single way of running apps locally, as it makes sense especially from a maintenance point of view, 

I do agree with this sentiment to have one unifying method

> I don't agree. I think having a simple way of running the applications on the host is valuable. 

I also think this is true.  

> Are we confident that everything can be run in docker? I thought there were some outliers, still. 

I was wondering if this was the case with all apps? I did think some had some known issues (I know `whitehall` has been mentioned which sounds like it's own rabbit hole, I wondered about `frontend`...)

> All apps are supported
> GOV.UK Docker compatibility incorrect for Frontend (is supported)

I was recently was starting up  `frontend` + docker.  I had expectations that this would work navigating through [`find-local-council`](https://www.gov.uk/find-local-council)  as I thought this would install the relevant `mapit` dependencies or utilise setup from a live api.  This doesn't seem to be the case?

> it's one hell of a requirement to install docker and download a database full of content in order to view what amounts to a static template full of hard coded html.

I can identify with this, eg with `/find-my-council`, I don't need all of the local postcodes across all of the UK installed locally in order to continue to work - I just need the dozen or so postcodes that lead to the variations of outcome pages.

>the effort to debug problems with GOV.UK Docker when things go wrong can be more involved

I'm a bit concerned about this when startup sometimes hasn't been as smooth as it could be

> I think that keeping the start-up scripts reduces the barrier to entry to get the apps which contain these scripts running, which is no bad thing.

&

> I'm in favour of keeping the startup scripts as they make it easy for anyone to quickly get the environment up.   

I tend to agree or deprecate once there's one unifying method across all, however the suggested way forward seems like a nice halfway.

> Change startup.sh scripts for frontend apps to always run with --live

This makes sense.

benthorner:
Thanks for all your comments everyone. Since the discussion is now very long, and the direction quite different from the beginning, I'm closing this one in favour of a (hopefully) more agreeable proposal / PR:

\>\>\>\>\>\>\>\>\>\> https://github.com/alphagov/govuk-developer-docs/pull/2897 <<<<<<<<<<<

## What's changed?

- The third commit is different (suggest, rather than deprecate).

- The PR description contains a bunch of associated actions.

benthorner:
@alex-ju @maxgds @chris-gds @chao-xian @injms hello 👋. I've got a PR up on Collections to implement the changes we discussed here. I'd appreciate your thoughts on one specific commit: https://github.com/alphagov/collections/pull/2362/commits/58d68104e5110069ff5f758da1704084a20c9166. 

For context, we've now rewritten the bulk of the READMEs for GOV.UK repos. The idea is now to make frontend repos  consistent with those. See here for the first use of the new content: https://github.com/alphagov/content-publisher/pull/2257.

Much of the PR is about getting the README into shape, and has already been reviewed. You don't need to look at the other commits. Once I've got some feedback from you, I'll look at rolling out the same change across other frontend repos. 

You're welcome to leave a comment on the PR, or here.

PR #2893, opened 2020-12-09
Remove github-markdown gem

ChrisBAshton:
This is unsupported and hasn't been updated since 2015.
Removing it doesn't appear to break anything, so I don't
think we use it anywhere anymore and it can be safely removed.


chao-xian (approved):
I came to the same conclusion! Thanks for raising the PR.

PR #2897, opened 2020-12-14
Clarify when to use "startup.sh" scripts

benthorner:
This is as much a proposal as a PR, so feel free to leave
comments about your own thoughts on this:

- Update various documentation:

   - Frontend app READMEs should Point to this documentation.
   - Other apps should point to the GOV.UK Docker README.
   - Fix: https://github.com/alphagov/govuk-docker/issues/444

- Remove `startup.sh` scripts for all non-frontend apps. These are [no different](https://github.com/alphagov/asset-manager/blob/master/startup.sh) to running `rails s` manually, so we're actually making things clearer / easier by removing them for non-frontend apps.

- Change `startup.sh` scripts for frontend apps to always run with `--live`, since this is the only way they currently work. Remove `startup.sh` scripts for certain frontend apps:

   - Remove for Email Alert Frontend (doesn't work / stateful, requires local Email Alert API).
   - Remove for Feedback (doesn't work / stateful, requires local Support API).

### DIFF
--- a/source/manual/local-frontend-development.html.md
+++ b/source/manual/local-frontend-development.html.md
@@ -85,7 +85,6 @@ gem 'govuk_publishing_components', path: '../govuk_publishing_components'
 ```
 
 ```shell
-bundle install
 ./startup.sh --live
 # Check the output to see what port the app is running on, e.g: localhost:3005
 ```
--- a/source/manual/local-frontend-development.html.md
+++ b/source/manual/local-frontend-development.html.md
@@ -68,7 +68,9 @@ govuk-docker-up app-live
 
 ## Using startup scripts
 
-If you are making changes to a frontend app and nothing else, you can view these changes by running the application's `./startup.sh` script. This example is for [government-frontend], but these instructions apply to any frontend app.
+**NOTE:** this approach only works for frontend development and relies on live APIs. You will need to find and install any dependencies yourself. Consider using the general GOV.UK Docker environment in the first instance.
+
+If you are making changes to certain frontend apps you can also view these changes by running the application's `./startup.sh` script - if it has one. This example is for [government-frontend], but these instructions may apply to other frontend apps.
 
 ```shell
 cd /govuk/government-frontend
--- a/source/manual/local-frontend-development.html.md
+++ b/source/manual/local-frontend-development.html.md
@@ -7,28 +7,6 @@ section: Frontend
 type: learn
 ---
 
-## Using startup scripts
-
-If you are making changes to a frontend app and nothing else, you can view these changes by running the application's `./startup.sh` script. This example is for [government-frontend], but these instructions apply to any frontend app.
-
-```shell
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
This looks reasonable to me 👍 

Happy that we're keeping the scripts around for the frontend apps, where I think there a useful reminder that there's no magic needed to run these apps. Also nice that they'll get a bit simpler once they only support `--live`.

source/manual/local-frontend-development.html.md
```diff
@@ -84,10 +62,34 @@ cd /govuk/govuk-docker
...
-govuk-docker up government-frontend-app-live
+govuk-docker-up app-live
 # You can now view the app on government-frontend.dev.gov.uk
 ```
 
+## Using startup scripts
+
+**NOTE:** this approach only works for frontend development and relies on live APIs. You will need to find and install any dependencies yourself. Consider using the general GOV.UK Docker environment in the first instance.
+
+If you are making changes to certain frontend apps, you can also view these changes by running the application's `./startup.sh` script, if it has one. This example is for [government-frontend], but these instructions apply to other frontend apps.
```
injms, line 73: Suggest a slight tweak to emphasise that the startup scripts might not be present and that they vary from app to app:

```suggestion
If you are making changes to certain frontend apps you can also view these changes by running the application's `./startup.sh` script - if it has one. This example is for [government-frontend], but these instructions may apply to other frontend apps.
```

source/manual/local-frontend-development.html.md
```diff
@@ -84,10 +62,34 @@ cd /govuk/govuk-docker
...
+
+If you want to test changes in [govuk_publishing_components] against a frontend app, you need to edit your frontend app Gemfile and then run the startup script:
+
+```ruby
+gem 'govuk_publishing_components', path: '../govuk_publishing_components'
+```
+
+```shell
+bundle install
+./startup.sh --live
```
injms, line 89: I think that most startup scripts include `bundle install` - so we could get rid of it here:
```suggestion
./startup.sh --live
```

source/manual/local-frontend-development.html.md
benthorner: Sure, I've updated the 3rd commit with your suggestion 👍.

benthorner:
@karlbaker02 @maxgds @ChrisBAshton @alex-ju @theseanything @chao-xian @DilwoarH @chris-gds thanks for giving your feedback on the previous proposal. I've raised a new one here, based on that feedback. 

Could you take a look and review this new PR / proposal?

theseanything (approved):

maxgds (approved):

DilwoarH (approved):
Thank you @benthorner

injms (approved):

benthorner:
I've written a card for the follow-up work: https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh.

PR #2898, opened 2020-12-16
Update docs for rebooting cache machines

bilbof:
These docs now reflect the new process for rebooting cache machines manually.


sihugh (approved):
This is ace @bilbof, a vast improvement over a very boring process

PR #2896, opened 2020-12-14
Update process for Gandi TLS certs renewal

fredericfran-gds:
Following changes were made:
1. remove info about reliability engineer having to make the changes
   since now all govuk developers are expected to be able to perform
   these tasks.
2. add that there is a Gandi cert for www.gov.uk in addition to the
   wildcard ones.
3. remove the puppet instructions since the relevant Carrenza infra
   has been shutdown and AWS is not configured via Puppet.
4. Clarify that for production, the certs need to be uploaded to AWS
   ACM in North Virginia for disaster recovery cloudfront.

Refs:
1. [integration renewal trello card](https://trello.com/c/EjVHpJdg/3947-zendesk-hostmaster-gandi-expiration-of-the-certificate-ssl-standard-integrationpublishingservicegovuk-in-29-days)
2. [www.gov.uk renewal trello card](https://trello.com/c/gdIGvJNL/3946-zendesk-hostmaster-gandi-expiration-of-the-certificate-ssl-standard-wwwgovuk-in-29-days)

PR #2899, opened 2020-12-16
Add requirement to add TLS certs to pass store

fredericfran-gds:
For TLS certificates that are renewed manually, there is a need to add
the certs details in the pass store so that these can be retrieved
easily later when configuring external systems.

This PR add this requirements in the doc which was missing in #2896


bilbof (approved):
Thank you!

PR #2900, opened 2020-12-16
Change most "Reliability Engineering" references to other things

cbaines:
I'm not sure where these "Reliability Engineering" references came
from, but it's never been particularly accurate as far as I
understand. Reliability Engineering has always been a vague way of
referencing either those who support specific things. In particular RE
Autom8 support Concourse and Logit and RE GOV.UK when it existed
supported GOV.UK and DNS. It's always been more helpful to be more
specific than Reliability Engineering.

This issue is more apparent now given that RE GOV.UK no longer exists,
and that the most similar team is the GOV.UK Replatforming team within
the GOV.UK program, which currently has some SRE's on loan. This means
that saying "Reliability Engineering" is now wrong in many cases, as
the documentation should reference the Replatforming team within
GOV.UK.

I think the changes in this commit make things less wrong. There's
still some more work to remove or change the documentation about
raising things with Reliability Engineering, as I think that was only
correct while the RE GOV.UK team existed.

(this is from a fork as I no longer have sufficient access to raise Pull Requests directly)

### DIFF
--- a/source/manual/alerts/renew-tls-certificate.html.md
+++ b/source/manual/alerts/renew-tls-certificate.html.md
@@ -19,19 +19,19 @@ expire.
 
 See [renew a TLS certificate for GOV.UK](/manual/renew-a-tls-certificate.html)
 for details of how to renew the relevant certificate. This is normally done by
-Reliability Engineering.
+GOV.UK Replatforming.
 
 ## Production www.gov.uk certificate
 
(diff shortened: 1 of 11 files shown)


richardTowers (approved):
Thanks Chris!

There are a couple of places where we'll want to do some further updates to try to weasel things off to GOV.UK's main support team, as we don't get as much support from SREs as we used to. But this is still much more accurate than what we have now, so thank you!

PR #2902, opened 2020-12-18
Update docs on Fastly, CDN source IPs, DoS mitigation.

sengi:
Add a new section about the type of DoS/spam mitigation which was used in November 2020 for this incident:
https://docs.google.com/document/d/12DzQsDeu7zUcICy9zVporjprX4qZFIrpOOWtYYRx-nk/edit#

Also various small updates including:

- Our origin moved from London to Dublin quite some time ago.
- Fix confusion between the Fastly cache-clearing ACL (access control list) and the origin ACL.
- The vSphere automation stuff in govuk-provisioning is never coming back, so let's delete the mention of it.
- Clarify that the Fastly IP safelist and alert in Carrenza is only relevant to Performance Platform.
- Fix a broken link to the IP address denylist.

### DIFF
--- a/source/manual/cdn.html.md
+++ b/source/manual/cdn.html.md
@@ -9,7 +9,7 @@ parent: "/manual.html"
 
 GOV.UK uses Fastly as a CDN. Public users aren't accessing GOV.UK servers directly, they connect via the CDN. This is better because:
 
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
LGTM 👍

PR #2905, opened 2020-12-21
Increase performance and readability of GitHub Actions

ChrisBAshton:
Commit 1 simplifies our caching strategy, but brings no performance benefit.

Commit 2 simplifies our code slightly and brings a 39% improvement in build time!

PR #2906, opened 2020-12-21
Use native 'ruby/setup-ruby' options for caching

ChrisBAshton:
Using the 'bundler-cache: true' option in 'ruby/setup-ruby'
removes our dependency on 'actions/cache' and is much more
readable config.

After the initial build [1], it leads to build times of just under
4 minutes [2], which is comparable to what we have on master
already [3]. So this is just about readability rather than
improving the build time.

[1]: https://github.com/alphagov/govuk-developer-docs/runs/1589338635
[2]: https://github.com/alphagov/govuk-developer-docs/runs/1589391081
[3]: https://github.com/alphagov/govuk-developer-docs/runs/1588704546

Thanks to @tomnatt for the tip:
https://tomnatt.blogspot.com/2020/12/a-first-look-at-github-actions.html

We've already applied this new config to the Developer Docs
themselves:
https://github.com/alphagov/govuk-developer-docs/pull/2905

### DIFF
--- a/source/manual/test-and-build-a-project-with-github-actions.html.md
+++ b/source/manual/test-and-build-a-project-with-github-actions.html.md
@@ -104,12 +104,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: ruby/setup-ruby@v1
-      - uses: actions/cache@v1
         with:
-          path: vendor/bundle
-          key: bundle-${{ hashFiles('**/Gemfile.lock') }}
-          restore-keys: bundle
-      - run: bundle install --jobs 4 --retry 3 --deployment
+          bundler-cache: true
       - run: bundle exec rake
 ```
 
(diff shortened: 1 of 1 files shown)


source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -145,20 +138,13 @@ jobs:
...
-          key: bundle-${{ matrix.ruby }}-${{ hashFiles('**/*.gemspec') }}
-          restore-keys: bundle-${{ matrix.ruby }}
-      - run: bundle install --jobs 4 --retry 3 --path vendor/bundle
       - run: bundle exec rake
 ```
 
 Notes:
 
-- The [deployment][bundler-deployment] flag for bundler cannot be used for gems
-  as they should not have a Gemfile.lock committed.
```
kevindew, line 161: Ah it's great that the action seems to take care of this automatically now: https://github.com/ruby/setup-ruby/blob/e5b6990d5186f1ac1d8074e46b72f9efb63aa305/index.js#L256

kevindew (approved):
Looks great - thanks Chris. That'll make for much nicer, simpler actions. 

Now to find an equivalent for Yarn (where we're still discussing the caching approach): https://github.com/alphagov/govuk_publishing_components/pull/1013#discussion_r542243089

source/manual/test-and-build-a-project-with-github-actions.html.md
ChrisBAshton: Thanks, I've fixed a trailing space issue & added this link in the commit :). Will also check out the Yarn discussion!

PR #2907, opened 2020-12-22
Improve 'outstanding security updates' documentation

ChrisBAshton:
1) Clarify that a missing log probably means a newly provisioned instance
2) Imply a clearer ordering of debug commands to run


kevindew (approved):
Nice one 👍

PR #2912, opened 2020-12-29
Improved App Pages

ChrisBAshton:
This is the last PR for the ["Consume all docs folders" Trello card](https://trello.com/c/UyfVvwYm/183-consume-all-docs-folders), and should mean we can close the epic. 🎉 

The aim here is to highlight key information about our apps, above the fold. As a developer, I want to quickly read a summary of what an app does, find its GitHub repo, and know about how it's deployed and where it is hosted. I may also want quick access to its Sentry project and Grafana dashboards, and to be able to run Rake tasks against it.

This sort of change is always going to be a bit subjective, but I think using the [summary list](https://components.publishing.service.gov.uk/component-guide/summary_list) component helps us to achieve this goal quite nicely.

## Before

Screenshot 2021-02-08 at 12 45 57

## After

Screenshot 2021-02-08 at 12 46 11


app/run_rake_task.rb
```diff
@@ -23,6 +23,24 @@ def self.links(application, rake_task = "")
...
+      href=\"#{application.rake_task_url('integration', rake_task)}\"
+      aria-label=\"Run #{rake_task_name} on Integration\">Integration</a>"
+    staging_link = "<a
+      href=\"#{application.rake_task_url('staging', rake_task)}\"
+      aria-label=\"Run #{rake_task_name} on Staging\">Staging</a>"
+    production_link = "<a
+      href=\"#{application.rake_task_url('production', rake_task)}\"
+      aria-label=\"Run #{rake_task_name} on Production\">⚠️ Production ⚠️</a>"
+
+    "<span aria-hidden='true'>Run #{rake_task_name} on</span> #{integration_link}, #{staging_link} or #{production_link}"
```
owenatgov, line 41: Big fan of this screen reader implementation, very elegant.

owenatgov (commented):
A couple of little nitpicks on design system/components gem usage. It's not as critical that we eat our own dog food for the dev docs as they're almost entirely internal but I don't think there's any harm in mitigating risk of our own docs being inaccessible. Tested this locally with voiceover and works very nicely.

source/partials/application/_links.html.erb
```diff
@@ -0,0 +1,41 @@
+<ul>
```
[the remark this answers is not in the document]
ChrisBAshton: I wasn't able to with this one, as the styling doesn't get applied as it's pulled into the Summary List component 🤔 we were losing the bullets for some reason. I've added the govuk-link classes though.

source/partials/application/_readme.html.erb
```diff
@@ -0,0 +1,17 @@
+<% if application.private_repo? %>
+  <span class="govuk-error-message">
```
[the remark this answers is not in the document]
ChrisBAshton: Could do - though perhaps a bit confusing to use a form error on a page which has no form or user input :) I'd swap it out if we were rolling out our own `govuk-error-message` class, but this is built into the tech docs gem, so it wouldn't give us an opportunity to delete any code. Personally think it's fine as-is - what do you think?

source/partials/application/_links.html.erb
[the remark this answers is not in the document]
owenatgov: Fine by me, feel free to resolve

source/partials/application/_readme.html.erb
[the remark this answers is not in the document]
owenatgov: That makes sense to me. The mitigation risk is that if the design system ever change their error messages for any reason then this would be an extra source of truth, although I imagine the design system team would project breaking changes like this very loudly anyway. Feel free to resolve.

Gemfile
```diff
@@ -18,6 +18,7 @@ gem "html-pipeline"
 gem "kramdown"
 gem "mdl"
 
+gem "govuk_publishing_components", "~> 23.12"
```
owenatgov, line 21: What's the reason for this qualifier?

owenatgov (approved):
One little query but looking great. Really excited to make docs more readable 😃

Gemfile
ChrisBAshton: Been having a bit of trouble getting my local setup to use a version of govuk_publishing_components higher than v21!

```
Downloading govuk_publishing_components-23.12.1 revealed dependencies not in the
  API or the lockfile (sprockets (< 4)).
  Either installing with `--full-index` or running `bundle update
  govuk_publishing_components` should fix the problem.
```

Build is currently failing with a similar message. The above was an attempt to fix the issue. I'll be making another force push shortly to hopefully fix it outright!

app/app.rb
```diff
@@ -12,7 +12,7 @@ def api_payload
       dependencies_team: dependencies_team,
       puppet_name: puppet_name,
       production_hosted_on: production_hosted_on,
-      continuously_deployed: continuously_deployed,
+      continuously_deployed: continuously_deployed?,
```
brucebolt, line 15: Good spot 👍

app/run_rake_task.rb
```diff
@@ -23,6 +23,27 @@ def self.links(application, rake_task = "")
     ERB
   end
 
+  def self.terse_links(application, rake_task = "")
+    application = find_application(application) if application.is_a? String
+
+    rake_task_name = rake_task.presence || "a Rake task"
+
+    integration_link = "<a
```
brucebolt, line 31: Are we able to use the [`link_to`](https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to) helper here?  It might mean we can remove the need to escape quotes.

source/partials/application/_hosting.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if application.production_hosted_on == "aws" %>
+<strong><%= application.hosting_name %></strong>, on the <strong><%= application.aws_puppet_class %></strong> machine class
```
brucebolt, line 2: A very minor thing, but the `if`, `elsif` and `else` code should be indented once

brucebolt (commented):
Changes look good to me, I've just added a few stylistic comments.

The new layout of the application page should make it easier to scan over the content without so much scrolling 👍

app/run_rake_task.rb
ChrisBAshton: I think I ran into two problems with that:
1. Trying to call it from outside the partial
2. It not supporting the `aria-label` property.

source/partials/application/_hosting.html.erb
ChrisBAshton: Ah yes - originally this file was parsed as Markdown, so the indentation caused it to be treated as a code block. I think I can safely fix up the indentation now.

PR #2915, opened 2020-12-31
Add entry for GOV.UK Accounts

chao-xian:
We didn't have anything in our main developer docs for these.

# Screenshots

Screenshot 2020-12-31 at 16 43 47

Screenshot 2020-12-31 at 16 36 48

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -1,3 +1,17 @@
+# GOV.UK Account
+- github_repo_name: govuk-account-manager-prototype
+  type: GOV.UK Account
+  team: "#govuk-accounts-tech"
+  production_hosted_on: paas
+  sentry_url: https://sentry.io/organizations/govuk/issues/?project=5370953
+  dashboard_url: false
+- github_repo_name: govuk-attribute-service-prototype
+  type: GOV.UK Account
+  team: "#govuk-accounts-tech"
+  production_hosted_on: paas
+  sentry_url: https://sentry.io/organizations/govuk/issues/?project=5371008
+  dashboard_url: false
+
 # publishing apps
 
 - github_repo_name: collections-publisher
--- a/data/dashboard.yml
+++ b/data/dashboard.yml
@@ -1,6 +1,7 @@
 # These sections are magically populated from applications.yml
 - name: "Applications"
   sections:
+    - name: GOV.UK Account
     - name: Publishing apps
     - name: APIs
     - name: Services
--- a/source/manual/govuk-account.html.md.erb
+++ b/source/manual/govuk-account.html.md.erb
@@ -1,0 +1,21 @@
+---
+owner_slack: "#govuk-accounts-tech"
+title: Use GOV.UK Account
+section: GOV.UK Account
+layout: manual_layout
+parent: "/manual.html"
+---
+
+GOV.UK Account is a centralised account infrastructure for GOV.UK. It gives GOV.UK users continuity across their different interactions with the government.
+
+For information on how to use GOV.UK Account, see the following:
+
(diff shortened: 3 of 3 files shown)


benthorner:
I wonder if we should do this, though? Accounts are meant to be separate from "GOV.UK", including their apps. So by listing them here, we kind of saying "we" maintain them. We should probably check with @barrucadu or @huwd about what makes sense here - definitely think there should be a clear statement about accounts in here, somewhere.

chao-xian:
@benthorner it was never clear to me that they are _that_ separate. @richardTowers some steer from you perhaps please? Today's situation made for an interesting event. The entries here are pointers/harvest their repos anyway.

chao-xian:
@benthorner maybe I can add entries to link to their manual at least from our Manual somewhere?

benthorner:
@chao-xian yup, I reckon we should have at least something in here that links to the other manual. I think it's really on the Accounts team to write that, though: we (all of us) don't currently have a shared understanding of the relationship between "GOV.UK" and "GOV.UK Accounts", why there's a separate manual for it, and what should go in each of them.

huwd:
Heya, mmm, yeah this is a good debate to be having.

Ben is correct, the approach we've been taking is that we might be more akin to something like pay/verify in future and so we've been following their playbook on docs.

So we've got this repo: https://github.com/alphagov/govuk-account-team-manual (currently exceedingly sparse!)
which is up here: https://docs.account.publishing.service.gov.uk/#gov-uk-accounts-technical-documentation

A link sounds like a good idea. We may add / remove / change those repos so we'd need to maintain those links in GOV.UK docs.

:thinking: That said...@chao-xian isn't wrong. I do question sometimes how much we are actually a separate thing from GOV.UK right now, or when that separation might actually occur in a meaningful way.

Currently we're growing rapidly - we may rely on 2ndline / on call in some circumstance, and the Account system currently is not effectively isolated from GOV.UK services (we made compromises especially with the email alert system which draws them closer together than they might otherwise be...).

This will be resolved in time, as accounts broadens to take on multiple services.

In the meantime a link to our docs seems like a good idea and then I think @brucebolt, @bevanloon, @barrucadu, @richardTowers and I should get together on an Accounts ticket to clarify:

- how integrated into GOV.UK this thing actually is during these prototype phases
- what our support expecation is / might be in the future
- given those, what we need docs to do and if they make more sense in a GDS model, or where GOV.UK folks might go looking for them?

That sound sensible to both of you @chao-xian / @benthorner?

chao-xian:
@huwd I was going by the user need of 2ndline helping Dana to deploy stuff on NYE. Which I reckon will help steer your conversations. I'll keep this PR open then. Thanks!

huwd:
Ah ha, looks like this is already a source of friction then.
[I've created a card,](https://trello.com/c/ZUG2kb49/629-reflect-are-we-following-govuk-or-gds-dev-models) we've got planning today so I'll discuss with the team.

We might need to wait until later in the week when everyone is back.

huwd:
Sorry @chao-xian I know this one has lingered.
We're due to have a team chat tomorrow which should help sort out some of the bigger questions here.

Then @jonathanglassman has been getting all our team docs into shape. We'll ensure some of that makes it back into the govuk dev docs once we've had those chats.

jonathanglassman:
Had a chat with @chao-xian yesterday where he showed me how the GOV.UK Developer Docs worked. I've added the GOV.UK Account team manual and tech docs to the manual section. However, the preview of these is not working so I'll grab some help from the GOV.UK Account team. Have linked the trello card below as per GOV.UK Account team agile practice.

https://trello.com/c/hGrGtMzS/584-add-account-tech-docs-links-to-main-govuk-developer-docs

huwd (approved):
Great work, thanks both!

PR #2920, opened 2021-01-08
Use app-console command when replacing assets

brucebolt:
The instructions say to ssh into a machine, then load an app console.  But we can do all this within gds-cli, which saves a step.


ChrisBAshton (approved):

brucebolt:
Closing this as I realise why the extra step is needed.

If we get an app-console for `asset-manager`, we could end up on any of the `backend` machines.  But we need to specifically be on the one we uploaded the file to.

PR #2926, opened 2021-01-13
Document how to reboot a bad Jenkins agent

ChrisBAshton:


source/manual/alerts/rebooting-machines.html.md
```diff
@@ -103,6 +103,34 @@ more info.
...
+gds govuk connect ssh -e integration ci_agent:6
+```
+
+Reboot the machine: `sudo reboot`.
+
+It will take a few minutes for Jenkins to start up on the agent. You can
+SSH back into the machine and monitor this if you wish: use `top` to list
+the running processes, then `>` to sort by memory usage. You should see a
+`java` process at the top of the list when Jenkins is running.
+
```
sengi, line 129: Since we noticed that we seem to have Jenkins configured to ssh into the node and start the agent rather than the agents registering with the master, I'm not sure whether this step is meaningful.

sengi (approved):

source/manual/alerts/rebooting-machines.html.md
ChrisBAshton: I've now removed that paragraph 👍 thanks Chris.

PR #2927, opened 2021-01-15
Document Sentry in detail

ChrisBAshton:
Trello: https://trello.com/c/0CsM2Z4l/2202-document-sentry-in-more-detail

### DIFF
--- a/source/manual/sentry.html.md
+++ b/source/manual/sentry.html.md
@@ -1,6 +1,6 @@
 ---
 owner_slack: "#govuk-developers"
-title: Error reporting with Sentry
+title: Sentry
 parent: "/manual.html"
 layout: manual_layout
 section: Monitoring
--- a/source/manual/sentry.html.md
+++ b/source/manual/sentry.html.md
@@ -7,12 +7,220 @@ section: Monitoring
 type: learn
 ---
 
-We use [Sentry][] to notify us of exceptions that happen in all environments.
+[Sentry][] is a service that collates errors that happen on GOV.UK in one
+place, regardless of which application, what environment or what machine
+the error occurred on. It is far more convenient than logging on to
+individual machines and querying their logs.
 
-We configure Sentry using [govuk-saas-config][]. It reads a list of apps from the docs and makes sure that all configuration is set up correctly.
+Useful links:
 
-Use the [govuk_app_config][] gem to configure your application to talk to Sentry.
+- [Sentry Organization Stats](https://sentry.io/organizations/govuk/stats/)
+  for an overview of how many errors are being sent/rate-limited site-wide.
+- [Sentry Projects](https://sentry.io/organizations/govuk/projects/)
+  for a per-project view, enabling you to see which ones are causing the
+  most errors right now.
+- Sentry Grafana dashboards:
+  [Production](https://grafana.production.govuk.digital/dashboard/file/sentry_errors_across_all_environments.json),
+  [Staging](https://grafana.staging.govuk.digital/dashboard/file/sentry_errors_across_all_environments.json)
+  and [Integration](https://grafana.integration.publishing.service.gov.uk/dashboard/file/sentry_errors_across_all_environments.json)
+  (all require VPN), for a per-environment breakdown of Sentry errors.
 
-[Sentry]: https://sentry.io/govuk
-[govuk-saas-config]: https://github.com/alphagov/govuk-saas-config/blob/5171b2803a7e211fff9536909b7d27c7fa5a4840/sentry/Rakefile
+[Sentry]: https://sentry.io/govuk/
+
+## Nomenclature
+
+- **Project**: each GOV.UK application has its own project on Sentry, where
+  errors originating from that application are consolidated. Example:
+  [publishing-api](https://sentry.io/organizations/govuk/issues/?project=202242)
+- **Issue**: a group of similar errors. For example, multiple instances of
+  `NoMethodError` from the same app are likely to be logged under the same
+  issue (see [example][example-issue]) in the project in Sentry. In practice,
+  there are a [variety of factors][anchor-error-grouping] that determine
+  whether errors are grouped under the same issue.
+- **Error/Exception**: a single occurrence of an Issue, containing details
+  such as stack trace, arguments, server IP and error message.
+  See [example][example-error].
+
+[anchor-error-grouping]: #fingerprinting
+[example-issue]: https://sentry.io/organizations/govuk/issues/2150349125/?project=202259&query=is%3Aunresolved
+[example-error]: https://sentry.io/organizations/govuk/issues/2155210524/events/aa4edd05a94c4d14b07d4c754a13c788/
+
+## How Sentry is integrated on GOV.UK
+
+Projects can be created and edited in the Sentry UI, but this risks creating
+inconsistencies or missing apps. We therefore configure projects using
+[govuk-saas-config][] (and its [associated rake tasks][rake-tasks]), which
+read a [list of apps][docs-apps] from govuk-developer-docs and make sure that
+all configuration is set up correctly.
+
+Apps are configured to talk to Sentry using the [govuk_app_config][] gem,
+which interfaces with Sentry via its [`GovukError` class][govukerror]. Apps
+call `GovukError.configure` - see [example][email-alert-api-example]. This
+[delegates][delegator-pattern] to the [sentry-raven][] gem under the hood,
+though the gem is now superseded by sentry-ruby, which we have
+[plans to migrate to][trello-migrate] in the future.
+
+Unhandled exceptions are automatically logged to Sentry, but you can also
+[manually report something to Sentry using `GovukError.notify`][manually-report].
+
+[docs-apps]: https://docs.publishing.service.gov.uk/apps.json
+[govuk-saas-config]: https://github.com/alphagov/govuk-saas-config/blob/5171b2803a7e211fff9536909b7d27c7fa5a4840/sentry/Rakefile#L1-L12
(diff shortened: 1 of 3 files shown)


source/manual/sentry.html.md
```diff
@@ -7,12 +7,215 @@ section: Monitoring
...
+
+We should only capture actionable errors in Sentry, i.e. errors representing
+an underlying issue that developers can fix. We should not capture 4XX
+responses, for example, as these indicate a client issue. In fact, sentry-raven
+ignores several exception types by default - see below.
+
+### Ignoring exception types
+
+sentry-raven has a [list of exceptions to ignore][sentry-raven-excluded_exceptions],
+which by by default do not get logged in Sentry.
```
edwardkerry, line 103: an extra "by" here

source/manual/sentry.html.md
```diff
@@ -7,12 +7,215 @@ section: Monitoring
...
+[groups based on stack trace][stack-trace-grouping]. If no stack trace is
+available, it will [group by exception type][exception-grouping] instead.
+Failing that, it will group by error message.
+
+Fingerprint rules can be applied at the project level by editing the
+"Issue Grouping" page under a project (see [example][example-issue-grouping]).
+For example, you can force all errors of the same exception type to have the
+same fingerprint. On the same screen, you can set stack trace rules, to
+remove or rename certain 'frames' in the trace. In practice, we don't
+currently apply any custom rules, after an in-depth [investigaton][]
```
edwardkerry, line 172: is there an investigation link for here?

ChrisBAshton: Yes - `[]` refers to the link text as its alias, which is `investigation`, defined here: https://github.com/alphagov/govuk-developer-docs/pull/2927/files#diff-f59e86b993e1bbe8d27722a2a604d3df960a1282ede39f6f89fa6bc3fc96315cR199

source/manual/sentry.html.md
ChrisBAshton: 🤦 oops! Thanks - fixed.

source/manual/sentry.html.md
edwardkerry: ah, it's another typo - sorry i spotted the link wasn't working but not the reason why.
`investigaton` => `investigation`

ChrisBAshton: Ahhhhh good spot! 👏

source/manual/sentry.html.md
```diff
@@ -7,12 +7,215 @@ section: Monitoring
...
+of fingerprinting, so that stack trace frames like:
+
+`app/views/welcome/view_error.html.erb in _app_views_welcome_view_error_html_erb__2807287320172182514_65600 at line 1`
+
+...get normalised to:
+
+`app/views/welcome/view_error.html.erb at line 1`
+
+This means if the same `ActionView::Template::Error` error happens twice, it
+is grouped into the same issue instead of erroneously treated as separate issues.
```
1pretz1, line 186: I found this section a little hard to follow - I think this paragraph would make sense underneath the one below?

source/manual/sentry.html.md
```diff
@@ -7,12 +7,215 @@ section: Monitoring
...
+
+sentry-raven uses a [customised version][custom-rails-backtrace] of
+`Rails::BacktraceCleaner` to do this. The
+[original `BacktraceCleaner`][native-rails-backtrace] removes any "framework trace"
+from the stacktrace, leaving only the "application trace" behind, which would mean
+we'd lose key diagnostic information. It would, however, be generally beneficial for
+grouping, as Sentry is very sensitive to stack traces that differ only slightly
```
1pretz1, line 193: Could we re-phrase this sentence slightly so its obvious that we're referring to the original approach and not the customised version when mentioning how it'd be beneficial?

1pretz1 (commented):
Fantastic work @ChrisBAshton! ⭐  Really interesting read and I've learnt a lot in the process. Thanks for the in-depth dive into sentry and into how we use it. 

Only minor comments, no blockers! 😄

source/manual/sentry.html.md
```diff
@@ -7,12 +7,215 @@ section: Monitoring
 type: learn
 ---
 
-We use [Sentry][] to notify us of exceptions that happen in all environments.
+[Sentry][] is a service that collates all of the errors that happen on GOV.UK
```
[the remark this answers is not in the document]
ChrisBAshton: Fair point 👍 have dropped the word "all".

source/manual/sentry.html.md
```diff
@@ -7,12 +7,215 @@ section: Monitoring
...
+responses, for example, as these indicate a client issue. In fact, sentry-raven
+ignores several exception types by default - see below.
+
+### Ignoring exception types
+
+sentry-raven has a [list of exceptions to ignore][sentry-raven-excluded_exceptions],
+which by default do not get logged in Sentry.
+
+However, govuk_app_config [overwrites, rather than appends][overwrite-warning]
+its own list of ignorable exceptions, so only that list is used. If your
```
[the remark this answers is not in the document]
ChrisBAshton: Good idea. I started writing this on the basis that it appends rather than overwrites, and was quite surprised when I dug a bit deeper 😂

source/manual/sentry.html.md
```diff
@@ -7,12 +7,215 @@ section: Monitoring
...
+application raises an exception on
+[govuk_app_config's `excluded_exceptions`][govuk_app_config-excluded_exceptions],
+it will not be logged in Sentry.
+
+Then there is a list of exceptions to ignore if they occur during the nightly
+data sync: [`data_sync_excluded_exceptions`][data_sync_excluded_exceptions].
+These are exceptions that are commonly raised while the databases and content
+store are periodically unavailable.
+
+For both of these ignore lists, the [exception chain is inspected][chain-inspection],
```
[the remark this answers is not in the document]
ChrisBAshton: Yes, I wanted to link to a code example explaining how the exception chain inspection works, but couldn't easily reference both places without rewriting the sentence completely. I appreciate this is confusing though. I'll rewrite this section.

source/manual/sentry.html.md
```diff
@@ -7,12 +7,215 @@ section: Monitoring
...
+available to your app is not on the
+[`active_sentry_environments`][active_sentry_environments] list, then none of
+its errors will be logged to Sentry.
+
+[active_sentry_environments]: https://github.com/alphagov/govuk_app_config/blob/b911c5bbef9bd1df6a92cf31eb5a8e0d3a91d851/lib/govuk_app_config/govuk_error/configure.rb#L14-L18
+
+### Advanced Sentry customisation
+
+For all of the above, you can easily
+[configure any of these properties by appending to them][error-configuration].
```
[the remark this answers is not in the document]
ChrisBAshton: Good point. I've moved the anchor link to point to https://github.com/alphagov/govuk_app_config/blob/master/README.md#error-reporting.

source/manual/sentry.html.md
ChrisBAshton: Hmmm I've just tried this, but have moved it back because I think it makes more sense where it is. Have another read and let me know what you think. I've structured the section a bit better now so hopefully it is less confusing.

source/manual/sentry.html.md
```diff
@@ -7,12 +7,215 @@ section: Monitoring
...
+
+[anchor-error-grouping]: #fingerprinting
+[example-issue]: https://sentry.io/organizations/govuk/issues/2150349125/?project=202259&query=is%3Aunresolved
+[example-error]: https://sentry.io/organizations/govuk/issues/2155210524/events/aa4edd05a94c4d14b07d4c754a13c788/
+
+## How Sentry is integrated on GOV.UK
+
+Projects can be created and edited in the Sentry UI, but this risks creating
+inconsistencies or missing apps. We therefore configure projects using
+[govuk-saas-config][] (and its [associated rake tasks][rake-tasks]), which
```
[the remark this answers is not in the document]
ChrisBAshton: That's true, and is something I was encouraging everyone to do this time last year! Then we found that [the shortcut syntax didn't work with the Redcarpet markdown library](https://trello.com/c/XSgGyudd/181-fix-url-markdown-parsing) we were using and I stopped suggesting it. The issue is now fixed (by [moving to Kramdown](https://github.com/alphagov/govuk-developer-docs/pull/2785)) but I've since grown to prefer the explicit `[]` style.

Sometimes, shortcut links are followed by parentheses - `[foo] (https://google.com)` - which is confusing to look at. Is this a broken markdown link to google.com, where we need to remove the space between the brackets, or is it a working markdown link to an aliased `[foo]: http://something.else` later on in the document? It's hard to say without scanning the whole document, so I prefer the `[]` for the disambiguation. It's not something we've specified a preference for in the GDS Way, so it's up to us to muddle our way through as best we can!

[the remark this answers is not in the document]
1pretz1: Ah, thanks for explaining on why you use the extra brackets, hadn't thought of the issue you mentioned - it makes sense!

> muddle our way through as best we can!

That's a common approach I use for general development! 😆

source/manual/sentry.html.md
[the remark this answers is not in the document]
1pretz1: Yeah it does seem quite confusing behaviour - good that it's documented!

source/manual/sentry.html.md
1pretz1: Sure 👍 With the amendments I think this section generally reads better and so I'm resolving this comment, thanks!

1pretz1 (approved):
Looks great - thank you for working through my comments! 🍏

PR #2928, opened 2021-01-18
Improve documentation about signon multiple parents

thomasleese:
This should help whoever next has to deal with a similar ticket understand exactly what to do.

[Trello Card](https://trello.com/c/7koTf4OD/2293-investigate-prevent-parent-organisation-being-swapped-for-another-in-signon-timebox-2-days)

### DIFF
--- a/source/manual/manage-sign-on-accounts.html.md
+++ b/source/manual/manage-sign-on-accounts.html.md
@@ -57,12 +57,26 @@ Then create the account:
 
 An account can have API tokens for multiple applications.
 
-## Changing parent organisation (APHA not appearing under DEFRA for users)
+## Organisations with multiple parents
 
-There is a bug in Signon for orgs that have multiple parent orgs in Whitehall. There is a [Trello card for this problem](https://trello.com/c/7koTf4OD).
+This [structure cannot currently be modelled in Signon][signon-multiple-parents-issue]
+which sometimes leads to Zendesk tickets where users cannot access things they should
+be able to.
 
-To address the problem for APHA, re-run the following [rake task](running-rake-tasks.html):
+This can be applied manually for the day (it will reset when the organisations are
+imported from Whitehall):
 
-`db:migrate VERSION=20200805142418`
+```rb
+> parent = Organisation.find_by!(slug: "...")
+> child = Organisation.find_by!(slug: "...")
+> child.update!(parent: parent)
+```
 
-Otherwise raise a PR similar to [https://github.com/alphagov/signon/pull/1481](https://github.com/alphagov/signon/pull/1481) to add a migration.
+Or if you think it makes sense to apply this permanently,
+[it can be added to the `OrganisationFetcher`][organisation-fetcher].
+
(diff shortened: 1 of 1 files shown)


source/manual/manage-sign-on-accounts.html.md
```diff
@@ -57,12 +57,23 @@ Then create the account:
...
 
-`db:migrate VERSION=20200805142418`
+```rb
+> parent = Organisation.find_by!(slug: "...")
+> child = Organisation.find_by!(slug: "...")
+> child.update!(parent: parent)
+```
 
-Otherwise raise a PR similar to [https://github.com/alphagov/signon/pull/1481](https://github.com/alphagov/signon/pull/1481) to add a migration.
+Or if it makes sense to fix this permanently,
```
ChrisBAshton, line 75: Will there ever be a case where it doesn't need fixing permanently? 🤔 I think I'd rephrase the docs along the lines of "You can quickly fix it by running this code ... and you should then make the fix permanent by adding it to the OrganisationFetcher".

thomasleese: I think there is. For example, if we consider APHA [where we've manually put in a fix](https://github.com/alphagov/signon/pull/1570) so that Defra ends up it's parent organisation. What if the Welsh or Scottish government need to work on APHA content (as they should be able to because they're parent organisations in Whitehall)? We wouldn't want to revert the permanent fix, but we could give them access temporarily.

ChrisBAshton: That makes sense 👍 I think that context is missing from the docs - could a line explaining this be added? And maybe `fixed` replaced with the word `applied`, as the former implies there's something wrong, when it sounds like often there isn't.

thomasleese: Makes sense, I've added a line about this. Do you think it goes far enough?

I don't want to be too prescriptive because as far as we know it's only happened once so far (with APHA) so there could be reasons we're not aware of at the moment for making the fix temporarily or permanently.

source/manual/manage-sign-on-accounts.html.md
```diff
@@ -57,12 +57,26 @@ Then create the account:
...
-## Changing parent organisation (APHA not appearing under DEFRA for users)
+## Organisations with multiple parents
 
-There is a bug in Signon for orgs that have multiple parent orgs in Whitehall. There is a [Trello card for this problem](https://trello.com/c/7koTf4OD).
+This [structure cannot currently be modelled in Signon][signon-multiple-parents-issue]
+which sometimes leads to Zendesk tickets where users cannot access things they should
+be able to.
 
-To address the problem for APHA, re-run the following [rake task](running-rake-tasks.html):
+This can be fixed manually for the day (it will reset when the organisations are
```
ChrisBAshton, line 66:
```suggestion
This can be applied manually for the day (it will reset when the organisations are
```

ChrisBAshton (approved):
LGTM 👍

PR #2930, opened 2021-01-20
Fix links that are relative to the site root

ChrisBAshton:
Imported links like `/public/foo.json` currently get converted to
https://docs.publishing.service.gov.uk/public/foo.json.
These links obviously 404.

Whilst technically they should be converted to
https://www.github.com/public/foo.json (if we were to take them
at face value), those links would also 404. What the developer
intends, and what GitHub obliges to do, is to convert them
relative to the repository, i.e.
https://github.com/alphagov/lipsum/blob/master/public/foo.json.
We must do the same.

This commit should fix a live issue I'm seeing on this page:
https://docs.publishing.service.gov.uk/apps/hmrc-manuals-api/extended_documentation.html#example-json

### DIFF
--- a/app/external_doc.rb
+++ b/app/external_doc.rb
@@ -59,17 +59,15 @@ class ExternalDoc
 
         href = element["href"].strip
         uri = URI.parse(href)
-        path = uri.path
 
         next if uri.scheme || href.start_with?("#")
 
-        base = if path.start_with? "/"
-                 base_url
-               else
-                 context[:subpage_url]
-               end
-
-        element["href"] = URI.join(base, href).to_s
+        if href.start_with?("/")
+          # remove preceding "/" to make links relative to the repository
+          # rather than to github.com. This is what GitHub does too.
+          href = href[1..]
+        end
+        element["href"] = URI.join(context[:subpage_url], href).to_s
       end
 
       doc
--- a/spec/app/github_repo_fetcher_spec.rb
+++ b/spec/app/github_repo_fetcher_spec.rb
@@ -12,9 +12,11 @@ RSpec.describe GitHubRepoFetcher do
 
   describe "#repo" do
     it "fetches a repo from cache if it exists" do
-      allow(CACHE).to receive(:fetch).with("all-repos", hash_including(:expires_in)) do
+      cache = double("CACHE")
+      allow(cache).to receive(:fetch).with("all-repos", hash_including(:expires_in)) do
         [public_repo]
       end
+      stub_const("CACHE", cache)
 
       repo = GitHubRepoFetcher.instance.repo("some-repo")
 
@@ -22,7 +24,9 @@ RSpec.describe GitHubRepoFetcher do
     end
 
     it "fetches a repo from GitHub if it doesn't exist in the cache" do
-      allow(CACHE).to receive(:fetch).and_yield
+      cache = double("CACHE")
+      allow(cache).to receive(:fetch).and_yield
+      stub_const("CACHE", cache)
 
       repo = GitHubRepoFetcher.instance.repo("some-repo")
 
@@ -40,7 +44,9 @@ RSpec.describe GitHubRepoFetcher do
     let(:repo_name) { "some-repo" }
 
     before :each do
-      allow(CACHE).to receive(:fetch).and_yield
+      cache = double("CACHE")
+      allow(cache).to receive(:fetch).and_yield
+      stub_const("CACHE", cache)
     end
 
     def readme_url
@@ -53,9 +59,11 @@ RSpec.describe GitHubRepoFetcher do
         .to_return(status: 200, body: "Foo")
 
       outcome = "pending"
-      allow(CACHE).to receive(:fetch) do |&block|
+      cache = double("CACHE")
+      allow(cache).to receive(:fetch) do |&block|
         outcome = block.call
       end
+      stub_const("CACHE", cache)
 
       GitHubRepoFetcher.instance.readme(repo_name)
       expect(outcome).to eq("Foo")
--- a/spec/app/github_repo_fetcher_spec.rb
+++ b/spec/app/github_repo_fetcher_spec.rb
@@ -10,13 +10,18 @@ RSpec.describe GitHubRepoFetcher do
   let(:private_repo) { double("Private repo", private_repo?: true) }
   let(:public_repo) { double("Public repo", private_repo?: false, default_branch: "master") }
 
+  def stub_cache
+    cache = double("CACHE")
+    allow(cache).to receive(:fetch).and_yield
+    stub_const("CACHE", cache)
+    cache
+  end
+
   describe "#repo" do
     it "fetches a repo from cache if it exists" do
-      cache = double("CACHE")
-      allow(cache).to receive(:fetch).with("all-repos", hash_including(:expires_in)) do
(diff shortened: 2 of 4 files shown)


app/external_doc.rb
```diff
@@ -59,17 +59,15 @@ def call
...
-                 base_url
-               else
-                 context[:subpage_url]
-               end
-
-        element["href"] = URI.join(base, href).to_s
+        if href.start_with?("/") && context[:subpage_url].to_s.start_with?("https://github.com")
+          # remove preceding "/" to make links relative to the repository
+          # rather than to github.com. This is what GitHub does too.
+          href = href[1..-1]
```
benthorner, line 68:
```suggestion
```

`URI.join` is agnostic of whether the suffix starts with a `/`.

app/external_doc.rb
```diff
@@ -59,17 +59,15 @@ def call
...
         next if uri.scheme || href.start_with?("#")
 
-        base = if path.start_with? "/"
-                 base_url
-               else
-                 context[:subpage_url]
-               end
-
-        element["href"] = URI.join(base, href).to_s
+        if href.start_with?("/") && context[:subpage_url].to_s.start_with?("https://github.com")
```
benthorner, line 65: The extra check here puzzles me, as we already seem to assume "external doc" == "GitHub doc" at the top of the file. If there's an edge case, could we clarify it with a test?

app/external_doc.rb
ChrisBAshton: Apparently it's not. If I don't remove the preceding `/`, it converts to `https://github.com/public/foo.json` (arguably correctly).

app/external_doc.rb
ChrisBAshton: That's a fair point. We _do_ only import GitHub docs, and that assumption is made in lots of places. I tacked this on as an extra fail-safe but we don't currently have a situation where we need it. I'll remove the extra check.

app/external_doc.rb
benthorner: Oh, that's strange. Just to check my understanding:

- `/public/foo.json` -> `https://github.com/public/foo.json`
- `public/foo.json` -> `https://github.com/alphagov/lipsum/blob/master/public/foo.json`

Is that right?

ChrisBAshton: That's right. Assuming `context[:subpage_url]` is something like `https://github.com/alphagov/foo`, and the document we're processing is `bar.html`, that would have a path of `https://github.com/alphagov/foo/bar.html`. Any links inside `bar.html` that begin with `/` will be relative to the site root, so `/fish` becomes `https://github.com/fish`.

And that's fine most of the time, but not in GitHub's case where people incorrectly use relative-to-root links :)

benthorner: Wow, that's surprisingly unintuitive - I only tried it with a root path. I'm seeing the same behaviour. Worse, it seems to rely on the `context[:subpage_url]` ending in a `/` 😠.

Would it actually be valid to just do `context[:subpage_url] + href[1..]`? We know the former ends in a `/`, and `href[1..]` doesn't, so I'm not sure what value `URI.join` is adding.

ChrisBAshton: It feels fairly intuitive to me - `URI.join` seems to be following the same URL interpretation rules I'd expect in browser. I'm surprised that `context[:subpage_url]` needs to end in a `/`, but on reflection I suppose it makes sense; if you're on `https://google.com/foo` and there's a link to simply `bar`, I'd expect it to take me to `/bar` rather than `/foo/bar`.

We could concatenate the strings manually, though that does risk adding lots of extraneous slashes (or worse, no slashes), depending on the two vars we're joining. I think we ought to leave it as-is.

I hadn't spotted that I could use `href[1..]` instead of `href[1..-1]` - I'll apply that! The StackOverflow I took it from probably assumed it needed to do some array housekeeping or something.

benthorner: > that does risk adding lots of extraneous slashes 

I don't think that's true. My comment here is because we're actually making two assumptions about the components being joined, and that those assumptions would be more clearly expressed with a simple concatenation.

- We're assuming `context[:subpage_url]` ends in a `/`. Otherwise the URL will be incorrect.
- We're assuming `href` begins with a `/`, as stated in the `if` clause.

So we know enough about the structure of the two strings to simply concatenate them. URI.join seems to be more focussed on replicating browser behaviour them helpfully constructing a valid URI, which is really what we're trying to do here. All of this is a relatively minor comment, though, so I agree we shouldn't get hung up on it.

spec/app/github_repo_fetcher_spec.rb
```diff
@@ -12,17 +12,21 @@
 
   describe "#repo" do
     it "fetches a repo from cache if it exists" do
-      allow(CACHE).to receive(:fetch).with("all-repos", hash_including(:expires_in)) do
+      cache = double("CACHE")
```
benthorner, line 15: The use of CACHE in this class seems like a smell, given that it's a Singleton - it shouldn't need to use a global constant to cache things. However, I see that CACHE gets used in quite a few places, so not really in scope here to change that.

I do think there's a small risk of this code regressing, if someone thinks the `allow` form is more concise in future. Perhaps we could DRY it up and preempt that by using a helper method in here?

benthorner (approved):
Nice fix - this will make the experience much more comprehensive for external docs. I've left a couple of final comments, but I think they're both minor.

I was a bit surprised and confused by the second commit, as it seems to be unrelated to the PR. But it does seem like a good change to make, as RSpec Mocks specifically advised using `stub_const` for this purpose vs. the `allow` that we had here.

spec/app/github_repo_fetcher_spec.rb
ChrisBAshton: > I was a bit surprised and confused by the second commit, as it seems to be unrelated to the PR.

Yes, totally understand that. I opened this PR for the relative links fix, then found the build failed, so tacked on a fix commit that had been on my to-do list for a while 😅 thanks for reviewing it.

ChrisBAshton:
> I was a bit surprised and confused by the second commit, as it seems to be unrelated to the PR.

Yes, totally understand that. I opened this PR for the relative links fix, then found the build failed, so tacked on a fix commit that had been on my to-do list for a while 😅 thanks for reviewing it.

app/external_doc.rb
ChrisBAshton: Hmm looking at the code again, it is true that we know what strings we're dealing with, so it would normally be safe to simply concatenate them. _However_, `context[:subpage_url]` is of type `URI::HTTP`, rather than a string, so it will still use the `URI.join` concatenation method even if we're concatenating a normal string:

```ruby
foo = URI.join("http://www.google.com", "foo")
=> #<URI::HTTP http://www.google.com/foo>
irb(main):002:0> foo + "bar"
=> #<URI::HTTP http://www.google.com/bar>

# (expected output to be `http://www.google.com/foobar`)
```

Given that we use `URI.join` for every URI creation in this file, I am inclined to leave this as-is.

spec/app/github_repo_fetcher_spec.rb
ChrisBAshton: I've just pushed up a commit to tidy this up 👍

benthorner: Thanks, that's much simpler - feel free to merge.

PR #2935, opened 2021-01-24
Replace mentions of `brew cask` with plain `brew install`

issyl0:
- Usage of `brew cask install` returns "Error: Calling brew cask install is disabled! Use brew install [--cask] instead." since Homebrew 2.6.0 (released in December). None of the software installed in these docs is both a cask and a formulae, so there's no need to explicitly pick the cask with `--cask`.

:sparkling_heart:


issyl0:
I can't reproduce the test failures locally. 🤔

kevindew (approved):
Hi @issyl0 👋   Hope you're well and enjoying life in GitHub towers. Thanks very much for this ✅ 

The test seems to be a flaky one - looks like state from previous tests is making it into an object. I've kicked off another run of your job to see if that builds green. I'll raise a quick issue for the problem while it is on my mind.

ChrisBAshton:
Thanks for the PR, Issy - and sorry about the flaky tests! Thanks for the write-up Kevin, I'll take a look.

issyl0:
No problem! I'm having a great time at GitHub, thanks, though it's crazy to realise it's been four months already!

PR #2941, opened 2021-01-28
Add section about new access controls

karlbaker02:
Add section to the docs describing the new access controls that Replatforming are putting in place to allow anyone with production access to assume admin, poweruser or user roles.

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -242,6 +242,19 @@ After your PR has been merged, someone from the `govuk-administrators`
 or `govuk-internal-administrators` group needs to deploy the
 `infra-security` project.
 
+#### Assuming all roles for users with production access
+
+In addition to the previous section's named roles, anyone with production access
+can assume `admin`, `poweruser` and `user` roles. For example, if you want to
+assume a `poweruser` role in `integration`, the role you would specify would be:
+
+- `govuk-integration-poweruser` if using the GDS CLI
+- `your.name-poweruser` if going directly through the AWS Console
+
+More generally, to assume a role directly through the AWS Console, the role you
+specify should be of the form `your.name-<role>`, where role is either `admin`,
+`poweruser` or `user`.
+
 ## 8. Use your AWS access
 
 ### First run


source/manual/get-started.html.md
```diff
@@ -242,6 +242,22 @@ After your PR has been merged, someone from the `govuk-administrators`
 or `govuk-internal-administrators` group needs to deploy the
 `infra-security` project.
 
+#### New access controls
```
richardTowers, line 245: I think having a "New access controls" section will age poorly... It's not going to be new forever!

I'd prefer to update the existing docs just to say how to access the new roles, and leave out the contextual bit around removing the RE out-of-hours rota.

source/manual/get-started.html.md
```diff
@@ -242,6 +242,22 @@ After your PR has been merged, someone from the `govuk-administrators`
...
+
+By allowing an admin to assume roles other than admin, we allow people to assume
+a role of least privilege for the work that they need to do. Similarly, allowing
+powerusers to assume an admin role allows them to address issues that would have
+previously been dealt with by an admin, such as those relating to AWS IAM.
+
+All you need to do to assume one of the specified roles is specify a role as you
+normally would. For example, if you want to assume a `poweruser` role in `integration`,
+the role you would specify is `govuk-integration-poweruser`, either directly through
+the AWS Console or via the GDS CLI.
```
richardTowers, line 259: I don't think this bit is quite right. If you're using the gds-cli, you can specify `govuk-integration-poweruser`, but if you're using the AWS Console it'll need to be `karl.baker-poweruser` (or something like that).

source/manual/get-started.html.md
karlbaker02: Sounds good 👍

source/manual/get-started.html.md
karlbaker02: This is true, good shout

karlbaker02: Done

source/manual/get-started.html.md
karlbaker02: Done

richardTowers (approved):

PR #2942, opened 2021-01-28
Update get-started docs on accessing AWS

richardTowers:
Recommend using the least privileged role available.

I think readers will be able to work out to replace integration with
staging / production, so they don't need that pointing out explicitly.


karlbaker02 (approved):
Well put, thanks Richard

PR #2948, opened 2021-02-02
Add instructions for scaling Mapit

brucebolt:
Scaling Mapit is different to other GOV.UK applications and the process has recently changed, so it should be documented.

Trello card: https://trello.com/c/W5I1wGG2


ChrisBAshton (approved):
LGTM 👍

PR #2952, opened 2021-02-03
Add instructions to fall back to google Cloud CDN

rtrinque:
Add instructions to fall back to google Cloud CDN


source/manual/fall-back-to-backup-cdn.html.md
```diff
@@ -0,0 +1,36 @@
...
+by static mirrors and not the origin of GOV.UK if you fallback to one of the backup CDN.
+
+## Procedure
+
+To switch to AWS Cloudfront:
+You will have to make 2 DNS changes to GOV.UK:
+
+1. you will have to ask JISC (very few people are authorised to make such changes) to change
+   the cname of `www.gov.uk` to the `www` AWS CloudFront distribution domain.
+   You can log into the AWS web console to find the `www` AWS CloudFront distribution domain.
```
richardTowers, line 22: Maybe we could say how to do this from the CLI, to save people having to work out exactly where to look for the domain?

I think this would do it:

```
gds aws govuk-production-readonly \
  aws cloudfront list-distributions --query "DistributionList.Items[?Comment=='WWW'].DomainName"
```

source/manual/fall-back-to-backup-cdn.html.md
```diff
@@ -0,0 +1,36 @@
...
+
+2. you will have to change the cname of `assets.publishing.service.gov.uk` to
+   the `assets` AWS CloudFront distribution domain using the usual gov.uk processes.
+   You can log into the AWS web console to find the `assets` AWS CloudFront distribution domain.
```
richardTowers, line 26:
```
gds aws govuk-production-readonly \
  aws cloudfront list-distributions --query "DistributionList.Items[?Comment=='Assets'].DomainName"

File: source/manual/fall-back-to-aws-cloudfront.html.md [hunk 3]
```diff
@@ -1,25 +0,0 @@
----
```
richardTowers, line 1: Could you put a redirect in `config/tech-docs.yml`, just in case there are any incoming links to this page from random google docs and such?

source/manual/fall-back-to-backup-cdn.html.md
```diff
@@ -0,0 +1,36 @@
...
+
+To switch to Google Cloud CDN
+To access the GCP console you will neeed to [set up your GCP account](/manual/set-up-gcp-account.html)
```
richardTowers, line 29: I've made a PR to the GCP account docs (#2955).

People reading these docs should already have access now that we've got the google group set up, so perhaps this doesn't need to call out the need to set  the access up anymore. Maybe something like:

> You can access the Google Cloud Platform console at [https://console.cloud.google.com/](https://console.cloud.google.com/). Everyone with production access should have access to GCP. If you don't, follow the docs to [set up your GCP account](/manual/set-up-gcp-account.html).

source/manual/fall-back-to-backup-cdn.html.md
```diff
@@ -0,0 +1,36 @@
...
+You will have to make 2 DNS changes to GOV.UK:
+
+1. you will have to ask JISC (very few people are authorised to make such changes) to change
+   the cname of `www.gov.uk` to point towards the loadbalancer associated with the govuk-mirror-www Cloud CDN.
+
+2. you will have to change the cname of `assets.publishing.service.gov.uk` to point towards the loadbalancer associated with the govuk-mirror-www Cloud CDN.
+   You can log into the AWS web console to find the `assets` AWS CloudFront distribution domain.
```
richardTowers, line 36: This looks like it needs updating for GCP...

richardTowers: I wonder if we could just link directly to:

https://console.cloud.google.com/net-services/loadbalancing/details/http/govuk-mirror-www?project=govuk-production

... it doesn't look like there's anything sensitive in that URL to me.

ChrisBAshton:
Superseded by [RFC 128](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-128-continuous-deployment.md).

PR #2955, opened 2021-02-04
Update Request GCP access docs

richardTowers:


source/manual/set-up-gcp-account.html.md
```diff
@@ -10,12 +10,15 @@ parent: "/manual.html"
...
-Permission to Google Cloud Platform (GCP) is managed by the GOV.UK
-Replatforming team. Access to GCP is granted when [permanent
-Production access](manual/rules-for-getting-production-access.html) is
-approved and merged to [GOV.UK user
-reviewer](https://github.com/alphagov/govuk-user-reviewer) repository.
+Permission to Google Cloud Platform (GCP) is managed through the [GOV.UK GCP
+access Google Group](https://groups.google.com/a/digital.cabinet-office.gov.uk/g/govuk-gcp-access).
 
-To request GCP access, [raise a Zendesk ticket with GOV.UK Replatforming](https://docs.publishing.service.gov.uk/manual/raising-issues-with-reliability-engineering.html#raising-a-zendesk-ticket-with-reliability-engineering). Include the user reviewer PR in the ticket body.
+Access to GCP is granted when [permanent Production access](manual/rules-for-getting-production-access.html) is
```
alex-ju, line 16: I'm afraid this may append an extra `manual` to the URL since current path will be `/manual/set-up-gcp-account.html` so it may need to be
```suggestion
Access to GCP is granted when [permanent Production access](rules-for-getting-production-access.html) is
```
or
```suggestion
Access to GCP is granted when [permanent Production access](/manual/rules-for-getting-production-access.html) is
```

alex-ju (approved):
Thanks for updating this, Richard! Makes sense to me.

source/manual/set-up-gcp-account.html.md
richardTowers: Wow, good reviewing! I actually copied this from line 15, but sure enough, it's [already broken in the live docs](https://docs.publishing.service.gov.uk/manual/set-up-gcp-account.html#request-gcp-access)!

PR #2957, opened 2021-02-05
Update reindex doc

1pretz1:
Updates the doc primarily regarding the reindex process by:

* Adding a link to the reindex Jenkins job
* Adding a list of indices available in Search API

Also corrects the strange numbered lists as they all currently start at
0.

### DIFF
--- a/source/manual/reindex-elasticsearch.html.md
+++ b/source/manual/reindex-elasticsearch.html.md
@@ -13,13 +13,13 @@ index before the new fields and types can be used.
 
 The reindexing process:
 
-0. Locks the Elasticsearch index to prevent writes to the index while data is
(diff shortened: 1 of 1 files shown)


thomasleese (commented):
Text changes look good to me! The `0.` numbering is actually a feature of Markdown, it'll automatically turn it into a correctly numbered list meaning you don't need to worry about changing the numbers if you need to reorder or add one in the middle, etc. (In fact it doesn't have to be `0.` it can be any number and it'll get automatically fixed)

1pretz1:
Ah, interesting @thomasleese. I did recognise that it was a markdown feature but assumed it could only started at 0. Updated to use lots of 1's 😄 

Thanks for pointing that out!

thomasleese (approved):

PR #2959, opened 2021-02-05
Change in out of hours

MahmudH:
As OOH rota no longer exists, this change has meant that the doc needs updating.


bilbof (commented):
Great stuff! I think the top half of this doc could do with a bit of correction too.

> Reliability Engineering are a programme in GDS

This is at the top of the doc; I'm not sure if it's correct given RE are under CDIO.

There's also reference to "the RE GOV.UK team" and channel re-govuk which we could fix.

urmishah (approved):
Look good to me

richardTowers (approved):

PR #2964, opened 2021-02-12
Update Transition docs to mention creating TLS certs

bilbof:
We can now have Fastly create a TLS certificate on our behalf
in order to transition sites to GOV.UK that already served over
HTTPS.

A doc already exists explaining how to do it:
https://docs.publishing.service.gov.uk/manual/request-fastly-tls-certificate.html

This change updates the transition docs to say we can actually
tranisition HTTPS sites to GOV.UK via Bouncer.


beccapearce (approved):
Looks good to me :)

PR #2966, opened 2021-02-16
Update doc for search analytics data

deborahchua:
We have seen this job failing on many occassions with the same issue,
so documenting this for future reference.

Trello card: https://trello.com/c/6fod7LEt/674-investigate-fix-fetch-analytics-data-for-search-api-alert

### DIFF
--- a/source/manual/alerts/fetch-analytics-data-for-search-failed.html.md
+++ b/source/manual/alerts/fetch-analytics-data-for-search-failed.html.md
@@ -22,7 +22,37 @@ This job has two parts:
    updated.
 
 If the job fails or gets stuck in part 1, you can cancel it and leave
-it to run the next day.
+it to run the next day. If you see in the Jenkins job output:
+
+```sh
+$ jenkins.util.io.CompositeIOException: Unable to delete '/var/lib/jenkins/workspace/search-api-fetch-analytics-data'.
+```
+
+The job runs a Docker container which uses `root`, and if it fails the
+files don't get cleaned up properly. You can check by running the following
+command on the machine:
+
+```sh
+$ sudo su - jenkins
+$ ls -al /var/lib/jenkins/workspace/search-api-fetch-analytics-data
+```
+
+```sh
+total 20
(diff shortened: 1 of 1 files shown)


source/manual/alerts/fetch-analytics-data-for-search-failed.html.md
```diff
@@ -22,7 +22,37 @@ This job has two parts:
...
 If the job fails or gets stuck in part 1, you can cancel it and leave
-it to run the next day.
+it to run the next day. If you see in the Jenkins job output:
+
+```sh
+$ jenkins.util.io.CompositeIOException: Unable to delete '/var/lib/jenkins/workspace/search-api-fetch-analytics-data'.
+```
+
+The job runs a Docker container which uses `root`, and if it fails the
+files don't get cleaned up properly. You can check by running the following
```
bilbof, line 32: Interesting! Do we have a piece of work tracked to fix this one?

I read in the [Trello card](https://trello.com/c/6fod7LEt/674-investigate-fix-fetch-analytics-data-for-search-api-alert) that "it had last passed on the 20th December 2020."  I had a look on Staging Jenkins and interestingly, there's a Docker container that [this job starts](https://github.com/alphagov/search-analytics/blob/master/nightly-run.sh#L11) that's been running for 7 weeks (since 20th December).
I guess that's expected, and Jenkins doesn't know about Docker so doesn't clean it up.

As you mentioned, the container runs as root.

We're running Docker version 17 on Jenkins. If we upgrade to version 20, we could run it in [rootless mode](https://docs.docker.com/engine/security/rootless/). The only problem is that it's only known to work on Ubuntu 16.04 and up (not on 14.04, which Jenkins is running 🤢), but we could give it a go.

The benefit would be that, even if the job fails, the files would be created by Jenkins, so could be cleaned up by Jenkins, too.

bilbof (approved):
Thanks for adding docs for this workaround. It's probably not worth fixing the code, given it happens only once every couple of months and we're planning on replacing it with an ECS ScheduledTask as part of replatforming over the next year.

PR #2967, opened 2021-02-16
Add manual fix for content-data-api

deborahchua:
We have an ongoing issue with content-data-api, as Platform Health
are not actively looking into this we should document the manual fix
2nd line are using for now.

Trello card: https://trello.com/c/d45mq4DS/1081-document-manual-fix-for-content-data-api-sidekiqretrysize-alert


beccapearce (approved):
Nice! ⭐

PR #2969, opened 2021-02-18
Remove SC clearance requirement for prod access

timblair:
GDS' position on blanket SC clearance has changed, and the default position is now that technologists require BPSS unless their specific role requires a higher level (which the majority of GOV.UK dev roles do not).


richardTowers (approved):
Looks good to me, thanks Tim!

VitalieMogoreanu:
should we mention BPSS then?

timblair:
No, I don't think so. BPSS checks are the _minimum_ required for anyone to be working for GDS / CO, so everyone will definitely have that.

PR #2975, opened 2021-02-22
Add documentation about Mapit caches

theseanything:


kevindew:
Thanks for putting this together @theseanything - and yikes that's a telnet command rather than some sort of task 😱  can it be done via a `manage.py`at all?

I'm thinking that this documentation is something you should *always* need to do when you update mapit's data so I wondered if it was better to transplant what you've wrote here as a step into https://github.com/alphagov/mapit/blob/master/IMPORTING-DATA.md#-2-export-new-database-to-s3. An alternative could be to link from that doc to this one, but I wonder if you need to worry about this cache outside of updating.

Very sorry for taking so long to review this 🙇

theseanything:
Thanks @kevindew!

> Thanks for putting this together @theseanything - and yikes that's a telnet command rather than some sort of task 😱 can it be done via a `manage.py`at all?

So nope there's no task. I can't see the telnet command being an issue - its nice you can clear the cache without having to worry about the application dependencies (python, etc).

> I'm thinking that this documentation is something you should _always_ need to do when you update mapit's data so I wondered if it was better to transplant what you've wrote here as a step into https://github.com/alphagov/mapit/blob/master/IMPORTING-DATA.md#-2-export-new-database-to-s3. An alternative could be to link from that doc to this one, but I wonder if you need to worry about this cache outside of updating.

Yup - I think I will add a link from those docs to this. I'm reluctant to add it in directly to the repository docs, as the instructions are specific for a our infrastructure.

kevindew:
> Thanks @kevindew!
> 
> > Thanks for putting this together @theseanything - and yikes that's a telnet command rather than some sort of task 😱 can it be done via a `manage.py`at all?
> 
> So nope there's no task. I can't see the telnet command being an issue - its nice you can clear the cache without having to worry about the application dependencies (python, etc).
> 
> > I'm thinking that this documentation is something you should _always_ need to do when you update mapit's data so I wondered if it was better to transplant what you've wrote here as a step into https://github.com/alphagov/mapit/blob/master/IMPORTING-DATA.md#-2-export-new-database-to-s3. An alternative could be to link from that doc to this one, but I wonder if you need to worry about this cache outside of updating.
> 
> Yup - I think I will add a link from those docs to this. I'm reluctant to add it in directly to the repository docs, as the instructions are specific for a our infrastructure.

The [importing data doc](https://github.com/alphagov/mapit/blob/master/IMPORTING-DATA.md) is entirely specific to our infrastructure so it's fine to be in the repo docs. The thing I find a little strange about this being in our developer-docs is that there's so much other stuff in the IMPORTING-DATA.md file already that having this in developer-docs probably misbalances things a bit.

theseanything:
> The [importing data doc](https://github.com/alphagov/mapit/blob/master/IMPORTING-DATA.md) is entirely specific to our infrastructure so it's fine to be in the repo docs. The thing I find a little strange about this being in our developer-docs is that there's so much other stuff in the IMPORTING-DATA.md file already that having this in developer-docs probably misbalances things a bit.

Hmm yeah - So I wonder whether we should move `IMPORTING-DATA.md` into the dev docs, as we have our other Mapit infra documentation there.

I'm hesitant about adding this to the repo docs because the existing documentation isn't well organised (no docs folder and I think this should be a separate doc file) and it won't be searchable within the dev docs (unless its in the README).

kevindew:
> Hmm yeah - So I wonder whether we should move `IMPORTING-DATA.md` into the dev docs, as we have our other Mapit infra documentation there.

Yeah, it'd be good if the importing docs was available here. I'm not sure it needs to moved directly - I think it'd be better placed in a docs folder.

> I'm hesitant about adding this to the repo docs because the existing documentation isn't well organised (no docs folder and I think this should be a separate doc file) and it won't be searchable within the dev docs (unless its in the README).

Sure, I think you trade a solution to those problems for adding to a growing problem in that case. We seem to be building up a pile of rather oddly specific mapit docs in this repo ([scale mapit](https://docs.publishing.service.gov.uk/manual/scale-mapit.html), [fix mapit database not available](https://docs.publishing.service.gov.uk/manual/mapit-database-not-available.html), and [unknown postgresql mapit error](https://docs.publishing.service.gov.uk/manual/alerts/unknown-error-with-postgresql-mapit.html)) and this one would seem to add to that disorganised pile.

I guess all we actually want is two Mapit docs: one on how to spin up a new instance and one to import data - which the more specific docs can slot into. But seems way out of the scope of this specific task.

theseanything:
@kevindew - I've got another [PR for Mapit](https://github.com/alphagov/mapit/pull/132) to improve doc organisation and add note to clear caches when importing data.

Agreed that a broader clean up of the docs would be good. Hopefully this an improvement, until someone has time to do that.

kevindew (approved):
Great, thanks for that 👍

PR #2976, opened 2021-02-23
Document Sentry issue actions

ChrisBAshton:
We ought to document how to merge, resolve and ignore issues, and when it is appropriate to do so.

### DIFF
--- a/source/manual/sentry.html.md
+++ b/source/manual/sentry.html.md
@@ -224,3 +224,46 @@ are configured in the [Alerts panel][].
 
 [Alerts panel]: https://sentry.io/organizations/govuk/alerts/rules/
 [sentry-configure-rate-limits]: https://sentry.io/settings/govuk/rate-limits/
+
+## Sentry issue actions
+
+In the Sentry UI, you can merge related issues together, resolve issues, or
+ignore issues (permanently or for a set time period).
+
+### Merging Sentry issues
+
+Duplicate issues can be merged by going into the project view, checking the
+boxes associated with each issue, and clicking "Merge". This should be done very
+carefully, as issues cannot be unmerged once they are merged, and the issues are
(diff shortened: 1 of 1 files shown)


source/manual/sentry.html.md
```diff
@@ -224,3 +224,46 @@ are configured in the [Alerts panel][].
...
+less noisy, but also has the advantage of marking it as a regression and emailing
+you if the issue occurs again.
+
+### Ignoring an issue
+
+When you've identified an issue and written it up as a Trello card, or are
+actively working on fixing it, it can be unhelpful for the issue to keep
+accumulating events (and potentially spamming your Slack channel). In these cases,
+you should [comment on the issue][anchor-comment] with a link to your card or PR,
+then ignore the issue.
```
thomasleese, line 259: Do we want to think about using the "Assignee" feature in Sentry as another way to determine if the issue is being worked on?

ChrisBAshton: We could do. I haven't used it a lot yet, as we already have visibility over that sort of thing on Trello, but it seems sensible to ask people to assign themselves to issues they're working on too. I'll add a line.

thomasleese: Cool, sounds good. It does feel like in many ways we're underutilising the features of Sentry, such as the performance measurements, automatic creation of GitHub/Trello cards, assigning apps to various teams, etc, etc... but this is definitely an improvement over what we had before!

ChrisBAshton: Yep, I'm hoping to set up a new GitHub integration (we have one already, but it's legacy) and start properly linking Sentry issues either with the git commit of the corresponding code, or the release tag. And hopefully having it notify the code author if there is an issue.

This PR should be ready for re-review btw :)

thomasleese (approved):

PR #2977, opened 2021-02-23
Add some of the Athena queries and snippets I frequently use

timblair:


source/manual/query-cdn-logs.html.md
```diff
@@ -183,6 +183,253 @@ ORDER BY count DESC
...
+bucket. E.g. `/some/path` --> `/www.gov.uk/some/path.html`
+
+```sql
+REPLACE(REPLACE(SPLIT_PART(url, '?', 1), '/www.gov.uk', ''), '.html', '') AS url_path
+```
+
+### Count cases of a thing
+
+```sql
+SUM(CASE WHEN ... THEN 1 ELSE 0 END)
```
kevindew, line 423: Pro-tip for if you do more presto queries, they have a [COUNT_IF](https://docs.data.world/documentation/sql/reference/aggregations/count_if.html) function to simplify these scenarios.

kevindew (approved):
Awesome thanks for sharing these Tim 👍

PR #2978, opened 2021-02-23
Replace instances of SUM(CASE WHEN ...) with COUNT_IF()

timblair:
Replace instances of `SUM(CASE WHEN ...)` with `COUNT_IF()` after @kevindew's [excellent suggestion][comment].

[comment]: https://github.com/alphagov/govuk-developer-docs/pull/2977#discussion_r581070710

### DIFF
--- a/source/manual/query-cdn-logs.html.md
+++ b/source/manual/query-cdn-logs.html.md
@@ -189,22 +189,16 @@ LIMIT 50
 SELECT
     DATE(request_received),
     COUNT(*) AS hits,
-
-                             SUM(CASE tls_client_protocol WHEN 'TLSv1'   THEN 1 ELSE 0 END) AS TLSv10,
-    ROUND(100.0 / COUNT(*) * SUM(CASE tls_client_protocol WHEN 'TLSv1'   THEN 1 ELSE 0 END), 4) AS TLSv10_perc,
-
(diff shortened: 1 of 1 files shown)


source/manual/query-cdn-logs.html.md
```diff
@@ -332,7 +325,7 @@ SELECT
     ROUND(APPROX_PERCENTILE(request_time, 0.95), 3) AS p95,
     ROUND(MAX(request_time), 3) AS p100
 FROM fastly_logs.govuk_www
-WHERE year = 2021 AND month = 2 AND date = X
+WHERE year = 2020 AND month = 12 AND date = 19
```
kevindew, line 328: Did you mean to replace the X here? I notice it was added in on some of the earlier queries such as https://github.com/alphagov/govuk-developer-docs/pull/2978/files#diff-76282ecd5f5d2c52fa5fc1d72876392a5547857ddf7688fe18c6ca03ce8c2af6R203. There's a couple more instances of this.

kevindew (commented):
Nice - that's a heck of a lot of `THEN 1 ELSE 0 END`'s removed 👍

source/manual/query-cdn-logs.html.md
timblair: Ah, that's an artefact of copying-and-pasting from a somewhere else. It's not really important, but I've pushed a change so things that don't need to be changed are left alone.

kevindew (approved):
Nice one, thanks

PR #2979, opened 2021-02-24
Update Whitehall scheduled publishing worker healthcheck docs

brucebolt:
In https://github.com/alphagov/govuk-puppet/pull/10963, we added a new healthcheck for the number of scheduled documents that are not queued. The end goal is to remove this from the generic Whitehall app healthcheck, so updating the documentation in preparation for this.

Trello card: https://trello.com/c/UAbbL13x

### DIFF
--- a/source/manual/alerts/whitehall-scheduled-publishing.html.md.erb
+++ b/source/manual/alerts/whitehall-scheduled-publishing.html.md.erb
@@ -51,3 +51,21 @@ These documents cannot be published and will cause a
 
 The situation is likely to be re-created the next day by the above process on
 the same documents/editions.
+
+## 'scheduled publications in Whitehall not queued'
+
+This alert means that the number of editions in the database which are
+scheduled to be published in the future is different from the number currently
+in the queue.
+
+Run the `publishing:scheduled:requeue_all_jobs` Rake task to requeue all
+scheduled editions:
+
+<%= RunRakeTask.links("whitehall", "publishing:scheduled:requeue_all_jobs") %>
+
+Historically, this error has happened in the staging and integration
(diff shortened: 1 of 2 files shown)


source/manual/alerts/whitehall-scheduled-publishing.html.md.erb
```diff
@@ -51,3 +51,21 @@ These documents cannot be published and will cause a
...
+## 'scheduled publications in Whitehall not queued'
+
+This alert means that the number of editions in the database which are
+scheduled to be published in the future is different from the number currently
+in the queue.
+
+Run the `publishing:scheduled:requeue_all_jobs` Rake task to requeue all
+scheduled editions:
+
+<%= RunRakeTask.links("whitehall", "publishing:scheduled:requeue_all_jobs") %>
```
1pretz1, line 64: Could we change this to point towards a rake task in Jenkins rather than a command through the console? I also notice that the `RunRakeTask` class isn't available through Whitehall's console.

brucebolt: [`RunRakeTask`](https://github.com/alphagov/govuk-developer-docs/blob/d26c216d80c3e0492ee16e1e8dc30335271bc10a/app/run_rake_task.rb#L4-L24) is a helper that generates the links to Jenkins automatically.  This should render in the dev docs as three links, one for each environment.

1pretz1: Ohh, fancy! Haven't seen that before.

1pretz1 (approved):

PR #2986, opened 2021-03-02
Document conditional request caching at the CDN

kevindew:
Trello: https://trello.com/c/P2Y4nGaw/156-document-asset-e-tags-in-the-cdn

This adds documentation on a caching process Fastly applies for
responses which return conditional headers. It can be a little
surprising that Fastly can modify a response from Origin, a 304, to
convert it to a 200.

This has been added following an incident on 21st January 2021 where
Asset Manager returned the wrong content-type. In order to resolve this
problem we had to remove 10,000 items from the Fastly cache. We
initially thought this incident was resolved after the Cache-Control
time was exhausted, however E-Tags kept the problem occurring.


JonathanHallam (approved):
Nice explanation, thanks 🎉

PR #2987, opened 2021-03-02
Bring asset documentation up to date

kevindew:
Trello: https://trello.com/c/P2Y4nGaw/156-document-asset-e-tags-in-the-cdn

This is updated to reflect that static assets are now served from
www.gov.uk and that uploaded assets aren't served from a share anymore.

I also decided to include the peculiar other thing hosted on
assets.publishing.service.gov.uk since it is rather a surprise to find
that one of our apps is proxied there as only a service to development
versions of GOV.UK. Hopefully I've explained it ok as it's a rather
confusing concept.

### DIFF
--- a/source/manual/assets.html.md
+++ b/source/manual/assets.html.md
@@ -5,10 +5,10 @@ section: Assets
 type: learn
 layout: manual_layout
 parent: "/manual.html"
-related_applications: [asset-manager]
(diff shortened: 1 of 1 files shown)


JonathanHallam (approved):
Nice one, thanks 🎉

PR #2989, opened 2021-03-03
Correct the 'Writing Pact tests' section & add 'Updating' section

ChrisBAshton:
I'd gotten my provider and consumer mixed up in #2982.
I've also added a section on changing the Provider's API, which
needs to be co-ordinated in a particular way.

### DIFF
--- a/source/manual/pact-broker.html.md
+++ b/source/manual/pact-broker.html.md
@@ -76,8 +76,21 @@ This has to be done in several stages:
 1. Write the provider pact ([example](https://github.com/alphagov/frontend/pull/2643)).
   - You'll need to write these two in conjunction with each other.
 1. Merge both of the above PRs.
-1. Configure the consumer to test against the provider as part of its deployment pipeline ([example](https://github.com/alphagov/frontend/pull/2644)).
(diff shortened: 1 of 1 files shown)


source/manual/pact-broker.html.md
```diff
@@ -76,8 +76,21 @@ This has to be done in several stages:
 1. Write the provider pact ([example](https://github.com/alphagov/frontend/pull/2643)).
   - You'll need to write these two in conjunction with each other.
 1. Merge both of the above PRs.
-1. Configure the consumer to test against the provider as part of its deployment pipeline ([example](https://github.com/alphagov/frontend/pull/2644)).
-  - You'll need to provide a way of testing against different branches of the provider, so that changes can be made to the API in future. See the example PR above.
+1. Configure the provider to test against the `master` consumer as part of its deployment pipeline ([example](https://github.com/alphagov/frontend/pull/2644)).
```
thomasleese, line 79:
```suggestion
1. Configure the provider to test against the default branch of the consumer as part of its deployment pipeline ([example](https://github.com/alphagov/frontend/pull/2644)).
```

Just to save us having to change this in the future - what do you think?

thomasleese (approved):

PR #2993, opened 2021-03-04
Improve playbook for S3 mirror sync alert.

sengi:
Improve the alert description, add some important context, describe the user impact and add a troubleshooting section.

The "service description" of the alert is being changed from "Mirror sync" to "Mirror GOV.UK content to S3" in alphagov/govuk-puppet#10971, hence the change of title while preserving the original URL.

### DIFF
--- a/source/manual/alerts/mirror-sync.html.md
+++ b/source/manual/alerts/mirror-sync.html.md
@@ -1,30 +1,69 @@
 ---
 owner_slack: "#govuk-2ndline"
-title: Mirror sync
+title: Mirror GOV.UK content to S3
 parent: "/manual.html"
 layout: manual_layout
 section: Icinga alerts
 ---
 
-This is [a script that runs hourly](https://github.com/alphagov/govuk-puppet/blob/99486124689b198120800572b331b38b87a18a6c/modules/govuk_crawler/manifests/init.pp#L220-L227) to sync the mirror contents to S3.
+## What this alert means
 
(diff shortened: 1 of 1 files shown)


source/manual/alerts/mirror-sync.html.md
```diff
@@ -1,30 +1,69 @@
...
+
+Provided www-origin and assets-origin are working perfectly, there should be no user-visible impact. If the origin has a problem however, we will fail to serve content which was published after the govuk_mirror_sync job stopped working.
+
+Fastly falls back to the S3 mirror on a per-request basis, so even intermittent unavailability or partial outages limited to specific GOV.UK features can result in user traffic being served from mirrors. In practice, even on a good day there is still a tiny proportion of user traffic which is served from the mirrors.
+
+The static mirrors are a key part of GOV.UK's business continuity plan so it's very important that they are up-to-date so that they're ready in case of a badly-timed outage.
+
+## Troubleshooting
+
+You can see [stats in Graphite](https://graphite.production.govuk.digital/dashboard/mirror_sync) of how long recent runs of govuk_mirror_sync have been taking.
```
sengi, line 25: I tried to get this working in Grafana btw, but I couldn't find a way to make it plot anything apart from the count (which is no good because we need the actual values). So this will have to do for now unless someone knows how to plot values from sparse Graphite metrics in our ancient Grafana :)

gclssvglx (approved):
Great work @sengi 👍 Very clear and detailed explanation

PR #2998, opened 2021-03-09
Update rubocop-govuk to 4.0.0.pre.1

kevindew:
Trello: https://trello.com/c/pqXbKZiC/215-test-rubocop-govuk-400pre1-prior-to-400-release

This bumps to the 4.0.0 pre release of rubocop-govuk which is part of [trialling the new rubocop-govuk release](https://github.com/alphagov/rubocop-govuk/issues/129). The gem pin to 4.0.0.pre.1 is intended to be removed once the final release of rubocop-govuk 4.0.0 is out.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ gem "html-pipeline"
 gem "kramdown"
 gem "mdl"
 
-gem "govuk_publishing_components", "~> 24.5"
+gem "govuk_publishing_components"
 gem "govuk_schemas"
 
 # GitHub API
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gem "capybara"
 gem "ffi"
 gem "rake"
 gem "rspec"
-gem "rubocop-govuk"
+gem "rubocop-govuk", "4.0.0.pre.1", require: false # Trialling pre-release
 gem "simplecov"
 gem "webmock"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -418,7 +418,7 @@ DEPENDENCIES
   faraday_middleware
   ffi
   git
-  govuk_publishing_components (~> 24.5)
+  govuk_publishing_components
   govuk_schemas
   govuk_tech_docs
   html-pipeline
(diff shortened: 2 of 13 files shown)


Gemfile
```diff
@@ -18,7 +18,7 @@ gem "html-pipeline"
 gem "kramdown"
 gem "mdl"
 
-gem "govuk_publishing_components", "~> 24.4"
+gem "govuk_publishing_components"
```
benthorner, line 21: This was introduced [here](https://github.com/alphagov/govuk-developer-docs/commit/8bacd1d9ea6048d0a8034f5a4c9e1111bf6abe8b), but I don't understand the explanation. Seems reasonable to remove.

benthorner (approved):
Looks good to me. 

(Can't say I'm a fan of things like `"/" + var -> "/#{var}"`, but it does reduce decisions.

kevindew:
> Looks good to me.
> 
> (Can't say I'm a fan of things like `"/" + var -> "/#{var}"`, but it does reduce decisions.

It's proving the biggest source of churn in this rollout - particularly on some of the madder string usages (Whitehall 😆 ). Like many rules, I'm ambivalent to it - wouldn't have picked it myself as essential but can see consistency benefits for code and with Rubocop defaults.

Thanks for the review 👍

Gemfile
kevindew: Interesting. Running `bundle install` will just pull what's from the Gemfile.lock and won't impact it.

The situation where it'd not default to installing the newest version with a `bundle update` is if there was a shared dependency of this app and govuk_publishing_components where there were version constraints. Bundler then needs to negotiate what it thinks is the appropriate version (for example if current version of govuk_publishing_components depended on an old version of Kramdown, and an older version had an unbounded dependency it might go for an old govuk_publishing_components to satisfy a new Kramdown) - these situations seem to be rather rare though.

I've just tried running a full `bundle update` to see if it downgrades, but it was taking ages so I'll check that on Monday.

kevindew: I found out - after leaving it running for ages (it was running for hours) - that the problem is because [govuk_tech_docs gem specifies sprockets 4](https://github.com/alphagov/tech-docs-gem/blob/master/govuk_tech_docs.gemspec#L49) or greater, so I guess it tries to find a govuk_publishing_components without the constraint. It turns out that constraint in govuk_tech_docs isn't needed so I'm going to open a PR to remove it.

benthorner: Oh that's interesting - nice catch! I guess that might explain the excessive run time.

PR #3001, opened 2021-03-09
Add republishing information for content-publisher

huwd:
This came up during a 2ndline shift where a govspeak component change
caused the contact element for press releases to render with different
HTML.

As a result we needed to go in and republish a lot of content publisher
content, and the how-to was missing from our docs.

I also took the chance to rearrange the complex links as footnotes
whilst I was there.


Rosa-Fox (approved):
I haven't actually seen markdown links written as variables like this before but I like it! 👍

PR #3004, opened 2021-03-10
Add instructions for hiding the Brexit link

sihugh:
If we need to use the emergency banner, then we have been cleared to remove the Brexit link in order to reduce the overall height of the banner section.


deborahchua (approved):

sihugh:
Thanks Deb!

PR #3009, opened 2021-03-11
Clarify how to deploy a govuk-puppet release

AlanGabbianelli:
- Remove mention to Carrenza (it's no more)
- Add useful links
- Reword and reorder instructions

### DIFF
--- a/source/manual/deploy-puppet.html.md
+++ b/source/manual/deploy-puppet.html.md
@@ -20,12 +20,18 @@ deploy.
 
 You can deploy Puppet to Staging and Production using the following steps:
 
-1. Deploy a newer version of govuk-puppet to staging by using the 'Deploy to Staging (Carrenza)'
-   and 'Deploy to Staging (AWS)' buttons in the Release app after clicking on the release
-   tag and reviewing the changes listed.
-1. You will either need to wait 30 minutes or read about [convergence](#convergence).
-   You should monitor Icinga and Smokey, and test anything you're concerned about.
+1. In the [Release app][release-app-puppet], select the release tag and review
+   the changes listed.
+1. If your change on integration appears as expected, deploy a newer version of
+   govuk-puppet to staging by using the **Deploy to Staging** button.
+1. Wait 30 minutes or read about [convergence](#convergence).
+   You should monitor [Staging Icinga][icinga-staging] and [Staging Smokey][smokey-staging],
+   and test anything you're concerned about.
 1. Repeat the above steps for Production.
+
+[release-app-puppet]: https://release.publishing.service.gov.uk/applications/puppet
+[icinga-staging]: https://alert.blue.staging.govuk.digital/
+[smokey-staging]: https://deploy.blue.staging.govuk.digital/job/Smokey
 
 Puppet is automatically deployed to integration by a combination of the [integration-puppet-deploy job on Jenkins CI](https://ci-deploy.integration.publishing.service.gov.uk/job/Deploy_Puppet/) and [Deploy Puppet job on Jenkins Deploy](https://deploy.integration.publishing.service.gov.uk/job/Deploy_Puppet/) jobs.
 


source/manual/deploy-puppet.html.md
```diff
@@ -20,13 +20,19 @@ deploy.
...
 You can deploy Puppet to Staging and Production using the following steps:
 
-1. Deploy a newer version of govuk-puppet to staging by using the 'Deploy to Staging (Carrenza)'
-   and 'Deploy to Staging (AWS)' buttons in the Release app after clicking on the release
-   tag and reviewing the changes listed.
-1. You will either need to wait 30 minutes or read about [convergence](#convergence).
-   You should monitor Icinga and Smokey, and test anything you're concerned about.
+1. In the [Release app][release-app-puppet], click on the release tag and review
+   the changes listed.
+1. If all looks right, deploy a newer version of govuk-puppet to staging by using
```
huwd, line 25:
```suggestion
1. If your change on integration appears as expected, deploy a newer version of govuk-puppet to staging by using
```

Take it or leave it, I remember Jon giving me some good advice about sentences that didn't make the subject explicit for clairity.

huwd (approved):

source/manual/deploy-puppet.html.md
```diff
@@ -20,13 +20,19 @@ deploy.
 
 You can deploy Puppet to Staging and Production using the following steps:
 
-1. Deploy a newer version of govuk-puppet to staging by using the 'Deploy to Staging (Carrenza)'
-   and 'Deploy to Staging (AWS)' buttons in the Release app after clicking on the release
-   tag and reviewing the changes listed.
-1. You will either need to wait 30 minutes or read about [convergence](#convergence).
-   You should monitor Icinga and Smokey, and test anything you're concerned about.
+1. In the [Release app][release-app-puppet], click on the release tag and review
```
jonathanglassman, line 23: To keep in line with GOV.UK style, replace "click on" with "select". This is because not all users "click", especially if they are using assistive tech.

source/manual/deploy-puppet.html.md
```diff
@@ -20,13 +20,19 @@ deploy.
...
+   the changes listed.
+1. If your change on integration appears as expected, deploy a newer version of 
+   govuk-puppet to staging by using the 'Deploy to Staging' button.
```
jonathanglassman, line 26: GOV.UK style point - change Deploy to Staging from code format to bold format. UI elements are always in bold text.

source/manual/deploy-puppet.html.md
```diff
@@ -20,13 +20,19 @@ deploy.
...
+1. Wait 30 minutes or read about [convergence](#convergence).
+   You should monitor [Staging Icinga][icinga-staging] and [Staging Smokey][smokey-staging],
+   and test anything you're concerned about.
 1. Repeat the above steps for Production.
```
jonathanglassman, line 30: I know this ins't part of your changes, but I don't think you need line 30. Line 21 already makes it clear that the following steps are for both environments.

jonathanglassman (changes requested):
@AlanGabbianelli looks good to me. Just a few minor style changes.

source/manual/deploy-puppet.html.md
AlanGabbianelli: If we removed line 30 we should change some wording and possibly add extra points for production because all these steps (including the links) are specific to staging.

AlanGabbianelli:
@jonathanglassman thanks for reviewing this. 👍 I addressed the two GOV.UK style points. ⭐

Regarding the line that says to repeat the steps for production, I'd like to keep it because otherwise, we should also either reword the previous steps or add new steps specific for production. Let me know what you think!

source/manual/deploy-puppet.html.md
jonathanglassman: ok no problem, leave it as is

jonathanglassman (approved):

PR #3016, opened 2021-03-17
Updated the diagram to the latest colourful version

VitalieMogoreanu:
Updated the diagram to the latest version, also changed the viewer from draw.io to diagrams.net to fix the iframe

### DIFF
--- a/source/manual/architecture.html.md
+++ b/source/manual/architecture.html.md
@@ -7,13 +7,13 @@ layout: manual_layout
 parent: "/manual.html"
 ---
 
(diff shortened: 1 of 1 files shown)


richardTowers:
Running this locally, I get asked to auth:

image

Once I have, it's a bit small:

image

... I wonder whether we should export it as an SVG instead... but maybe that's a yak shave too far.

VitalieMogoreanu:
> Running this locally, I get asked to auth:
> 
> image
> 
> Once I have, it's a bit small:
> 
> image
> 
> ... I wonder whether we should export it as an SVG instead... but maybe that's a yak shave too far.

yeah it does that it's to authenticate diagram.net, not your user... 
it is too small, there is a link to open full screen - we could remove the iframe and perhaps this whole page. I don't think there is an easy way to navigate to this page anymore

maxf:
@VitalieMogoreanu - shall we close this, it's been a while? Otherwise I'm happy to look into it

PR #3019, opened 2021-03-18
Add howto rename master to main branches

chao-xian:


source/manual/howto-switch-default-branch.html.md
```diff
@@ -0,0 +1,26 @@
+---
+owner_slack: "#govuk-developers"
+title: How to switch a git default branch from master to main
```
thomasleese, line 3:
```suggestion
title: How to change the default branch of a Git repo
```

It might be nice to keep the docs generic enough that they can stay around - but maybe that's unnecessary...?

source/manual/howto-switch-default-branch.html.md
```diff
@@ -0,0 +1,26 @@
...
+---
+
+These steps are for switching the default branch of a repo to `main`.
+
+1. Rename the `master` branch in GitHub to `main` for that repo. See [instructions from Github for renaming an existing default branch](https://github.com/github/renaming#renaming-existing-branches).
+1. Update the [Release app](https://release.publishing.service.gov.uk/applications/):
+  - Go to the app entry from the [listing](https://release.publishing.service.gov.uk/applications/).
+  - Hit the `Edit` tab.
+  - Switch under the dropdown for `GitHub repository default branch`.
+1. [Reset the Jenkins build number](https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-on-jenkins-ci.html#fixing-the-build-number).
```
thomasleese, line 16: We might want to be explicit here on where to get the build number from, i.e. it has to be the latest `master` build number plus 1.

source/manual/howto-switch-default-branch.html.md
```diff
@@ -0,0 +1,26 @@
...
+1. Delete the master branch locally, otherwise you might rebase the wrong branch by mistake!
```
thomasleese, line 17: GitHub includes some instructions on how to do this that it might be useful to link to: https://docs.github.com/en/github/administering-a-repository/renaming-a-branch#updating-a-local-clone-after-a-branch-name-changes

Rather than deleting the `master` branch locally, it renames it to `main` and keeps it up to date with origin.

source/manual/howto-switch-default-branch.html.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Additional things to check
+
+- Update references in Jenkins to pulling down the `master` branch of the app.
```
thomasleese, line 21: It might be worth adding some instructions on how to use the GitHub search feature to find these. (https://github.com/alphagov/govuk-puppet/search?l=HTML%2BERB&q=master)

thomasleese (commented):
I've left some comments, I don't know how much effort we want to put into this docs since in theory they're only likely to be needed once. Perhaps share this to #govuk-developers and see what other people think (as in, would they find it easy/hard to follow through the docs?).

source/manual/howto-switch-default-branch.html.md
kevindew: We'd need to update release to accept any input for default branch if we choose to do that. It's a pretty trivial UI change though. I decided against having it being free text as it seemed very unlikely to need other options.

thomasleese: That's true, it's probably not worth it really.

thomasleese: I'd still suggest capitalisation of `Git` and use the phrase `change` rather than `switch` though (but this is personal preference).

chao-xian: Ok left it specific, but upper case G on Git.

source/manual/howto-switch-default-branch.html.md
chao-xian: Great suggestion :+1:

source/manual/howto-switch-default-branch.html.md
chao-xian: Ooooh! 😍

source/manual/howto-switch-default-branch.html.md
chao-xian: Nice! Done

chao-xian:
@thomasleese apart from updating Pact tests and other Jenkins references, I got this to work for short-url-manager by following these steps myself. Probably not worth asking others if they can follow along then :)

Thanks for your review!

source/manual/howto-switch-default-branch.html.md
thomasleese: I think you may have added `(the last `master` build +1)` to the wrong line, I think it should be this one rather than the one above.

chao-xian: 🤦

thomasleese:
Looks good, I just had a final thought, could we link to https://trello.com/c/xVhGGzOE/225-change-default-branch-to-main somewhere to remind people to check this to see which apps need switching, and to update it once they have switched an app over.

chao-xian:
@thomasleese good idea - done.

thomasleese (approved):
Cool this looks good, we can always amend it later if we find developers need extra pointers! I'm hoping we won't need this page to stay around for too long though.

PR #3023, opened 2021-03-22
Update link to Special Route Publisher's Jenkins task

brucebolt:
This previously linked to the "publish all special routes" task in staging. Updating to link to both available tasks in all environments.

Trello card:  https://github.com/alphagov/govuk-developer-docs/pull/new/update-special-routes

### DIFF
--- a/source/manual/publish-special-routes.html.md
+++ b/source/manual/publish-special-routes.html.md
@@ -12,4 +12,7 @@ The [Special Route Publisher](https://github.com/alphagov/special-route-publishe
 
 All new special routes should be created in this tool, and existing special routes should be moved as necessary.
 
-The [Publish special routes Jenkins job](https://deploy.blue.staging.govuk.digital/job/Publish_Special_Routes/) will re-publish all defined special routes to the Publishing API.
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
Looks good to me, although there is a small accessibility issue whereby the link text is not unique for each link. It might be good to see if we can reuse the [`terse_links` method](https://github.com/alphagov/govuk-developer-docs/blob/e3677f7342b6247946617c2985edf1d135a320e9/app/run_rake_task.rb#L26-L45), but for custom Jenkins jobs?

(I suspect it's more work than it's worth, so am approving as-is).

brucebolt:
> Looks good to me, although there is a small accessibility issue whereby the link text is not unique for each link. It might be good to see if we can reuse the [`terse_links` method](https://github.com/alphagov/govuk-developer-docs/blob/e3677f7342b6247946617c2985edf1d135a320e9/app/run_rake_task.rb#L26-L45), but for custom Jenkins jobs?
> 
> (I suspect it's more work than it's worth, so am approving as-is).

Nice idea.  I'll see if I can make this with a simple change (as another PR).

PR #3024, opened 2021-03-22
Add helper to render links to Jenkins jobs

brucebolt:
We already have `RunRakeTask` that renders links to the Jenkins rake task job in all environments, so this adds a similar feature for other Jenkins jobs.


ChrisBAshton (commented):
This is looking great! I wonder if we could push up one final commit, turning `RunRakeTask` into a proxy that calls `RunJenkinsTask`?

brucebolt:
> This is looking great! I wonder if we could push up one final commit, turning `RunRakeTask` into a proxy that calls `RunJenkinsTask`?

I've tried a few possible ways of doing this, but everything is making the code more complex.  The main issue is that the URLs for Rake tasks are application dependant, whereas they aren't for Jenkins jobs, so we end up with lots of conditional logic.

ChrisBAshton (approved):
Thanks for looking 👍  happy to approve, maybe we'll find a way of DRYing these classes later.

PR #3033, opened 2021-03-24
Fix rendering of app pages

ChrisBAshton:
In 21b47c9a2ac6d4805fdaabd6a5480c706377cea3, we switched back to
Redcarpet for markdown rendering, removing the Kramdown dependency.

Redcarpet takes into account spacing when rendering HTML; if it
sees some indented HTML, it assumes it is markdown and attempts to
render it as a code block. This was leading to a very broken page:

> borked

We now call `.strip` to remove indentation from the output, which
fixes Redcarpet's parsing of the HTML:

> good


sihugh (approved):
Nice catch, thanks 👍

PR #3035, opened 2021-03-24
Remove fabric tasks from a few docs

barrucadu:
A lot of the fab tasks are one or two commands, and the only utility the script gives is not needing to SSH into a machine.

For those cases, just replace the task with the actual commands it runs.

I've left references to `fab` where the behaviour is quite complex (eg, [mysql.reset_slave](https://github.com/alphagov/fabric-scripts/blob/master/mysql.py#L105)), or where it's used to run a task across a `class` of machines.


edwardkerry (approved):
sorry to see `bodge_unicorn` go though

PR #3038, opened 2021-03-25
Document some gotchas with role names when deploying Terraform.

sengi:
`gds-cli` has some usability issues in this area which have caused confusion recently (to another user and to me when troubleshooting the issue).

For now, document the issue so that at least it's easier to discover that there are similar, easily-confused role names and where to find the canonical source of which one to use.

Tested: ran locally; formatting looks ok.


sengi:
I think this is obsolete since the roles were consolidated, and at any rate the issue hasn't come up again since.

PR #3057, opened 2021-04-12
Improve introductory paragraph on translations

thomasleese:
We want to draw attention to the specific syntax of the tags to ensure the right ones are chosen. Hopefully this should avoid the issue we experienced recently where we had to change one of the live tags on the site: https://github.com/alphagov/publishing-api/pull/1910

### DIFF
--- a/source/manual/add-translation-whitehall.html.md
+++ b/source/manual/add-translation-whitehall.html.md
@@ -6,11 +6,14 @@ layout: manual_layout
 section: Publishing
 ---
 
-Useful links:
+Before starting, it’s worth checking the language tag that is being added is correct: it should be compatible with the [IETF Language Tag syntax][] and use correct [ISO 639][] codes (with optional [ISO 15924][] or [ISO 3166][] subtags where appropriate). This is important to ensure maximum compatibility with accessibility feature such as screen readers. Some other useful reading includes [Rails Translation Manager][] and [Whitehall Internationalisation guide][].
 
-- [List of ISO 639-1 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)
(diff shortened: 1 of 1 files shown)


source/manual/add-translation-whitehall.html.md
```diff
@@ -6,11 +6,19 @@ layout: manual_layout
...
 ---
 
-Useful links:
-
-- [List of ISO 639-1 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)
-- [Rails Translation Manager](https://github.com/alphagov/rails_translation_manager)
-- [Internationalisation guide](https://github.com/alphagov/whitehall/blob/master/docs/internationalisation_guide.md)
+Before starting, it’s worth checking the language tag that is being added is correct:
+it should be compatible with the [IETF Language Tag syntax][] and use correct
+[ISO 639][], [ISO 15924][] or [ISO 3166][] codes. This is important to ensure maximum
```
ChrisBAshton, line 11: I was confused as to why we were linking off to three different standards. My understanding from the [linked Wikipedia page](https://en.wikipedia.org/wiki/IETF_language_tag#Syntax_of_language_tags) is that we always provide the ISO-639 two-letter language code, and that we sometimes need additional subtags on top of that?

```suggestion
[ISO 639][] codes (with optional [ISO 15924][] or [ISO 3166][] subtags where appropriate). This is important to ensure maximum
```

thomasleese: That's fair, I wasn't completely happy about the wording of that either, I'll use your suggestion.

ChrisBAshton (approved):
LGTM 👍 thanks

PR #3059, opened 2021-04-12
These alerts have been removed

MuriloDalRi:
These alerts have been removed in https://github.com/alphagov/govuk-puppet/pull/11034

https://trello.com/c/nfzw7FvL/2458-remove-app-memory-alerts-2


edwardkerry (approved):

thomasleese:
I think it would be worth changing the wording of the commit message here, I think if someone saw "These alerts have been removed" it wouldn't be clear exactly which alerts they are referring to. Perhaps worth including a link to https://github.com/alphagov/govuk-puppet/pull/11034 in the description of the commit as well.

PR #3061, opened 2021-04-13
Update docs to reflect route updates within Router

karlbaker02:
This PR updates the docs to reflect the recent changes to Router and Router API. Router API previously called the `/reload` endpoint of every Router instance to perform a full route reload; this has now been superseded with [Router polling for and applying updates itself](https://github.com/alphagov/router/commit/811ed82d0ecc23b8077ed5dd2c460ccdfe2808ab), which was a result of [RFC 135](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-135-updating-routes-in-router.md).

### DIFF
--- a/source/manual/architecture-deep-dive.html.md
+++ b/source/manual/architecture-deep-dive.html.md
@@ -145,10 +145,12 @@ from a MongoDB database of all known routes.
 
 The MongoDB database is written to via [router-api] whenever a route is
 added or changed: the Content Store (which we'll cover later) talks to
-Router API [directly][register-route] to do this. Router API then
-[sends POST requests to every instance of Router][router-api-updating-router],
-whereupon each [Router then reloads all routes][router-reloading-routes]
-from MongoDB.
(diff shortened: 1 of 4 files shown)


fredericfran-gds (commented):
looks good, shall we warn about small inconsistency period between the router instances or the polling period is so small that the probability this is experienced is very small?

karlbaker02:
> looks good, shall we warn about small inconsistency period between the router instances or the polling period is so small that the probability this is experienced is very small?

This is a good point, though I would say it's a) no different than before (as Router API would call Router instances sequentially and inevitably result in variations when updates occur) and b) there was quite a bit of discussion around this [in the RFC](https://github.com/alphagov/govuk-rfcs/pull/135/files#diff-1d27edce5f0ef80cff38cf26ea71bad0690f4157b81c6efb51c8ca79a99e8a57R27).

fredericfran-gds (approved):
cool, thanks for answering my questions.

PR #3062, opened 2021-04-13
Nicer ssh port forwarding instructions

richardTowers:
This gets rid of two uses of `ssh -A` (which is bad practice) and simplifies things significantly.

There's one more use of `ssh -A` in `source/manual/clone-mysql-slave.html.md`, but I don't understand that doc so I'm leaving it alone.


sengi (approved):
LGTM. Thanks for updating these.

PR #3058, opened 2021-04-12
Rewrite README and simplify GitHub Token setup

benthorner:
This follows the changes we're applying to other GOV.UK repos. My
original motivation for this change was just to simplify the GitHub
setup, so if the rest of this PR proves contentious, we could just
focus on that commit instead.

PR #3073, opened 2021-04-21
Ignore .env

ChrisBAshton:
Follow-on from #3058.

We have instructions in the README advising devs to create a .env
file for local development. If they're not careful, this file
could be committed to GitHub and they'd compromise their access
token. We should ignore it to ensure that can't happen.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@
 # Ignore bundler config
 /.bundle
 
+# Ignore sensitive files
+.env
+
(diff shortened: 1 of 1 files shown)


brucebolt (approved):

benthorner:
Nice catch, thanks @ChrisBAshton!

(I have a global `.gitignore` for this one, so I didn't catch it in my original change.)

Issue #3083, opened 2021-04-28
Broken link on 'Get started on GOV.UK'


ESKYoung:
# Problem with ['Get started on GOV.UK'](https://docs.publishing.service.gov.uk/manual/get-started.html)

There's a broken link (link text `hieradata/integration.yaml`) in the ["Create a user to SSH into integration"](https://docs.publishing.service.gov.uk/manual/get-started.html#create-a-user-to-ssh-into-integration) section.

AlanGabbianelli:
Fixed in https://github.com/alphagov/govuk-developer-docs/pull/3085

PR #3085, opened 2021-04-29
Remove reference to non-existent file

AlanGabbianelli:
The `hieradata/integration.yaml` file is no more (we completely migrated to AWS).


ChrisBAshton (approved):
Resolves #3083.

PR #3115, opened 2021-06-01
Update Zendesk documentation to reflect changes to DGU support

benjamineskola:
GOV.UK 2nd line no longer provides first-line support for data.gov.uk, updating the documentation to reflect that.

### DIFF
--- a/source/manual/zendesk.html.md
+++ b/source/manual/zendesk.html.md
@@ -7,8 +7,7 @@ section: 2nd line
 type: learn
 ---
 
-2nd line Zendesk tickets are technical errors reported by our users - including government publishers -
-and general queries from users of data.gov.uk ("DGU"), for whom 2nd line provides 1st line support.
+2nd line Zendesk tickets are technical errors reported by our users, including government publishers.
 
 ## Get started
 


ChrisBAshton (approved):
Looks good to me. Do we know where this leaves documentation such as https://docs.publishing.service.gov.uk/manual/data-gov-uk-2nd-line.html? Is it a case of User Support referring to this documentation (and thus we still have to maintain it) or do they have their own docs, in which case we could consider its removal?

(I'd err on the side of keeping the docs for a few months longer anyway, whilst we monitor how many tickets get referred back to 2nd line if they're deemed too complex).

benjamineskola:
> Do we know where this leaves documentation such as https://docs.publishing.service.gov.uk/manual/data-gov-uk-2nd-line.html?

I'm not certain — perhaps it's handy to keep around just in case, at least for a while. I don't really know which (if any) of those things we'll ever need to do in future. Maybe a note at the top of that page indicating that we no longer do first-line support, so it's clear it wouldn't be needed in most situations?

ChrisBAshton:
> I'm not certain — perhaps it's handy to keep around just in case, at least for a while. I don't really know which (if any) of those things we'll ever need to do in future. Maybe a note at the top of that page indicating that we no longer do first-line support, so it's clear it wouldn't be needed in most situations?

Probably worth checking with our DM before making any decisions on the support docs - it could be that this has been discussed already. But I think this PR is good to go as-is.

PR #3116, opened 2021-06-01
Accuracy updates to applications list

kevindew:
This change:

- Removes govuk-alert-tracker application, since it is no longer a running tool
- Adds special-route-publisher as an application, since that utility seems a surprising omission from this list despite being on the borderline of whether it could be considered to be an app
- Removes the broken "Deploy script" link for apps that have a production hosting of none
- Removes a duplicated entry of step-by-step-browser in the apps list
- Fixes a comment on what belongs in "Utilities", that no longer reflected usage.

Further details in the commits.

### DIFF
--- a/app/app.rb
+++ b/app/app.rb
@@ -123,7 +123,7 @@ class App
   end
 
   def deploy_url
-    return if app_data["deploy_url"] == false || production_hosted_on == "heroku"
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
Great stuff 👍

PR #3117, opened 2021-06-01
Remove Carrenza and UK Cloud as hosting providers

kevindew:
We no longer have any apps hosted in these environments, nor plans to add new ones, thus the code for them can be removed.


ChrisBAshton (approved):
LGTM 👍 thanks for doing this.

PR #3124, opened 2021-06-03
Change instructions for logging into Fastly

nicholsj:
We won't have a shared 2nd Line account soon, so removing mention of it

### DIFF
--- a/source/manual/request-fastly-tls-certificate.html.md
+++ b/source/manual/request-fastly-tls-certificate.html.md
@@ -10,7 +10,7 @@ related_applications: [bouncer, transition]
 When transitioning HTTPS domains, these are the steps to request a TLS
 certificate from Fastly.
 
-1. Use the 2nd line account to login to Fastly.
+1. Login to Fastly.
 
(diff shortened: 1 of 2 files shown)


source/manual/request-fastly-tls-certificate.html.md
```diff
@@ -10,7 +10,7 @@ related_applications: [bouncer, transition]
 When transitioning HTTPS domains, these are the steps to request a TLS
 certificate from Fastly.
 
-1. Use the 2nd line account to login to Fastly.
+1. Login to Fastly.
```
leenagupte, line 13: How does a new person go about getting a Fastly account?

nicholsj: Good point. Seems its fairly standard to say "speak to a member of the senior tech team" so I'll add that

thomasleese (approved):
Changes seem good, I think we'll also need to add some instructions on how to deploy changes to the CDN using your own API key rather than the shared one, but I'll write a card for that.

PR #3128, opened 2021-06-10
Set Corona services as dependencies team for 3 app

GDSNewt:
It's more helpful that #govuk-corona-services-tech gets to see information about
these Dependabot PRs.

[trello](https://trello.com/c/XtUQU9Dj/426-assign-this-team-as-dependabot-maintainers-of-collections-collections-publisher-and-local-links-manager)

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -20,7 +20,7 @@
   continuously_deployed: true
   type: Publishing apps
   team: "#govuk-platform-health"
-  dependencies_team: "#govuk-platform-health"
+  dependencies_team: "#govuk-corona-services-tech"
   production_hosted_on: aws
   actors:
   - Publisher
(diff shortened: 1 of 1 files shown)


gclssvglx (commented):
Nice one - thanks @GDSNewt!

Just checking - if we only change the `dependencies_team` field then we'd see 
`#govuk-platform-health owns the app. #govuk-corona-services-tech is responsible for updating its dependencies.` [on the apps manual page](https://docs.publishing.service.gov.uk/apps/collections.html) according to the logic [here](https://github.com/alphagov/govuk-developer-docs/blob/ec5e0afaa3df4daefc2d1cebc8a34b4872d83486/source/partials/application/_ownership.html.erb).  Is that what we want to do, or do we want that page to just say `Ownership #govuk-corona-services-tech`?  If so, we'll change both `team` and `dependencies_team`.

gclssvglx (approved):
Ok - I can see that we just want to change the `dependencies_team`.  Thanks @GDSNewt .

PR #3129, opened 2021-06-14
Add documentation for modifying a Whitehall changelog

maxgds:
Updating a change note is a task that comes up on 2nd Line from time to time and there was no guidance on how to action it if the content is in Whitehall. I figured out how to make it happen with a combination of this Publishing API documentation and the existing [change note remover](https://github.com/alphagov/whitehall/blob/827a41d05358ad9bc36cb443ea1d93af7ab73cae/lib/data_hygiene/change_note_remover.rb)

While this may not be the ideal approach it's a start, it works, it can be iterated on and it didn't require me making Whitehall run on my machine or learning about migrations.

### DIFF
--- a/source/manual/howto-modify-change-note.html.md
+++ b/source/manual/howto-modify-change-note.html.md
@@ -1,13 +1,14 @@
 ---
 owner_slack: "#govuk-2ndline"
-title: Modify a change note in Publishing API
+title: Modify a change note in Publishing API or Whitehall
 section: Publishing
 layout: manual_layout
 parent: "/manual.html"
 ---
 
-Spelling mistakes can creep into change notes, and we are often asked
-to correct them.
+Spelling mistakes can creep into change notes, and we are often asked to correct them. The instructions below cover doing this task in the Publishing API and in Whitehall - if your change note lives in Whitehall and is updated in the Publishing API it could be overwritten later.
+
+## Publishing API
 
 Connect to the Publishing API console:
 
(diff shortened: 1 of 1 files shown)


source/manual/howto-modify-change-note.html.md
```diff
@@ -1,13 +1,14 @@
...
-title: Modify a change note in Publishing API
+title: Modify a change note in Publishing API or Whitehall
 section: Publishing
 layout: manual_layout
 parent: "/manual.html"
 ---
 
-Spelling mistakes can creep into change notes, and we are often asked
-to correct them.
+Spelling mistakes can creep into change notes, and we are often asked to correct them. The instructions below cover doing this task in the Publishing API and in Whitehall - if your change note lives in Whitehall and is updated in the Publishing API it could be overwritten later.
```
theseanything, line 9: Not sure I understand what's meant by "if your change note lives in Whitehall and is updated in the Publishing API it could be overwritten later." - as in if you change it Publishing API but not Whitehall, the change not in Publishing API will overwritten by the note in whitehall?

source/manual/howto-modify-change-note.html.md
```diff
@@ -28,3 +29,51 @@ $ d.live.change_note.update(note: "Foo")
...
+
+## Whitehall
+
+Connect to the Whitehall console (test your change on Integration first):
+
+```
+govuk-connect app-console -e integration whitehall_backend/whitehall
+```
+
+This finds the document, finds the correct edition within it, and updates it. You need the document ID, a string to search for to identify the right change log entry (search is fuzzy, be precise to get the right one), and the updated content
```
theseanything, line 41: Whats the first "This" referring too? (`This finds the document...`). I was slightly confused because I thought you meant the command above.

theseanything (approved):
Few readability comments - but overall looks good! Thanks for adding!

source/manual/howto-modify-change-note.html.md
maxgds: Yep, that's the impression I got from Richard. My understanding is tenuous, tbh!

source/manual/howto-modify-change-note.html.md
maxgds: Ah, it's the whole list of commands, I'll update to clarify.

PR #3142, opened 2021-06-21
Document recommended fix and link to card

ChrisBAshton:
Replatforming have this on their radar. Running the command is more
efficient than reprovisioning the instance, so I've moved it to the
top of the suggestions.


rtrinque (approved):
Looks good

PR #3147, opened 2021-06-23
Document Alert: GOV.UK Account SLOs - Error rates

huwd:
[Trello](https://trello.com/c/Crj6Z5SW/841-add-slack-alerts-for-when-we-drop-to-50-and-25-remaining-error-budget)

Documentation that will appear on the Icinga alerts for the GOV.UK
Account Error & Latency rate warnings.

See [govuk-puppet##11171][1]

[1]: https://github.com/alphagov/govuk-puppet/pull/11171

### DIFF
--- a/source/manual/alerts/slo-rate-checks-account.html.md
+++ b/source/manual/alerts/slo-rate-checks-account.html.md
@@ -1,0 +1,44 @@
+---
+owner_slack: "#govuk-accounts-tech"
+title: SLO Error Rate on GOV.UK Accounts has exceeded an acceptable threshold across the last 10 minutes
+section: Icinga alerts
+layout: manual_layout
+parent: "/manual.html"
+---
+
+The GOV.UK Account team has defined Service Level Objectives (SLOs) for
+reliability of the `account-api`.
+
+You can read more about these in [GOV.UK Accounts Service Level
+Objectives](https://docs.google.com/document/d/1HVXlyy2vdy9U0O-_7Nw8_Oso7VXQeKgJ2_-bi4pKlz0/edit)
+
+## HTTP Error on GOV.UK Account has exceeded an acceptable threshold across the last 10 minutes
+
+This warning triggers when over 1% of all HTTP status codes are deemed "bad
+responses" as a proportion of all responses within a rolling 10 minute time
+period.
+
+This is an early warning that, if these errors continue at a consistent rate,
+the GOV.UK Account team would spend its entire error budget across the full
+monitoring period.
+
+This warning automatically triggers a message to the `#govuk-accounts-tech`
+slack channel.
+
+No further action is required from GOV.UK 2ndline or any other team.
+
+## Latency on GOV.UK Account has exceeded an acceptable threshold across the last 10 minutes
+
+This warning triggers when over 1% of all HTTP responses are deemed
+"too slow" as a proportion of all responses within a rolling 10 minute time
+period.
+
+This is an early warning that, if these errors continue at a consistent rate,
+the GOV.UK Account team would spend its entire error budget across the full
+monitoring period.
+
+This warning automatically triggers a message to the `#govuk-accounts-tech`
+slack channel.
+
+No further action is required from GOV.UK 2ndline or any other team.
+


source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,45 @@
...
+reliability of the `account-api`.
+
+You can read more about these in [GOV.UK Accounts Service Level
+Objectives](https://docs.google.com/document/d/1HVXlyy2vdy9U0O-_7Nw8_Oso7VXQeKgJ2_-bi4pKlz0/edit)
+
+## Warning - "HTTP Error on GOV.UK Accounts has exceeded an acceptable threshold across the last 10 minutes"
+
+This triggers when, within a rolling 10 minute window, over 1% of all HTTP
+status codes were deemed "bad responses", as a proprotion of all responsed in
+that period.
```
jonathanglassman, line 19: Suggest following change: "This warning triggers when over 1% of all HTTP status codes are deemed "bad responses" as a proportion of all responses within a rolling 10 minute time period."

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,45 @@
...
+
+This is intended as an early warning that, if these errors continued at a
+consistent rate, the GOV.UK Accounts team is on route to spend it's entire
```
ChrisBAshton, line 22:
```suggestion
consistent rate, the GOV.UK Accounts team is en route to spend its entire
```

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,45 @@
...
+error budget of 1% across the full 4 week monitoring period.
```
jonathanglassman, line 23: Suggest following change: "This is an early warning that, if these errors continue at a consistent rate, the GOV.UK Account team would spend its entire error budget of 1% across the full 4 week monitoring period."

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,45 @@
+---
+owner_slack: "#govuk-account-tech"
+title: GOV.UK Account SLO error rate exceeded 1% in 10 minute window
+section: Icinga alerts
```
ChrisBAshton, line 4: Is this an Icinga alert, or do they stem from Prometheus?

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,45 @@
...
+This warning automatically triggers a message to the `#govuk-accounts-tech`
+slack channel.
+
+No further action is required from GOV.UK 2ndline or any other team.
+
+## Warning - "Latency on GOV.UK Accounts has exceeded an acceptable threshold across the last 10 minutes"
+
+This triggers when, within a rolling 10 minute window, over 1% of all HTTP
+responses occur with insufficently quickly, as a proprotion of all responsed in
+that period.
```
jonathanglassman, line 34: See change in previous section.

I think there's a word missing or an extra word in line 33: "...responses occur with insufficiently quickly..."

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,45 @@
...
+
+This is intended as an early warning that, if these errors continued at a
+consistent rate, the GOV.UK Accounts team is on route to spend it's entire
+error budget of 1% across the full 4 week monitoring period.
```
jonathanglassman, line 38: See change in previous section.

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,45 @@
+---
+owner_slack: "#govuk-account-tech"
+title: GOV.UK Account SLO error rate exceeded 1% in 10 minute window
+section: Icinga alerts
+layout: manual_layout
+parent: "/manual.html"
+---
+
+The GOV.UK Accounts team has defined Service Level Objectives (SLOs) for
```
jonathanglassman, line 9: I think it's GOV.UK Account team rather than GOV.UK Account**s** team.

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,45 @@
...
+reliability of the `account-api`.
+
+You can read more about these in [GOV.UK Accounts Service Level
```
jonathanglassman, line 12: I think it's GOV.UK Account team rather than GOV.UK Account**s** team.

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,45 @@
...
+Objectives](https://docs.google.com/document/d/1HVXlyy2vdy9U0O-_7Nw8_Oso7VXQeKgJ2_-bi4pKlz0/edit)
+
+## Warning - "HTTP Error on GOV.UK Accounts has exceeded an acceptable threshold across the last 10 minutes"
```
jonathanglassman, line 15: I think it's GOV.UK Account team rather than GOV.UK Account**s** team.

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,45 @@
...
+This is intended as an early warning that, if these errors continued at a
+consistent rate, the GOV.UK Accounts team is on route to spend it's entire
+error budget of 1% across the full 4 week monitoring period.
+
+This warning automatically triggers a message to the `#govuk-accounts-tech`
+slack channel.
+
+No further action is required from GOV.UK 2ndline or any other team.
+
+## Warning - "Latency on GOV.UK Accounts has exceeded an acceptable threshold across the last 10 minutes"
```
jonathanglassman, line 30: I think it's GOV.UK Account team rather than GOV.UK Account**s** team.

ChrisBAshton (commented):
Thanks for raising this, @huwd.

Tangential perhaps, but I was actually hoping to look at docs around Accounts alerts yesterday, as I found it difficult to find information when we saw the spike in 5xx's last week! For example, the PagerDuty Slack alert looked a bit like this:

> GovukAccounts_AppRequestsExcess5xx # [5628](https://governmentdigitalservice.pagerduty.com/incidents/PAV29JD)
> GovukAccounts_AppRequestsExcess5xx govuk-accounts (govuk-account-manager b25f6491-57fa-4857-9ebb-0f142a76e...

The [link to the associated dashboard](https://github.com/alphagov/prometheus-aws-configuration-beta/blob/d64898eb3b80eddaa171a68672677bf19410513d/terraform/modules/prom-ec2/alerts-config/alerts/govuk-accounts-alerts.yml#L12) didn't appear in the alert, nor did the contextual information of "[App govuk-accounts has 5xx errors in excess of 25% of total requests](https://github.com/alphagov/prometheus-aws-configuration-beta/blob/d64898eb3b80eddaa171a68672677bf19410513d/terraform/modules/prom-ec2/alerts-config/alerts/govuk-accounts-alerts.yml#L11)". Is there some work we can do to ensure all of that useful information is sent to and displayed by PagerDuty?

I found out after the fact that clicking `#5628` above takes you to the PagerDuty incident page which has more information - possibly my lack of familiarity with PagerDuty. Useful to know for the future, but not obvious at the time - perhaps we can document somewhere where to click when the alert fires.

Finally, it's probably worth having a canonical 'accounts' page in the Dev Docs - probably [this one](https://docs.publishing.service.gov.uk/apps/govuk-account-manager-prototype.html), which pulls in the README etc - and should link to https://docs.account.publishing.service.gov.uk/ and https://team-manual.account.publishing.service.gov.uk/. I see now that these links are tucked at the bottom of the page - perhaps they should be brought to the top, and/or much of the README content moved out into a `docs` directory?

Just some thoughts/feedback, hope it's useful 👍

jonathanglassman (changes requested):
Looks good, just a few suggested changes and 1 question

source/manual/alerts/slo-rate-checks-account.html.md
huwd: I think Jon helped me with some better wordingfor this bit, thanks!

source/manual/alerts/slo-rate-checks-account.html.md
huwd: So... here's the config:
https://github.com/alphagov/govuk-puppet/pull/11171

This is my first venture into this stuff, but I'm pretty sure we're not touching prometheus.
My understanding is that it'd got:
Graphite  query -> Icinga warning?

huwd:
@ChrisBAshton Interesting - yeah I think we should think this through a bit more, I know @chao-xian has also mentioned out docs have left him high and dry once or twice too. Which isn't ideal.

I think here's a tiny bit of context here which might help - our earlier stuff (the [govuk-account-manager-prototype ](https://github.com/alphagov/govuk-account-manager-prototype/) and [govuk-attribute-service-prototype](https://github.com/alphagov/govuk-attribute-service-prototype/)) are prototypes we put on PaaS. Partly this was for speed at a time when we were trying to move quickly. Partly also because these were early placeholders that allowed us to progress our experiement whilst the DI folks build the "real thing".

So those two repos are a bit unusual for GOV.UK I think, which is why PaaS prometheus is involved and we have a different way of alerting to pager duty.

These repos are still doing important things however, so we'll need to reconsider how to make docs and alerts transparent to 2ndline and on-call.... however in the grand scheme of things they're on the way out.

We're gradually moving features over to repos like [account-api](https://github.com/alphagov/account-api/) which sits on GOV.UK proper, and uses the same infra any other GOV.UK API would. The alerts in this doc apply to that repo.

In the end the prototypes should become fairly thin placeholders for a third party authentication service, that is fairly generic so we can easily unplug it and replace it with the the DI team's work.

Hopefully that explains why there's a bit of a difference between our repos around things like icinga vs prometheus alerts?

Generally though i think we should reassess our docs situation, @jonathanglassman / @barrucadu perhaps we could get a ticket together for this. Initially we'd intended [these docs](https://docs.account.publishing.service.gov.uk/#gov-uk-accounts-technical-documentation) to be like API documentation for third party departments trying to integrate with us. I'm thinking that's a scenario which has now moved outside our team?

Perhaps we should also consider what should live in the team manual, and what should live in https://docs.publishing.service.gov.uk/ (or a /docs folder that gets pulled in by the GOV.UK Docs repo?).

Perhaps we should also do a bit of dev user research? Take a few of our alerts and see if we can pair with another GOV.UK dev, watch how they go about looking for docs in solution and where any sticking points might be?

Would that get some way to solving the issue you encountered @ChrisBAshton?

source/manual/alerts/slo-rate-checks-account.html.md
barrucadu: Yeah, these are Icinga  alerts

ChrisBAshton:
> our earlier stuff (the govuk-account-manager-prototype and govuk-attribute-service-prototype) are prototypes we put on PaaS... which is why PaaS prometheus is involved and we have a different way of alerting to pager duty. We're gradually moving features over to repos like account-api which sits on GOV.UK proper, and uses the same infra any other GOV.UK API would. The alerts in this doc apply to that repo.

Thanks Huw, that helps 👍 probably not worth investing too much time investigating how to make PagerDuty alerts / PaaS Prometheus linkage better. As things move across to the usual GOV.UK infrastructure, it will become easier to deal with incidents etc.

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,44 @@
+---
+owner_slack: "#govuk-account-tech"
```
barrucadu, line 2: `#govuk-accounts-tech` - not `#govuk-account-tech`

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,44 @@
...
+title: GOV.UK Account SLO error rate exceeded 1% in 10 minute window
```
barrucadu, line 3: 1% is part of the SLO, if we change the SLO then we don't really want to have to update the title of this page too. 
 Also, it's good if this title is close to the Icinga alert titles, so it's easier to find in search.

How about:

```
SLO Error Rate on GOV.UK Accounts has exceeded an acceptable threshold across the last 10 minutes
```

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,44 @@
...
+parent: "/manual.html"
+---
+
+The GOV.UK Account team has defined Service Level Objectives (SLOs) for
+reliability of the `account-api`.
+
+You can read more about these in [GOV.UK Accounts Service Level
+Objectives](https://docs.google.com/document/d/1HVXlyy2vdy9U0O-_7Nw8_Oso7VXQeKgJ2_-bi4pKlz0/edit)
+
+## Warning - "HTTP Error on GOV.UK Account has exceeded an acceptable threshold across the last 10 minutes"
```
barrucadu, line 15: I don't think the `Warning - "..."` is adding anything here, just use `...` directly

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,44 @@
...
+
+This warning triggers when over 1% of all HTTP status codes are deemed "bad
+responses" as a proportion of all responses within a rolling 10 minute time
+period.
+
+This is an early warning that, if these errors continue at a consistent rate,
+the GOV.UK Account team would spend its entire error budget of 1% across the
```
barrucadu, line 22: Here (and below) there's no need to mention the 1% again, just say

```
the GOV.UK Account team would spend its entire error budget across the
```

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,44 @@
...
+full 4 week monitoring period.
```
barrucadu, line 23: Can probably drop the length of the monitoring period too.

source/manual/alerts/slo-rate-checks-account.html.md
```diff
@@ -0,0 +1,44 @@
...
+
+This warning automatically triggers a message to the `#govuk-accounts-tech`
+slack channel.
+
+No further action is required from GOV.UK 2ndline or any other team.
+
+## Warning - "Latency on GOV.UK Account has exceeded an acceptable threshold across the last 10 minutes"
+
+This warning triggers when over 1% of all HTTP responses are deemed
+"insufficiently quick" as a proportion of all responses within a rolling 10
```
barrucadu, line 33: "too slow" is better than "insufficiently quick"

jonathanglassman (approved):

source/manual/alerts/slo-rate-checks-account.html.md
jonathanglassman: good point @barrucadu, i thought "insufficiently quick" was the term used in Icinga. Whatever term we use in the docs should reflect what the user sees.

barrucadu (approved):

PR #3150, opened 2021-06-24
Document fall back to CloudFront as a manual process

richardTowers:
Deploying these changes using Jenkins and Terraform is too slow and
unreliable to do during an incident.

To be honest, I think this would be better as a shell script. I'll raise
this as a PR to get people's opinions.

https://trello.com/c/LasnNuz4/561-rationalise-use-of-gcp-credentials-eg-for-dns-deployments

### DIFF
--- a/source/manual/fall-back-to-aws-cloudfront.html.md
+++ b/source/manual/fall-back-to-aws-cloudfront.html.md
@@ -9,28 +9,47 @@ parent: "/manual.html"
 There is a backup Content Delivery Network (CDN) that can be used if Fastly is down.
 This backup CDN is currently provided by AWS CloudFront.
 
-**Important** At the time of writing, GOV.UK will be served
-by static mirrors and not the origin of GOV.UK if you fallback to the backup CDN.
+> **Important** GOV.UK will be served by static mirrors and not the origin of GOV.UK if you fallback to the backup CDN.
+> This is significantly degraded service compared to the primary CDN. Major features such as search, smart answers, and
+> postcode lookups will not work when using the secondary CDN.
 
-## Procedure
+## Fail over checklist
 
-You will have to make 2 DNS changes to GOV.UK:
+- Confirm that Fastly is the cause of the incident (check [https://status.fastly.com/](https://status.fastly.com/)
+  and keep an eye on twitter - if there's a major Fastly outage there will be a lot of noise)
+- Sign in to the AWS console as an admin (`gds aws govuk-production-admin -l`, or however you prefer to sign in to AWS)
+- Sign in to [the GCP console](https://console.cloud.google.com/home/dashboard?project=govuk-production)
+- Open the following four pages as separate tabs:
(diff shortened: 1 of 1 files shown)


sengi:
It's an improvement. I agree a script would be better. On the other hand, having working instructions in the meantime is no bad thing.

source/manual/fall-back-to-aws-cloudfront.html.md
```diff
@@ -9,29 +9,45 @@ parent: "/manual.html"
...
+- Open the following four pages as separate tabs:
+  - [GCP Cloud DNS www-cdn.production.govuk.service.gov.uk](https://console.cloud.google.com/net-services/dns/zones/govuk-service-gov-uk/rrsets/www-cdn.production.govuk.service.gov.uk./CNAME/edit?project=govuk-production)
+  - [AWS Route 53 govuk.service.gov.uk](https://console.aws.amazon.com/route53/v2/hostedzones#ListRecordSets/Z22RPYZA77J620)
+  - [GCP Cloud DNS assets.publishing.service.gov.uk](https://console.cloud.google.com/net-services/dns/zones/publishing-service-gov-uk/rrsets/assets.publishing.service.gov.uk./CNAME/edit?project=govuk-production)
+  - [AWS Route 53 publishing.service.gov.uk](https://console.aws.amazon.com/route53/v2/hostedzones#ListRecordSets/Z3SBFBO09PD5HF)
+- Because the secondary CDN provides degraded service we have a grace period
+  before failing over
+  - For a major Fastly outage (more than 30% of requests failing) wait 15
+    minutes (from the start of the incident) before failing over
+  - For a minor Fastly outage (less than 30% of requests failing) wait at least
```
fredericfran-gds, line 31: This point is confusing to me, is it:
1. minor outage, wait 15 minutes, failover and contact SMT or
2. minor outage, wait 15 minutes, contact SMT -> they decide whether to failover?

richardTowers: Yeah. I need to think about this more, it's too vague right now.

I think I was thinking "start a conversation with SMT immediately about how long to wait. Wait at least 15 minutes".

richardTowers:
I had a stab at writing a script for this, but it turned into a bit of a nightmare:

https://github.com/alphagov/govuk-dns/commit/86b0936f4bb636865e0027ee0f03cdc1d0332a14

I now think a better approach would be to come up with a way to mirror the terraform providers so we can deploy things "the normal way", even when the terraform registry is down. Sadly that's a bit of a yak shave because we're still on terraform 0.11, and all the cool registry mirror stuff is terraform >= 0.13.

In the meantime, documenting a manual process is still probably a good intermediate step.

source/manual/fall-back-to-aws-cloudfront.html.md
richardTowers: I've clarified this point now (several months later 😬 )

sihugh (approved):
Much clearer, thanks 👍

PR #3151, opened 2021-06-25
Document how to deploy Fastly from our individual accounts

ChrisBAshton:
We no longer have a shared 2nd line account for Fastly, so can
no longer rely on a centrally stored API key for deployments. This
PR documents how to create an access token linked to your own
Fastly account, and how to use it to deploy the CDN configuration
for `assets` and `www`. (There is also a third option,
`performanceplatform`, but as this is retired, I assume it should
be removed and we've no need to document it here).

Trello: https://trello.com/c/ShvEW3om/2560-add-documentation-on-how-to-use-individual-fastly-accounts-2

### DIFF
--- a/source/manual/cdn.html.md
+++ b/source/manual/cdn.html.md
@@ -20,12 +20,36 @@ The CDN is responsible for retrying requests against the [static mirror](/manual
 
 Most of the CDN config is versioned and scripted:
 
-- [CDN configuration](https://github.com/alphagov/govuk-cdn-config/)
-- [CDN config secrets](https://github.com/alphagov/govuk-cdn-config-secrets)
-
-These are deployed to [integration][integration_cdn], [staging][staging_cdn] and [production][production_cdn].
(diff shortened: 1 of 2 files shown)


source/manual/cdn.html.md
```diff
@@ -20,13 +20,37 @@ The CDN is responsible for retrying requests against the [static mirror](/manual
...
+1. Click "[Personal API tokens](https://manage.fastly.com/account/personal/tokens)"
+1. Click "Create Token"
+
+Then follow the steps to create a token with the minimum privileges required for the task at hand:
+
+1. The token name is unimportant; you'll be deleting it shortly.
+1. For "Service Access", choose the service you'll be deploying,
+   e.g. "Integration GOV.UK" (for `www`) or "Integration Assets" (for `assets`).
+1. For "Scope", you'll need to pick "Global API access" (i.e. the most permissive)
+1. Under "Expiration", choose tomorrow's date
```
thomasleese, line 48: Nice!

thomasleese (approved):

PR #3154, opened 2021-06-29
Document Elasticsearch JSON example

ChrisBAshton:
This took a lot of trial and error on my part this morning, which would have been helped immeasurably by a solid example in the Dev Docs!


1pretz1 (approved):
Looks good to me, thanks for adding an example! 🍏

PR #3155, opened 2021-06-29
Update TLS certificate docs

richardTowers:


source/manual/renew-a-tls-certificate.html.md
```diff
@@ -6,52 +6,63 @@ layout: manual_layout
...
+interface. It is managed entirely through Fastly support.
+
+Credentials for the Fastly Zendesk support site are in the [2nd line password store](https://github.com/alphagov/govuk-secrets/blob/master/pass/2ndline/fastly).
+
+## Renewing wildcard certificates
+
+Wilcard certificates for `*.publishing.service.gov.uk`, `*.staging.publishing.service.gov.uk`
+and `*.integration.publishing.service.gov.uk` are managed by AWS ACM.
+
+For AWS ACM to issue a certificate you must prove ownership of the domain using DNS.
```
fredericfran-gds, line 26: minor, missing comma: For AWS ACM to issue a certificate**,** you must prove ownership of the domain using DNS.

fredericfran-gds (approved):
looks good, I think there are some leftover certs here: https://github.com/alphagov/govuk-secrets/tree/main/pass/2ndline/certificates that can be deleted?

source/manual/renew-a-tls-certificate.html.md
```diff
@@ -6,52 +6,63 @@ layout: manual_layout
...
+4. Find the certificate you wish to renew and click Renew.
 5. Go through the steps on the renewal form until you reach a page requesting a
    Certificate Signing Request.
 6. Upload the CSR to Gandi by pasting the contents of the .csr file into the
    text box.
 7. Next, choose DNS validation to validate it and follow the instructions to add
    the relevant DNS records.
-   For DNS validation of certificate request for `www.gov.uk`, you need to ask a
-   senior GOV.UK member to request JISC to add the new DNS record.
 8. Pay for it - we don't have a stored payment method, so find the person with
```
sengi, line 62: Something I ran into when renewing domains, which might also be relevant here: there's a step near the beginning of the checkout process where it asks which user you want to invoice. You have to choose the one called "GOV.UK (Public Body)" as opposed to "govuk (Individual)" otherwise VAT gets applied incorrectly. Not 100% sure if the UX is the same when buying certs as opposed to domains, but something to check perhaps?

sengi (approved):
Thanks for updating this.

PR #3159, opened 2021-07-01
Rewrite template for Rails app READMEs

benthorner:
https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

This follows on from the work we've done to revise READMEs for most
GOV.UK repos. Please see the commits for more details.

PR #3172, opened 2021-07-13
Rewrite template for general READMEs

benthorner:
https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

This continues the changes from https://github.com/alphagov/govuk-developer-docs/pull/3159.

### DIFF
--- a/source/manual/readmes.html.md
+++ b/source/manual/readmes.html.md
@@ -7,7 +7,7 @@ type: learn
 parent: "/manual.html"
 ---
 
-This is a guide to writing and maintaining README documents for GOV.UK's public repositories.
+This is a guide to writing and maintaining README documents for GOV.UK's public repositories. It's based on [guidance we have in the GDS Way for READMEs](https://gds-way.cloudapps.digital/manuals/readme-guidance.html#writing-readmes).
 
 READMEs are for a technical audience. This could be a new starter or an existing developer. Using the following template will help to ensure each README is consistent, correct, short and actionable.
 
--- a/source/manual/readmes.html.md
+++ b/source/manual/readmes.html.md
@@ -62,17 +62,11 @@ Examples READMEs that follow the above structure:
 
 - [General app: Email Alert API](https://github.com/alphagov/email-alert-api/blob/451481ce0b6335bb1f640ef52fa0e8305f38d09c/README.md#technical-documentation)
 
-## One-line summaries on GitHub repositories
+## GitHub repo description
 
-Who are these for? These are for people who scan the list in [alphagov](https://github.com/alphagov/) and need a quick overview.
+These appear in [search results](https://github.com/alphagov/). A good description should use simple language and make it clear the repository is part of GOV.UK. Example: "GOV.UK filtered search of public content".
 
-- Include information about whether it's on GOV.UK
-- Try to tone down technical language - 'Filtered search of content' is better than 'Faceted search interface'
-- Use the GitHub link field to link to the developer docs page for the repository, or otherwise the live service (if public facing)
-
-A good example:
-
-> GOV.UK filtered search of public content
+Use the GitHub link field to link to the developer docs page for the repository, or otherwise the live service (if it's public facing).
 
 ## CONTRIBUTING.md
 
--- a/source/manual/readmes.html.md
+++ b/source/manual/readmes.html.md
@@ -11,6 +11,8 @@ This is a guide to writing and maintaining README documents for GOV.UK's public 
 
 READMEs are for a technical audience. This could be a new starter or an existing developer. Using the following template will help to ensure each README is consistent, correct, short and actionable.
 
+Unless it helps someone to get started, any other content should be a separate file in the `docs/` directory for a repository. This also makes the content easier to discover in external search.
+
 ## Template for new READMEs
 
 ```markdown
@@ -74,15 +76,6 @@ A good example:
 
 ## Things we shouldn't put in the README
 
-- Full documentation of the API - if the app is complex or has a complex API, use the `docs` directory liberally, and link out to it from the `README`
-  - If the API is small try to keep it in the main README and don't link to docs
-  - Some bad examples - too much documentation in the README:
-    - [smart-answers (older version)](https://github.com/alphagov/smart-answers/blob/06dfeb854ec9728374186a2e37c1eb62a5aaa49a/README.md)
-    - [govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit/blob/edd834de8d2f4e854475f5d226aa0cd260795cef/README.md)
-  - Good examples:
-    - [content-store](https://github.com/alphagov/content-store/blob/b244620f505f248fc93d8556eedad14b5cba1187/README.md) - links out to more docs as required
-    - [smart-answers (newer version)](https://github.com/alphagov/smart-answers/blob/c0b4580d18ccc5004abfa7015017d26e1a73f2aa/README.md) - links out to more docs as required
-  - For Ruby libraries always use [yardoc](http://yardoc.org/) - then Gems will automatically have docs built
 - Contributors - we can use GitHub's native graphing tools for this
 - Contribution guidelines - use GitHub's [CONTRIBUTING.md](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) guidelines for this
   - When writing this document, always link to the [GOV.UK pull request style guide](https://github.com/alphagov/styleguides/blob/master/pull-requests.md)
--- a/source/manual/readmes.html.md
+++ b/source/manual/readmes.html.md
@@ -19,58 +19,49 @@ This is a guide to writing and maintaining README documents for GOV.UK's public 
 ```markdown
 # App name
 
-One paragraph description and purpose.
+One or more paragraphs describing the app:
 
-## Screenshots (if there's a client-facing aspect of it)
+- What it's used for.
+- Any unusual dependencies it has.
 
-## Live examples (if available)
+## Live examples
 
-- [gov.uk/thing](https://www.gov.uk/thing)
+(Only applies to frontend apps.)
+
+A list of links to example pages rendered by the app.
 
 ## Nomenclature
 
-- **Word**: definition of word, and how it's used in the code
+A list of definitions for unusual terms in the code.
 
 ## Technical documentation
 
-Write a single paragraph including a general technical overview of the app.
-Example:
+What goes here depends on the type of app.
 
-This is a Ruby on Rails application that maps RESTful URLs onto a persistence
-layer. It's only presented as an internal API and doesn't face public users.
-
-### Dependencies
-
-- [alphagov/other-repo]() - provides some downstream service
-- [redis]() - provides a backing service for work queues
-
-### Running the application
-
-`./one-line-command.sh`
-
-Documentation for where the app will appear (default port, vhost, URL etc).
+See the links below for example content to put here.
 
 ### Running the test suite
 
-`./one-line-command.sh`
+Give one command to run all the tests, linting, etc.
 
-Include any other edge cases, e.g parallel test runner in Whitehall
+You can also add other commands e.g. to run JS tests.
 
-### Any deviations from idiomatic Rails/Go etc. (optional)
+### Further documentation
 
-### Example API output (optional)
+A list of links to key files in docs/.
 
-`one-line-curl-command with JSON response after`
-
-Keep this section limited to core endpoints - if the app is complex link out to `/docs`.
+You can also just link to the docs/ directory itself.
 
 ## Licence
 
-[MIT License](LICENCE)
-
-## Versioning policy (for Gems only)
-
+Link to your LICENCE file.
 ```
+
+Examples READMEs that follow the above structure:
+
+- [Frontend app: Collections](https://github.com/alphagov/collections/blob/13e53b7b63b2a9c1e618ba309756523341befc5b/README.md#technical-documentation)
+
+- [General app: Email Alert API](https://github.com/alphagov/email-alert-api/blob/451481ce0b6335bb1f640ef52fa0e8305f38d09c/README.md#technical-documentation)
 
 ## Benefits of using this template
 
--- a/source/manual/readmes.html.md
+++ b/source/manual/readmes.html.md
@@ -9,10 +9,7 @@ parent: "/manual.html"
 
 This is a guide to writing and maintaining README documents for GOV.UK's public repositories.
 
-## Who are READMEs for (aka what's the user need)?
-
-- New starters to GDS
-- Developers on a team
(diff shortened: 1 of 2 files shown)


source/manual/readmes.html.md
```diff
@@ -74,10 +74,12 @@ A good example:
...
+## CONTRIBUTING.md
 
-- Contributors - we can use GitHub's native graphing tools for this
-- Contribution guidelines - use GitHub's [CONTRIBUTING.md](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) guidelines for this
-  - When writing this document, always link to the [GOV.UK pull request style guide](https://github.com/alphagov/styleguides/blob/master/pull-requests.md)
-  - Document any weirdness here (eg when to use Cucumber over something else)
-- Full licence - again, follow [GitHub convention](https://help.github.com/articles/open-source-licensing/#where-does-the-license-live-on-my-repository)
+**Don't add this unless the repository gets a lot of external contributions.**
+
+Use GitHub's [CONTRIBUTING.md](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) guidelines for this. Always link to the [GOV.UK pull request style guide](https://github.com/alphagov/styleguides/blob/master/pull-requests.md).
```
benthorner, line 75: It looks like the `styleguides` repo is archived, but I'm not sure what a like-for-like replacement should be.

source/manual/readmes.html.md
```diff
@@ -7,105 +7,73 @@ type: learn
...
 
 ```markdown
 # App name
 
-One paragraph description and purpose.
+One or more paragraphs describing the app:
 
-## Screenshots (if there's a client-facing aspect of it)
+- What it's used for.
+- Any unusual dependencies it has.
```
jonathanglassman, line 25: Suggest augmenting slightly with information from https://github.com/alphagov/readme-template/blob/main/README.md:

"# App name

- What the app does 
- Who the app is for
- Any unusual dependencies the app has
- Limitations of the app
- Links to other things as necessary

source/manual/readmes.html.md
```diff
@@ -7,105 +7,73 @@ type: learn
...
 
-## Live examples (if available)
+## Live examples
 
-- [gov.uk/thing](https://www.gov.uk/thing)
+(Only applies to frontend apps.)
+
+A list of links to example pages rendered by the app.
 
 ## Nomenclature
```
jonathanglassman, line 32: Can we say "Defined terms" or something more plain English than "Nomenclature"?

source/manual/readmes.html.md
```diff
@@ -7,105 +7,73 @@ type: learn
...
 
 ### Running the test suite
 
-`./one-line-command.sh`
-
-Include any other edge cases, e.g parallel test runner in Whitehall
+Give one command to run all the tests, linting, etc.
 
-### Any deviations from idiomatic Rails/Go etc. (optional)
+You can also add other commands e.g. to run JS tests.
```
jonathanglassman, line 46: Not sure I understand this? Is the point to be specific on the type of testing? E.g. "To run javascript tests, run the following... To run linting, run the following..."

source/manual/readmes.html.md
```diff
@@ -7,105 +7,73 @@ type: learn
...
-This is a Ruby on Rails application that maps RESTful URLs onto a persistence
-layer. It's only presented as an internal API and doesn't face public users.
-
-### Dependencies
-
-- [alphagov/other-repo]() - provides some downstream service
-- [redis]() - provides a backing service for work queues
-
-### Running the application
+What goes here depends on the type of app.
```
jonathanglassman, line 38: Would distinguish between running for the first time ("hello world") and running in general

source/manual/readmes.html.md
```diff
@@ -7,105 +7,73 @@ type: learn
...
-
-Documentation for where the app will appear (default port, vhost, URL etc).
+See the links below for example content to put here.
 
 ### Running the test suite
 
-`./one-line-command.sh`
-
-Include any other edge cases, e.g parallel test runner in Whitehall
+Give one command to run all the tests, linting, etc.
```
jonathanglassman, line 44: Would also include information about error messages and suggested fixes

source/manual/readmes.html.md
jonathanglassman: If applicable, include a link to any editable project configuration files and environment variables

source/manual/readmes.html.md
```diff
@@ -7,105 +7,73 @@ type: learn
...
+You can also add other commands e.g. to run JS tests.
 
-### Example API output (optional)
+### Further documentation
 
-`one-line-curl-command with JSON response after`
+A list of links to key files in docs/.
 
-Keep this section limited to core endpoints - if the app is complex link out to `/docs`.
+You can also just link to the docs/ directory itself.
```
jonathanglassman, line 53: Would include a section of where users can get support on this app

jonathanglassman (changes requested):
Would recommend expanding the content as per comments

source/manual/readmes.html.md
benthorner: Sure, we can add some of those in. I think it's worth agreeing what a valid example for each of them would be first, so we have a common understanding of what we mean by them.

> Links to other things as necessary

We generally put links in naturally where it's appropriate, throughout the document, so I'm not sure we need a specific prompt here. If we just mean bullet point links, we've got a "Further docs" section for that.

Do you think this would prompt someone to add something they otherwise wouldn't?

> Who the app is for

I think this is usually covered by "what the app does", which usually includes the phrase "frontend app" or "publishing app" or "API" (not really _for_ anyone). Frontend apps are for the general public, and publishing apps are for government publishers. This is common knowledge on GOV.UK, so I don't think we need to repeat it in each README.

Can you think of any exceptions?

> Any unusual prerequisites for the app to work

This sounds the same as the one above it - dependencies and prerequisites are synonyms to me. Most of the dependencies are automated away by our development environment, but maybe we could be clearer about what we mean by "unusual" what do you think?

> Limitations of the app

All apps are fundamentally limited 😉. What would be an example of this for you?

source/manual/readmes.html.md
benthorner: How about just "Definitions"?

("Defined terms" leads me to question where the "Undefined terms" are listed...)

source/manual/readmes.html.md
benthorner: Yes, that's right. We could tweak it to be clearer perhaps. How about:

```
You can also suggest other commands e.g. to run JS tests, run ....
```

source/manual/readmes.html.md
benthorner: We've actually removed this section in almost all READMEs because it's the same for every app: [first time and ongoing running is standardised by our development environment](https://github.com/alphagov/govuk-docker#usage), which also handles config files / env vars. While a general app definitely needs this sort of documentation, on GOV.UK we'd just be repeating the same long spiel in every README, so we've shortened it to the blurb about "go and read the central GOV.UK Docker guidance".

Do you think we need to say more in each README?

source/manual/readmes.html.md
benthorner: I agree that can be helpful, but only if the errors are frequent. We don't really have an example of that though, because if an error is really that bad then we would just go and fix it. For less frequent errors, we could suggest a separate "troubleshooting" guide in the "further docs" section. I'm keen not to make people read a bunch of stuff that's only relevant once in a blue moon. A troubleshooting guide has the advantage of single, clear purpose.

How does that sound?

source/manual/readmes.html.md
benthorner: Could you help me understand what you mean here? 

(I'm not sure who the users here would be.)

source/manual/readmes.html.md
jonathanglassman: "Definitions" works for me

source/manual/readmes.html.md
jonathanglassman: works for me

source/manual/readmes.html.md
jonathanglassman: If it is a standardised GOV.UK Docker process, then agree that the shortened content on reading the docker guidance is fine

source/manual/readmes.html.md
jonathanglassman: I would recommend against a troubleshooting section. It's content best practice to have issues in the section of docs that those issues relate to. For example, https://docs.notifications.service.gov.uk/java.html#error-codes.

What I would recommend is to have a clear route for users to support on the less frequent errors, as per my next comment. That way, when a user does have an error, they can go down the support route and get it fixed.

source/manual/readmes.html.md
jonathanglassman: See above

source/manual/readmes.html.md
jonathanglassman: - Re link to other things, I thought maybe links to other related apps or programs would be useful
- Re who the app is for, I cannot think of any exceptions right now, but it is best practice to specify who should use the thing you are documenting
- Re dependencies or prerequisites, i would say include anything that is not automated away by the development environment
- Re limitations, I don't know about an example. I guess it would something like "You can use this data analysis app with this type of data in this location, however it is not suitable for this other type of data"

source/manual/readmes.html.md
benthorner: I've added a commit to change it to that 👍.

source/manual/readmes.html.md
jonathanglassman: Sounds good to me.

source/manual/readmes.html.md
benthorner: Cool. Are you comfortable if we defer changing this to a follow-up PR? At the moment all our READMEs use the current wording, so ideally we should change them so the template is consistent with reality.

There's already a task on [the follow-up card](https://trello.com/c/VU3R1B8X/240-make-govuk-readmes-consistent-with-the-gds-way) to rename "Technical documentation" to "Getting started", which will involve touching every app, so we could do it as part of that.

jonathanglassman: sure

source/manual/readmes.html.md
benthorner: Ah, I think I see what you mean. GOV.UK apps don't have direct users - everything is provided as a service to publishers and to the public. So the only people we need to "support" are the same devs who wrote the code.

Content about error messages or general usage instructions would definitely be appropriate for a library, but this template is specific to apps, so I don't think we need to worry about it in here.

This is perhaps one example of where having a GOV.UK app-specific template can be helpful to avoid ambiguity.

Does clarifying the users help at all?

jonathanglassman: ok, that does help. no need to have information about errors and fixes then

source/manual/readmes.html.md
benthorner: Great, I've added another commit with those changes.

source/manual/readmes.html.md
benthorner: OK. I've made a note on the follow-up card, and will leave this comment unresolved.

source/manual/readmes.html.md
benthorner: Is this the same as https://github.com/alphagov/govuk-developer-docs/pull/3172#discussion_r681909953?

jonathanglassman: yes, so can ignore my comment

benthorner:
@jonathanglassman I think I've addressed all your comments (with a follow-up PR planned for one of them).

Do you think this is ready to go?

I might ask a GOV.UK dev to review as well, so we've got a couple of perspectives.

jonathanglassman:
> @jonathanglassman I think I've addressed all your comments (with a follow-up PR planned for one of them).
> 
> Do you think this is ready to go?
> 
> I might ask a GOV.UK dev to review as well, so we've got a couple of perspectives.

@benthorner LGTM, thanks for your time and expertise

jonathanglassman (approved):

source/manual/readmes.html.md
```diff
@@ -74,10 +74,12 @@ A good example:
...
-- Contributors - we can use GitHub's native graphing tools for this
-- Contribution guidelines - use GitHub's [CONTRIBUTING.md](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) guidelines for this
-  - When writing this document, always link to the [GOV.UK pull request style guide](https://github.com/alphagov/styleguides/blob/master/pull-requests.md)
-  - Document any weirdness here (eg when to use Cucumber over something else)
-- Full licence - again, follow [GitHub convention](https://help.github.com/articles/open-source-licensing/#where-does-the-license-live-on-my-repository)
+**Don't add this unless the repository gets a lot of external contributions.**
+
+Use GitHub's [CONTRIBUTING.md](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) guidelines for this. Always link to the [GOV.UK pull request style guide](https://github.com/alphagov/styleguides/blob/master/pull-requests.md).
+
+## LICENCE.md
```
ChrisBAshton, line 81: Is it `LICENCE.md` or simply `LICENCE`? E.g. Content Publisher has no file extension: https://github.com/alphagov/content-publisher/blob/main/LICENCE

ChrisBAshton (approved):
LGTM 👍 only a couple of points below.

source/manual/readmes.html.md
```diff
@@ -19,58 +19,49 @@ This is a guide to writing and maintaining README documents for GOV.UK's public
...
+A list of links to key files in docs/.
 
-Keep this section limited to core endpoints - if the app is complex link out to `/docs`.
+You can also just link to the docs/ directory itself.
 
 ## Licence
 
-[MIT License](LICENCE)
+Link to your LICENCE file.
+```
```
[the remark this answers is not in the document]
benthorner: Good spot, I've amended in the second commit.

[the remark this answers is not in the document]
benthorner: Ah, scratch that: the "\`\`\`" here was the end of the "\`\`\`markdown", so I've put it back.

source/manual/readmes.html.md
benthorner: Oh dear, another can of worms. Analysis of the projects I have locally:

- 5 have `LICENSE.md`
- 4 have `LICENCE.md`
- 42 have `LICENCE`
- 24 have `LICENSE`

🙄. I've amended to `LICENCE` because it reflects the majority and the plaintext content of the file. I've added a task on [the follow-up card](https://trello.com/c/VU3R1B8X/240-make-govuk-readmes-consistent-with-the-gds-way) to decide if it's worth renaming the others.

As a side-note, quite a few of our repos also have a `CONTRIBUTING.md` (happily there are no other variants!). I've added another follow-up task to decide if we want to delete them (all).

kevindew:
This looks great, nice work 👍

PR #3173, opened 2021-07-13
Add `app-console` to Getting Started docs

brucebolt:
This is an important thing for developers to be aware of as it saves a few extra steps, so we should include this in the Getting Started docs.


beccapearce (approved):
Nice!

PR #3179, opened 2021-07-19
Update links to rake task

leenagupte:
The original links were to a Jenkins job that can no longer be accessed.

Also added a note about using single quotes. If double quotes are used the task fails with:

Don't know how to build task 'reslug:organisation[national-probation-service,' (See the list of available tasks with `rake --tasks`)


karlbaker02 (approved):
Looks good 🥇

PR #3180, opened 2021-07-20
Add govuk_schemas to applications list

ChrisBAshton:
Following the convention established in 64127d3417788aba3c2c39a7e7c9ce1dbd458356,
we want to make sure that all active GOV.UK utilities are included
in the Developer Docs, to increase visibility and discoverability
of their documentation.

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -441,6 +441,12 @@
   sentry_url: false
   dashboard_url: false
   deploy_url: https://github.com/alphagov/govuk-secrets/blob/main/puppet_aws/deploy.sh
+- github_repo_name: govuk_schemas
+  team: "#govuk-developers"
+  production_hosted_on: none
+  type: Utilities
+  sentry_url: false
+  dashboard_url: false
 - github_repo_name: govuk-zendesk-display-screen
   management_url: https://dashboard.heroku.com/apps/govuk-zendesk-display-screen
   production_hosted_on: heroku


brucebolt (approved):

kevindew:
Should we create a section called utilities for things that aren't applications? It seems strange to have gems in the list of applications given they are flagged as "hosted on" places.

It doesn't feel like the data we record for these none-apps is appropriate for their needs (e.g. dashboard_url, sentry_url) and I know the none-apps caused a bunch of confusion recently when the list of applications was exported to a spreadsheet for product people (since there's a handful of none-apps amongst there).

ChrisBAshton:
@kevindew that sounds like a good idea 👍 would tidy up the configuration somewhat. It needs a bit of thought, as we have other things hooked into this `applications.yml` config, such as the Dependapanda, which checks whether or not apps are continuously deployed and adjusts its message accordingly. Obviously doesn't apply to gems, but could have some quirky side-effects we're not expecting.

I'll write up a Trello card on the DevTools backlog, and continue with this for now.

kevindew:
Thanks @ChrisBAshton - yes, it's caused problems in the past that applications.yml powers dependapanda and that only lists (or listed) running apps. 

Do be cautious populating this too much as I do think it creates confusion adding non-apps to this list. I imagine it has knock-on's of duplication with the dashboard.yml file.

PR #3183, opened 2021-07-28
Add Redis memory debugging docs

bilbof:
We recently went through an incident involving Redis running out of memory. The two immediate solutions were to:
1. Increase the size of the Redis nodes, to buy time/memory
   while we investigated the root cause
2. Find the key using large amounts of memory (maps to a
   an app's queue), and fix the underlying app bug.

I have update the doc to reflect the various steps we took
for future incidents.


ChrisBAshton (approved):

smford (approved):
lgtm

PR #3189, opened 2021-08-03
Add more troubleshooting steps for GPG key updating

ChrisBAshton:
The CLI instructions did not work for me for updating two distinct keys.
I'm not sure why using the Mac GPG Keychain application is any different, as if I go into its "Preferences", it appears to use `hkps://keys.openpgp.org`.
If I run `gpg --keyserver keys.openpgp.org --recv-keys THE_KEY_ID` (with or without the protocol prefix), I get `gpg: keyserver receive failed: General error`.

### DIFF
--- a/source/manual/encrypted-hiera-data.html.md
+++ b/source/manual/encrypted-hiera-data.html.md
@@ -513,7 +513,11 @@ cat ~/govuk/govuk-secrets/puppet_aws/gpg_recipients/production_hiera_gpg.rcp | c
 
 You can check the output for any keys that may have expired. (Or pipe the output into grep ` | grep expired`)
 
-If a key has expired you need to ask the owner to renew that key and publish to a keyserver. Once the key has been renewed you can get the renewed key by:
+If a key has expired, it's possible that your local copy of the key is out of date. Follow the steps below, and if that doesn't resolve the issue, contact the owner and ask them to extend the expiry date or renew their key, then repeat the steps below.
+
+#### Get someone's latest GPG key
+
+If you're on a Mac, open the "GPG Keychain" app, right click on the name of the person in the list, then choose "Update from key server". Alternatively, on the command line you can run:
 
 ```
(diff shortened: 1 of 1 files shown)


source/manual/encrypted-hiera-data.html.md
```diff
@@ -513,7 +513,11 @@ cat ~/govuk/govuk-secrets/puppet_aws/gpg_recipients/production_hiera_gpg.rcp | c
 
 You can check the output for any keys that may have expired. (Or pipe the output into grep ` | grep expired`)
 
-If a key has expired you need to ask the owner to renew that key and publish to a keyserver. Once the key has been renewed you can get the renewed key by:
+If a key has expired, it's possible that your local copy of the key is out of date. Follow the steps below, and if that doesn't resolve the issue, contact the owner and ask them to renew their key and publish to a keyserver, then repeat the steps below.
```
leenagupte, line 516: Are there docs on how to renew a key? I was advised to extend the expiry date of mine which I think is a bit simpler: https://superuser.com/questions/813421/can-you-extend-the-expiration-date-of-an-already-expired-gpg-key. Also I forgot about the sub-keys and did that through the app. Maybe we just need a GPG troubleshooting section?

ChrisBAshton: I can't see any docs on how to renew a key. A doc for that would be good, next time someone does it - so far, I only know of you and @MuriloDalRi who have done it, perhaps one of you could draft a PR?

In the meantime, it sounds like a tweak to this troubleshooting section might be sufficient:

```suggestion
If a key has expired, it's possible that your local copy of the key is out of date. Follow the steps below, and if that doesn't resolve the issue, contact the owner and ask them to extend the expiry date or renew their key, then repeat the steps below.
```

leenagupte (approved):
👍 It's good to have this documented.

PR #3217, opened 2021-09-02
Set up user monitor before GitHub access

kevindew:
This is changed around so that a user is set-up in the user monitor before GitHub, this is to prevent people (like myself) who add someone to GitHub and not realise 2nd line are about to be nagged.


sihugh:
Thanks @kevindew ⭐

sihugh (approved):

PR #3219, opened 2021-09-03
Provide links to Jenkins for Elasticsearch Rake tasks

brucebolt:
There are a number of Rake tasks listed in the Elasticsearch reindexing documentation that suggest running the task directly on a machine.

This is generally discouraged, with Rake tasks preferred to be run as a Jenkins job.

Therefore updating the documentation to provide links to Jenkins for each environment, which also reduces the amount of copy and pasting needed.


sihugh (approved):
You are the hero we need ⚡

PR #3221, opened 2021-09-03
Rename Elasticsearch reindex docs to `.erb`

brucebolt:
Missing the `.erb` suffix means that the `RunRakeTask.links` helper method failed to work and the code was rendered within the page.


sihugh (approved):
Mea culpa

PR #3222, opened 2021-09-03
Add details of MongoDB for Licensify developers

benjamineskola:


source/manual/access-to-licensing-for-third-parties.html.md
```diff
@@ -26,6 +26,10 @@ This is a prerequisite for accessing Jenkins (which is used for deployments), an
 Follow [the VPN guidance for non-GDS devices ("BYOD")](https://docs.google.com/document/d/150JX1xiWdXY29ahcYUMb05Si-hEAZvtkGAKojT9Rjis/edit)
 to set up the VPN. You will need to sign into your `@digital.cabinet-office.gov.uk` Google account to access this document.
 
+## Accessing MongoDB
+
+Licensify uses a MongoDB cluster hosted by AWS (DocumentDB). The database hosts in use by a particular Licensify instance can be found in `/etc/licensing/gds-licensing-config.properties` on the `licensing_backend` machines, in the `mongo.database.*` keys.
```
jonathanglassman, line 31: So, what does a user to actually access MongoDB? Use the database hosts in some way? It would be good to clarify the action here if possible, even if we assume that the user will know what to do with this information.

PR #3226, opened 2021-09-07
Update A/B test with 0% advice

injms:
## What

Adds extra advice about whether turning on an A/B test at 0% should be considered.

Card: https://trello.com/c/iubAAacJ

## Why

The Explore team recently ran an A/B test, and in the retro it was dicussed that it could've been deployed but not switched on to allow other teams to be aware of the impact that the test has on their work. Documenting it here so that it can be considered for future A/B tests.

### DIFF
--- a/source/manual/run-ab-test.html.md
+++ b/source/manual/run-ab-test.html.md
@@ -34,14 +34,13 @@ everything is working OK.
 
 ### Decide how the variants should be split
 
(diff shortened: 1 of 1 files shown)


owenatgov (approved):
🅰️ 🅱️

chao-xian:
💤

PR #3242, opened 2021-09-20
Add jQuery removal doc

andysellick:
- provides guidance for removing jQuery from our JavaScript, which is an ongoing task on GOV.UK

Trello card: https://trello.com/c/2T8YA7dv/529-write-down-some-javascript-stuff-make-it-generic

### DIFF
--- a/source/manual/how-to-remove-jquery-from-javascript.html.md
+++ b/source/manual/how-to-remove-jquery-from-javascript.html.md
@@ -1,0 +1,228 @@
+---
+owner_slack: "#govuk-frontenders"
+title: How to remove jQuery from JavaScript
+section: Frontend
+layout: manual_layout
+parent: "/manual.html"
+related_applications:
+ - govuk_publishing_components
+---
+
+GOV.UK has a dependency on jQuery 1.12.4. This is an old and unsupported version and we are trying to remove it.
+
+This is a guide to removing jQuery from JavaScript (JS). First read the documentation for our [JS modules](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/javascript-modules.md).
+
+## General advice
+
+Modules are passed an element to act on. When that changes from a jQuery object to a JS node, all of the rest of the code is likely to error. This means that most tasks to remove jQuery from our modules cannot be done in small (easily reviewable) commits.
+
+All of our JS modules are started centrally by `modules.js` in the components gem. That means you don’t need to worry about manually calling a bit of JS to make a module run - if an element on a page has a `data-module="name-of-module"` attribute, it’ll get started. However if the code wasn’t a module to begin with, it may be manually started somewhere in the application, and this will need to be removed.
+
+## Rewriting tests
+
+We use Jasmine for testing our JS and our tests often use jQuery. Jasmine includes its own copy of jQuery which isn’t included in the public site, so jQuery does not have to be removed from tests. However some rewriting is necessary, for example:
+
+- module initialisation inside the tests will need to change from `start` to `init` (see below)
+- elements passed to the module from the test will need to be JS nodes, not jQuery elements. This can be achieved by changing e.g. `$('.element')` to `$('.element')[0]`
+
+In many tests the module is kept as a variable in order to check things, this may need to be restructured.
+
+```javascript
+// jQuery (old)
+var module = new GOVUK.Modules.nameOfModule.start(jQueryElement)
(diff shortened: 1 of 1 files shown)


source/manual/how-to-remove-jquery-from-javascript.html.md
```diff
@@ -0,0 +1,202 @@
...
+
+```javascript
+var jqExample = $('div')
+jqExample.addClass('test')
+```
+
+Using JS to find an element returns a node, which can have JS functions applied to it e.g.
+
+```javascript
+jsExample = document.querySelector('div')
```
injms, line 67:
```suggestion
var jsExample = document.querySelector('div')
```

andysellick: Done, thanks.

source/manual/how-to-remove-jquery-from-javascript.html.md
```diff
@@ -0,0 +1,202 @@
...
+  // do something
+}
+```
+
+### When to use querySelector and querySelectorAll
+
+In jQuery an element can be found on the page using `.find()`. There are two equivalent functions in JS.
+
+- [querySelector](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector) finds only the first element that matches the given query.
+- [querySelectorAll](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll) finds all elements that match that query, and returns an array of them.
```
injms, line 114: I'm going to be picky here - `querySelectorAll()` doesn't return an array but a non-live NodeList. A minor point, as the NodeList can be treated as an array in _most_ instances, but one that I think it worth clarifying as it can trip things up when `forEach()` is used in older browsers.

source/manual/how-to-remove-jquery-from-javascript.html.md
```diff
@@ -0,0 +1,202 @@
...
+Attempting the same thing with JS and `querySelector` will return `null` if no matching element is found, which means any attempt to do anything with that result will cause an error.
+
+This means more checks may have to be added to the code.
+
+```javascript
+var example = document.querySelector('.this')
+
+if (example) {
+  // do something
+}
```
injms, line 106: When using `querySelectorAll()` the length of the NodeList needs to be tested rather than the presence as an empty NodeList is still truthy; is it worth mentioning this?

```suggestion
```javascript
var example = document.querySelector('.this')

if (example) {
  // do something
}

var multipleElements = document.querySelectorAll('div')

if (multipleElements.length > 0) {
  // do something
}
```

source/manual/how-to-remove-jquery-from-javascript.html.md
```diff
@@ -0,0 +1,202 @@
...
+
+jQuery uses the same functions to both get and set attributes on elements.
+
+```javascript
+.attr('class') // replace with getAttribute('class')
+.attr('class', 'value') // replace with setAttribute('class', 'value')
+
+.data('item') // replace with getAttribute('data-item')
+.data('item', 'value') // replace with setAttribute('data-item', 'value')
+```
```
injms, line 132: Can the JavaScript equivalents be put in their own code block? This would follow the pattern set in the examples in "Chaining functions" and "jQuery objects are different from JS nodes".

source/manual/how-to-remove-jquery-from-javascript.html.md
andysellick: Updated, thanks.

source/manual/how-to-remove-jquery-from-javascript.html.md
andysellick: Updated, thanks.

source/manual/how-to-remove-jquery-from-javascript.html.md
andysellick: As a reference guide I thought it was clearer to keep them together, so readers can easily see the jQuery and JS equivalent. Happy to change to something else though if you can come up with an alternative - I can't think of a better way to do this.

andysellick: @injms I've had a rethink on this and tried your suggestion - please take a look.

injms: That's super and exactly what I had in mind - thank you

injms (approved):
📜 👍

PR #3243, opened 2021-09-21
Update govuk-alerts guidance to match new config

benthorner:
https://www.pivotaltracker.com/story/show/179482902

Previously the Integration and Staging versions of GOV.UK Alerts
both pointed at the same backend (in Notify's Staging environment).
Now that Notify has a dedicated backend for GOV.UK Integration to
point to, this updates the guidance to reflect that.


ChrisBAshton (approved):

benthorner:
Adding "DNM" until this change has been applied in Integration.

PR #3247, opened 2021-09-23
Consolidate and update troubleshooting information for SSH

brucebolt:
There was a page titled 'Unable to SSH into a machine' which explained the solution to a single problem. There was also a section titled 'Troubleshooting' in the how to SSH guide.

Consolidating these into a single manual page to make troubleshooting easier.

Also adding instructions on how to access an AWS EC2 instance's system log, which can be useful in diagnosing Puppet errors.


ChrisBAshton (approved):
⭐

PR #3278, opened 2021-11-02
Update DNS docs to notify senior tech and CDDO

richardTowers:
CDDO colleagues monitor the DNS records for gov.uk, and they would like
us to notify them before making changes so they know to ignore the
alarms we set off.

We also need to tell the GOV.UK lead technologists because they're the
people who'll get a call from JISC to confirm. And also it's generally
good for them to know what's going on.

### DIFF
--- a/source/manual/dns.html.md
+++ b/source/manual/dns.html.md
@@ -128,6 +128,7 @@ access.
 
 You should also make sure that the following groups of people are aware before
 requesting any changes:
+
 - 2nd line (via email)
 - GOV.UK's Head of Tech and the lead technologists
 - The CDDO domains team (the lead technologists can contact them)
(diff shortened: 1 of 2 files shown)


source/manual/dns.html.md
```diff
@@ -126,6 +126,12 @@ list. Speak to a senior technologist member of GOV.UK or someone in
 GOV.UK Replatforming if you need to make a change and don't have
 access.
 
+You should also make sure that the following groups of people are aware before
+requesting any changes:
```
ChrisBAshton, line 130: This should fix the linter:

```suggestion
requesting any changes:

```

ChrisBAshton (approved):

PR #3283, opened 2021-11-03
Update use of `low` priority Zendesk queue

deborahchua:
We have been using this queue for product related tickets but we now
have a new group to triage them too. This means 2nd line can use the
low queue for other purposes.

### DIFF
--- a/source/manual/zendesk.html.md
+++ b/source/manual/zendesk.html.md
@@ -23,7 +23,7 @@ relevant and actionable - and start with the oldest by date. If external assista
 line is required, assign them on the ticket and move on to the next ticket (keeping the ticket in the
 2nd line queue).
 
-Ignore 'Low' priority tickets - these have been triaged by 2nd line for GOV.UK product managers to review.
+'Low' priority tickets are for tickets that are `on-hold` and are not being actively worked on.
 
 New tickets are 'Normal' priority by default and require triaging (i.e. a developer reading and categorising
 it).
@@ -44,6 +44,7 @@ appropriate place.
 
 Below are some of the groups we often triage tickets to:
 
+* Product requests - `3rd Line--GOV.UK Product Requests`
 * Content support - `2nd Line--GOV.UK Content Triage, Gov't`
 * GDPR and Privacy enquiries - `3rd Line--GDS Privacy`
 * Licensing - `1st Line--Licensing Support`
@@ -82,16 +83,16 @@ You can use Internal Notes to keep a log of actions you've taken so far: this ca
 staff to pick up where you left off if you're unable to solve the ticket yourself. Please note that departments
 can see internal notes, so don’t write anything you wouldn’t say to someone publicly.
 
-If you need more information from the user, or confirmation that everything is now working as expected, fill
-in the "Public reply" as appropriate and click "Submit as Pending".
+If you need more information from the user, fill in the "Public reply" as appropriate and click "Submit as Pending".
 
 ## Closing a ticket
 
-Once you've resolved a ticket, click "Submit as Solved".
+Once you've resolved a ticket, click "Submit as Solved". You do not need to wait for user confirmation to
+close the ticket.
 
 ## Tickets left pending without a response
 
-Note that if a ticket has been resolved and pending a response for 5 or more days with no response from the
+Note that if a ticket has been pending a response for 5 or more days with no response from the
 requester, you can submit it as solved with a message telling them why and that if they still need help they
 can get in touch. Do this by clicking the "Apply macro" button at the bottom of the ticket screen, and
 choosing the "GOV.UK 2nd line tech: pending for 5+ days" option.


source/manual/zendesk.html.md
```diff
@@ -23,7 +23,10 @@ relevant and actionable - and start with the oldest by date. If external assista
 line is required, assign them on the ticket and move on to the next ticket (keeping the ticket in the
 2nd line queue).
 
-Ignore 'Low' priority tickets - these have been triaged by 2nd line for GOV.UK product managers to review.
+'Low' priority tickets are for tickets that are:
+• `on-hold` or are waiting to be handed over to another team (unless they are product related, they
+  can be assigned to `3rd line product`)
+• tickets that have been dealt with and can be closed if the user doesn't reply in 5 working days
```
leenagupte, line 29: I don't think the bullet points are rendering correctly.

deborahchua: Thanks @leenagupte , I think that's fixed now!

source/manual/zendesk.html.md
```diff
@@ -23,7 +23,10 @@ relevant and actionable - and start with the oldest by date. If external assista
 line is required, assign them on the ticket and move on to the next ticket (keeping the ticket in the
 2nd line queue).
 
-Ignore 'Low' priority tickets - these have been triaged by 2nd line for GOV.UK product managers to review.
+'Low' priority tickets are for tickets that are:
+* `on-hold` or are waiting to be handed over to another team (unless they are product related, they
+  can be assigned to `3rd Line--GOV.UK Product Requests`)
+* tickets that have been dealt with and can be closed if the user doesn't reply in 5 working days
```
leenagupte, line 29: This looks good. I'm not sure how it fits with this diagram though, which is what I use to triage tickets: https://docs.google.com/presentation/d/1EotoM2CVtqlnx54Qz5bP7OyIx5c9ji_GptUuymHkBrc/edit

ChrisBAshton: Thanks for updating the diagram Leena linked to above - looks good 👍 

Just to check my understanding: until now, 1) when a ticket is waiting on more info from the user, we'd mark it as "pending user response" and keep it in the High queue. And 2) when we solve issues, we 'optimistically close' the ticket with a comment, rather than wait for a user response asking them to _confirm_ it's resolved.

Are we now saying that 1) should be moved to the Low queue? **Or** are we saying that the way I've been doing 2) is incorrect (quite possible!) and that we should always wait for user confirmation before closing the ticket?

Depending on the answer above, I'd tweak the wording to either:

```suggestion
* tickets that have been resolved, now awaiting user confirmation ([can be closed](#tickets-left-pending-without-a-response) if no response in 5 working days)
```

...or:

```suggestion
* tickets that are waiting for a response from the user ([can be closed](#tickets-left-pending-without-a-response) if no response in 5 working days)
```

source/manual/zendesk.html.md
```diff
@@ -23,7 +23,10 @@ relevant and actionable - and start with the oldest by date. If external assista
 line is required, assign them on the ticket and move on to the next ticket (keeping the ticket in the
 2nd line queue).
 
-Ignore 'Low' priority tickets - these have been triaged by 2nd line for GOV.UK product managers to review.
+'Low' priority tickets are for tickets that are:
+* `on-hold` or are waiting to be handed over to another team (unless they are product related, they
+  can be assigned to `3rd Line--GOV.UK Product Requests`)
```
nicholsj, line 28: >can be assigned to `3rd Line--GOV.UK Product Requests`)

I don't think this is needed? Tickets should simply be assigned, not left in Low

source/manual/zendesk.html.md
deborahchua: Thanks @ChrisBAshton , that's a really good point. I've personally left tickets as `pending` and wait for the user to confirm it's been resolved, but contradicts [this part of the doc](https://docs.publishing.service.gov.uk/manual/zendesk.html#closing-a-ticket) - `Once you’ve resolved a ticket, click “Submit as Solved”.` I think we should be trying to close tickets as early as possible by default and would help fulfil our [SLAs](https://docs.publishing.service.gov.uk/manual/zendesk.html#service-level-agreements) , what do you think @benjamineskola ?

benjamineskola: I think closing early makes sense — if we're able to reproduce the issue and have been able to confirm for ourselves that it's fixed, I don't know if it's strictly necessary to wait for the user to respond before closing it. And if it turns out we're mistaken and it wasn't fixed, then it's reopened, no harm done.

If we do need the user to confirm that it's fixed, then that could wait around as `pending`, but that shouldn't be necessary in all cases.

benjamineskola (approved):

leenagupte (approved):
👍 Also, thanks for updating the flow chart as well ⭐

PR #3290, opened 2021-11-12
Make following the indices-fixing instructions easier

ChrisBAshton:
- Link to Jenkins job with pre-filled parameters
- Reminder of which machine to SSH to
- Reminder of which directory to CD into before attempting to run rake task
- Fix typo

### DIFF
--- a/source/manual/fix-out-of-date-search-indices.html.md
+++ b/source/manual/fix-out-of-date-search-indices.html.md
@@ -15,21 +15,22 @@ and `unpublish` messages that have not been processed need to be resent.
 Content in the `govuk` index is populated from the [Publishing API message queue][queue].
 Missing documents can be recovered by resending the content to the message queue. In the
 Publishing API, run the following rake task (including the quotes) to replay traffic between
-two datestamps:
+two datestamps (after SSH'ing into the `publishing_api` machine):
 
 ```
+cd /var/apps/publishing-api && \
 govuk_setenv publishing-api bundle exec \
 rake 'represent_downstream:published_between[2018-12-17T01:02:30, 2018-12-18T10:20:30]'
 ```
 
-You can also run this task from Jenkins.
+You can also [run this task from Jenkins](https://deploy.blue.production.govuk.digital/job/run-rake-task/parambuild/?TARGET_APPLICATION=publishing-api&MACHINE_CLASS=publishing_api&RAKE_TASK=represent_downstream:published_between[2018-12-17T01:02:30,%202018-12-18T10:20:30]).
 
 [Other replay options are available](https://github.com/alphagov/publishing-api/blob/main/lib/tasks/represent_downstream.rake), for example replaying all traffic for a single publishing app or doctype.
 Be aware that these options will replay the entire Publisher API history for that app or doctype, and may take some time.
 
 ## `government`/`detailed` indexes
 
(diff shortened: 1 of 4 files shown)


source/manual/fix-out-of-date-search-indices.html.md
```diff
@@ -15,21 +15,22 @@ and `unpublish` messages that have not been processed need to be resent.
 Content in the `govuk` index is populated from the [Publishing API message queue][queue].
 Missing documents can be recovered by resending the content to the message queue. In the
 Publishing API, run the following rake task (including the quotes) to replay traffic between
-two datestamps:
+two datestamps (after SSH'ing into the `publishing_api` machine):
 
 ```
+cd /var/apps/publishing-api && \
 govuk_setenv publishing-api bundle exec \
 rake 'represent_downstream:published_between[2018-12-17T01:02:30, 2018-12-18T10:20:30]'
```
1pretz1, line 23: Should we now remove the command line option from the docs, as the standard / preferred approach is to run Rake tasks from Jenkins?

source/manual/fix-out-of-date-search-indices.html.md
```diff
@@ -15,21 +15,22 @@ and `unpublish` messages that have not been processed need to be resent.
...
 ```
 
-You can also run this task from Jenkins.
+You can also [run this task from Jenkins](https://deploy.blue.production.govuk.digital/job/run-rake-task/parambuild/?TARGET_APPLICATION=publishing-api&MACHINE_CLASS=publishing_api&RAKE_TASK=represent_downstream:published_between[2018-12-17T01:02:30,%202018-12-18T10:20:30]).
```
1pretz1, line 27: We could add a link to each environment which may be helpful: https://github.com/alphagov/govuk-developer-docs/blame/322891a0379b180e10e2cb6593e502a76bf729bd/source/manual/howto-change-slug-and-create-redirect.html.md#L21-L28

1pretz1 (commented):
Looks good Chris - just left two suggestions.

source/manual/fix-out-of-date-search-indices.html.md
ChrisBAshton: Agreed. I've tacked on a couple of commits to tidy this up.

source/manual/fix-out-of-date-search-indices.html.md
ChrisBAshton: We _could_, but we do this very inconsistently throughout the docs. It also requires changing the file type (to end in `.erb`) which is a pain for version control diffs. I think I'll just leave this one 👍

1pretz1: The example I linked above just uses Markdown, rather than a Ruby helper so the file type could stay the same, but no worries, I'm not too fussed!

1pretz1 (approved):
Thanks for applying Jenkins > console to a few other places, and updating the rake task documentation. Above and beyond - great work @ChrisBAshton!

source/manual/fix-out-of-date-search-indices.html.md
ChrisBAshton: Ah yes, I did see that, but would suggest we use the helper for consistency, if we were to do this at all 👍

PR #3294, opened 2021-11-15
Update gnupg setup docs

benjamineskola:
People having trouble getting GPG working for access to `govuk-secrets` seems to be a recurring issue, so I'm trying to fill in some of the gaps in the documentation.

https://trello.com/c/VZyB4RIu/1242-gpg-secrets-documentation-out-of-date

### DIFF
--- a/source/manual/create-a-gpg-key.html.md
+++ b/source/manual/create-a-gpg-key.html.md
@@ -12,8 +12,6 @@ We use GPG keys to encrypt our secrets. Documentation for using your GPG key can
 
 Install `gpg` if you don't already have it. Use `brew install gpg-suite` to install the graphical [GPG Suite](https://gpgtools.org/).
 
-Once installed, you will likely have both `gpg` and `gpg2` on your machine. Always use `gpg2`.
-
 ## Creating a GPG key (using the GUI)
 
 [GPGtools](https://gpgtools.org/) comes with a GUI which can perform most of the operations you need.
@@ -30,7 +28,7 @@ See below for checking your passphrase.
 
 ## Creating a GPG key (using the command line)
 
-Create a gpg key with `gpg2 --gen-key` using your
+Create a gpg key with `gpg --gen-key` using your
 digital.cabinet-office.gov.uk email address. Defaults for the questions
 should be fine, although you should choose a 4096-bit key.
 
@@ -38,13 +36,13 @@ should be fine, although you should choose a 4096-bit key.
 >
 > You should also generate a [revocation
 > certificate](http://www.dewinter.com/gnupg_howto/english/GPGMiniHowto-3.html#ss3.4)
-> with `gpg2 --gen-revoke` and store it in a safe place (*not* on your
+> with `gpg --gen-revoke` and store it in a safe place (*not* on your
 > laptop, maybe a USB stick in your locker).
 
 ### Working out your key ID and fingerprint
 
 ```
-gpg2 --fingerprint firstname.lastname@digital.cabinet-office.gov.uk
+gpg --fingerprint firstname.lastname@digital.cabinet-office.gov.uk
 ```
 
 Should look something like this.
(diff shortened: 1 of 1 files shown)


benjamineskola:
An additional potential change is to replace the recommendation of `keyserver.ubuntu.com` with `keys.openpgp.org`.

There are [potential reliability issues with the SKS keyservers](https://code.firstlook.media/the-death-of-sks-pgp-keyservers-and-how-first-look-media-is-handling-it), although in practice I'm not sure we're really affected by it.

The downsides to switching, as far as I can see:

1. `keys.openpgp.org` isn't federated, so not all of the keys we use can be found there; that would mean a simple one-off task to upload them.
2. the privacy feature of not sharing unverified email addresses is not supported by stock gnupg (including the version in homebrew). However it works fine with the version from gpg-suite, which is what we recommend anyway.

source/manual/create-a-gpg-key.html.md
```diff
@@ -12,8 +12,6 @@ We use GPG keys to encrypt our secrets. Documentation for using your GPG key can
 
 Install `gpg` if you don't already have it. Use `brew install gpg-suite` to install the graphical [GPG Suite](https://gpgtools.org/).
 
-Once installed, you will likely have both `gpg` and `gpg2` on your machine. Always use `gpg2`.
```
ChrisBAshton, line 15: Nice find 👍 I can confirm mine seems to be v2:

```
$ gpg --version
gpg (GnuPG/MacGPG2) 2.2.24
```

source/manual/create-a-gpg-key.html.md
```diff
@@ -61,10 +61,10 @@ The key ID is `90E65803`, and the fingerprint is `37CC 021A C5C2
 
 ### Upload your GPG key to a keyserver
 
-Before doing this, make sure that your default keyserver is `http://keyserver.ubuntu.com/`. You can do this by changing the default keyserver in `~/.gnupg/gpg.conf`:
+Before doing this, make sure that your default keyserver is `https://keyserver.ubuntu.com/`. You can do this by changing the default keyserver in `~/.gnupg/gpg.conf`:
 
 ```
-keyserver http://keyserver.ubuntu.com/
+keyserver https://keyserver.ubuntu.com/
```
ChrisBAshton, line 65: Good change 👍 as you've said elsewhere in the PR, this server is problematic, and we've had more luck recently with keys.openpgp.org. This is what I have in my `~/.gnupg/dirmngr.conf`, which works well for me:

```
keyserver hkps://keys.openpgp.org
```

I agree it is probably worth changing our recommended server to this keys.openpgp.org one, in a new commit that justifies the reasoning.

ChrisBAshton (commented):
Looking good to me 👍 just a few thoughts/questions below.

source/manual/create-a-gpg-key.html.md
```diff
@@ -50,14 +50,14 @@ gpg2 --fingerprint firstname.lastname@digital.cabinet-office.gov.uk
 Should look something like this.
 
 ```
-pub   2048R/90E65803 2013-02-08
-      Key fingerprint = 37CC 021A C5C2 4E27 C4D9  5735 9B0E 9DD1 90E6 5803
-      uid                  my name <my.name@digital.cabinet-office.gov.uk>
-      sub   2048R/FDD27DBE 2013-02-08
```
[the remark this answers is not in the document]
benjamineskola: No, I only get one — I think any subkeys should be prefixed with `sub` (and actually, in the version I'm using it doesn't show any key ID for the subway at all).

I'm guessing the output format changed at least in part to discourage use of the 8-character version of the key ID, but it's a shame that it no longer seems to prefix the fingerprint with `fingerprint` like it used to.

source/manual/create-a-gpg-key.html.md
```diff
@@ -59,10 +59,12 @@ and the key ID is `90E65803` — the last 8 characters of the fingerprint.
 
 ### Upload your GPG key to a keyserver
 
-Before doing this, make sure that your default keyserver is `hkps://keys.openpgp.org/`. You can do this by changing the default keyserver in `~/.gnupg/gpg.conf`:
+Before doing this, make sure that your default keyserver is `hkps://keys.openpgp.org/`. You can do this by changing the default keyserver in `~/.gnupg/dirmngr.conf`:
 
 ```
 keyserver hkps://keys.openpgp.org/
+
+You should also ensure that there is no `keyserver` directive in `~/.gnupg/gpg.conf`.
```
ChrisBAshton, line 67: This should be moved below the ``` on line 68.

ChrisBAshton (approved):
LGTM 👍  thanks!

PR #3300, opened 2021-11-18
Standardise how we represent repositories in the Dev Docs

ChrisBAshton:
This is a work in progress.
TBC: what to do about the `apps` portion of URLs like https://docs.publishing.service.gov.uk/apps/govuk-jenkinslib.html (e.g. should this now be https://docs.publishing.service.gov.uk/repos/govuk-jenkinslib.html)?
Also TODO: figure out whether we can cut config down to just `github_repo_name`, `type` and `team`.

## What

Reorganises the old `applications.yml` to allow for non-apps to be added to the Dev Docs.

NB you may need to run `ulimit -n 2048` before running `GITHUB_TOKEN=$(cat ~/github_token.txt) bundle exec rake`. Will need to document this.

## Why

There is valuable documentation that govuk-developer-docs pulls from other repositories that are not deployable applications (e.g. govuk-jenkinslib, govuk-docker etc). The only option prior to this PR was to add such repos to `applications.yml`, which is confusing as they're not applications, and many of the configuration options (`hosted_on`, etc) don't apply.


ChrisBAshton:
Superseded by #3369.

PR #3302, opened 2021-11-19
Change the govuk-publishing to govuk-publishing-tech

beccapearce:
Change the name of the team owning the publishing applications so they 
will appear in the correct slack channel when we receive dependency 
updates from the https://github.com/alphagov/govuk-dependencies application.

Trello card: https://trello.com/c/S6LHALdf/40-set-up-alerts-to-slack-for-dependencies-for-the-apps-we-own

### DIFF
--- a/data/applications.yml
+++ b/data/applications.yml
@@ -16,25 +16,25 @@
 
 - github_repo_name: collections-publisher
   type: Publishing apps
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Nice one 👍

PR #3304, opened 2021-11-22
Add Assitiv Labs documentation

chris-gds:
## What 

Add some information about [AssitivLabs](https://assistivlabs.com/)

- what it is, what it does
- how to request access
- highlight that it can be used locally

## Why

- For reference & future developers.
- Highlight the use of AssitivTunnel for local testing.

## Visuals

image

## Anything else

[GDS Way](https://gds-way.cloudapps.digital/manuals/accessibility.html#testing-with-assistive-technologies) has also been referenced and linked.

### Helpful startup commands:
```
govuk-docker-up env GITHUB_TOKEN=$(cat ~/github_token.txt)
GITHUB_TOKEN=$(cat ~/github_token.txt) ./startup.sh
```

### DIFF
--- a/source/manual/how-to-test-with-assitive-technology.html.html.md
+++ b/source/manual/how-to-test-with-assitive-technology.html.html.md
@@ -1,0 +1,38 @@
+---
+owner_slack: "#govuk-frontenders"
+title: How to test with Assistiv Labs
+section: Frontend
+layout: manual_layout
+parent: "/manual.html"
+---
+
+GOV.UK has to meet government [accessibility requirements](https://www.gov.uk/service-manual/helping-people-to-use-your-service/making-your-service-accessible-an-introduction), digital services must work on the most commonly used assistive technologies - including screen magnifiers, screen readers and speech recognition tools
+
+[Assistiv Labs](https://assistivlabs.com/) remotely connects you to real assistive technologies, like NVDA, JAWS, and Windows High Contrast Mode, using any modern web browser.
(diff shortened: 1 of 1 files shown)


source/manual/how-to-test-with-assitive-technology.html.html.md
```diff
@@ -0,0 +1,38 @@
...
+section: Frontend
+layout: manual_layout
+parent: "/manual.html"
+---
+
+GOV.UK has to meet government [accessibility requirements](https://www.gov.uk/service-manual/helping-people-to-use-your-service/making-your-service-accessible-an-introduction), digital services must work on the most commonly used assistive technologies - including screen magnifiers, screen readers and speech recognition tools
+
+[Assistiv Labs](https://assistivlabs.com/) remotely connects you to real assistive technologies, like NVDA, JAWS, and Windows High Contrast Mode, using any modern web browser.
+
+## Get setup
```
maxgds, line 13: Tiny nitpick - here and I think in one other instance you say "setup" without a space which is a noun, you should be using "set up" as a verb.

maxgds (approved):
One tiny change requested, otherwise this looks like a useful addition, thanks!

source/manual/how-to-test-with-assitive-technology.html.html.md
chris-gds: Huh, did not know this - updated

PR #3316, opened 2021-11-29
Add Signon access to Get Started guide

brucebolt:
Getting access to Signon allows new developers to see the user-facing side of their team's applications and allow them to replicate bugs without needing to access the production environment.

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -319,6 +319,15 @@ If you forget your `aws-vault` password, you must reset that password.
 1. Delete the `aws-vault` keychain by running `rm ~/Library/Keychains/aws-vault.keychain-db` in the command line.
 1. Re-initialise the `gds-cli` by opening `~/.gds/config.yml` and changing `initialised: true` to `initialised: false`.
 
+## 9. Get a Signon account for integration
+
+[Signon](https://docs.publishing.service.gov.uk/apps/signon.html) is the application used to control access to the
+GOV.UK Publishing applications.
+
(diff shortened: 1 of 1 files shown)


source/manual/get-started.html.md
```diff
@@ -319,6 +319,15 @@ If you forget your `aws-vault` password, you must reset that password.
...
 1. Re-initialise the `gds-cli` by opening `~/.gds/config.yml` and changing `initialised: true` to `initialised: false`.
 
+## 9. Get a Signon account for integration
+
+[Signon](https://docs.publishing.service.gov.uk/apps/signon.html) is the application used to control access to the
+GOV.UK Publishing applications.
+
+Ask someone with production access (e.g. your tech lead or buddy) to [create an account for the integration
+Signon](https://signon.integration.publishing.service.gov.uk/users/invitation/new), with permission to access
+the applications that your team are likely to work on.
```
ollietreend, line 329: Is any particular level of access required for somebody to create new accounts? Or is this something that anybody with integration Signon access should be able to do?

beccapearce (approved):
Nice

ollietreend (approved):
This looks good 🙌🏻

I've added one question, but possibly more just for my own understanding. It might not affect what goes into the page content.

source/manual/get-started.html.md
brucebolt: It requires "superadmin" level in Signon, which all GOV.UK developers with ["production access"](https://docs.publishing.service.gov.uk/manual/rules-for-getting-production-access.html#what-production-access-means) are given.

PR #3318, opened 2021-11-29
Update 'getting started' docs for accessing AWS

ollietreend:
The CLI command given in step '8. Access AWS for the first time' attempted to log the user in to AWS as a power user. However the user will only have read-only permissions at this point (as granted in step 7).


kevindew (approved):
Nice catch 👍  thanks

PR #3328, opened 2021-12-02
Document how to test with current PostgreSQL and MySQL versions

kevindew:
Trello: https://trello.com/c/oRrpsveM/64-investigating-running-current-version-of-postgresql-and-mysql-on-govuk-ci

This explains the connection strings and set-up needed to utilise the
Docker hosted versions of the databases in CI. These were installed via
Docker as a means to work around the fact we run CI off a very old
operating system (Ubuntu Trusty).

### DIFF
--- a/source/manual/test-and-build-a-project-on-jenkins-ci.html.md
+++ b/source/manual/test-and-build-a-project-on-jenkins-ci.html.md
@@ -18,7 +18,7 @@ Application tests run in a continuous integration (CI) environment.
 
 You need to add your repo to the `govuk_ci::master::pipeline_jobs` in [govuk-puppet][pipeline_jobs]. This will create the job in Jenkins.
 
-[pipeline_jobs]: https://github.com/alphagov/govuk-puppet/blob/master/hieradata/common.yaml
+[pipeline_jobs]: https://github.com/alphagov/govuk-puppet/blob/main/hieradata_aws/common.yaml
 
 ### 2. Add a Jenkinsfile
 
@@ -63,7 +63,7 @@ Many GOV.UK applications test against the
 [content schemas](https://github.com/alphagov/govuk-content-schemas/).
 
 To test your application for each PR on govuk-content-schemas, add it to the [govuk-content-schemas
-Jenkinsfile](https://github.com/alphagov/govuk-content-schemas/blob/master/Jenkinsfile).
+Jenkinsfile](https://github.com/alphagov/govuk-content-schemas/blob/main/Jenkinsfile).
 
 ## Specifying which database to use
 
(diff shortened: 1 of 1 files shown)


source/manual/test-and-build-a-project-on-jenkins-ci.html.md
```diff
@@ -63,6 +63,31 @@ Many GOV.UK applications test against the
 To test your application for each PR on govuk-content-schemas, add it to the [govuk-content-schemas
 Jenkinsfile](https://github.com/alphagov/govuk-content-schemas/blob/master/Jenkinsfile).
 
+## Using Docker databases
+
+We have utilised Docker as a way to provide access to different versions of
+databases which can be used as part of your build. These can be accessed with
+the following connection strings:
```
ollietreend, line 70: I'd be tempted to reframe this section to be tech-agnostic – for example, "Specify which database version to use" or similar.

```suggestion
## Specifying which database version to use

There are different versions of the MySQL and Postgres database engines which
can be used as part of your build. These can be accessed with
the following connection strings:
```

The fact that the different database versions run using docker seems like an implementation detail that probably isn't relevant here.

source/manual/test-and-build-a-project-on-jenkins-ci.html.md
```diff
@@ -63,6 +63,31 @@ Many GOV.UK applications test against the
...
+
+- MySQL 8: `mysql2://root:root@127.0.0.1:33068/<your-database-name>`
+- PostgreSQL 13: `postgresql://postgres@127.0.0.1:54313/<your-database-name>`
+
+Typically for these to work with a Rails application you'll need to set an
+environment variable. For example, in your Jenkinsfile, for an application
+that uses `TEST_DATABASE_URL`, you can set:
```
ollietreend, line 77: It might also be nice to explicitly detail what the _default_ database versions are, and what will happen if you don't set the connection string.

```suggestion
- MySQL ?.? (default): `mysql2://root:root@127.0.0.1:3306/<your-database-name>`
- MySQL 8: `mysql2://root:root@127.0.0.1:33068/<your-database-name>`
- Postgres 9.6 (default): `postgresql://postgres@127.0.0.1:5432/<your-database-name>`
- Postgres 13: `postgresql://postgres@127.0.0.1:54313/<your-database-name>`

Typically for these to work with a Rails application you'll need to set an
environment variable. If no environment variable is present, the app
will use the default database version.

For example, in your Jenkinsfile, for an application
that uses `TEST_DATABASE_URL`, you can set:
```

kevindew:
Thanks @ollietreend, great shouts. I don't want to dig too much into documenting the current(/past) approach on this as that's a bit of a rabbit hole, so I've just tried to neaten it up a bit.

ollietreend (approved):
Looks good @kevindew 👍🏻

PR #3335, opened 2021-12-07
Remove obsolete docs about offsite Carrenza backups and alerts

richardTowers:
All of the offsite backups have been replaced by govuk_env_sync in the AWS
environment. Carrenza has been completely decommissioned for some time
now, so these docs could only cause confusion.

The alert about postgres replication isn't relevant in the AWS environment either.


sengi (approved):
Thanks!

PR #3343, opened 2021-12-10
Document the update SpeedCurve lux.js process

kevindew:
This is documentation to pair with
https://github.com/alphagov/govuk-puppet/pull/11339
It documents what steps need to be taken to update lux.js file. This is
so that anyone who sees the alert can react to it.

I was a little surprised how nuanced this process was and expect that
it'll need some input from someone experienced in the tool whenever it
takes place.


alex-ju (approved):
Thanks for documenting the process! 🙌  Nothing I can come up with as suggestions.

PR #3356, opened 2021-12-16
Document error messages and workarounds for importing GPG keys

ChrisBAshton:
I could not find these error messages in the docs, and the
eventual workaround was not signposted very well. Hopefully this
documentation change will help the next developer who has to
import a GPG key.


karlbaker02 (approved):
Much clearer, thanks Chris 💯

PR #3359, opened 2021-12-20
Update manual to include guidance and tools for digital accessibility…

jon-kirwan:
## What
https://trello.com/c/ozOtWR3S/945-signpost-the-guidance-and-tools-for-digital-accessibility-page-for-devs

Update manual to include a new section / page for "guidance and tools for digital accessibility"

## Why
So GOV.UK teams know where to find accessibility resources/tools and are empowered to create accessible products and address accessibility issues.

### DIFF
--- a/source/manual/guidance-and-tools-for-digital-accessibility.html.md
+++ b/source/manual/guidance-and-tools-for-digital-accessibility.html.md
@@ -1,0 +1,10 @@
+---
+owner_slack: "#govuk-frontenders"
+title: Guidance and tools for digital accessibility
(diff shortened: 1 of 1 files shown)


maxgds (approved):
This is fine to go in, although it feels a little sparse. It might be worth expanding later with a short paragraph to explain why a frontender on gov.uk needs to care about accessibility and that the advice is all centralised behind that link. Don't let that stop you merging, but see what the content side of the team think and if they might be able to craft a short paragraph to that effect if they agree.

PR #3362, opened 2021-12-22
Update asset removal instructions

brucebolt:
Makes it clearer how to determine whether an asset is a Whitehall asset or from another application.

Also updates a few of the other steps where the process has changed.

### DIFF
--- a/source/manual/manage-assets.html.md.erb
+++ b/source/manual/manage-assets.html.md.erb
@@ -11,24 +11,47 @@ parent: "/manual.html"
 If you need to remove an asset manually from `assets.publishing.service.gov.uk`,
 follow these steps:
 
-1. Mark the asset as deleted. Sometimes it may be necessary to completely scrub an asset from our systems e.g. if it contains secret information. In this case, add `,true` to the appropriate rake task:
+1. Identify whether the asset was published using Whitehall or another application.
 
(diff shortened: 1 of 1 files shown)


source/manual/manage-assets.html.md.erb
```diff
@@ -11,24 +11,47 @@ parent: "/manual.html"
...
-  <%= RunRakeTask.links("asset-manager", "assets:whitehall_delete[LEGACY_URL_PATH]") %>
+   > An example Whitehall asset URL is `https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/12345/filename.pdf`.
+   >
+   > An example non-Whitehall asset URL is `https://assets.publishing.service.gov.uk/media/abcd1234efgh/filename.pdf`.
 
-1. Add a cache bust and check that the asset responds with a 404 not found
+   Decide if the asset can be marked as deleted or whether all traces need to be
+   removed (e.g. if it contains secret information). If all traces need to be
+   removed, append `, true` in the square brackets when running the rake tasks
+   beloe.
```
deborahchua, line 23: I think that should be `below`

brucebolt: Thanks.  Have corrected this.

PR #3367, opened 2022-01-04
Simplify 'Deploy Terraform' docs

ChrisBAshton:
The previous docs described in detail how to use the gds-cli and
deploy.rb script to deploy Terraform, but both of these methods
simply trigger the same parameterised Jenkins build under the
hood. There's no particular advantage to this abstraction layer,
and furthermore, no way of triggering a build of a branch of
changes, which can only be done by configuring the job in Jenkins
manually. As this is quite a common use case for the job, it
makes sense to put the Jenkins instructions first, and to highlight
this as the primary means of deploying Terraform.

With this as the preferred approach, we no longer need to
document how to get a GitHub token (it isn't required by the job),
and we don't need to repeat the definition of each parameter either.

I've kept a short section at the end of the doc to make a reference
to these alternative means of triggering the jobs, for sake of
completion.

### DIFF
--- a/source/manual/deploying-terraform.html.md
+++ b/source/manual/deploying-terraform.html.md
@@ -27,81 +27,11 @@ poweruser role.
 
 ### 2. Install `gds-cli`
 
-[`gds-cli`][gds-cli] is the [preferred way](/manual/get-started.html) of obtaining
-AWS credentials.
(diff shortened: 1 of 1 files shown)


baisa (approved):

baisa (commented):
This will be helpful 👍

PR #3369, opened 2022-01-10
Standardise how we represent repositories in the Dev Docs

ChrisBAshton:
## What

This PR changes our old `applications.yml` file to a new `repos.yml` file, meaning we can now explicitly add non-apps to the Developer Docs, such as `govuk_publishing_components`. The old `Applications` class is now `Repos`. URLs have changed too, i.e. `/apps/*.html` => `/repos/*.html`. An attempt to rename `App` to `Repo` was attempted, but proved problematic, so will continue in a separate PR: #3373.

Now, a repo is only considered to be an app if it is hosted somewhere. There are a couple of visual changes as part of this PR:

### [Applications on GOV.UK](https://docs.publishing.service.gov.uk/apps.html) page

| Before | After |
|--------|------|
|Screenshot 2022-01-11 at 16 44 49|Screenshot 2022-01-11 at 16 44 34|

### Individual app pages

| Before | After |
|--------|------|
|Screenshot 2022-01-11 at 17 22 10| Screenshot 2022-01-11 at 17 21 58|

## Why

The former `applications.yml` file was intended for just applications, but over time, [non-apps were added](https://github.com/alphagov/govuk-developer-docs/blob/fdf2d43d5b0109dfb3955759f909917aead647a1/data/applications.yml#L452-L457), in order to benefit from the [documentation centralisation](https://docs.publishing.service.gov.uk/apps/govuk-docker.html#imported-documents) that the GOV.UK Developer Docs provides.

This situation was somewhat confusing for non-developer stakeholders and developers alike, when trying to ascertain how many apps there are and who is responsible for them.

The definition of an 'app' is somewhat woolly. On GOV.UK, an 'app' tends to mean something that is deployed to the standard GOV.UK infrastructure, e.g. a publishing or frontend app. This is in contrast to other things that could be considered an 'app', e.g. the [Seal](https://docs.publishing.service.gov.uk/apps/seal.html), and other things that could _not_ be considered an 'app', e.g. [govuk_app_config](https://docs.publishing.service.gov.uk/apps/govuk_app_config.html).

I initially considered breaking these down into separate YAML files, in #3300, to better name each group. However, this opened a can of worms - what do we name each group, how granularly do we break down groups, do we maintain alphabetical or chronological order, or subgroupings preceded with a special comment? - and it was decided in a Developer Tools meeting to simply stick to one list of 'repos' instead.


app/repo.rb
```diff
@@ -35,39 +35,45 @@ def production_hosted_on_aws?
...
   end
 
   def private_repo?
-    app_data["private_repo"]
+    repo_data["private_repo"]
   end
 
   def page_title
     if retired?
       "Application: #{app_name} (retired)"
```
maxgds, line 67: This looks like things turn back into apps when they get retired regardless of whether they were beforehand?

maxgds (commented):
I like this idea but the code seems to have a lot of lingering references to app_name which is potentially confusing, and I'm not wedded to the definition of an app being something that is hosted. APIs don't feel like apps, while the seal does. But I accept that this is hard and I'm not going to dig my heels in over any of it.

app/repo.rb
ChrisBAshton: Good spot. Fixed (see final commit) 👍

ChrisBAshton:
@maxgds I can look at removing references to `app_name` in #3373 (where I will be renaming the `App` class to `Repo`). I had attempted to do so here, but it broke the build and is likely to require further changes like the `app_name` change you've suggested, so I think would be best continued in a separate PR. Quite keen to get this PR merged first, due to its size and also because any changes to the existing application.yml in the meantime will require a multi-commit rebase.

NB, I agree APIs don't feel like apps, but they are treated as apps in the context of our infrastructure, the fact we deploy them, etc. Because it's so subjective/contextual, I've opted to treat everything as a 'repo' as far as possible, rather than get caught up in naming and grouping (as I originally tried to do in #3300).

huwd:
Both the ratonale and the code look good to me - I like the idea of repos and apps being separated out.

I am also a tad confused about some of the remaining variable names, but I agree given the size and momentum here perhaps we can merge this as the user-facing change and document a need for a separate tidy up ticket in issues?

You want me to approve or to get a few more 👀 on this first?

ChrisBAshton:
Thanks @huwd! I'd appreciate the ✅ so it's ready to go, but I'll give it another day or so before I merge (and will pop a reminder for people in Slack tomorrow if they want to review).

spec/app/repos_spec.rb
```diff
@@ -0,0 +1,49 @@
+RSpec.describe Repos do
+  before :each do
+    allow(Repos).to receive(:all) do
+      applications.map(&:stringify_keys).map { |app_data| App.new(app_data) }
+    end
+  end
+
+  describe "public" do
```
kevindew, line 8: It looks like this is halfway to using the conventional define method approach for RSpec: https://www.betterspecs.org/#describe so this should be `describe ".public"` Looking at the source code I think this project is almost using the regular `describe "#instance_method"` or `describe ".class_method"` style so we probably want to bring this file into line.

kevindew (approved):
Amazing work Chris 👏  Thanks for chipping away at this long term lingering issue.

I agree with some of the other comments about bits of "app" remaining that should be thought about and I've left a few comments to that effect myself. But since this is already pretty big it seems best to iterate the changes in.

Only other thing I left is an issue with the test naming, but I'm going to approve as I don't mind if you fix that in a separate PR (I think it may affect a couple of other files).

ChrisBAshton:
Thanks Kevin - I'll attempt to address those comments in #3373. 👍

PR #3374, opened 2022-01-12
Remove references to fabric from failover docs

huwd:

Fabric previously ran sets of commands across multiple machines.
It has now been archived and is no longer in use due to difficulty in
upgrading python versions.

This begins a process of removing references to fab from our docs. Until
we find a new tool that allows us to run commands across multiple
machines, they will need to be a bit more verbose where a command must
be run on every machine.

### DIFF
--- a/source/manual/fall-back-to-mirror.html.md
+++ b/source/manual/fall-back-to-mirror.html.md
@@ -70,16 +70,23 @@ To avoid Fastly traffic hitting Origin when Origin is down (potentially making t
 
 Alternatively, we can stop [Nginx](https://www.nginx.com/) on the cache machines, which will prevent requests hitting GOV.UK applications. Fastly will automatically retry these failed requests against the mirror.
 
-To stop Nginx on the cache machines, run the following command:
+SSH into each cache machine (you can increment box number after the colon to hit each one in turn):
 
 ```bash
-$ fab $environment class:cache incident.fail_to_mirror
+$ gds govuk connect -e production ssh cache:1
 ```
 
-To disable this fallback:
+Stop Nginx to force use of mirrors:
 
 ```bash
-$ fab $environment class:cache incident.recover_origin
+$ govuk_puppet --test --disable "fail_to_mirror task (by $USER)"
+$ sudo service nginx stop
+```
+
+When required you can re-enable puppet, which will restart Nginx:
+
+```bash
+$ govuk_puppet --test --enable
 ```
 
 ## Emergency publishing content using the GOV.UK mirror


source/manual/fall-back-to-mirror.html.md
```diff
@@ -63,20 +63,36 @@ cd /mnt/crawler_worker/www.gov.uk
...
+gds govuk connect -e production ssh cache:1
 ```
 
-To disable the fallback:
+and run:
 
-```bash
-$ fab $environment class:cache incident.recover_origin
+```
+  sudo govuk_puppet --test --disable "fail_to_mirror task (by $USER)"
```
barrucadu, line 79: There's no need to use `sudo` with `govuk_puppet`

Also it would be good to give the command to run to re-enable puppet as well

huwd:
Seems someone recently made a change to this file too 🤔 
My bad for letting it go stale. Will compare and see if there's still a change for me to make here.

source/manual/fall-back-to-mirror.html.md
```diff
@@ -73,13 +73,20 @@ Alternatively, we can stop [Nginx](https://www.nginx.com/) on the cache machines
 To stop Nginx on the cache machines, run the following command:
 
 ```bash
-$ fab $environment class:cache incident.fail_to_mirror
+$ gds govuk connect -e production ssh cache:1
```
benthorner, line 76: This isn't equivalent - probably needs splitting into the Puppet step and the Nginx step as they both happen in different contexts.

huwd: I think I'm missing something here - do you mean I need to be clearer about ssh-ing into multiple boxes?
Then... run disable puppet on all first, _then_ disable nginx?

Looking [at this](https://github.com/alphagov/fabric-scripts/blob/master/incident.py#L9-L12), seemed these were run one after another... on the same cash box?

benthorner: Oops, I was half talking rubbish, sorry.

The steps look good, just the intro is confusing: "to stop Nginx, run...ssh" ❌.

Suggest replacing with "SSH onto each cache machine e.g.".

huwd: gotcha! Yeah that's a good suggestion I'll amend now - thanks 😄!

source/manual/fall-back-to-mirror.html.md
```diff
@@ -70,16 +70,23 @@ To avoid Fastly traffic hitting Origin when Origin is down (potentially making t
...
 
-To stop Nginx on the cache machines, run the following command:
+SSH into each cache machine (you can increment box number after the colon to hit each one in turn):
 
 ```bash
-$ fab $environment class:cache incident.fail_to_mirror
+$ gds govuk connect -e production ssh cache:1
 ```
 
 To disable this fallback:
```
benthorner, line 79:
```suggestion
Stop Nginx to force use of mirrors
```

This used to be paired with _starting_ Nginx.

source/manual/fall-back-to-mirror.html.md
```diff
@@ -70,16 +70,23 @@ To avoid Fastly traffic hitting Origin when Origin is down (potentially making t
...
 
 ```bash
-$ fab $environment class:cache incident.recover_origin
+$ govuk_puppet --test --disable "fail_to_mirror task (by $USER)"
+$ sudo service nginx stop
+```
+
+When required you can re-enable puppet with the following command:
```
benthorner, line 86:
```suggestion
When required you can re-enable puppet, which will restart Nginx:
```

benthorner (approved):
Thanks Huw! I think we can still be more accurate when we describe what each set of commands does, but the main thing is that the commands are correct 👍.

PR #3391, opened 2022-01-25
Document how to get access to Sentry

ChrisBAshton:


maxgds (commented):
Thanks - who might our Sentry Owner or Manager be? If not names, roles. Would be good to know who to chase if nothing happens.

ChrisBAshton:
@maxgds looking at https://sentry.io/settings/govuk/members/:

'Owners' include the lead developers, head of technology, and a couple of other senior tech.
'Managers' and 'Admins' seem to both be a fairly small, random mix of senior developers.
We should probably do a better job at ensuring all tech leads and/or all seniors are either a Manager or an Admin.

PR #3394, opened 2022-01-26
Remove references to blue-postgresql-primary

kevindew:
Trello: https://trello.com/c/6fLV2SRM/103-review-documentation-for-references-to-old-database-architecture

We no longer have this RDS instance. I've replaced the specific link
(Email Alert API) as a direct one to the instance, for the generic one
I'm just not including a database variable so Grafana picks the first
one and the user of the graph can determine the appropriate one.


ChrisBAshton (approved):
Makes sense 👍

PR #3398, opened 2022-01-26
Fix references to singular DB Admin machine

kevindew:
Trello: https://trello.com/c/6fLV2SRM/103-review-documentation-for-references-to-old-database-architecture

This goes through the places where we're referencing DB Admin machines and adapts the docs to reflect that there are now multiple. I also cleaned up some of the cruft I found along the way.

Best way to review is to go through commit by commit.


source/manual/non-ha-node-classes.html.md
```diff
@@ -14,14 +14,12 @@ Installation of a critical software component on a non-redundant node class has
...
 - asset-master
 - jumpbox
 - ckan
 - apt
 - mirrorer
 - licensing_backend
+- db-admin or any of the \*-db-admin machines
 
-If in doubt, please discuss your requirements with RE GOV.UK (#govuk-2ndline on Slack).
+If in doubt, please discuss your requirements with GOV.UK Platform Reliability (#govuk-platform-reliability-team on Slack).
```
ChrisBAshton, line 25: 👍

source/manual/govuk-env-sync.html.md
```diff
@@ -18,12 +18,12 @@ In this context:
...
+The environment synchronisation is achieved by granting cross-account access of the `govuk-<environment>-database-backups` S3 buckets.
 
-![Schematic of the data flow of the govuk_env_sync data synchronisation](images/govuk_env_sync.png)
-Figure 1:  Schematic of the data flow of the govuk_env_sync data synchronisation.
-
-In practice, some Integration databases pull from Production (with data sanitisation - see below) rather than from Staging. This is done to reduce the number of cron job timing dependencies and to save resources where large databases do not need to be backed up from Staging.
+* In production, databases _push_ to the `govuk-production-database-backups` S3 bucket.
+* In staging, databases are replaced by _pulling_ data from `govuk-production-database-backups`.
+* In integration, databases are replaced by _pulling_ data from `govuk-production-database-backups` - some databases have a [data sanitisation process](#data-sanitisation).
+* In integration, databases are _pushed_ to `govuk-integration-database-backups` nightly - these are produced for developers to use [real data on their local machines](/repos/govuk-docker/how-tos.html#how-to-replicate-data-locally).
```
ChrisBAshton, line 26: Nice one for explaining why integration databases are backed up, as I know this came up a couple of times in the recent DB upgrade work.

Is it worth explicitly stating that staging doesn't have a mechanism for pushing its data to a `govuk-staging-database-backups` bucket (as implied in its absence above)?

ChrisBAshton (approved):
LGTM ⭐

source/manual/govuk-env-sync.html.md
kevindew: I don't think I really want to dig into that too much as it ends up a bit confusing with the special case for Whitehall. I feel like I'll either leave a lie or it be overly verbose.

PR #3399, opened 2022-01-27
Update CDN mirror docs

jonathanglassman:
## What

Review documentation of the processes to update the CDN mirror, and improve them if applicable

First step is to collate/compare the current docs - then check for accuracy/completeness with help from developers. 

## Why

Each time they're looked at, there are questions. We need this documentation to be clear so it can be understood, especially in an urgent situation

## More info

Relevant docs at:
https://docs.publishing.service.gov.uk/manual/fall-back-to-mirror.html#updates-to-the-mirror
https://docs.publishing.service.gov.uk/manual/alerts/mirror-sync.html
https://docs.publishing.service.gov.uk/manual/architecture-deep-dive.html#failover

Plus the first part of [Chris's comment here](https://trello.com/c/RmUQlkKO/2798-fix-incorrect-mime-type-for-govuk-homepage-being-served-from-static-mirror-5#comment-61b70312244fe77b1e79c660) is a helpful summary

Trello card: https://trello.com/c/d30upIju/2832-improve-documentation-of-cdn-mirror-process

### DIFF
--- a/source/manual/fall-back-to-mirror.html.md
+++ b/source/manual/fall-back-to-mirror.html.md
@@ -1,92 +1,92 @@
 ---
 owner_slack: "#govuk-2ndline"
-title: Fall back to the static mirrors
+title: GOV.UK CDN static mirrors
 section: Deployment
 layout: manual_layout
 parent: "/manual.html"
 ---
 
-We maintain a static copy of most of the site, which gets used by the content delivery
-network (CDN) whenever origin (the application server) times out or serves an error
-response.
+The GOV.UK mirror is a static copy of GOV.UK.
 
-This process is handled by our CDN config and is entirely transparent to us and
-our users. It happens multiple times a day, for lots of different reasons.
+## When is the GOV.UK mirror used?
 
-This is why we refer to switching off Nginx on the origin cache machines as
-"falling back to the mirrors".
+Most requests to GOV.UK are handled by the [Fastly content delivery network (CDN)](/manual/cdn.html), which sits between the user and [GOV.UK Origin](/manual/architecture-shallow-dive.html#a-user-visits-the-gov-uk-homepage).
 
-## Hosting
+When a user requests a GOV.UK page, Fastly retrieves that page from its cache, or fetches the page from GOV.UK Origin if Fastly does not have the page in its cache.
 
-GOV.UK is mirrored to two cloud providers:
+Sometimes, GOV.UK Origin may time out or return a 5xx error response. When that happens, Fastly automatically fetches the page from the GOV.UK mirror instead.
 
-- Amazon S3: the static mirror is hosted in a bucket `govuk-<environment>-mirror` and the
-  content is retrieved by the CDN using an API.  This bucket is replicated to
-  `govuk-<environment>-mirror-replica` in another AWS region.
+If Fastly goes down, we would manually [switch to AWS CloudFront](https://docs.publishing.service.gov.uk/manual/fall-back-to-aws-cloudfront.html) instead of Fastly. Where Fastly makes requests to GOV.UK Origin, AWS CloudFront instead makes all its requests to the GOV.UK mirror.
 
-- Google GCS: the static mirror is hosted in a bucket `govuk-<environment>-mirror` and the
-  content is retrieved by the CDN using an API.
+## GOV.UK mirror locations and access
 
-## Access
+The GOV.UK mirror is hosted in an [Amazon Web Services (AWS) S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html). The bucket contains copies of GOV.UK HTML files. The mirror is static, meaning dynamic pages such as search pages will not work.
 
-Access to the:
+The term "GOV.UK mirror" actually refers to 3 separate mirrors:
 
-1. Amazon S3 buckets are restricted to Fastly, Office and Pingdom IP addresses for read-only access
-   and authenticated users in AWS web console.
+- the main `govuk-<environment>-mirror` S3 bucket in one AWS region
+- the `govuk-<environment>-mirror-replica` S3 bucket in another region, so if the first AWS region is down, we can fall back to this other region
+- the `govuk-<environment>-mirror` bucket in Google Cloud Storage (GCS), so if AWS overall is down, we can fall back to this mirror
 
-1. Google GCS buckets are restricted by secret keys in `govuk-secrets` and authenticated users
-   in Google GCP web console.
+Access to the Amazon S3 buckets is restricted. If you have a Fastly, Office or Pingdom IP address, you have read-only access. If you’re an authenticated AWS web console user, you have read-write access.
 
-## Updates to the mirror
+Access to the GCS bucket is also restricted. You can access the GCS bucket if you can access the secret keys in `govuk-secrets`, or if you’re an authenticated Google Cloud Platform (GCP) web console user.
 
-Every day at 20:00, the [govuk_seed_crawler][] on the Mirrorer machine adds all GOV.UK URLs listed
-within the sitemaps (in the magnitude of hundreds of thousands) to a message queue. The [govuk_crawler_worker][]
-on the Mirrorer machine consumes these URLs, retrieves the HTML returned by these URLs, saves the content
-to disk and adds any new URLs found on those pages to the back of the queue.
+## Updates to the GOV.UK mirror
 
-Every hour, the static copy of the site is copied from the Mirrorer machine to the primary AWS S3 bucket
-`govuk-<environment>-mirror`. This primary bucket is automatically replicated to another S3 bucket
-(named `govuk-<environment>-mirror-replica`) in another region by AWS.
+Automatic scripts update the GOV.UK mirror every day.
 
-In addition, the primary AWS S3 bucket `govuk-<environment>-mirror` is synced to the Google GCS bucket
-of the same name daily at 12:00.
+The [`govuk_seed_crawler`](https://github.com/alphagov/govuk_seed_crawler) is a script on the GOV.UK “mirrorer” machine. Every day at 8:00pm, this script fetches all of the URLs in the [GOV.UK sitemap](https://www.gov.uk/sitemap.xml) and adds them to a [RabbitMQ message queue](/manual/rabbitmq.html).
 
-The crawler is entirely independent of the mirrors. Stopping the crawler means no new updates are made
-to the mirrors, but it will not stop the mirrors from working.
+Then the [`govuk_crawler_worker`](https://github.com/alphagov/govuk_crawler_worker) on the mirrorer machine:
 
-To inspect the contents of the mirror:
+- consumes the fetched GOV.UK URLs from the RabbitMQ message queue
+- retrieves the GOV.UK HTML files returned by these URLs
+- saves these HTML files to a `/tmp` folder on the mirrorer machine
+- adds any new URLs found on those pages to the back of the RabbitMQ message queue
 
-```bash
+Every hour, the [`govuk_sync_mirror`](https://github.com/alphagov/govuk-puppet/blob/86d1480c6e081313c415246063d5931af24473da/modules/govuk_crawler/manifests/init.pp#L109) script runs on the mirrorer machine. This script copies the GOV.UK HTML files from the mirrorer machine to the main `govuk-<environment>-mirror` AWS S3 bucket. AWS then copies this main bucket to the replica `govuk-<environment>-mirror-replica` S3 bucket in another region.
+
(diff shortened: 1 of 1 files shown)


source/manual/fall-back-to-mirror.html.md
```diff
@@ -1,70 +1,75 @@
...
+If this is the first time the page has been requested, Fastly will:
 
-This is why we refer to switching off Nginx on the origin cache machines as
-"falling back to the mirrors".
+- forward the request to the GOV.UK servers, otherwise known as Origin
+- cache the response locally
+- return the response to the user
 
-## Hosting
+Subsequent requests for the same resource will never hit the origin server, as long as Fastly keeps the response in its cache. Instead, Fastly will return what it has in its cache. This means that Fastly absorbs a lot of traffic and reduces the load on Origin.
```
nicholsj, line 18: I think everything above here should be covered elsewhere (eg https://docs.publishing.service.gov.uk/manual/cdn.html). 

Otherwise there's a risk of duplication/diluting this page. It should only be about the mirror

jonathanglassman: ok, will remove and add link to that page

source/manual/fall-back-to-mirror.html.md
```diff
@@ -1,90 +1,84 @@
...
-response.
+GOV.UK uses the [Fastly content delivery network (CDN)](/manual/cdn.html) to handle most requests to [GOV.UK Origin](/manual/architecture-shallow-dive.html#a-user-visits-the-gov-uk-homepage). However, there are 2 scenarios where the user requests will not get to Fastly.
 
-This process is handled by our CDN config and is entirely transparent to us and
-our users. It happens multiple times a day, for lots of different reasons.
+First and most commonly, user requests to GOV.UK may time out or return an error response.
 
-This is why we refer to switching off Nginx on the origin cache machines as
-"falling back to the mirrors".
+Second, if Fastly goes down, user requests to Fastly will fail with a 5xx error code (the problem was caused by the server). GOV.UK servers will still be able to serve content to users, but user requests will not be able to get to those servers.
```
nicholsj, line 13: Small style thing, but should this be a numbered list (could help with scanning) and the sentence above ending with a `:`?

jonathanglassman: In terms of style, we use numbered steps to guide a user through a process. I agree with your point re scanning, so perhaps an unordered list would help in this situation i.e.

"However, user requests will not get to Fastly in the following scenarios:

- first and most commonly, user requests to GOV.UK may time out or return an error response
- second, if Fastly goes down, user requests to Fastly will fail with a 5xx error code (the problem was caused by the server)"

@nicholsj how does that sound?

source/manual/fall-back-to-mirror.html.md
```diff
@@ -1,90 +1,83 @@
...
+GOV.UK uses the [Fastly content delivery network (CDN)](/manual/cdn.html) to handle most requests to [GOV.UK Origin](/manual/architecture-shallow-dive.html#a-user-visits-the-gov-uk-homepage). However, user requests will not get to Fastly in the following scenarios:
 
-This process is handled by our CDN config and is entirely transparent to us and
-our users. It happens multiple times a day, for lots of different reasons.
+- first and most commonly, user requests to GOV.UK may time out or return an error response
+- second, if Fastly goes down, user requests to Fastly will fail with a 5xx error code (the problem was caused by the server)
 
-This is why we refer to switching off Nginx on the origin cache machines as
-"falling back to the mirrors".
+When user requests cannot get to Fastly, those requests will instead go to a static copy of GOV.UK known as the GOV.UK mirror.
```
ChrisBAshton, line 14: This isn't quite right. If Fastly goes down, no requests are ever going to reach any part of GOV.UK (i.e. neither the origin nor the mirror). We have to [manually fail over to AWS CloudFront](https://docs.publishing.service.gov.uk/manual/fall-back-to-aws-cloudfront.html) in such a scenario, so that user requests go to AWS rather than to Fastly. AWS will, in turn, make requests to the mirror.

source/manual/fall-back-to-mirror.html.md
```diff
@@ -1,90 +1,83 @@
...
 
+## “Mirror GOV.UK content to S3” alert
+
+If the govuk_sync_mirror cronjob has not succeeded for 24 hours, it triggers the “Mirror GOV.UK content to S3” alert. See the [Mirror GOV.UK content to S3 alert documentation](/manual/alerts/mirror-sync.html) for more information.
+
 ## Forcing failover to the static mirrors
 
-Because the CDN will retry every request against the mirrors automatically if origin is unavailable,
-[stopping Nginx on the cache machines with Fabric][fab-fail] will result in falling back to mirrors:
+Because Fastly will retry every request against the mirrors automatically if Origin is unavailable, [stopping Nginx on the cache machines with Fabric][fab-fail] will result in user requests going to the mirror. Run the following command to stop Nginx on the cache machines:
```
ChrisBAshton, line 62: We're no longer recommending Fabric as a tool, so this section will eventually need a small rewrite. That can happen after this PR though. @huwd is gradually working through the docs to remove references to Fabric.

source/manual/fall-back-to-mirror.html.md
```diff
@@ -1,90 +1,83 @@
...
+Run the following to inspect the contents of the GOV.UK mirrorer machine:
+
+```
 gds govuk connect -e production ssh mirrorer
 cd /mnt/crawler_worker/www.gov.uk
 ```
 
+## “Mirror GOV.UK content to S3” alert
+
+If the govuk_sync_mirror cronjob has not succeeded for 24 hours, it triggers the “Mirror GOV.UK content to S3” alert. See the [Mirror GOV.UK content to S3 alert documentation](/manual/alerts/mirror-sync.html) for more information.
```
ChrisBAshton, line 58:
```suggestion
If the `govuk_sync_mirror` cronjob has not succeeded for 24 hours, it triggers the “Mirror GOV.UK content to S3” alert. See the [Mirror GOV.UK content to S3 alert documentation](/manual/alerts/mirror-sync.html) for more information.
```

ChrisBAshton (commented):
Thanks @jonathanglassman - this PR looks good, just one inaccuracy in the introduction to fix:

source/manual/fall-back-to-mirror.html.md
jonathanglassman: Thanks @ChrisBAshton. How about the following:

"When user requests to GOV.UK time out or return an error response, these requests instead go to a static copy of GOV.UK known as the GOV.UK mirror.

If Fastly has gone down, we [manually fall back to AWS CloudFront](/manual/fall-back-to-aws-cloudfront.html), the backup CDN. AWS will then send user requests to the GOV.UK mirror."

source/manual/fall-back-to-mirror.html.md
jonathanglassman: Whoops! thank you, will add to next commit

source/manual/fall-back-to-mirror.html.md
jonathanglassman: Sounds good, will leave for now

source/manual/fall-back-to-mirror.html.md
```diff
@@ -1,125 +1,113 @@
...
-our users. It happens multiple times a day, for lots of different reasons.
+- first and most commonly, user requests to GOV.UK may time out or return an error response
+- second, if Fastly goes down, user requests to Fastly will fail with a 5xx error code (the problem was caused by the server)
 
-This is why we refer to switching off Nginx on the origin cache machines as
-"falling back to the mirrors".
+When user requests to GOV.UK time out or return an error response, these requests instead go to a static copy of GOV.UK known as the GOV.UK mirror.
 
-## Hosting
+If Fastly has gone down, we [manually fall back to AWS CloudFront](/manual/fall-back-to-aws-cloudfront.html), the backup CDN. AWS then sends user requests to the GOV.UK mirror.
```
ChrisBAshton, line 16: Ah, I was sort of expecting this section to be changed a bit in light of https://github.com/alphagov/govuk-developer-docs/pull/3399#discussion_r796362313. Apologies if that wasn't clear.

It took me a while to figure out the purpose of these introductory paragraphs. I think it could be made clearer by becoming its own section:

```suggestion
The GOV.UK mirror is a static copy of GOV.UK.

## When is the static mirror used?

Most requests to GOV.UK are handled by the [Fastly content delivery network (CDN)](/manual/cdn.html), which sits between the user and [GOV.UK Origin](/manual/architecture-shallow-dive.html#a-user-visits-the-gov-uk-homepage). When a user requests a GOV.UK page, Fastly retrieves it from its cache, or fetches it from GOV.UK Origin if it does not have it in its cache.

Sometimes, GOV.UK Origin may time out or return a 5xx error response. When that happens, Fastly automatically fetches the page from the GOV.UK mirror instead.

Finally, if Fastly goes down, we'd need to manually [switch to AWS CloudFront](https://docs.publishing.service.gov.uk/manual/fall-back-to-aws-cloudfront.html) instead of Fastly. Where Fastly makes requests to GOV.UK Origin, AWS CloudFront instead makes all its requests to the static mirror.
```

We refer to 'GOV.UK mirror' and 'static mirror' interchangeably throughout this document. It might be best to stick with one (I usually hear it referred to as the latter, but might be worth making a Slack poll to gather a consensus from the developers).

source/manual/fall-back-to-mirror.html.md
```diff
@@ -1,125 +1,113 @@
...
+## GOV.UK mirror locations and access
 
-- Amazon S3: the static mirror is hosted in a bucket `govuk-<environment>-mirror` and the
-  content is retrieved by the CDN using an API.  This bucket is replicated to
-  `govuk-<environment>-mirror-replica` in another AWS region.
+The GOV.UK mirror is a copy of GOV.UK hosted in an [Amazon Web Services (AWS) S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html). The bucket contains copies of GOV.UK HTML files. The mirror is static, meaning dynamic pages such as search pages will not work.
 
-- Google GCS: the static mirror is hosted in a bucket `govuk-<environment>-mirror` and the
-  content is retrieved by the CDN using an API.
+There are 3 mirrors:
```
ChrisBAshton, line 22: Might be worth calling this out:

```suggestion
Whilst "the static mirror" implies only one mirror, it collectively refers to 3 mirrors:
```

Not happy with my wording, but hopefully you see what I'm getting at!

source/manual/fall-back-to-mirror.html.md
```diff
@@ -1,125 +1,113 @@
...
 
+## `Mirror GOV.UK content to S3` alert
+
+If the `govuk_sync_mirror` cronjob has not succeeded for 24 hours, it triggers the ‘Mirror GOV.UK content to S3’ alert. See the [Mirror GOV.UK content to S3 alert documentation](/manual/alerts/mirror-sync.html) for more information.
+
 ## Forcing failover to the static mirrors
 
-Because the CDN will retry every request against the mirrors automatically if origin is unavailable,
-[stopping Nginx on the cache machines with Fabric][fab-fail] will result in falling back to mirrors:
+If Origin is unavailable, Fastly will automatically retry every request against the mirrors. This means stopping [Nginx](https://www.nginx.com/) on the cache machines will result in user requests going to the mirror. To stop Nginx on the cache machines, run the following command:
```
ChrisBAshton, line 64: Hmmm I've just noticed that there are two approaches to this in the dev docs - the one here, and the one described in https://docs.publishing.service.gov.uk/manual/fall-back-to-aws-cloudfront.html. I.e. switching to AWS CloudFront gives us the "failover to the static mirrors" behaviour, although it's a blunter tool than disabling Nginx on specific machines.

Given the fabric scripts are deprecated, switching to CloudFront is probably the preferred approach here. That said, that's only because AWS CloudFront currently only serves static content; there's nothing stopping us having AWS CloudFront making requests to GOV.UK Origin in future, for example. So this might only be a short term best practice:

```suggestion
If Origin is unavailable, Fastly will automatically retry every request against the mirrors. To avoid Fastly traffic hitting Origin while its down (potentially exacerbating its problems), we can [fall back to AWS CloudFront](/manual/fall-back-to-aws-cloudfront.html), which serves all content via the static mirrors.

Alternatively, we can stop [Nginx](https://www.nginx.com/) on the cache machines, which will prevent requests hitting GOV.UK applications. Fastly will automatically retry these failed requests against the mirror.

To stop Nginx on the cache machines, run the following command:
```

ChrisBAshton (commented):
Thanks @jonathanglassman. Apologies this isn't an approval yet ✅  - I've just spotted a couple of things that I think should be edited before merge.

ChrisBAshton (approved):
LGTM - please use the "Squash and merge" option when merging 👍

PR #3400, opened 2022-01-31
Add page about new alert

MuriloDalRi:
https://trello.com/c/SMuSi4bX/2841-add-documentation-about-new-alert

### DIFF
--- a/source/manual/alerts/mirror-sync.html.md
+++ b/source/manual/alerts/mirror-sync.html.md
@@ -67,3 +67,7 @@ COMMAND  PID          USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
 bash    5221 govuk-crawler    3wW  REG   0,19        0  348 /run/govuk_sync_mirror.lock
 s3cmd   5233 govuk-crawler    3w   REG   0,19        0  348 /run/govuk_sync_mirror.lock
 ```
+
+## The static mirror has not been updated in n seconds
+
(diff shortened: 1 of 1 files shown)


source/manual/alerts/mirror-sync.html.md
```diff
@@ -67,3 +67,7 @@ COMMAND  PID          USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
 bash    5221 govuk-crawler    3wW  REG   0,19        0  348 /run/govuk_sync_mirror.lock
 s3cmd   5233 govuk-crawler    3w   REG   0,19        0  348 /run/govuk_sync_mirror.lock
 ```
+
+## The static mirror has not been updated in n seconds
+
+This alert means that the S3 static mirror has not been updated for over 24h. The govuk_sync_mirror script [writes a file](https://github.com/alphagov/govuk-puppet/blob/main/modules/govuk_crawler/templates/govuk_sync_mirror.erb#L44) to the S3 mirror bucket every time it runs, this alert checks when that file was last modified. It indicates the script is not running as it should, otherwise it would either succeed or throw an error.
```
ChrisBAshton, line 73: I'd change the link to https://github.com/alphagov/govuk-puppet/blob/fffb8bbde66a0af7f18a75b837ae3090f95488e0/modules/govuk_crawler/templates/govuk_sync_mirror.erb#L44, so that future changes to the code don't change the line that this paragraph is pointing to.

(Retrieved by visiting your URL and using the `y` keyboard shortcut)

"otherwise it would either succeed or throw an error" seems a potentially confusing way of ending this, too, as if for example the script was throwing an error, this alert would still end up firing after 24 hours. So the presence of the alert doesn't mean the absence of an error. I think we can omit this last sentence altogether.

MuriloDalRi: That y shortcut just blew my mind!

I'll push an update now.

ChrisBAshton (approved):

PR #3402, opened 2022-02-01
Outline further ruby testing steps

ChrisBAshton:
Taking inspiration from https://reinteractive.com/posts/499-no-app-left-behind-upgrade-your-application-to-ruby-3-0-and-stay-on-the-upgrade-path

Trello: https://trello.com/c/ZMrLumn8/2837-upgrade-ruby-for-our-apps-5

### DIFF
--- a/source/manual/ruby.html.md
+++ b/source/manual/ruby.html.md
@@ -60,7 +60,13 @@ and then on the respective machine that your app will run on (e.g. `backend`).
 We use [upgrade-ruby-version][] to automatically raise pull requests in GOV.UK repositories which use Ruby.
 See an [example PR](https://github.com/alphagov/upgrade-ruby-version/pull/1) for how to specify the repos and versions.
 
-Once the PRs are raised, each app should have that branch built to Integration, where it should be tested to ensure that the new Ruby version hasn't caused any problems. Our monitoring tools, such as Grafana and Sentry, should be checked too. Once fully tested, the PR should be safe to merge.
+Once the PRs are raised, you should test the changes before merge, as follows:
+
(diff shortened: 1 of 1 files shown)


source/manual/ruby.html.md
```diff
@@ -60,7 +60,13 @@ and then on the respective machine that your app will run on (e.g. `backend`).
 We use [upgrade-ruby-version][] to automatically raise pull requests in GOV.UK repositories which use Ruby.
 See an [example PR](https://github.com/alphagov/upgrade-ruby-version/pull/1) for how to specify the repos and versions.
 
-Once the PRs are raised, each app should have that branch built to Integration, where it should be tested to ensure that the new Ruby version hasn't caused any problems. Our monitoring tools, such as Grafana and Sentry, should be checked too. Once fully tested, the PR should be safe to merge.
+Once the PRs are raised, they should be tested before merge, as follows:
+
+1. The app's automated CI build should run a test suite, which should be checked for any Ruby deprecation warnings. If there are any warnings, they should be fixed.
+1. Then, each app should have its ruby upgrade branch built to Integration, where its features should be manually tested to ensure that the new Ruby version hasn't caused any problems.
```
nicholsj, line 66: Would this be better as something like "then you should build..." - making it clear that this isn't automated like the previous item

nicholsj: Similarly, item 1 could be reworded something like ", and you should check.."?

ChrisBAshton: Yeah, I've been avoiding second person language for some reason, but I think it would benefit from being reworded 👍 will push up a change.

MuriloDalRi (approved):

PR #3405, opened 2022-02-01
Account docs cleanup

huwd:
[Trello](https://trello.com/b/TlvJ4j59/doing-accounts)

A bit of a clear up as we look to hand over some of our stuff.
- Increase visibility of the GOV.UK Personalisation Gem, folks can use that for future personalisation needs
- Remove a GOV.UK Account section, it was empty and probably should remain so.

See commit messages for further thinking.

## What it looks like
image

image

### DIFF
--- a/data/dashboard.yml
+++ b/data/dashboard.yml
@@ -57,6 +57,7 @@
         - govuk_document_types
         - govuk_message_queue_consumer
         - govuk_navigation_helpers
+        - govuk_personalisation
         - govuk_publishing_components
         - govuk_sidekiq
         - govuk_taxonomy_helpers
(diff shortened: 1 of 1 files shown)


data/dashboard.yml
```diff
@@ -1,7 +1,6 @@
 # These sections are magically populated from repos.yml
```
ChrisBAshton, line 1: I'm going to remove this misleading line! The sections do seem to be curated manually 😅

ChrisBAshton (approved):
Looks good - thanks!

data/dashboard.yml
ChrisBAshton: https://github.com/alphagov/govuk-developer-docs/pull/3407

PR #3408, opened 2022-02-02
Remove tech-docs-monitor

ChrisBAshton:
This is not a GOV.UK repo, so should not be imported as a GOV.UK
app in the GOV.UK docs.


kevindew (approved):
Thanks

PR #3411, opened 2022-02-03
Remove requirement to purge caches during emergency banner deploy

bilbof:
We have implemented the proposal in [RFC 144] and can now remove
the requirement to clear the caches.

Fastly and Varnish Caches will expire within 5 minutes, and memcached
will expire in 1 minute, so we can ask folks to wait a minute and
then do some cache bust testing.

Within 10 minutes the banner should be displayed on every page,
as tested by Ken this morning.

[RFC 144]: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-144-consistent-cache-expiry.md


sengi (approved):
🎉

PR #3414, opened 2022-02-03
Skip loading of proxy pages with an env var

richardTowers:
Often I find myself wanting to work on a few pages of the dev docs which
are local to this repository.

It's annoying having to wait for my laptop to download all of GitHub
before middleman will start serving the entirely local content I want to
edit.

Adding this flag gives me a way to run the middleman server and work on
these pages without needing to make requests to GitHub.

Unfortunately, middleman server will still download stuff that's
requested via .erb files when you go to the homepage / apps page etc. I
had a look at disabling all of this, but I'd need to do a big refactor
to make it clean so this small change was the best I could come up with.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,23 @@ You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-do
 
 ### Running the app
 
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
Thanks - this has bothered me too 😀

PR #3415, opened 2022-02-03
Update deploy licensing with jenkins docs

richardTowers:
The previous docs assumed a fair bit of GOV.UK knowledge, and also that
things (like the automatic deployment to integration) actually worked.

This made it very difficult for third parties to follow the docs, unless
they'd been shown all the idiosyncracies a few times.

I've updated the docs so they're much more explicit. This means making
more Jenkins URLs public, but these are already effectively in the
public domain since govuk-puppet and govuk-aws are already open source.

Our colleagues at Apto still don't have access to staging or production
Jenkins', but I'm looking into that separately.

### DIFF
--- a/source/manual/access-to-licensing-for-third-parties.html.md
+++ b/source/manual/access-to-licensing-for-third-parties.html.md
@@ -61,7 +61,7 @@ Usually, GOV.UK developers coordinate deployments through these Jenkins instance
 [the Release app](https://release.publishing.service.gov.uk/applications). This shows which releases are deployed to
 which environments, and has useful buttons to take you to the correct Jenkins instance.
 
-Access to the release app is controlled through GOV.UK Signon. If you don't have access, you can request it from the
+Access to the Release app is controlled through [GOV.UK Signon](https://github.com/alphagov/signon). If you don't have access, you can request it from the
 GOV.UK developers.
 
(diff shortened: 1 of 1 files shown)


source/manual/access-to-licensing-for-third-parties.html.md
```diff
@@ -46,20 +46,48 @@ Enter password: REDACTED
...
+
+Access to Jenkins is controlled through GitHub teams. Users in the "GOV.UK" team have full access to the CI and
+Integration Jenkins instances, and read only access to the Staging and Production Jenkins instances. Users in the "
+GOV.UK Production" team have full access in all environments.
+
+Usually, GOV.UK developers coordinate deployments through these Jenkins instances using
+[the Release app](https://release.publishing.service.gov.uk/applications). This shows which releases are deployed to
+which environments, and has useful buttons to take you to the correct Jenkins instance.
+
+Access to the release app is controlled through GOV.UK Signon. If you don't have access, you can request it from the
```
ChrisBAshton, line 64: Nitpick:

```suggestion
Access to the Release app is controlled through [GOV.UK Signon](https://github.com/alphagov/signon). If you don't have access, you can request it from the
```

ChrisBAshton (approved):
LGTM 👍

PR #3416, opened 2022-02-04
Update modify change history doc

1pretz1:
This commit:

* Adds both ways to update a change note in Publishing API. Previously,
only one method was documented and didn't actually include the last
step which is to send the document downstream.

* Updates Whitehall guidance to check change note as FuzzyMatch can
sometimes return the wrong note if searching by substring.

We could / should probably turn this documentation into rake tasks at
some point.

Trello:
https://trello.com/c/Xrz8kzbW/1043-documentation-update-for-modifying-a-change-note


alex9smith (approved):
Looks good to me, thanks!

PR #3420, opened 2022-02-08
Add note to restart apps when upgrading Ruby version

edwardkerry:
When upgrading Ruby versions apps need to be restarted following a deployment. This PR adds a note and some instructions on how to do this.


ChrisBAshton (approved):
⭐

PR #3421, opened 2022-02-08
Update retagging docs

beccapearce:


source/manual/government-changes.html.md.erb
```diff
@@ -141,14 +141,21 @@ order):
...
+```
+$ gds govuk connect scp-push --environment [integration|staging|integration] name-of-machine[:1|:2|:3] path/to/file.csv /tmp/
+```
+The rake task can then be run from the command line following [these instructions][running-rake-tasks-on-the-command-line].
+
 There is a similar [Rake task to change the organisations for Manuals][manuals-bulk-update-organisation].
 A data migration is required to change the organisations for (Mainstream) Publisher documents. This is an example PR: [Migrate Publisher docs][publisher-bulk-update-organisation-example]
 
-[whitehall-bulk-update-organisation]: https://github.com/alphagov/whitehall/blob/917085dca58fb58dbf65c7e226ad445cc1f27bdd/lib/tasks/data_hygiene.rake#L37
+[whitehall-bulk-update-organisation]: https://github.com/alphagov/whitehall/blob/main/lib/tasks/data_hygiene.rake#L16
```
ollietreend, line 157: Did you intend for this to change from being a 'permalink' to pointing at the `main` branch?

There's a risk that `data_hygiene.rake` gets updated in the future, and then this link to Line 16 no longer highlights to what you thought it would.

beccapearce: yep. Since the old link was pointing to commit if the task is deleted it will still remain linked in the docs It's happened in a few other docs.

If there a better way of linking to the rake task obviously the line number is not the best choice but trying to way what it will be linked to our current code base.

beccapearce: Could link to the file and give the task name?

ollietreend: That makes sense – I think this seems a fair approach to avoid the link becoming frozen in time!

ollietreend (approved):

PR #3443, opened 2022-03-04
Update documentation about rotating the puppet server encryption key

fredericfran-gds:
Update the documentation about rotation the puppet server encryption key to reflect that the GPG key and other relevant artefacts are no longer stored in the pass store but in the `puppet_servers_gpg` directory because these artefacts were not created by the pass tool.

 Ref:
1. [govuk-secrets change pr](https://github.com/alphagov/govuk-secrets/pull/1244)


sengi (approved):
Thanks!

PR #3445, opened 2022-03-04
Add docs for GOV.UK Account cross domain tracking

alex9smith:
This is a special case of cross domain tracking where we use the same
GA property on both domains, so the tracking setup is more complicated
to pass cookie consent and GA client IDs between domains.

---
image
---

[Trello](https://trello.com/c/sf1T9BNR/1274-check-that-the-cross-domain-google-analytics-implementation-is-covered-in-the-govuk-developer-docs)

### DIFF
--- a/source/manual/analytics.html.md
+++ b/source/manual/analytics.html.md
@@ -67,6 +67,18 @@ The `GOVUK.Modules` library will attempt to initialise and start a module for ev
 In the case of cross domain tracking this can be undesirable if more than one element contains the `cross-domain-tracking` data-module because the underlying `ga.js` library from Google will not accept multiple calls to register a linked tracker and will throw javascript errors.
 See [this static commit](https://github.com/alphagov/static/commit/c03c11a84f86deb83ed3b7a4d16ad2e6de3f1d95) for how we mitigate against this for multiple govspeak buttons with cross domain tracking enabled.
 
+### Cross domain tracking for the GOV.UK Account
+
+The GOV.UK Account shares the account home and management pages between `www.gov.uk` and `account.gov.uk`. We send tracking data to the main GOV.UK GA property from both domains and use cross domain tracking to see the whole user journey.
+
+User journeys between these domains involve both users clicking links or buttons and server side redirects, so the standard cross domain tracking won't suffice on its own. To handle these journeys we append a (`cookie_consent`) parameter recording if they have engaged with the cookie banner ([see ADR](https://github.com/alphagov/account-api/blob/77b36bf9959813ab3a15a85e7939fcf034adac5c/docs/adr/003-cookie-consent.md#cookie-consent)). If the user has consented to tracking, we also send the user's Google Analytics client ID (`_ga`). We listen for these when a user lands on a page. To achieve this, we have:
+
+1. [A JavaScript module in `govuk_publishing_components`](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js) that adds the tracking parameters to a link or form action. Any links or forms that may result in a user being redirected to sign in must have the `data-module="explicit-cross-domain-links"` attribute so this module can add the event listener.
+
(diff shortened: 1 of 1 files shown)


source/manual/analytics.html.md
```diff
@@ -67,6 +67,18 @@ The `GOVUK.Modules` library will attempt to initialise and start a module for ev
...
+
+The GOV.UK Account shares the account home and management pages between `www.gov.uk` and `account.gov.uk`. We send tracking data to the main GOV.UK GA property from both domains and use cross domain tracking to see the whole user journey.
+
+User journeys between these domains involve both users clicking links or buttons and server side redirects, so the standard cross domain tracking won't suffice on its own. To handle these journeys we append parameters to the URL that send the user's Google Analytics client ID (`_ga`) and if they've accepted additional cookies (`cookie_consent`) and listen for these parameters when the user lands on a page. To achieve this, we have:
+
+1. [A JavaScript module in `govuk_publishing_components`](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js) that adds the tracking parameters to a link or form action. This must be added to forms, links or buttons that may result in a user being redirected to sign in by using the `"explicit-cross-domain-links"` data attribute so the parameters are available to any server side redirects.
```
danacotoran, line 76: >  using the `"explicit-cross-domain-links"` data attribute

I'm not sure this bit is 100% clear. To be more specific it would be the `data-module` attribute with a value of "explicit-cross-domain-links" (i.e `data-module="explicit-cross-domain-links"`) but I'm no wordsmith so I'm not sure how you'd explain that in a non-clunky way.

alex9smith: True, and actually the first half of that sentence isn't completely correct either - we're not adding the module to a link or form, we need to indicate that the module should be run on a link or a form. How about:

> Any links, forms or buttons that may result in a user being redirected to sign in must have the `data-module="explicit-cross-domain-links"` attribute so this module can add the event listener.

danacotoran: Yep that sounds much clearer @alex9smith!
Only thing to note is that it's only links and forms that can have the attribute, it won't do anything on a button unless it's a sneaky link disguised as a button.

alex9smith: Thanks! I'll change it to just links and forms

source/manual/analytics.html.md
```diff
@@ -67,6 +67,18 @@ The `GOVUK.Modules` library will attempt to initialise and start a module for ev
...
 See [this static commit](https://github.com/alphagov/static/commit/c03c11a84f86deb83ed3b7a4d16ad2e6de3f1d95) for how we mitigate against this for multiple govspeak buttons with cross domain tracking enabled.
 
+### Cross domain tracking for the GOV.UK Account
+
+The GOV.UK Account shares the account home and management pages between `www.gov.uk` and `account.gov.uk`. We send tracking data to the main GOV.UK GA property from both domains and use cross domain tracking to see the whole user journey.
+
+User journeys between these domains involve both users clicking links or buttons and server side redirects, so the standard cross domain tracking won't suffice on its own. To handle these journeys we append parameters to the URL that send the user's Google Analytics client ID (`_ga`) and if they've accepted additional cookies (`cookie_consent`) and listen for these parameters when the user lands on a page. To achieve this, we have:
```
huwd, line 74:
```suggestion
User journeys between these domains involve both users clicking links or buttons and server side redirects, so the standard cross domain tracking won't suffice on its own. To handle these journeys we append a (`cookie_consent`) parameter recording if they have engaged with the cookie banner ([see ADR](https://github.com/alphagov/account-api/blob/77b36bf9959813ab3a15a85e7939fcf034adac5c/docs/adr/003-cookie-consent.md#cookie-consent)). If the user has consented to tracking, we also send the user's Google Analytics client ID (`_ga`). We listen for these when a user lands on a page. To achieve this, we have:
```

huwd (approved):

danacotoran (approved):

PR #3447, opened 2022-03-04
Replace restart harvesting instructions referencing fab

huwd:

Fabric has been deprecated, we are replacing refrences with commands
that can be run directly on each box

image


ChrisBAshton (approved):
LGTM 👍 might be worth linking to https://github.com/alphagov/fabric-scripts/blob/master/ckan.py from the commit message.

PR #3448, opened 2022-03-04
Replace references to fabric in purge cache docs

huwd:

We have deprecated fab and are removing references to it from all our
dev docs.

image


benthorner (approved):
Confirmed this matches the current fabric script behaviour ✅.

PR #3454, opened 2022-03-08
Shipping auth logs to Cyber's Splunk

sihugh:
We have enabled log shipping to Splunk. These are some basic docs, which I expect to be expanded when we find useful things that can be done with them!

https://trello.com/c/3PEIdbKp/2861-ship-ssh-logs-from-jumpboxes-to-splunk-5


ChrisBAshton (approved):
LGTM 👍  A general point: the rest of the file doesn't use hard wrapping for sentences, so I'd suggest removing the hard newlines from your additions.

sihugh:
> LGTM 👍 A general point: the rest of the file doesn't use hard wrapping for sentences, so I'd suggest removing the hard newlines from your additions.

Good point, fixed - thanks!

PR #3457, opened 2022-03-10
Add doc on single page notification exemptions

danacotoran:
The GOVUK Accounts team have just rolled out single page notifications to 157k+ publications on GOVUK but we will not be the owners of this feature in the future. 

This doc is here to document what should be done in the event that there is a request to add or remove a publication from the single page notification button exemption list.

https://trello.com/c/XdtP8gTH


alex9smith (approved):
📖 👍🏻

PR #3458, opened 2022-03-11
Update hiera data to handle network is unreachable

kentsanggds:
## What

Fix gpg when you get the error - `Network is unreachable`

### DIFF
--- a/source/manual/encrypted-hiera-data.html.md
+++ b/source/manual/encrypted-hiera-data.html.md
@@ -441,6 +441,19 @@ to fail when using GPG version 1 with lots of credentials.
 If you see this error:
 
 ```
+keyserver receive failed: Network is unreachable
+```
+
+Run this command to add `standard-resolver` to `dirmngr.conf`
+
+```sh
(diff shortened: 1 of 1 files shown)


source/manual/encrypted-hiera-data.html.md
```diff
@@ -440,6 +440,18 @@ to fail when using GPG version 1 with lots of credentials.
...
 If you see this error:
 
+```
+keyserver receive failed: Network is unreachable
+```
+
+Run this command to add `standard-resolver` to `dirmngr.conf`
+
+```sh
+echo "standard-resolver" >> ~/.gnupg/dirmngr.conf
```
benjamineskola, line 450: It's possible that this settings change won't take effect because there may be a background process that doesn't reload its configuration. Config can be reloaded as follows:

```suggestion
echo "standard-resolver" >> ~/.gnupg/dirmngr.conf
gpgconf --kill all
```

kentsanggds: 👍

benjamineskola (approved):

PR #3460, opened 2022-03-14
Remove mentions of manuals-frontend

KludgeKML:
## WHAT
Remove reference to manuals-frontend, update docs for apps that use it to mention government-frontend instead.

## WHY
We'd like to migrate the rendering so we can retire manuals-frontend as on trello ticket.

### Trello
https://trello.com/c/CfH8SX6D/1130-manuals-update-developer-documents

## RELATED
https://github.com/alphagov/hmrc-manuals-api/pull/664
https://github.com/alphagov/manuals-publisher/pull/1844

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -829,9 +829,13 @@
   type: Data science
 
 - repo_name: manuals-frontend
+  retired: true
   type: Frontend apps
   team: "#find-and-view-tech"
   production_hosted_on: aws
+  description: |
+    Used to host manuals from manuals-publisher and hmrc-manuals-api
+    (now hosted by government-frontend)
 
 - repo_name: manuals-publisher
   type: Publishing apps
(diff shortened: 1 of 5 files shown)


data/repos.yml
```diff
@@ -481,11 +481,6 @@
   dashboard_url: false
   deploy_url: false
 
-- github_repo_name: manuals-frontend
```
edwardkerry, line 484: I think that, rather than removing it from this list, we add a 
`retired: true` and add a description about what it was and why it's gone
e.g https://github.com/alphagov/govuk-developer-docs/blob/c3f24e2cc7100b1d4b13c29da3cdee1548bf2d15/data/repos.yml#L59
then there is still a record and a link to the repo in the dev docs

KludgeKML: Thanks! Fixed.

PR #3461, opened 2022-03-14
Removes mention of RE GOV.UK team, and clarifies 2nd Line numbers

nicholsj:
A few changes to the 2nd Line welcome docs to:
-clarify that there won't *always* be a shadow
-remove mention of RE:GOV.UK team which no longer exists


richardTowers (approved):
Nice one :)

PR #3462, opened 2022-03-14
Update conventions for Jasmine JS testing

ollietreend:
This commit updates the Rails conventions manual to reflect the new approach we've taken for running Jasmine tests.

This change has come about because the `jasmine` gem has been deprecated, and it doesn't support Rails 7. We've therefore had to introduce this new approach as an enabler for our applications to run on Rails 7.

Ticket: https://trello.com/c/21VMx1NV/264-migrate-apps-to-use-jasmine-browser-runner


kevindew (approved):
Thank you ⭐

PR #3481, opened 2022-03-29
Allow all repos to be access via JSON and CSV

kevindew:
This is to create a https://docs.publishing.service.gov.uk/repos.json and https://docs.publishing.service.gov.uk/repos.csv file which contain more than just the apps which run (as is available in /apps.(json|csv)).

The motivation behind this is to adapt govuk-dependencies to consume the repos.json [instead of apps.json](https://github.com/alphagov/govuk-dependencies/blob/0f35d6003f0346d0aa7c14ad1e94f7a74b1587e8/lib/gateways/team.rb#L7), so that dependabot alerts can be made for more projects than just running applications.


ChrisBAshton (approved):
Nice one! ⭐

PR #3492, opened 2022-04-12
Clarify policy on non-GOV.UK Docker local development usage

ChrisBAshton:
This is an action arising from the Developer Survey Winter 2021.

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -128,6 +128,8 @@ We use a `govuk-docker` Docker environment for local development.
 
 To set up GOV.UK Docker, see the [installation instructions in the `govuk-docker` GitHub repo][govuk-docker].
 
+You can also try developing outside of Docker, using tools like [rbenv](https://github.com/rbenv/rbenv) directly. [This approach generally works for frontend apps](https://docs.publishing.service.gov.uk/manual/local-frontend-development.html) but not for other apps with databases, etc., and is not officially supported.
+
 [govuk-docker]: https://github.com/alphagov/govuk-docker/blob/master/README.md
 
 ## 6. Get SSH access to integration


source/manual/get-started.html.md
```diff
@@ -128,6 +128,8 @@ We use a `govuk-docker` Docker environment for local development.
 
 To set up GOV.UK Docker, see the [installation instructions in the `govuk-docker` GitHub repo][govuk-docker].
 
+You are allowed to use `rbenv`/`bundle` instead of GOV.UK Docker, but this won't work well with apps that are dependent on other apps, and is not officially supported.
```
benthorner, line 131:
```suggestion
You can also try developing outside of Docker, using tools like [rbenv](https://github.com/rbenv/rbenv) directly. [This approach generally works for frontend apps](https://docs.publishing.service.gov.uk/manual/local-frontend-development.html) but not for other apps with databases, etc., and is not officially supported.
```

Relaxing the language slightly (anyone is allowed to do anything really).

I also want to try and emphasise that (GOV.UK) Docker is just a wrapper around tools people may be familiar with already - not an alternative to them.

My impression is it's mostly frontend apps where non-Docker dev is popular. An experienced backend dev probably doesn't need this guidance to know the particular circumstances when they can avoid it, so we may as well make the caveat specific to frontend apps.

benthorner, line 131:
```suggestion
You can also try developing outside of Docker, using tools like [rbenv](https://github.com/rbenv/rbenv) directly. [This approach generally works for frontend apps](https://docs.publishing.service.gov.uk/manual/local-frontend-development.html) but not for other apps with databases, etc., and is not officially supported.
```

Relaxing the language slightly (anyone is allowed to do anything really).

I also want to try and emphasise that (GOV.UK) Docker is just a wrapper around tools people may be familiar with already - not an alternative to them.

My impression is it's mostly frontend apps where non-Docker dev is popular. An experienced backend dev probably doesn't need this guidance to know the particular circumstances when they can avoid it, so we may as well make the caveat specific to frontend apps.

MuriloDalRi (approved):

PR #3494, opened 2022-04-14
Expand on docker_management reboot reasons

sihugh:
I was really tempted to just reboot this machine, and needed a bit more context about the warning.

The docker_management machine runs etcd which is used to provide a lock for other machines that wish to perform an unattended reboot. For that reason, we need to be a bit careful about rebooting it.

The etcd host is specified by the [etcd_endpoints](https://github.com/alphagov/govuk-puppet/blob/2a1b4084d8692707a8af72f4ed0f59f194705c86/hieradata_aws/common.yaml#L1532) configuration.

There is further context available in:
- [puppet-unattended_reboot Github repo](https://github.com/alphagov/puppet-unattended_reboot)
- [a presentation by Matt Bostock (@mattbostock)](https://www.dotconferences.com/2015/06/matt-bostock-coordinating-unattended-reboots-using-a-distributed-mutex)

### DIFF
--- a/source/manual/alerts/rebooting-machines.html.md
+++ b/source/manual/alerts/rebooting-machines.html.md
@@ -243,7 +243,9 @@ sudo reboot
 
 ### Rebooting docker-management
 
-It is safe to reboot while no other unattended reboot is underway:
+It is only safe to reboot while no other unattended reboot is underway. This is because it is used to manage locks for unattended reboots of other machines. If this machine is down, then multiple machines in high availability groups may choose to reboot themselves at the same time.
+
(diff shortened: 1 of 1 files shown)


deborahchua (approved):

mattbostock:
Hopefully a good thing :-) While I'm in the peanut gallery here, have you
ruled out making the automated reboots fail closed, i.e. refuse to reboot
if etcd is unreachable?

sihugh:
Errr, probably not (though I'm not sure) - [we're moving off VMs to EKS](https://technology.blog.gov.uk/2021/07/23/automatically-creating-and-rotating-gov-uk-application-secrets/) though, so it's likely that we'll be swapping this particular problem for another set

PR #3497, opened 2022-04-22
Remove deprecated repos and sections

ChrisBAshton:
We linked to a number of Archived repositories, which doesn't seem very useful.
We also had an empty section for 'Specialist apps', which we should just remove.

### DIFF
--- a/data/dashboard.yml
+++ b/data/dashboard.yml
@@ -7,7 +7,6 @@
     - name: Frontend apps
     - name: data.gov.uk apps
     - name: Licensing apps
-    - name: Specialist apps
 
 - name: Community
   sections:
@@ -55,7 +54,6 @@
         - govspeak
         - govuk_document_types
         - govuk_message_queue_consumer
-        - govuk_navigation_helpers
         - govuk_personalisation
         - govuk_publishing_components
         - govuk_sidekiq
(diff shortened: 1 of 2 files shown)


MuriloDalRi (approved):

data/dashboard.yml
```diff
@@ -68,7 +66,6 @@
         - govuk-rfcs
         - rubocop-govuk
         - publishing-e2e-tests
-        - styleguides
```
AgaDufrat, line 71: Would it make sense to link to GDS Way in the intro (which is where this now lives)?

data/dashboard.yml
```diff
@@ -90,12 +87,10 @@
     - name: Configuration
       repos:
         - govuk-puppet
-        - fabric-scripts
```
AgaDufrat, line 93: Should we remove all the other links from the docs as a part of this PR? There are few manuals that are still linking to the fabric scripts (they'll probably need updating not just removing the links) and styleguides https://github.com/alphagov/govuk-developer-docs/blob/main/source/manual/readmes.html.md?plain=1#L79

ChrisBAshton: Definitely - that's a larger piece of work though, captured by https://trello.com/c/AGdifoDb/250-remove-fabric-and-replace-references-in-docs. It's actually looking for a volunteer, if you'd like the job!

data/dashboard.yml
ChrisBAshton: Good shout. I've added on a commit.

PR #3508, opened 2022-04-29
Document GOV.UK mirrors responses for search and non-existent pages

AgaDufrat:
It's worth mentioning how those pages are handled by the mirrors in the docs as
it might not be expected to get a 403 response for an originally 404 page. 

This behaviour is currently documented in [Smokey tests](https://github.com/alphagov/smokey/blob/aaa0522666d49db116b939276f0e0969357e3ed4/features/mirror.feature#L23-L30) which were identified as
redundant in the [Smokey audit](https://docs.google.com/spreadsheets/d/1i2k4WexywoJurdFddOxiNICN-Elv1D5Mb2sdaCjF1t0/edit#gid=0&fvid=1582574436) and will be removed.

### DIFF
--- a/source/manual/fall-back-to-mirror.html.md
+++ b/source/manual/fall-back-to-mirror.html.md
@@ -20,7 +20,7 @@ If Fastly goes down, we would manually [switch to AWS CloudFront](https://docs.p
 
 ## GOV.UK mirror locations and access
 
-The GOV.UK mirror is hosted in an [Amazon Web Services (AWS) S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html). The bucket contains copies of GOV.UK HTML files. The mirror is static, meaning dynamic pages such as search pages will not work.
+The GOV.UK mirror is hosted in an [Amazon Web Services (AWS) S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html). The bucket contains copies of GOV.UK HTML files. The mirror is static, meaning dynamic pages such as search pages will not work. S3 will return 403 Forbidden response for non-existent pages instead of 404, because we don't allow the ListBucket permission. This also affects requests for features such as Search, which are not covered by the mirrors.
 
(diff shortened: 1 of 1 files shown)


source/manual/fall-back-to-mirror.html.md
```diff
@@ -20,7 +20,7 @@ If Fastly goes down, we would manually [switch to AWS CloudFront](https://docs.p
 
 ## GOV.UK mirror locations and access
 
-The GOV.UK mirror is hosted in an [Amazon Web Services (AWS) S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html). The bucket contains copies of GOV.UK HTML files. The mirror is static, meaning dynamic pages such as search pages will not work.
+The GOV.UK mirror is hosted in an [Amazon Web Services (AWS) S3 bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html). The bucket contains copies of GOV.UK HTML files. The mirror is static, meaning dynamic pages such as search pages will not work. Search and non-existent pages return 403 Forbidden response.
```
sengi, line 23: This is a bit ambiguous and potentially misleading - it's unclear to the reader whether this is saying that the 403 is intended behaviour from the end-user perspective when using the mirrors (it's not) or whether it's talking about just the S3->Fastly bit (i.e. an implementation detail).

If we're not already changing these 403s to 404s in Fastly when serving from mirrors, then that sounds like something we should fix rather than document.

sengi: I'd recommend clarifying that the 403 is an S3 thing, for example something like: `S3 will return `403 Forbidden` for non-existent pages instead of 404, because we don't allow ListBucket permission. This also affects requests for features such as Search, which are not covered by the mirrors.`

AgaDufrat: Thanks, that's a very good point! I'll change it to your description 🌟

sengi (approved):

PR #3372, opened 2022-01-11
Refactor Applications, ApplicationsByTeam, AppData and Hosts

ChrisBAshton:
See commits for details. 

Consolidating these classes will make it easier to implement repository standardisation in #3369.

Trello: https://trello.com/c/oUsgzZnk/247-standardise-how-we-represent-repositories-in-the-dev-docs

PR #3509, opened 2022-05-02
Remove final erroneous references to 'apps' in favour of 'repos'

ChrisBAshton:
In #3372 and #3369, we did the work to allow importing all GOV.UK repositories into the Developer Docs. This meant changing a lot of references from 'App' to 'Repo', to genericise the templates. However, a lot of references to 'App' remained, with a view to tidying things up properly later. This PR is that attempt at tidying-up!

These changes are purely non-visual, but they did unlock some improvements by way of PRs 3510 and 3511. This should be reviewed/merged in isolation first though.


app/repos.rb
```diff
@@ -19,20 +19,4 @@ def self.active
...
-    Repos.grouped_by_team.keys.reject { |team| team == UNKNOWN }
-  end
-
-  def self.for_team(team)
-    Repos.grouped_by_team.fetch(team, []).map(&:app_name)
-  end
-
-  def self.no_known_owner
-    Repos.for_team(UNKNOWN)
-  end
```
ChrisBAshton, line 37: Note that I spent some time trying to re-create ActiveRecord style 'scope' queries for these, before concluding that actually many of these methods can just be removed.

Here is the commit to the 'scope' spike work, for future reference: https://github.com/alphagov/govuk-developer-docs/commit/1a19c5bc1f60a698b383ebcc6736fc3a8abb6503

ChrisBAshton: Note to self: split this PR into maybe 3:

1 for the general refactoring
1 for the Apps/Repos pages changes
1 for the Repo overview page changes

app/related_things.rb
```diff
@@ -7,8 +7,8 @@ def initialize(manual, current_page)
   end
 
   def related_applications
```
edwardkerry, line 9: Do we want to rename this method to related_repos also? It can be used to refer to gems e.g
[govuk_publishing_components](https://github.com/alphagov/govuk-developer-docs/blob/d08e38bf969fba6875c95c9a0585e4469632b13f/source/manual/how-to-remove-jquery-from-javascript.html.md?plain=1#L8)

ChrisBAshton: Good spot 👍  I've pushed up a commit.

edwardkerry (approved):
Looks good to me, always appreciate your clean and clear commit history!

PR #3512, opened 2022-05-03
Replace cache-clearing service in suggested purging process

kevindew:
Trello: https://trello.com/c/TX9wLihh/101-cache-clearing-app-only-clears-path-not-query-strings

This updates the guidance on purging a page from the cache to be less
focused on the cache-clearing service and to reflect current GOV.UK
reality (where most pages are cached for 5 mins)

It replaces the instructions on how to clear the cache to use the
Jenkins task that is provided for this (which was updated in
https://github.com/alphagov/govuk-puppet/pull/11663 for this purpose)

This has been done in order to reduce dependencies on the cache-clearing
service to prepare for it's potential retirement.

### DIFF
--- a/source/manual/purge-cache.html.md
+++ b/source/manual/purge-cache.html.md
@@ -8,38 +8,49 @@ important: true
 ---
 
 The `www.gov.uk` domain is served through Fastly, which honours the
-cache control headers sent by Varnish. When new content is published, the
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
LGTM 👍

PR #3538, opened 2022-05-23
Remove redundant whitehall healthcheck doc

benthorner:
https://trello.com/c/fRihmgQ8/266-add-top-level-guidance-about-where-we-write-tests-on-govuk

This is no longer linked to from alerts [^1].

[^1]: https://github.com/alphagov/govuk-puppet/pull/11688


KludgeKML (approved):
Looks good.

PR #3541, opened 2022-05-23
Update the 5xx docs to point to ARP cache docs

leenagupte:
Update the 5xx docs to point to potential areas to check. One of these is flushing the
ARP cache.


source/manual/alerts/high-nginx-5xx-rate.html.md
```diff
@@ -19,6 +19,13 @@ such as Whitehall can have spikes - if you can determine this is a spike
 it is best to acknowledge the alert and let a team that is working on the app
 know (or alert Platform Reliability).
 
+## Multiple applications reportng errors
```
MuriloDalRi, line 22: Missing an i in 'reporting'

leenagupte: Oops! Thanks!

PR #3542, opened 2022-05-23
Add how to run commands on many machines

sihugh:
The lack of instructions here was mentioned in an incident review. It's rarely needed, but is sometimes useful.

Better ways undoubtedly exist!

### DIFF
--- a/source/manual/howto-run-ssh-commands-on-many-machines.html.md
+++ b/source/manual/howto-run-ssh-commands-on-many-machines.html.md
@@ -1,0 +1,130 @@
+---
+owner_slack: "#govuk-2ndline-tech"
+title: Run commands on many machines
+section: AWS
+layout: manual_layout
+parent: "/manual.html"
+---
+We used to rely heavily on Fabric scripts to help us do things like running commands on all machines in a class. This approach has been deprecated, but we still need to do those things occasionally.
+
+> **WARNING** Please test your command before unleashing it on production machines.
+
+Imagine you need to enable puppet on all cache machines. Here's one way to do that:
+
+```
+for host in $(gds govuk connect ssh -e integration jumpbox -- govuk_node_list -c cache)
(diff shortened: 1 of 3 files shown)


source/manual/howto-run-ssh-commands-on-many-machines.html.md
```diff
@@ -0,0 +1,130 @@
...
+
+### SSH to a host and run a command:
+
+```
+gds govuk connect ssh -e integration $host -- govuk_puppet --enable
+```
+
+This connects to the current `host` specified in the for loop, via the `integration` jumpbox.
+
+The `--` separator is used to tell govuk-connect to send all the arguments afterwards to `ssh`. The command to run on the remote host is provided after that.
```
ChrisBAshton, line 54: I always forget this is a thing - super useful!

source/manual/howto-run-ssh-commands-on-many-machines.html.md
```diff
@@ -0,0 +1,130 @@
...
+### Disable Puppet
+
+```
+for host in $(gds govuk connect ssh -e integration jumpbox -- govuk_node_list -c cache)
+do
+   gds govuk connect ssh -e integration $host -- govuk_puppet --disable \'Han Solo testing a new power inverter\'
+done
+```
+
+Note the backslashes escaping the quotes in the puppet command.
```
ChrisBAshton, line 91: I always use double quotes for the `--disable` command, have never had to escape it. Any reason why we can't do that here? Feels like it would simplify the instructions a bit.

ChrisBAshton (approved):
LGTM ⭐

leenagupte (approved):
Looks good to me! 💯  I love the link from the ARP cache page to this new one. It'd probably be handy to add something similar to the puppet pages later on.

PR #3555, opened 2022-05-30
Deprecate arp cache flushing instructions

sihugh:
Since [upgrading the kernels on our machines](https://github.com/alphagov/govuk-puppet/pull/11696) we don't expect this to be required any more. We'd like to know if this isn't the case.

### DIFF
--- a/source/manual/flush-the-arp-cache.html.md
+++ b/source/manual/flush-the-arp-cache.html.md
@@ -6,7 +6,7 @@ layout: manual_layout
 section: 2nd line
 ---
 
-> These instructions should no longer be required. Please alert #govuk-platform-reliability if you have to use them.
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
Agree it's sensible to keep these instructions for now 👍

source/manual/flush-the-arp-cache.html.md
```diff
@@ -6,6 +6,8 @@ layout: manual_layout
 section: 2nd line
 ---
 
+> These instructions should no longer be required. Please alert #govuk-platform-reliability if you have to use them.
```
nicholsj, line 9:
```suggestion
> These instructions should no longer be required. Please alert #govuk-platform-reliability-team if you have to use them.
```

nicholsj (commented):
One small change needed to Slack channel, then looks good

PR #3561, opened 2022-06-06
Ignore GitHub API token file (as per the README)

benthorner:
I've left the precise name of the file flexible. I doubt we will
have any other top-level files ending in "...token.txt".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@
 Staticfile.auth
 coverage
 tmp
+
+# Ignore GitHub API token
+*token.txt


.gitignore
```diff
@@ -20,3 +20,6 @@
 Staticfile.auth
 coverage
 tmp
+
+# Ignore GitHub API token
+*token.txt
```
ChrisBAshton, line 25: The README actually seems to have two ways of specifying the GitHub token:

1) store the token in a `.env` file
2) `cat` the token from `~/github_token.txt` (i.e. a text file in the home directory)

That's not to say that the developer can't choose to put the token in a text file in the root of the cloned govuk-developer-docs repository, but this change does feel a bit arbitrary/anti-[YAGNI](https://martinfowler.com/bliki/Yagni.html) (especially as it uses a wildcard).

I'm wondering what the benefit for this approach is over 1 or 2?

benthorner: Oh yes good point - agree (1) is better and there's no benefit to (2), which is unconventional among our apps. Anyone can still take that approach if they want, but we don't need them both here.

I've replaced this PR with https://github.com/alphagov/govuk-developer-docs/pull/3565, where I've also adjusted that whole section to better match the README structure in other repos, as well as making it more Docker/non-Docker cross-compatible.

PR #3562, opened 2022-06-06
Tidy up contents section of the home page

benthorner:
It was unclear what the itention was for this page: the headings
seem arbitrary and ones like "Transition" don't make sense without
(a lot) more context - we have a manual for it [^1].

This removes all links to individual repos as these are now provided
more robustly by the "Repos" and "Apps" pages. I've also removed the
link to Sentry as we have a manual about it [^2].

I'm not sure exactly what should be on this page, but I suspect it's
most useful to someone who's new to GOV.UK and never seen these docs
before, so let's start tailoring it to that use case.

## Screenshots

| Before | After |
| - | - |
| screencapture-docs-publishing-service-gov-uk-2022-06-06-10_13_59 | screencapture-localhost-4567-2022-06-06-10_14_19 |

[^1]: https://docs.publishing.service.gov.uk/manual/transition-architecture.html
[^2]: https://docs.publishing.service.gov.uk/manual/sentry.html

### DIFF
--- a/data/dashboard.yml
+++ b/data/dashboard.yml
@@ -1,15 +1,11 @@
-- name: "Applications"
+- name: Joining GOV.UK
   sections:
-    - name: Publishing apps
-    - name: APIs
-    - name: Services
-    - name: Supporting apps
-    - name: Frontend apps
-    - name: data.gov.uk apps
-    - name: Licensing apps
-
-- name: Community
-  sections:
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
Love this 🎉 completely agree the homepage needed a review/cull.

PR #3563, opened 2022-06-06
Add guide to viewing publisher activity for a CKAN dataset

nicholsj:
This adds information and an example for viewing a log of publisher activity on a specific dataset in CKAN. It can be useful when troubleshooting. As far as I know, the activity log isn't available through any links in CKAN.


hannako (approved):
Very useful!

PR #3565, opened 2022-06-06
Clarify approach to storing GitHub tokens

benthorner:
In response to: https://github.com/alphagov/govuk-developer-docs/pull/3561#discussion_r890023456

I've also removed the part about avoiding certain pages, as these
are unlikely to generate rate limit errors on their own - I don't
see any issues using them locally.


1pretz1 (approved):
Looks good 👍

PR #3566, opened 2022-06-07
Change GitHub access process to emailing owners

kevindew:
We've found that people posting in #govuk-tech-leads to ask for GitHub access isn't particularly helpful as we often get a message from someone asking to be added and don't know who they are. 

This change makes the process instead the same as the one used for a content designer joining where we request an email sent to a list of people who can access the request (https://docs.publishing.service.gov.uk/manual/give-a-content-designer-access-to-github.html)

I was undecided on whether we should suggest the new starter emails or whether their tech lead should do it. It seemed slightly simpler to ask the new starter.


sihugh (approved):
A decent improvement 👍

PR #3570, opened 2022-06-09
Update dependencies owner for collections

leenagupte:
Changes the dependency owner to match the application owner. The coronavirus team doesn't exist anymore so the dependencies are not being monitored.


leenagupte:
> if the `dependencies_team` is the same as the `team`, you don't need to specify it. It can just be removed.

@nicholsj I didn't know that, thanks! I've also updated the entry for local-links-manager and fixed up.

nicholsj (approved):

PR #3574, opened 2022-06-16
Update steps for changing Whitehall change note

brucebolt:
In https://github.com/alphagov/whitehall/pull/6618, we implemented a rake task to modify change notes in Whitehall.

This updates the documentation to reflect the usage of this task, instead of logging into a production console.


beccapearce (approved):
Looks great!

PR #3589, opened 2022-06-28
Update command example to clarify doesn't need square brackets

edwardkerry:
This command example was a little confusing, hopefully this clarifies that it needs just the node IP address and not wrapped in square brackets


gclssvglx (approved):
Much clearer - thanks @edwardkerry 👍

PR #3594, opened 2022-07-05
Add instructions on setting Athena query location

beccapearce:


source/manual/query-cdn-logs.html.md
```diff
@@ -38,6 +38,12 @@ integration and staging). To access,
 [Athena](https://eu-west-1.console.aws.amazon.com/athena) and select the
 `fastly_logs` database.
 
+You need to set a query result location before running queries in Athena. To do
+this go to the settings tab on the Athena query editor page and click manage.
+Then choose a location to save the results to; this will be a path starting with
+`s3://govuk-athena-results/`, you can use the s3 browser to find where in the
+`govuk-athena-results` folder you would like to save your results.
```
ChrisBAshton, line 45: I think you can simplify this to just "use `s3://govuk-athena-results/` as the value".
I don't think it needs to be categorised into subfolders - doubt anyone will ever go into the bucket and look!

PR #3600, opened 2022-07-12
Add docs on rate-limiting

KludgeKML:
Rate-limiting page with information about infrastructure level and app-level rates currently in place.


AgaDufrat (approved):
Nice addition, thank you!
Would it be worth adding a link to your Tech fortnightly presentation (after restricting view permissions to our organisation, since it may have some sensitive details)?

KludgeKML:
> Nice addition, thank you! Would it be worth adding a link to your Tech fortnightly presentation (after restricting view permissions to our organisation, since it may have some sensitive details)?

Ah yes, good idea. Have added, and restricted permissions.

PR #3622, opened 2022-07-21
Add shared workflow for publishing gems

kevindew:
We now prefer using a shared workflow for publishing rubygems, this updates this documentation to reflect that and explain how a repository gains a secret.


ChrisBAshton (approved):
Thanks Kevin

PR #3632, opened 2022-07-29
Clarify the internals of the Licensify/UKCloud alert.

sengi:
Add some technical details about what happens when the Licensify->UKCloud VPN is down. This hopefully clarifies how the alert is implemented so that troubleshooting is easier.

An assumption about the implementation of the vpn-down alert led to some unfortunate confusion and delay in resolving an incident recently, hence adding the extra info here.


sihugh (approved):
⭐

PR #3635, opened 2022-08-03
Update CDN cache purge docs to use auth tokens

robinjam:
[Trello card](https://trello.com/c/O5Wy2X9w/2946-authenticate-fastly-purge-requests-with-an-api-token-3)
[Rendered version of the updated docs](https://github.com/alphagov/govuk-developer-docs/blob/update-cdn-purge-docs/source/manual/purge-cache.html.md)

We are updating our Fastly configuration to authenticate cache purge requests using API tokens instead of an IP address allowlist.[^1]

This PR updates the documentation about purging the CDN cache to help developers authenticate their requests using Fastly API tokens stored in the `govuk-secrets` repo. By authenticating our requests we also remove the requirement to `ssh` onto a `cache` machine before issuing the request, so that step has been removed (though it is still necessary to connect from on-premises or via the VPN due to an account-level allowlist on Fastly).

[^1]: https://docs.fastly.com/en/guides/authenticating-api-purge-requests

### DIFF
--- a/source/manual/purge-cache.html.md
+++ b/source/manual/purge-cache.html.md
@@ -54,18 +54,39 @@ GOV.UK internal varnish.
 
 ## Purging a page from Fastly manually (e.g. if GOV.UK Production is dead)
 
-To purge content on the Fastly cache nodes, SSH onto a `cache` machine and use the
-PURGE method against the URL you wish to purge. For instance:
+To purge content on the Fastly cache nodes, use the PURGE method against the URL
+you wish to purge, passing the correct Fastly API token for the environment.
+
+First, connect to the GDS VPN if you're out of the office or on GovWiFi. Our
+Fastly account is configured with an IP allowlist, preventing access to the
+Fastly website or API from outside the GDS network.
+
+Next, extract the Fastly API token. Check out the `govuk-secrets` repo, install
+dependencies with `bundle install`, `cd` into the `puppet_aws` directory,
+and run the following command, replacing `production` with a different
+environment if required:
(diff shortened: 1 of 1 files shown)


source/manual/purge-cache.html.md
```diff
@@ -54,18 +54,35 @@ GOV.UK internal varnish.
...
+Fastly account is configured with an IP allowlist, preventing access to the
+Fastly website or API from outside the GDS network.
+
+Next, extract the Fastly API token. Check out the `govuk-secrets` repo, install
+dependencies with `bundle install`, `cd` into the `puppet_aws` directory,
+and run the following command, replacing `production` with a different
+environment if required:
+
+```sh
+$ FASTLY_API_TOKEN=$(bundle exec rake 'eyaml:decrypt_value[production,govuk_jenkins::jobs::clear_cdn_cache::fastly_api_token]')
```
ChrisBAshton, line 70: I like the shorthand here, but it didn't work for me. Echoing the variable, I had empty output.

When I ran the `bundle exec rake ...` command directly, I found out why: I needed to enter the password for my keychain.

The `echo` statement below should help developers catch that, but it might be safer to just say "retrieve the token using `bundle exec ...` and then leaving it up to the developer to set the `$FASTLY_API_TOKEN` alias or just hardcode the token when they do the `curl`.

robinjam: Good point!

ChrisBAshton (approved):
⭐

PR #3641, opened 2022-08-09
Add command to log into detached instances

KludgeKML:
Add some information on logging in to detached instances after they've been detached.

https://trello.com/c/dXJ3PMbu/1460-drill-detach-an-instance


AgaDufrat (approved):
noice 👍

PR #3671, opened 2022-09-02
Add fix for unhealthy nodes

MartinJJones:
Update the documentation to include a fix that was used to resolve an Icinga alert.


KludgeKML (approved):
LGTM

PR #3672, opened 2022-09-02
Add docs on using a restored database in an app

deborahchua:
Trello card: https://trello.com/c/uqU7EHja/2949-figure-out-how-to-get-apps-to-point-to-new-database

### DIFF
--- a/source/manual/howto-backup-and-restore-in-aws-rds.html.md
+++ b/source/manual/howto-backup-and-restore-in-aws-rds.html.md
@@ -204,7 +204,43 @@ gds-cli aws govuk-<environment>-admin aws route53 change-resource-record-sets --
 
 To see the changes to AWS Route 53, Log into AWS Console > Route 53 > Hosted Zones > Check the Public and private link to <environment>.govuk-internal.digital and search for the restored DB.
 
-## How to Delete a Database Instance
+## Restore an RDS instance via the AWS CLI
+
+Once you have restored a database in AWS RDS, you can now use this in your app.
+
+### 1. Make a change to the database contents
+
+Through the app's user interface, or via the app console or database console, make
+a change that you can use as a sense check to verify that the database switch has
+been successful.
+
+For example, you might create a draft edition of something, or modify or delete
+a record. After switching to the restored database, your changes should be undone.
+
+In this example we want to make a change to the database `local-links-manager_production`
+such as delete an old record:
+
+```
+sudo psql -U aws_db_admin -h local-links-manager-postgres -d local-links-manager_production
+```
+
+### 2. Connect to the restored backup database
+
+This requires updating the CNAME for `local-links-manager-postgres`.
+
+1. In AWS Route53 navigate to `Route 53 > Hosted zones > integration.govuk-internal.digital`
+2. In the list search for the hostname and select it to edit the record.
+3. Make a note of the current value if you are planning on reconnecting to the original database afterwards e.g. if you're carrying this out as a drill
+4. Replace the value with the new RDS backup and save your changes. This takes about 60 seconds, you can click "view status" for updates. Once updated it will say `INSYNC`.
+5. SSH back into the machine and query for the record you deleted. If the record is back this should verify the app is now using the backup database.
+
+### 3. Connect back to the original database
+
+Repeat the steps covered in [Step 5](#connect-to-the-restored-backup-database).
+
+> If you need to find the endpoint again for the original database, navigate to Amazon RDS, find the database in the list and look for `Endpoint & Port` under the `Connectivity & Security` tab.
+
+### 4. Delete the restored backup database
 
 > PLEASE BE CAREFUL WHEN EXECUTING THIS COMMAND AS IT CAN NOT BE UNDONE
 


source/manual/howto-use-restored-db-in-app.html.md
```diff
@@ -0,0 +1,83 @@
+---
+owner_slack: "#govuk-2ndline-tech"
+title: How to use a restored database in an app
+section: Backups
+layout: manual_layout
+parent: "/manual.html"
+---
+
+Once you have [restored a database in AWS RDS](/manual/howto-backup-and-restore-in-aws-rds.html),
```
ChrisBAshton, line 9: We'll need to point to this new doc from the one above ^.
Or, there's an argument for keeping them both in the same doc - I can't think of a situation where you wouldn't want to point the app to the restored DB.

source/manual/howto-use-restored-db-in-app.html.md
```diff
@@ -0,0 +1,83 @@
...
+you can now use this in your app.
+
+> The steps outlined below are for what to do during an incident where you may need to
+temporarily switch to use a backup. For a more permanent solution this will require
+work in Terraform as any new instances will still use the original database.
```
ChrisBAshton, line 14: This leads the reader down a different path than intended - we're not suggesting they use Terraform, we're suggesting they restore another DB to the original DB name and then point back to that. I think we can just remove this pullout above.

source/manual/howto-use-restored-db-in-app.html.md
```diff
@@ -0,0 +1,83 @@
...
+This will help us determine whether we are using the backup database which will still
+contain this record.
+
+Select and delete an old user e.g.
+
+```
+select * from users where id=6;
+delete from users where id=6;
+DELETE 1
+```
```
ChrisBAshton, line 63: I think we can be less prescriptive here - sections 1-4 are all about making a change we can use to verify whether the DB switch has happened. I think that can be simplified to something like:

```md
## 1. Make a change to the database contents

Through the app's user interface, or via the app console or database console, make a change that you can use as a sense check to verify that the database switch has been successful.

For example, you might create a draft edition of something, or modify or delete a record. After switching to the restored database, your changes should be undone.
```

This has the advantage of being generic enough to apply to MySQL restores too - what you have so far will only work for Postgres.

source/manual/howto-use-restored-db-in-app.html.md
```diff
@@ -0,0 +1,83 @@
...
+4. Replace the value with the new RDS backup and save your changes. This takes about 60 seconds, you can click "view status" for updates. Once updated it will say `INSYNC`.
+5. SSH back into the machine and query for the record you deleted. If the record is back this should verify the app is now using the backup database.
+
+### 6. Connect back to the original database
+
+Repeat the steps covered in [Step 5](#connect-to-the-restored-backup-database).
+
+> If you need to find the endpoint again for the original database, navigate to Amazon RDS, find the database in the list and look for `Endpoint & Port` under the `Connectivity & Security` tab.
+
+### 7. Delete the restored backup database
```
ChrisBAshton, line 81: Nitpick: all of these should be heading level 2.

ChrisBAshton (commented):
Nice one for figuring out how to do this 👏 
Easy to follow documentation 👍 
Some comments below.

source/manual/howto-use-restored-db-in-app.html.md
nicholsj: Yep I think it makes sense to incorporate this as a section into /manual/howto-backup-and-restore-in-aws-rds.html, rather than it being a separate page

deborahchua: Good point both! Will move this into the backup and restore doc.

source/manual/howto-backup-and-restore-in-aws-rds.html.md
```diff
@@ -204,6 +204,71 @@ gds-cli aws govuk-<environment>-admin aws route53 change-resource-record-sets --
...
+sudo psql -U <user> -h <host> -d <database name>
+```
+
+In this example we want to connect to the database `local-links-manager_production`:
+
+```
+sudo psql -U aws_db_admin -h local-links-manager-postgres -d local-links-manager_production
+```
+
+### 4. Make a change to the database contents
```
ChrisBAshton, line 247: Now we have this step, we can get rid of steps 1-3.

source/manual/howto-backup-and-restore-in-aws-rds.html.md
```diff
@@ -204,6 +204,71 @@ gds-cli aws govuk-<environment>-admin aws route53 change-resource-record-sets --
...
+2. In the list search for the hostname and select it to edit the record.
+3. Make a note of the current value if you are planning on reconnecting to the original database afterwards e.g. if you're carrying this out as a drill
+4. Replace the value with the new RDS backup and save your changes. This takes about 60 seconds, you can click "view status" for updates. Once updated it will say `INSYNC`.
+5. SSH back into the machine and query for the record you deleted. If the record is back this should verify the app is now using the backup database.
+
+### 6. Connect back to the original database
+
+Repeat the steps covered in [Step 5](#connect-to-the-restored-backup-database).
+
+> If you need to find the endpoint again for the original database, navigate to Amazon RDS, find the database in the list and look for `Endpoint & Port` under the `Connectivity & Security` tab.
```
ChrisBAshton, line 270: Wasn't there something here about deleting the 'temporary' database? Might be useful to point out that that should be the next (final) step ("see below").

deborahchua: I missed this in my copy & paste - that's added back in now.

source/manual/howto-backup-and-restore-in-aws-rds.html.md
```diff
@@ -204,6 +204,46 @@ gds-cli aws govuk-<environment>-admin aws route53 change-resource-record-sets --
...
+
+### 4. Delete the restored backup database
+
+See [how to delete a database instance](#how-to-delete-a-database-instance).
+
 ## How to Delete a Database Instance
```
ChrisBAshton, line 247:
```suggestion
```

I think we can delete this section title and just have `4. Delete the restored backup database` lead directly to the instructions for deleting.

ChrisBAshton (approved):
LGTM ⭐ one small comment.

PR #3676, opened 2022-09-07
Fixes to the 'restore RDS instance from snapshot' documentation

ChrisBAshton:
The previous iteration was geared towards how to drill the process. This iteration expands on the instructions to account for what the process would be outside of a drill.

In writing this, I've arrived at the conclusion that the "manually getting back to where we were" steps are quite laborious/risky/error-prone, so documenting a Terraform approach would be better. I'm not quite sure what the implementation steps would be, but that can be documented later. Crucially, we have now documented the immediate steps to follow during an incident, so any tidy-up steps can be discussed post-incident and the docs updated. Hopefully the non-optimal route documented here is still a net improvement on what was there before.

NB I've also fixed a couple of broken headings and links.

https://trello.com/c/uqU7EHja/2949-figure-out-how-to-get-apps-to-point-to-new-database

### DIFF
--- a/source/manual/howto-backup-and-restore-in-aws-rds.html.md
+++ b/source/manual/howto-backup-and-restore-in-aws-rds.html.md
@@ -200,9 +200,9 @@ Please be aware that if you changed the file path in the previous step, remember
 
 To see the changes to AWS Route 53, Log into AWS Console > Route 53 > Hosted Zones > Check the Public and private link to <environment>.govuk-internal.digital and search for the restored DB.
 
-## Restore an RDS instance via the AWS CLI
+## Point the application to the new RDS instance
 
-Once you have restored a database in AWS RDS, you can now use this in your app.
+Once you have restored a database in AWS RDS, you now need to point the corresponding app towards it.
 
 ### 1. Make a change to the database contents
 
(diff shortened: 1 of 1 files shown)


source/manual/howto-backup-and-restore-in-aws-rds.html.md
```diff
@@ -230,13 +230,26 @@ This requires updating the CNAME for `local-links-manager-postgres`.
...
+*For the purposes of drilling*, it's quicker and easier to do the latter. Simply repeat the steps to [point the application to the new RDS instance](#point-the-application-to-the-new-rds-instance), but this time connect to the original database. If you need to find the endpoint again for the original database, navigate to Amazon RDS, find the database in the list and look for `Endpoint & Port` under the `Connectivity & Security` tab.
+
+*If this is not a drill*, then you would not want to connect to the original database again - you've created a database from a backup for a reason!
+The safest approach would be to update Terraform to refer to the new database.
 
-Repeat the steps covered in [Step 5](#connect-to-the-restored-backup-database).
+Alternatively, you could:
 
-> If you need to find the endpoint again for the original database, navigate to Amazon RDS, find the database in the list and look for `Endpoint & Port` under the `Connectivity & Security` tab.
+1. [Delete the original database](#delete-an-obsolete-database)
```
deborahchua, line 246: Would we want to carry out this step last, after the Terraform is updated?

source/manual/howto-backup-and-restore-in-aws-rds.html.md
```diff
@@ -230,13 +230,26 @@ This requires updating the CNAME for `local-links-manager-postgres`.
 4. Replace the value with the new RDS backup and save your changes. This takes about 60 seconds, you can click "view status" for updates. Once updated it will say `INSYNC`.
 5. SSH back into the machine and query for the record you deleted. If the record is back this should verify the app is now using the backup database.
 
-### 3. Connect back to the original database
+> This is only a temporary solution, to be used in an incident. You should continue onto the next section for a permanent solution.
+
+## Ensure your setup will continue to work if infrastructure is reprovisioned
+
+If new infrastructure is provisioned, then the "Point the application to the new RDS instance" solution above will break, as Terraform would fall out of sync with the manual changes. We either need to update Terraform with our changes, or manually get our infrastructure back to how it used to be.
```
sihugh, line 237: Should we try and think of a process to block deploying certain terraform projects? Is it hard to state here which ones they might be? (I guess different apps might have a different set of deps)

source/manual/howto-backup-and-restore-in-aws-rds.html.md
ChrisBAshton: In this numbered list of instructions, we're not touching any Terraform.
We'd need to delete the original database first because we'll then be creating a new database under the same hostname (can't have two databases with the same name). Unless I've misunderstood how it works? 🤔

source/manual/howto-backup-and-restore-in-aws-rds.html.md
ChrisBAshton: Probably better/easier to make the Terraform changes rather than build in a process for blocking deployments.

source/manual/howto-backup-and-restore-in-aws-rds.html.md
deborahchua: Ah yes that makes sense!

deborahchua (approved):

PR #3689, opened 2022-09-16
Try make it clearer who the getting started guidance is for

kevindew:
It seems that it's very common to get misunderstandings of this document with people new to GDS working in other programmes (DI is most common).

I've also removed the reference to CDDO as we don't have technical people in there who need this guidance.


sihugh (approved):
BINGPOT

PR #3690, opened 2022-09-16
Add WorldLocationNews to republishable content types

brucebolt:
Now that [`WorldLocationNews` has been moved out of the `WorldLocation` model into a model of its own](https://github.com/alphagov/whitehall/pull/6722), we have removed the special rake task to publish this content type. This type of document can now be republished in the same way as all other document types.

### DIFF
--- a/source/manual/republishing-content.html.md.erb
+++ b/source/manual/republishing-content.html.md.erb
@@ -58,6 +58,7 @@ To republish all instances of the following document types, run the following ra
 - TopicalEvent
 - TopicalEventAboutPage
 - WorldLocation
+- WorldLocationNews
 - WorldwideOrganisation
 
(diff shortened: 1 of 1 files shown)


source/manual/republishing-content.html.md.erb
```diff
@@ -58,6 +58,7 @@ To republish all instances of the following document types, run the following ra
 - TopicalEvent
 - TopicalEventAboutPage
 - WorldLocation
```
JonathanHallam, line 60: Does world location need removing if it's not a proper content type any more?

brucebolt: It needs to stay as the International Delegation pages can only be republished by republishing the `WorldLocation` model.

BeckaL (approved):

PR #3706, opened 2022-10-03
Document expectations of standby developers

ChrisBAshton:
This is the policy as informally agreed between Lead Developers and 2nd line delivery manager and tech leads. Note that until now, standby developers have not been allowed to attend All Staff events in person, so as well as general documentation, this is also documenting a small change in policy.


deborahchua (approved):

kevindew (approved):
Looks good. Thanks Chris. Sorry for taking so long

PR #3710, opened 2022-10-06
Update docs that mention mapit where is no longer used

catalinailie:
Mapit is no longer used by Frontend.
We need to update the postcode-related mentions in the dev docs.

Trello card: https://trello.com/c/6TO6GDJM/2934-switch-frontend-to-use-locations-api-5

### DIFF
--- a/source/manual/incorrect-postcode-data.html.md
+++ b/source/manual/incorrect-postcode-data.html.md
@@ -6,16 +6,36 @@ layout: manual_layout
 section: 2nd line
 ---
 
-We use postcode data from the [ONS Postcode Directory](https://geoportal.statistics.gov.uk/search?q=ONS%20Postcode%20Directory%20) (ONSPD) and Ordnance Survey for boundary
-line data. We typically [import the data into Mapit](https://github.com/alphagov/mapit/blob/master/IMPORTING-DATA.md) every 6 months.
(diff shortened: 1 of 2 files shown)


source/manual/incorrect-postcode-data.html.md
```diff
@@ -6,16 +6,34 @@ layout: manual_layout
...
 
-We sometimes receive Zendesk tickets about postcodes not returning the right results.
-You can check the accuracy of the postcode by:
+In case the data is incorrect, a first step would be checking the Locations API result with the OS Places API result.
 
-* Looking up the postcode on Mapit, for example https://mapit.mysociety.org/postcode/E18QS
-  and viewing the parents. There can be more than one parent which can be the case
-  if the postcode sits between two boundaries
-* Get the latitude/longitude and look it up on Google Maps
+```shell
```
ChrisBAshton, line 13: I'd split this into two code blocks, one for `shell` and one for `ruby`, to get the correct highlighting in both.

ChrisBAshton (approved):
LGTM ⭐

source/manual/incorrect-postcode-data.html.md
catalinailie: Yes, good point. Will do this now.

PR #3716, opened 2022-10-10
Add reminder not to remove user from .mailmap

KludgeKML:
When removing a user from govuk-user-reviewer, don't remove them from the .mailmap file. Updated 2L docs.

### DIFF
--- a/source/manual/removing-a-user-from-puppet.html.md
+++ b/source/manual/removing-a-user-from-puppet.html.md
@@ -21,12 +21,14 @@ will remain on our servers forever more, [unless you perform a workaround](#what
 1. Create another PR for Puppet that:
   - Removes the user manifest file
   - Removes the user from [Integration users][integration-users]
+  - DO NOT remove the user's email from the [.mailmap file][mailmap] if present - this allows the user to keep a nicer git commit history.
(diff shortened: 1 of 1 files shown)


source/manual/removing-a-user-from-puppet.html.md
```diff
@@ -21,12 +21,14 @@ will remain on our servers forever more, [unless you perform a workaround](#what
 1. Create another PR for Puppet that:
   - Removes the user manifest file
   - Removes the user from [Integration users][integration-users]
+  - DO NOT remove the user's email from the [.mailmap file][mailmap] if present - this allows use to keep a nicer git commit history once people have left.
```
ChrisBAshton, line 24: A typo I think - here's what I think you meant:

```suggestion
  - DO NOT remove the user's email from the [.mailmap file][mailmap] if present. It allows the user to keep a nicer git commit history.
```

AgaDufrat (approved):

source/manual/removing-a-user-from-puppet.html.md
KludgeKML: Oh, I actually meant "this allows *us*", but your change is also good. Will fix.

PR #3724, opened 2022-10-17
Mark datagovuk_reference as retired

ChrisBAshton:
The repo is now archived, and http://reference.data.gov.uk/ is unreachable.


nicholsj (approved):
LGTM thanks

PR #3731, opened 2022-10-21
Add dependabot auto-merge workflow

MuriloDalRi:
This workflow will auto-merge Dependabot PRs for selected internal dependencies if they're minor or patch version bumps.

https://trello.com/c/uwoMBinS/3015-create-proof-of-concept-for-auto-merging-internal-prs-5

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -1,0 +1,25 @@
+name: Dependabot auto-merge
+on: pull_request
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1.3.4
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Enable auto-merge for Dependabot PRs
+        if: ${{contains(env.AUTO_MERGE, steps.metadata.outputs.dependency-names) && contains(env.SEMVER, steps.metadata.outputs.update-type)}}
+        run: gh pr merge --auto --merge "$PR_URL"
+        env:
+          PR_URL: ${{github.event.pull_request.html_url}}
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          AUTO_MERGE: ("govuk_publishing_components")
+          SEMVER: ("version-update:semver-patch")


.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,25 @@
...
+on: pull_request
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
```
ChrisBAshton, line 11: Interesting `[bot]` syntax - I guess that's how GitHub exposes the actor name, to prevent someone just creating an account called 'dependabot'?
Also looks like `actor` is the right one to use [instead of `triggering_actor`](https://docs.github.com/en/actions/learn-github-actions/contexts) 👍

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1.3.4
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Enable auto-merge for Dependabot PRs
+        if: ${{contains(env.AUTO_MERGE, steps.metadata.outputs.dependency-names) && contains(env.SEMVER, steps.metadata.outputs.update-type)}}
```
ChrisBAshton, line 19: Hmmm we have to be a bit careful here... my understanding of [`contains`](https://docs.github.com/en/actions/learn-github-actions/expressions#contains) is that it will return `true` if the dependency is present, even if other ones we _don't_ want to merge are present. I.e.

```
env.AUTO_MERGE = ["govuk_publishing_components", "govuk_tech_docs"]
steps.metadata.outputs.dependency-names == ["govuk_publishing_components", "something_risky"]

contains(...) # returns true
```

In _practice_, I don't think this should be an issue 🤔 Dependabot may bump other _subdependencies_, but we trust that risk to be managed appropriately by use of semver (i.e. a risky subdependency upgrade would constitute a 'major' upgrade to `govuk_publishing_components`, so not be auto-merged).
Otherwise, I can't think of a situation where Dependabot would raise a PR that bumps two or more direct dependencies.

That said, is there any way we could tighten the scope so that it's doing something like this?

```ruby
if direct_dependencies_updated.count == 1
  && env.AUTO_MERGE.include?(direct_dependencies_updated.first)
  && (is_patch_update? || is_minor_update?)
    auto_merge!
else
  # leave it to the dev to manually review and merge
end
```

ChrisBAshton (approved):
Nice one 👏 succinct, elegant solution, and will be a good spike prior to an RFC.
I'm interested to see this in action and watch it auto-merge a govuk_publishing_components PR next week!

One comment below about being a bit less permissive in the "should this be auto-merged?" checks, but probably too complex to express using the [GitHub Actions expressions](https://docs.github.com/en/actions/learn-github-actions/expressions). Interested to get your thoughts though.

.github/workflows/dependabot-auto-merge.yml
MuriloDalRi: Good point, I didn't consider the possibility that Dependabot might raise a PR to update multiple direct dependencies. I'll see if I can make that change.

MuriloDalRi: I've had a look at this and I don't think Dependabot ever opens PRs to update more than one direct dependency at a time, so probably not something we need to check explicitly.

The other issue of a major change to an indirect dependency should be covered by our SEMVER restriction. As you can see [here](https://github.com/dependabot/fetch-metadata/blob/d9606730415777cc0dc46d64c4ce0e16624bd714/src/dependabot/output.test.ts#L67), this action should always return the highest value for dependency type and update type when multiple dependencies are updated.

This means that if Dependabot opens a PR to make a minor change to a direct dependency, which includes a major update to an indirect dependency, it would be classified as a major change to a direct dependency and therefore it would not be merged.

The only possible problem I can think of is when a PR is updating a dependency that isn't on our list, but an indirect dependency also being updated is on our list and it gets merged because of the `contains` method returning true. 

However, in my testing `steps.metadata.outputs.dependency-names` always returns the direct dependency being updated and nothing else.

@ChrisBAshton does that make sense?

ChrisBAshton: Makes sense 👍 thanks for looking into it.

My approval ✅ still stands then - pending senior tech approval!

ChrisBAshton (approved):
LGTM.

PR #3762, opened 2022-11-15
Mark govuk_taxonomy_helpers as retired

AgaDufrat:
This project was archived in October 2022, the functionality has been ported into content-tagger in https://github.com/alphagov/content-tagger/pull/1391

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -737,10 +737,12 @@
   dashboard_url: false
 
 - repo_name: govuk_taxonomy_helpers
-  team: "#navigation-and-presentation-govuk"
+  retired: true
   type: Utilities
-  sentry_url: false
-  dashboard_url: false
+  description: |
+    govuk_taxonomy_helpers was a gem that provided code for working with the
+    govuk-taxonomy. It was only used by Content Tagger and was retired as a gem
+    after the code was merged into Content Tagger.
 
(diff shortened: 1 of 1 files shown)


data/repos.yml
```diff
@@ -738,6 +738,7 @@
 
 - repo_name: govuk_taxonomy_helpers
   team: "#navigation-and-presentation-govuk"
+  retired: true
   type: Utilities
   sentry_url: false
   dashboard_url: false
```
kevindew, line 744:
```suggestion
  retired: true
  type: Utilities
  description: |
    govuk_taxonomy_helpers was a gem that provided code for working with the govuk-taxonomy. It was only used by Content Tagger and was retired as a gem after the code was merged into Content Tagger.
```

kevindew (commented):
Thanks, for catching this. I missed it when retiring the gem.

There's a couple of small changes we should make as we don't need as much information and we should keep a description. Here's a similar example: https://github.com/alphagov/govuk-developer-docs/blob/b3abfe28c59514d24840f34fad3efc0704885422/data/repos.yml#L82-L88

kevindew (approved):
Thank you.

I have a paired PR for this if you have a sec too: https://github.com/alphagov/govuk-developer-docs/pull/3772

PR #3765, opened 2022-11-17
Update licensify docs now that the Jenkins build is fixed

richardTowers:
Also the default branch is main now, so fix that too.


dj-maisy (approved):
Looks good to me! 👍

PR #3767, opened 2022-11-18
Remove references to fabric

GDSNewt:
### **What**

Fabric is an old tool that can no longer be used by [GOV.UK](http://gov.uk/) Devs. We should therefore remove all references to it from our docs and replace them with alternative commands

### **Why**

Fabric used to let us apply an ssh command across lots of machines. However in recent years the tool has become difficult for many devs to install.

It relies on an old deprecated form of python, and its dependencies cannot be installed today in a way that function.

It has still been referred to as a command in numerous places in [GOV.UK](http://gov.uk/) developer documentation, causing confusion for 2ndline shifts which are unable to use the tool.

This PR therefore seeks to remove said references. 

Trello: https://trello.com/c/AGdifoDb/250-remove-fabric-and-replace-references-in-docs

### DIFF
--- a/source/manual/alerts/rabbitmq-no-consumers-listening.html.md
+++ b/source/manual/alerts/rabbitmq-no-consumers-listening.html.md
@@ -15,12 +15,12 @@ the alert.
 
 For information about how we use RabbitMQ, see [here][rabbitmq_doc].
 
-### `No consumers listening to queue`
+### No consumers listening to queue
 
 This check reports a critical error when **no** consumers are listening to the
 queue, meaning messages entering the queue will never be processed.
 
-### `No activity for X seconds: idle times are [X, Y, Z]`
+### No activity for X seconds: idle times are [X, Y, Z]
 
 This checks whether the RabbitMQ consumers for a queue have been active in the
 last 5 minutes. Consumers in an `idle` state are listening to the queue but are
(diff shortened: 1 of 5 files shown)


source/manual/alerts/rabbitmq-no-consumers-listening.html.md
```diff
@@ -74,12 +74,7 @@ messages then this could indicate an issue with the consumer.
 
 ### Troubleshooting steps
 
-1. You could try restarting the consumer applications. After restarting, check
-   to see if the problem is solved. E.g for email-alert-service, run:
-
-```bash
-$ fab $environment class:email_alert_api app.restart:email-alert-service
-```
+1. You could try restarting the application [on all the machines in a class](https://docs.publishing.service.gov.uk/manual/howto-run-ssh-commands-on-many-machines). After restarting, check to see if the problem is solved.
```
ChrisBAshton, line 77: Suggest you signpost the `email_alert_api` machine class here.

```suggestion
1. You could try restarting the application [on all the `email_alert_api` machines](https://docs.publishing.service.gov.uk/manual/howto-run-ssh-commands-on-many-machines). After restarting, check to see if the problem is solved.
```

source/manual/deploy-puppet.html.md
```diff
@@ -49,13 +49,7 @@ before the release takes effect. This has an implication on how
...
 
 If you'd rather not wait and you're able to safely determine from the
-diff what classes of machines the change will affect, you can use
-[Fabric](https://github.com/alphagov/fabric-scripts) to force a run of
-Puppet. For example:
-
-```
-fab $environment class:frontend_lb class:backend_lb puppet
-```
+diff what classes of machines the change will affect, [you can force a run of puppet](/manual/howto-run-ssh-commands-on-many-machines.html#run-puppet).
```
ChrisBAshton, line 52:
```suggestion
diff what classes of machines the change will affect, [you can force a run of puppet](/manual/howto-run-ssh-commands-on-many-machines.html#run-puppet) on the relevant machine classes.
```

ChrisBAshton (commented):
LGTM, a couple of QA comments below

ChrisBAshton (approved):

PR #3771, opened 2022-11-23
Add policy on Ruby versions for Rubygems

kevindew:
This adds documentation on the approach we take for managing Ruby versions with gems, this is added as documentation is this is not widely known. I've also added some links to documentation that may be used when updating Ruby to hopefully bring this to peoples attention (thanks for the suggestions Murilo, Ollie and Jonathan).

With our Ruby apps we aim to run them, in dev, CI and production on the newest Ruby version available. However with a Rubygem we aim for it to be compatible across the range of supported Ruby versions - this allows our apps to upgrade Ruby versions when the teams are ready and is friendly to anyone outside GDS who also uses the gems.

When supporting a number of Ruby versions it is better to run on the lowest version by default in development. This is because it immediately prevents using features that aren't available in all of the supported Rubies and it stops the linter trying to make suggestions to utilise those new features.

### DIFF
--- a/source/manual/publishing-a-ruby-gem.html.md
+++ b/source/manual/publishing-a-ruby-gem.html.md
@@ -47,6 +47,35 @@ The default Jenkins build script will automatically detect the presence of a
 [Jenkinsfile for govuk_app_config](https://github.com/alphagov/govuk_app_config/blob/master/Jenkinsfile)
 for an example.
 
+## Ruby version compatibility
+
+Our policy is that our Ruby gems are compatible with all [currently supported
+minor versions of Ruby][supported-rubies]. For example, in November 2022, There
+are supported Ruby releases of 2.7, 3.0 and 3.1, thus we expect gems to be
+compatible with each of those and be tested against them (there is
+[documentation][testing-gems] on the approach to test them).
+
+We specify the minimum Ruby version supported in the [gemspec file][gemspec-ruby-version]
+and expect the `.ruby-version` to match that version. For example, if Ruby 2.7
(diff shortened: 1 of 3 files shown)


source/manual/publishing-a-ruby-gem.html.md
```diff
@@ -47,6 +47,35 @@ The default Jenkins build script will automatically detect the presence of a
...
+greater and the `.ruby-version` file to reference the most recent Ruby 2.7
+release (which in November 2022 is 2.7.6).
+
+When new minor versions of Ruby are released (typically each Christmas) we
+update gems to test against the new version. For example, when Ruby 3.2
+is released our gem test matrices should be expanded to test against Ruby 3.2.
+
+When Ruby versions reach end-of-life (typically April) we update gems
+to drop support for that Ruby version and update the `.ruby-version` files to
+the newest version. For example, when Ruby 2.7 reaches end of life, we drop 2.7
```
ChrisBAshton, line 70: "newest" seems the wrong choice of word here, as that would imply Ruby 3.1:

```suggestion
the next supported version. For example, when Ruby 2.7 reaches end of life, we drop 2.7
```

ChrisBAshton (approved):
Looks good ⭐ one nitpick comment.

source/manual/publishing-a-ruby-gem.html.md
kevindew: Thanks, good catch. It's rather tricky describing how should take the newest release from the oldest supported branch.

PR #3779, opened 2022-11-28
Add tips about finding an instance, and docker_management locks.

KludgeKML:
Add some tips about how to find an instance in AWS, plus reasons you might want to find on and why rebooting docker_management might clear unattended reboot locks.

https://gds.slack.com/archives/CAB4Q3QBW/p1668428029358079

### DIFF
--- a/source/manual/find-an-instance-in-aws.html.md
+++ b/source/manual/find-an-instance-in-aws.html.md
@@ -1,0 +1,19 @@
+---
+owner_slack: "#govuk-2ndline-tech"
+title: Identify an EC2 instance in AWS
+section: 2nd line
+layout: manual_layout
+parent: "/manual.html"
+---
+
+Sometimes you may have references (in logs or in a [locksmith log file](/manual/alerts/rebooting-machines.html#checking-locking-status)) to an EC2 instance by its internal name (for example ip-10-13-4-254.eu-west-1.compute.internal), and need to work out what it is.
+
+The numbers after IP are the internal IP address in AWS, which will allow you to narrow down where it should be:
+
+1. You can tell which environment it's in by looking at the first two octets of the IP (because we use /16s for environments) - production uses 10.13.0.0/16
+2. You can tell which subnet it's in by the first three octets of it's IP address. 10.13.4.0/24 is govuk_private_a. You can find details for the subnets in all enviroments by looking in [govuk-aws-data](https://github.com/alphagov/govuk-aws-data/tree/main/data/infra-networking).
+3. Not everything that gets a private IP address is an Ec2 instance - things like NAT gateways and RDS instances get them too (although RDS instances have their own subnet, so they won't be in `govuk_private_{a,b,c}`)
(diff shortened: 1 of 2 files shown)


source/manual/find-an-instance-in-aws.html.md
```diff
@@ -0,0 +1,19 @@
...
+section: 2nd line
+layout: manual_layout
+parent: "/manual.html"
+---
+
+Sometimes you may have references (in logs or in a [locksmith log file](/manual/alerts/rebooting-machines.html#checking-locking-status)) to an EC2 instance by its internal name (for example ip-10-13-4-254.eu-west-1.compute.internal), and need to work out what it is.
+
+The numbers after IP are the internal IP address in AWS, which will allow you to narrow down where it should be:
+
+1. You can tell which environment it's in by looking at the first two octets of the IP (because we use /16s for environments) - production uses 10.13.0.0/16
```
richardTowers, line 13: Is it worth linking to the code in govuk-aws-data where the other subnets are defined? This tells you that production uses 10.13/16, but it doesn't say how to work out what integration, staging and test are.

source/manual/find-an-instance-in-aws.html.md
```diff
@@ -0,0 +1,19 @@
...
+2. You can tell which subnet it's in by the first three octets of it's IP address. 10.13.4.0/24 is govuk_private_a
+3. Not everything that gets a private IP address is an Ec2 instance - things like NAT gateways and (maybe?) RDS instances get them too
```
richardTowers, line 15:
```suggestion
3. Not everything that gets a private IP address is an Ec2 instance - things like NAT gateways and RDS instances get them too (although RDS instances have their own subnet, so they won't be in `govuk_private_{a,b,c}`)
```

source/manual/find-an-instance-in-aws.html.md
```diff
@@ -0,0 +1,19 @@
...
+4. Searching in Network Interfaces should get them all, and you can search by a particular subnet - e.g. https://eu-west-1.console.aws.amazon.com/ec2/home?region=eu-west-1#NIC:v=3;subnetId=subnet-0e44c1f27dbaf5af7;sort=desc:privateIpAddress
+5. "Amazon reserves the first four (4) IP addresses and the last one (1) IP address of every subnet for IP networking purposes." but I'm not sure if this means 10.13.4.255 is reserved (I guess as the broadcast address?) or if it means that 10.13.4.254 is reserved in your /24
```
richardTowers, line 17:
```suggestion
5. "Amazon reserves the first four (4) IP addresses and the last one (1) IP address of every subnet for IP networking purposes."
```

richardTowers: I think it is the broadcast address, for what it's worth. But this detail is probably too much unless you're specifically looking for `.254`

KludgeKML:
@richardTowers Have updated with your suggestions.

richardTowers (approved):

PR #3781, opened 2022-11-28
Retire the Knowledge Graph and Accessibility Reports Generator

maxf:
Both were owned by data products. The Knowledge Graph still exists but is now hosted on CPTO's Gcloud account. The accessibility reports generator used data generated for the knowledge graph so needed to be migrated as well, however the service is no longer being used, so was not migrated.


nacnudus:
Thanks, well spotted in the YAML.

What should we do about these files?

- source/manual/datalabs-govuk-knowledge-graph.html.md
- source/manual/knowledge-graph.html.md
- source/manual/datalabs-onboarding-guidance.html.md

nacnudus (approved):

PR #3782, opened 2022-11-29
Document GitHub Actions process to release a gem

kevindew:
Trello: https://trello.com/c/QrxjshEm/270-move-ruby-gems-from-jenkins-to-github-actions

This supersedes the Jenkins approach. As of today only a couple of applications are using Jenkins to release gems. Soon 
this will be zero and that functionality will be removed.


ChrisBAshton (approved):
LGTM 👍

PR #3783, opened 2022-11-29
Pact documentation for GitHub Actions

kevindew:
Trello: https://trello.com/c/QrxjshEm/270-move-ruby-gems-from-jenkins-to-github-actions

With GDS API Adapters moving to GitHub Actions, most of the CI specific documentation is out of date. This change updates it to reflect the changes incurred by GitHub Actions.

There are a couple of notable changes with GitHub Actions:

1) There is not the concept of a parameterised build that can change CI
   status, as per Jenkins, which means we can't do some delicate custom
   builds to get a green tick. GitHub Actions does have the workflow
   dispatch option, but this won't change a PR status. Therefore I've
   suggested running locally
2) Pact runs aren't required checks pre-merge which will allow a
   developer to merge in breaking pact changes.
3) We've switched from both consumer and provider having their own
   verify tasks to the provider having one that the consumer can call.

### DIFF
--- a/source/manual/pact-testing.html.md
+++ b/source/manual/pact-testing.html.md
@@ -28,7 +28,7 @@ bundle exec rake test
 Or they can be run on their own:
 
 ```sh
-bundle exec rake test TEST=test/pacts/**/*.rb
+bundle exec rake pact_test
 ```
 
 ### For a provider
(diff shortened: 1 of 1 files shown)


source/manual/pact-testing.html.md
```diff
@@ -61,21 +61,17 @@ _If the app already had some Pact tests, follow [the steps for changing existing
...
   - [Provider example](https://github.com/alphagov/imminence/pull/644).
 
-1. Check the consumer tests pass locally.
-  - CI won't do this for you yet because it doesn't know about the new app.
+1. Check the tests pass locally for both provider and consumer
+  - CI won't be able to test them yet as they won't be pushed to the pact broker.
 
 1. Merge the consumer tests for the new app.
-  - The build will pass because it runs with the old Jenkinsfile.
+  - This will cause the pact tests to be published to the pact broker, given the [project has been configured to do so](https://github.com/alphagov/gds-api-adapters/blob/bcc8e58eccf69dd37657d13156cbe11c07535844/.github/workflows/ci.yml#L37-L51).
```
ChrisBAshton, line 68: Seems unnecessarily wordy - how about:

```suggestion
  - This will cause the [pact tests to be published to the pact broker](https://github.com/alphagov/gds-api-adapters/blob/bcc8e58eccf69dd37657d13156cbe11c07535844/.github/workflows/ci.yml#L37-L51).
```

source/manual/pact-testing.html.md
```diff
@@ -86,12 +82,16 @@ Follow these steps in order to change the provider and consumer in tandem.
 
 1. Make the change to the pactfile in the consumer (e.g. GDS API Adapters), in a branch.
   - Its build should fail at the Pact test stage, because it is testing against the default branch of the provider.
+1. Confirm locally that the provider can verify against the published consumer pact
+  - `env PACT_CONSUMER_VERSION=branch-<branch-of-gds-api-adapters> bundle exec rake pact:verify`
+
+1. Merge the provider change
+  - The failing pact test should not block merging.
```
ChrisBAshton, line 89: Subtle change, makes it a bit clearer this is by choice and not coincidence:

```suggestion
  - We've configured failing Pact tests to not block merging, so you should still be able to merge.
```

ChrisBAshton (approved):
LGTM 👍 a couple of nitpicks below.

kevindew:
Thanks @ChrisBAshton, good suggestions

ChrisBAshton (approved):

PR #3787, opened 2022-12-02
Document ruby version error fix

JamesCGDS:
This PR documents a fix for an error that can occur following setup when running an application where the installed Ruby version is inconsistent with the Ruby version specified in the application's `Gemfile`.

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -423,7 +423,7 @@ ruby -v (which should display version Y)
 
 If `ruby -v` returns a different version, try restarting the terminal. Otherwise, if the error persists:
 
-1. Add the `rbenv` Ruby binary, and initialisation command, to your `~/.zshrc` file as shown below:
+1. If you're using `zshell`, add the `rbenv` Ruby binary and initialisation command to your `~/.zshrc` file as shown below:
 
 ```
 export PATH="$HOME/.rbenv/bin:$PATH"
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -409,6 +409,29 @@ Signon](https://signon.publishing.service.gov.uk/users/invitation/new), at 'Norm
 the 'Release' app only. No permissions should be given for other applications, until [production access](/manual/rules-for-getting-production-access.html)
 is granted.
 
+## Troubleshooting
+
+### Your Ruby version is [version X], but your Gemfile specified [version Y]
+
+To fix this, run the following command in your terminal:
+
+```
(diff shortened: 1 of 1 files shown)


source/manual/get-started.html.md
```diff
@@ -409,6 +409,29 @@ Signon](https://signon.publishing.service.gov.uk/users/invitation/new), at 'Norm
...
+
+```
+rbenv install [version Y]
+rbenv local [version Y]
+ruby -v (which should display version Y)
+```
+
+If `ruby -v` returns a different version, try restarting the terminal. Otherwise, if the error persists:
+
+1. Add the `rbenv` Ruby binary, and initialisation command, to your `~/.zshrc` file as shown below:
```
andysellick, line 426: Is everyone on GOV.UK using zshell? If we are then fine, but alternatives are available. It shouldn't block this PR, but  maybe worth adding something like `if you're using zshell...`

andysellick (approved):
Great addition 👍

source/manual/get-started.html.md
JamesCGDS: Cool thanks Andy, have updated

kevindew:
Sorry to be a pain, but I'm not sure this should be added to our docs (or at least this page) as I think it risks causing some confusion.

I think it could cause confusion because it conflicts with https://github.com/alphagov/govuk-developer-docs/blob/main/source/manual/get-started.html.md#6-set-up-govuk-docker that states govuk-docker is the supported dev environment and rbenv is not, whereas these commands are not for govuk-docker.

The other point of confusion is that I'm not sure where this resolution fits into this getting started guide, as I'm not aware that the steps listed in this page require setting up Ruby so I don't think you should experience the issue the fix is documented when following the steps.

I also think, should we had a good spot for rbenv tips, that we'd want to link directly to the rbenv docs on how to install: https://github.com/rbenv/rbenv#using-package-managers. As this instruction is slightly off (you don't need the path step) and the rbenv docs can handle other shells.

ChrisBAshton:
Hi @JamesCGDS, just doing a pre-Christmas tidy-up! 👋 

Did you see Kevin's comment above? Please respond if this is still in progress, or close the PR if you don't intend to merge it.

JamesCGDS:
Hi @ChrisBAshton / @kevindew - sorry yes I saw this comment before GIFT week and it slipped my mind. I'll address this today!

JamesCGDS:
Thanks for your comments @kevindew - I'll close this PR as I need to have a think about what the best alternative would be.

PR #3795, opened 2022-12-09
Update Content Security Policy for new reporter

kevindew:
This does an overall update of our Content Security Policy documentation, which hasn't really been changed since 2019. It's updated to reflect the current status of the work and the somewhat long history.

We now have a new report system [1] which means we can now query reports again, this includes documentation on how to query the reports.

[1]: https://github.com/alphagov/govuk-aws/pull/1664

### DIFF
--- a/source/manual/content-security-policy.html.md
+++ b/source/manual/content-security-policy.html.md
@@ -14,21 +14,74 @@ the browser.
 CSP can be run in two modes - *report only*, where violations of the policy are reported to a given endpoint but
 allowed to execute, and *enforcement*, where violations are blocked.
 
+## GOV.UK CSP History
+
+As of 2022, GOV.UK has been working, on and off, towards adding a CSP to the public www.gov.uk website for a number of
+years. We have configured one that has been running on the frontend applications, in report only mode, since 2019.
+We track this incomplete implementation as [tech debt](https://trello.com/c/lxxx5XLZ/178-govuk-has-a-half-implemented-content-security-policy-csp).
+
+We continue to aspire to have the CSP set to enforcement mode and intend to enhance it to forbid [`unsafe-inline`][unsafe-inline]
+scripts and styles (in applicable applications) before launching it.
+
+[unsafe-inline]: https://content-security-policy.com/unsafe-inline/
+
 ## How the policy is set
 
-The specific policy that is sent with a request to the browser as an HTTP header is defined in the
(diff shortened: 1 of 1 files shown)


source/manual/content-security-policy.html.md
```diff
@@ -14,21 +14,74 @@ the browser.
...
+
+As we receive high volumes of false positive alerts, it is likely we will remove the reporting functionality once
+a CSP is enforced.
+
+[lambda]: https://github.com/alphagov/govuk-aws/blob/main/terraform/lambda/CspReportsToFirehose/index.mjs
+
+### Querying violations
+
+Athena is available through the AWS control panel. To access, [log into AWS](/manual/get-started.html#sign-in-to-aws),
+navigate to [Athena](https://eu-west-1.console.aws.amazon.com/athena) and select the `csp_reports` database. The
```
KludgeKML, line 57: When you go to athena, there's a pre-query landing page where it took me a good 30 seconds to work out that I had to choose a query before I got to the point where I could choose "csp_reports" - possibly mention that?

(Also, on production this works fine, but on integration you can't run a query yet because "Before you run your first query, you need to set up a query result location in Amazon S3.")

kevindew: Ooh what do you get when you visit https://eu-west-1.console.aws.amazon.com/athena ? I get the query editor 

Yeah I'm trying to get the latter sorted in all environments with: https://github.com/alphagov/govuk-aws/pull/1668

KludgeKML: Ah, _now_ I get the query editor, but the first time I did it I got bounced to the Athena landing page (https://eu-west-1.console.aws.amazon.com/athena/home?region=eu-west-1#/landing-page). So it's obviously only a thing if you've never used Athena before (but I might argue most people suddenly reading this on 2L or wherever might not have used Athena before).

kevindew: Great thanks for sharing. I'll update.

kevindew: I've pushed [d5e5b55](https://github.com/alphagov/govuk-developer-docs/pull/3795/commits/d5e5b55366420c71079ee7686359a38daa42c816) which now uses an anchor to link directly to the query editor.

KludgeKML (approved):
LGTM

PR #3797, opened 2022-12-12
Rewrite docs around DNS management

ChrisBAshton:
Removes some of the duplication and out of date content.

### DIFF
--- a/source/manual/dns.html.md
+++ b/source/manual/dns.html.md
@@ -9,7 +9,7 @@ parent: "/manual.html"
 
 GOV.UK is responsible for managing several DNS zones.
 
-By default, zones are hosted by AWS (Route 53) and Google Cloud Platform (Cloud DNS).
+By default, zones are hosted by AWS (Route 53) and Google Cloud Platform (Cloud DNS). We use both for redundancy.
 
 As of December 2022, there are 61 hosted zones. A list is retrievable from a terminal using:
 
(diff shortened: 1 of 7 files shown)


source/manual/dns.html.md
```diff
@@ -31,80 +27,34 @@ We use a few domains:
...
-2. `apply` in each DNS provider until you see that your changes has been applied.
-    There are [circumstances where terraform is not able to apply the changes](#google-based-caveats)
-    in 1 run and need **multiple** runs.
+1. Make the changes in govuk-dns-config (see [example](https://github.com/alphagov/govuk-dns-config/pull/851))
+1. Deploy the changes via the [Deploy_DNS Jenkins job](https://deploy.blue.production.govuk.digital/job/Deploy_DNS/). You'll need production admin access to do this.
 
-Deployment is done by first obtaining (first time set up) your GitHub credentials
-by creating a read-only GitHub personal access token. This [GitHub personal access token](https://github.com/settings/tokens) should be created with the `read:org`
-scope only.
+Deploying is a multi-step process:
```
MartinJJones, line 40: Content looks good to me, I think we just need to bump up the list numbers however, all steps are currently showing as "1".

```markdown
1. Run a `plan` of the deployment, against the `aws` provider.
   - You can get the necessary AWS credentials by running `gds aws govuk-production-admin -e`.
2. Check that the output is what you expect.
3. Retrieve a Google OAuth access token.
   - You'll need the gcloud CLI (`brew install --cask google-cloud-sdk`).
   - Then run the following - you'll be prompted to login to your Google account to allow Google Cloud SDK access your Google Account. The token will be printed in the terminal:
   - `gcloud config set project govuk-production; gcloud auth login --brief; gcloud auth print-access-token`
   - Use this token for the `GOOGLE_OAUTH_ACCESS_TOKEN` field in the next step.
4. Now run a `plan` of the deployment against the `gcp` provider.
   - You'll still need to provide all the AWS credentials as per step 1. This is because the Terraform state is held in an S3 bucket.
5. Check that the output is what you expect.
   - It's normal to see changes in TXT records relating to escaping of quotes. You can safely ignore these if they don't change any of the content of the record. This is a bug in the way we handle splitting long TXT records between AWS and GCP in our [YAML -> Ruby -> Terraform process](https://github.com/alphagov/govuk-dns).
6. Finally, run an `apply` deployment for both `aws` AND `gcp`. (The order doesn't matter).
   - Sometimes, the GCP deployment requires multiple runs. This is because, in order to change a DNS record, the Google provider deletes and re-adds that record. This can cause a [race condition](https://github.com/alphagov/govuk-dns/issues/67) where Google tries to create the new one before it has successfully deleted the old one. In this case, the build will fail, and you just need to re-run the GCP `apply` job.
```

ChrisBAshton: Thanks Martin! That's deliberate, as GitHub's markdown parser will automatically increment the numbers (saves you having to rewrite all the numbers if you reorder or remove any of them)

MartinJJones: Thanks very much, I didn't know that 👍

MartinJJones (approved):
Looks good to me 👍

PR #3805, opened 2022-12-14
Improve instructions for editing changenotes

ChrisBAshton:
1. Expand on the Whitehall instructions. Though we shouldn't have to do this ourselves anymore, it's useful to know the feature exists.
2. Add Content Publisher instructions. It's not immediately clear that 'change notes' are a concept that exist in multiple publishing apps, so this serves as a useful reminder of that.
3. Frontload the warning about not making changes directly in Publishing API unless absolutely necessary. It's arguable that we should remove this section entirely, but it seems useful to know about the two different approaches to editing changenotes within Publishing API.

### DIFF
--- a/source/manual/howto-modify-change-note.html.md.erb
+++ b/source/manual/howto-modify-change-note.html.md.erb
@@ -10,8 +10,8 @@ Spelling mistakes can creep into [change notes](https://www.gov.uk/guidance/cont
 
 ## Whitehall
 
-Whitehall supports modifying change notes directly from the UI, though the feature is somewhat hidden from publishers.
-The publisher can modify the change note themselves by:
+Whitehall supports modifying change notes directly from the UI. The feature is hidden from publishers, but available to anyone with `GDS Admin` permissions.
+Content 2nd Line can modify the change note themselves by:
 
 * Appending `/change_notes` to the URL of the admin screen, so that the URL looks something like:
   `https://whitehall-admin.publishing.service.gov.uk/government/admin/editions/<edition-id>/change_notes`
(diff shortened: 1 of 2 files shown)


source/manual/howto-modify-change-note.html.md.erb
```diff
@@ -1,53 +1,43 @@
...
 parent: "/manual.html"
 ---
 
-Spelling mistakes can creep into [change notes](https://www.gov.uk/guidance/content-design/writing-for-gov-uk#change-notes), and we are often asked to correct them. The instructions below cover doing this task in the Publishing API and in Whitehall - if your change note lives in Whitehall and is updated in the Publishing API it could be overwritten later.
+Spelling mistakes can creep into [change notes](https://www.gov.uk/guidance/content-design/writing-for-gov-uk#change-notes), and we are often asked to correct them. The instructions below cover doing this task in Whitehall, Content Publisher and Publishing API.
 
 ## Whitehall
 
-To modify a change note in Whitehall, visit the following URL, replacing `<edition-id>` with the ID of the edition:
+Whitehall supports modifying change notes directly from the UI, though the feature is somewhat hidden from publishers.
```
brucebolt, line 13:
```suggestion
Whitehall supports modifying change notes directly from the UI. The feature is hidden from publishers, but visible to anyone with `GDS Admin` permissions.
```

source/manual/howto-modify-change-note.html.md.erb
```diff
@@ -1,53 +1,43 @@
...
+The publisher can modify the change note themselves by:
```
brucebolt, line 14:
```suggestion
Content 2nd Line can modify the change note themselves by:
```

brucebolt (approved):

PR #3809, opened 2022-12-19
Create new "Role specific policies" section

ChrisBAshton:
We have a few special rules depending on the role of the person on 2nd line. It seems helpful to consolidate them into their own section, and link to the new-and-improved flow diagram.

This PR also clarifies that apprentices should only be given Production Deploy access.


leenagupte (approved):
👍

PR #3810, opened 2022-12-19
Remove query result location instruction

kevindew:
This step is no longer needed since
https://github.com/alphagov/govuk-aws/pull/1668.
We have now created a bucket and configured it for the primary workgroup so queries run without extra set-up.


ChrisBAshton (approved):
Ooh this is super helpful - thank you!

PR #3812, opened 2022-12-20
Update docs for contract testing

BeckaL:
This is an attempt to clarify some confusions that existed in the docs, and present a clearer explanation of contract testing.

[Trello](https://trello.com/c/RMgXVXhO/164-publishing-api-contract-tests-are-different-to-all-other-repos)

### DIFF
--- a/source/manual/pact-testing.html.md
+++ b/source/manual/pact-testing.html.md
@@ -7,7 +7,14 @@ layout: manual_layout
 parent: "/manual.html"
 ---
 
-[Pact](https://docs.pact.io/) is a tool we use for *contract testing*. Contract testing involves creating a set of tests that are shared between an API (the "provider") and its users ("consumers") using some kind of "broker". For example, the Publishing API has a "pact" or "contract" with GDS API Adapters:
+[Pact](https://docs.pact.io/) is a tool we use for *contract testing*. Contract testing involves creating a set of tests that are shared between an API (the "provider") and its users ("consumers"). Contract tests allow consumers to define a contract that it expects its provider to uphold. This means that a consumer can specify what response (both status code and body) they expect from a provider given a certain http call.
+This is useful to ensure that changes to the provider do not break the interaction between it and its consumers.
+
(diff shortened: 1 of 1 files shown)


source/manual/pact-testing.html.md
```diff
@@ -93,8 +101,11 @@ Follow these steps in order to change the provider and consumer in tandem.
...
+As is noted above, GDS API Adapters is often used as the consumer for pact tests. This is because many api calls made within our system are made
+via GDS API Adapters. Therefore treating GDS API Adapters as the consumer saves us from needing to write pact tests for the same api call in multiple
+different applications.
 
-The tests are run in the same way as other consumers and providers (see above).
+When an application calls an API directly, not via GDS API Adapters, this application will be the consumer. For example, Publishing API
+is the consumer in the pact tests between it and Content Store, since it makes direct calls to Content store.
```
JonathanHallam, line 111: might be worth a link here to where this call is being made

source/manual/pact-testing.html.md
```diff
@@ -7,7 +7,14 @@ layout: manual_layout
 parent: "/manual.html"
 ---
 
-[Pact](https://docs.pact.io/) is a tool we use for *contract testing*. Contract testing involves creating a set of tests that are shared between an API (the "provider") and its users ("consumers") using some kind of "broker". For example, the Publishing API has a "pact" or "contract" with GDS API Adapters:
+[Pact](https://docs.pact.io/) is a tool we use for *contract testing*. Contract testing involves creating a set of tests that are shared between an API (the "provider") and its users ("consumers"). Contract tests allow consumers to define a contract that it expects its provider to uphold. This means that a consumer can specify what response (both status code and json body) they expect from a provider given a certain http call.
```
brucebolt, line 10:
```suggestion
[Pact](https://docs.pact.io/) is a tool we use for *contract testing*. Contract testing involves creating a set of tests that are shared between an API (the "provider") and its users ("consumers"). Contract tests allow consumers to define a contract that it expects its provider to uphold. This means that a consumer can specify what response (both status code and body) they expect from a provider given a certain http call.
```
Just a little pedantic note: the body can be any format, it doesn't necessarily need to be JSON.

brucebolt (approved):

PR #3814, opened 2022-12-21
Make auto-merger wait on auto-approval of PR before merging

ChrisBAshton:
Currently, the auto-merger waits for CI to pass, before attempting to merge. However, the merge fails, as PRs require at least one approving review before merge.

We've already established the conditions under which an auto review is sufficient: a patch upgrade of govuk_publishing_components, where the associated CI checks pass. Therefore we already approve of the PR, we just need a mechanism to make that happen. We're using https://github.com/hmarr/auto-approve-action to use a GitHub Actions 'bot' to approve the PR.

I've tested this approach in:
https://github.com/ChrisBAshton/test-auto-merge/pull/2

Trello: https://trello.com/c/uwoMBinS/3015-create-proof-of-concept-for-auto-merging-internal-prs-5

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -6,8 +6,8 @@ permissions:
   pull-requests: write
 
 jobs:
-  dependabot:
-    name: Auto merge qualifying Dependabot PR
+  auto_approval:
+    name: Auto approve qualifying PR
     runs-on: ubuntu-latest
     if: ${{ github.actor == 'dependabot[bot]' }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-auto-merge.yml
```diff
@@ -18,12 +18,22 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
       - name: Dependabot metadata
         id: metadata
         uses: dependabot/fetch-metadata@v1.3.4
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
       - name: Enable auto-merge for Dependabot PRs
+        # TODO: figure out how to de-duplicate the conditional below
```
ChrisBAshton, line 36: Was hoping to define a separate job (e.g. `check_qualifies`) with its own `if` that consolidates everything, and then make both the `auto_approval` and `dependabot` jobs have a `needs` dependency on the `check_qualifies` job.

Had a play with that approach [here](https://github.com/ChrisBAshton/test-auto-merge/pull/4) but didn't have much luck, so this feels like an optimisation we can make later.

MahmudH (approved):
Great! In the past the only way I found to auto-merge was to use the @dependabot merge command through a comment, I didn't know auto-approve-action was available. This looks like a way cleaner approach. 👍🏽

PR #3816, opened 2022-12-21
Stop building PRs twice

ChrisBAshton:
We trigger our CI on both push (to branches, irrespective of there being an associated PR) and on `pull_request` (which triggers on any push to a branch which has a PR open - not just when the PR is first opened).

Consequently, we run two CI builds for each push, which is unnecessary and can slow us down (as sometimes one build langushes behind the other quite noticeably).

There should be no disadvantage to disabling build-on-push to branches which have no associated PR, except that any lint failures won't be brought to light. In practice, many branches are created directly from the GitHub interface, and a PR created in the same transaction, so this would only affect developers who create docs branches locally. In such cases, developers can always run their linter locally too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,4 @@
 on:
-  push:
   pull_request:
   schedule:
     # Deploy hourly between 9am and 7pm on weekdays


catalinailie (approved):

robinjam:
Good idea, but maybe we should continue to allow CI on pushes to `main` (only)?
Otherwise the test suite won't run after a PR is merged (which might have ramifications if the PR wasn't rebased before merging)

kevindew:
You’ll lose the ability to have a CI build prior to opening a PR with this which can be annoying. It’s also a different build when its PR to push, the PR one builds against a merged branch whereas push is the exact commit.

There was a debate on the GitHub action RFC about this which led to the push and pull_request being our convention:  https://github.com/alphagov/govuk-rfcs/blob/main/rfc-123-github-actions-ci.md

I find myself torn on it. I’d really miss having pre PR CI builds but do also agree that the two builds can feel somewhat superfluous.

ChrisBAshton:
Ah, thanks for the comments, both. Perhaps I should have reached out more widely first 😬 
Will have a think - especially about re-enabling on pushes to `main`.

kevindew:
> Ah, thanks for the comments, both. Perhaps I should have reached out more widely first 😬 Will have a think - especially about re-enabling on pushes to `main`.



Great, we could do with some outreach to work out a convention again. Replatforming didn't utilise the existing convention which means are conventions are now quite inconsistent.

I imagine the push on main run is essential though.

With regards to all pushes, if it's just me who uses CI before opening a PR then seems fine to change, my workflow is obviously not > everyone elses. But I do worry that we don't actually know how many people do use CI as part of their pre-PR workflow.

PR #3826, opened 2022-12-30
Remove UK Cloud VPN from docs (and fix references to other VPNs)

richardTowers:
The UK Cloud VPN has been removed now, so we don't need the alert or the documentation describing what it does.

I decided to keep the VPN docs to note that the VPNs to UK Cloud and Carrenza are gone, because I don't expect everyone to realise they're gone, and it's a better experience searching for "VPN" and getting a "the VPN is gone" page than searching for "VPN" and getting nothing.

I've also fixed a bunch of links that should have been pointing to instructions on connecting to the GDS VPN. One of these was a 404, one pointed at the UK Cloud VPN, and the other pointed at a google doc that looked reasonable so I've standardised on that.


ChrisBAshton (approved):
LGTM 👍 agreed about keeping a note about the VPN that used to exist.

PR #3840, opened 2023-01-09
Update 'exception' text

ChrisBAshton:
This was reduced from two exceptions to one exception, in https://github.com/alphagov/govuk-aws/commit/430e8fd7d579694560f80ea11174140d8f966364


KludgeKML (approved):
LGTM

PR #3842, opened 2023-01-09
Upgrade ruby 3

leenagupte:


spec/pages/manual_page_spec.rb
```diff
@@ -1,10 +1,7 @@
 Dir.glob("source/manual/**/*.md").each do |filename|
   RSpec.describe filename do
     raw = File.read(filename)
-    # rubocop:disable Security/YAMLLoad
```
kevindew, line 4: Nice 👍

source/apps.html.md.erb
```diff
@@ -35,8 +35,8 @@ Apps can be loosely grouped into different types, including:
 
 | Hosting | Count | Applications |
 | --- | --- | --- |
-<% Hosts::hosters_descending(Repos.active_apps).each do |id, label:, apps:| %>
```
kevindew, line 38: Re your commit message, this is failing because Ruby 2.7 would expand a hash into the keyword arguments, whereas Ruby 3.1 won't and without something to convert the hash to kwargs.

Your change looks like an improvement in my opinion.

kevindew (approved):
Looks good.

source/apps.html.md.erb
leenagupte: Oh, thank you! That's makes sense.

PR #3844, opened 2023-01-10
Clarify that env sync pulls from Staging if data-sanitisation happens

ChrisBAshton:
In putting together a presentation on data backups strategy, I dug into this a bit and it appears as though [Whitehall Integration pulls from Staging](https://github.com/alphagov/govuk-puppet/blob/f9c6b136d058b6e0e20fba5d3716b36e60462e2f/hieradata_aws/class/integration/whitehall_db_admin.yaml#L12), where the [data has already been sanitised](https://github.com/alphagov/govuk-puppet/blob/f9c6b136d058b6e0e20fba5d3716b36e60462e2f/hieradata_aws/class/staging/whitehall_db_admin.yaml#L32). The current docs are a bit misleading in that they suggest Integration always pulls from Production directly.


brucebolt (approved):
This is also my understanding of the data sync.  Thanks for updating.

PR #3847, opened 2023-01-11
Add GTM change process

andysellick:
Adds a change process for analysts and developers making changes to Google Tag Manager on GOV.UK.

Trello card: https://trello.com/c/WdqsuhbE/231-create-gtm-change-process

### DIFF
--- a/source/manual/gtm-change-process.html.md
+++ b/source/manual/gtm-change-process.html.md
@@ -1,0 +1,61 @@
+---
+owner_slack: "#govuk-frontenders"
+title: 'Google Tag Manager change process'
+section: Frontend
+layout: manual_layout
+type: learn
+parent: "/manual.html"
+related_repos: [govuk_publishing_components]
+---
+
+GOV.UK has implemented Google Analytics 4 (GA4) using Google Tag Manager (GTM) to track user journeys through the site. This page describes the process of making changes to GTM to support improvements to our analytics.
+
+This process exists because some changes can increase the size of the JavaScript included from GTM in the public frontend of GOV.UK. We do not want to unnecessarily degrade the performance of GOV.UK.
+
+## Roles and responsibilities
+
+Performance analysts:
+
+- to make changes in GTM and share the change with a succinct explanation of what it does and why it's needed
+- to fully document the details of the change
+
+Developers:
+
+- to review and publish the change in GTM
+- to check that the changes don't increase the JavaScript size
+
+## Process
+
+Analyst creates a change in the GTM web interface.
+
+- change is created in a new workspace
+- change must include a clear description
+- change can be checked in preview mode (optional)
+
+Analyst gets a second analyst to check the changes.
+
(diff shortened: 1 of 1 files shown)


source/manual/gtm-change-process.html.md
```diff
@@ -0,0 +1,62 @@
...
+
+Performance analysts:
+
+- to make changes in GTM and share the change with a succinct explanation of what it does and why it's needed
+- to fully document the details of the change
+
+Developers:
+
+- to review and publish the change in GTM
+- to confirm that the changes won't break anything
```
gclssvglx, line 25: Just thinking, if we go the normal route: Integration -> Staging -> Production, is this step is necessary. Wouldn't any breaking change be caught in Integration? I guess it's worth keeping in, but I wouldn't expect anything to break in Staging or Production if it's ok in integration.

source/manual/gtm-change-process.html.md
```diff
@@ -0,0 +1,62 @@
...
+
+- change is created in a new workspace
+- change must include a clear description
+- change can be checked in preview mode (optional)
+
+Analyst gets a second analyst to check the changes.
+
+Analyst raises a request to have the change approved.
+
+- can be raised in the `#govuk-ga4` slack channel
```
gclssvglx, line 40: I wonder if we can ask for the notification to be put into #user-experience-measurement-govuk, or maybe use a 'here' in #govuk-ga4 instead? I know I've missed some of these requests in the past.

source/manual/gtm-change-process.html.md
```diff
@@ -0,0 +1,62 @@
...
+- `Activity history` shows previous changes
+- ensure no custom HTML or JS included in tags
+
+Developer approves and publishes the change
+
+- publish to the required environment - `Production`, `Staging` or `Integration`
+- publishing defaults to the `Production` environment, if it needs to be deployed to a different environment that must be selected first
+- publishing may involve more than one publishing act e.g. publish version 35 to `Production`, publish version 36 to `Integration`
+
+Analyst tests that change has been successful
```
gclssvglx, line 58: I wonder if it might be worth saying somehow that Integration -> Staging -> Production is the normal (and expected) path to Live. So, release 34 should not go from Integration then directly to Production for example. What do you think? Overkill?

gclssvglx (commented):
Great stuff @andysellick 👍 Beautifully written, succinct and simple explanation. Just have a couple of thoughts.

source/manual/gtm-change-process.html.md
andysellick: That's a good point, have changed the channel.

source/manual/gtm-change-process.html.md
andysellick: Yeah, tricky not to go into unnecessary detail, I've added a slight change, see what you think.

gclssvglx (approved):
Great stuff @andysellick - thanks for this 👍

PR #3848, opened 2023-01-11
Refine the docs around blocking traffic at edge

ChrisBAshton:
See commits for details.


sengi (approved):
Cool, this is looking great - thanks!

PR #3858, opened 2023-01-17
Update owner of govuk-browser-extension

ChrisBAshton:
This is best placed in the publishing team, as it is most used by content people to navigate to and from the content they're working on in publishing tools.


ollietreend (approved):
Looks good to me - thanks @ChrisBAshton

PR #3861, opened 2023-01-18
Improve DNS deploy docs

richardTowers:
Previously we didn't mention the gds cli automation we have that wraps triggering the Jenkins build. There's a minimum of four Jenkins builds required (plan aws, plan gcp, apply aws, apply gcp) to deploy a change, so this automation makes life immensely easier.

I've documented this, and made it clear it's the preferred option. I've also pulled out the manual steps into their own section, and simplified the deployment instructions (now that they're not interleaved with instructions about getting credentials).

I've also recommended the poweruser role instead of the admin role, as DNS changes don't touch IAM so they don't need admin.

### DIFF
--- a/source/manual/dns.html.md
+++ b/source/manual/dns.html.md
@@ -56,14 +56,7 @@ USAGE:
    gds govuk dns [command options] [arguments...]
 
 OPTIONS:
-   --provider value, -p value            DNS provider to use
-   --zone value, -z value                DNS zone for
-   --action value, -a value              action to apply. e.g. plan
-   --github-username value, -u value     GitHub username used to log into Jenkins, can be set via env variable GITHUB_USERNAME [$GITHUB_USERNAME]
-   --role value, -r value                Name of the AWS role used to get credentials
-   --assume-role-ttl value, --art value  Expiration time for the assumed role. Most roles are configured to permit up to 60m. (default: "60m")
-   --skip-mfa, --sm                      Don't ask for MFA (default: false)
-   --help, -h                            show help (default: false)
+...
 ```
 
 For example, to have Jenkins run a terraform plan of changes to service.gov.uk against AWS, run:
--- a/source/manual/dns.html.md
+++ b/source/manual/dns.html.md
@@ -14,7 +14,7 @@ By default, zones are hosted by AWS (Route 53) and Google Cloud Platform (Cloud 
 As of December 2022, there are 61 hosted zones. A list is retrievable from a terminal using:
 
 ```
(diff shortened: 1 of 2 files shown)


chao-xian:
Unofficial request for changes to add the Mouse Trap pic for illustration.

source/manual/dns.html.md
```diff
@@ -35,25 +35,75 @@ Note that we __do not__ manage any other DNS records: if you get a request conce
...
+
+OPTIONS:
+   --provider value, -p value            DNS provider to use
+   --zone value, -z value                DNS zone for
+   --action value, -a value              action to apply. e.g. plan
+   --github-username value, -u value     GitHub username used to log into Jenkins, can be set via env variable GITHUB_USERNAME [$GITHUB_USERNAME]
+   --role value, -r value                Name of the AWS role used to get credentials
+   --assume-role-ttl value, --art value  Expiration time for the assumed role. Most roles are configured to permit up to 60m. (default: "60m")
+   --skip-mfa, --sm                      Don't ask for MFA (default: false)
+   --help, -h                            show help (default: false)
```
chao-xian, line 66: I think an example of running the command would suffice as otherwise this would go out of date when the tool gets updated.

chao-xian (approved):

leenagupte (commented):
Looks great!  Since you're here, do you mind also updating the gcloud install instructions on t https://docs.publishing.service.gov.uk/manual/google-cloud-platform-gcp.html as well?

I believe it should be: 

```
brew install --cask google-cloud-sdk
```

richardTowers:
>  Unofficial request for changes to add the Mouse Trap pic for illustration.

For posterity: 
image

* The red slammer represents the gds-cli, which you invoke like `gds govuk dns ...`
* This kicks the yellow bucket, which represents a Jenkins job
* The jenkins job runs a ruby script, represented by the silver ball
* This script generates some terraform config, which is placed in a second yellow bucket, raised up, and placed in a red bathtub
* The ball drops, and the jenkins job runs terraform apply
* Finally, terraform (represented by the green man) makes the required API calls, diving into the final yellow bucket and releasing the mousetrap which actually updates the DNS config in our cloud providers
* (Then you have to do the whole thing again for the other cloud provider - this setup does AWS or GCP, but one at a time)

source/manual/dns.html.md
richardTowers: I've gone for a half way house and removed the OPTIONS section (the rest of the help should be reasonably stable, and somehow I find its presence here reassuring).

PR #3864, opened 2023-01-20
Add `accessible-autocomplete-multiselect`

DilwoarH:
This add `accessible-autocomplete-multiselect` repo to the list.

This is currently a fork of `accessible-autocomplete` which is maintained by the GOVUK Publishing Service team.

https://trello.com/c/iNuMxfk4/893-adopt-kevins-fork-of-accessible-autocomplete

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -1,3 +1,7 @@
+- repo_name: accessible-autocomplete-multiselect
+  type: Utilities
+  team: "#govuk-publishing-experience-tech"
+
 - repo_name: account-api
   type: APIs
   team: "#tech-interaction-and-personalisation"


data/repos.yml
```diff
@@ -1,3 +1,7 @@
+- repo_name: accessible-autocomplete-multiselect
+  type: Node module
```
ollietreend, line 2: It looks like you're introducing a new repo type here – I'm not sure if that's supported, or if we have to stick to the existing types. Do you know?

It looks like 'utilities' is the usual dumping ground for miscellaneous repos – including `miller-columns-element`:

https://github.com/alphagov/govuk-developer-docs/blob/40411b6a840f59e27c8c96ccaa22d1699b25a4bc/data/repos.yml#L881-L883

I think we should either:

1. introduce a 'Node module' repo type, and recategorise other repos in the process (e.g. miller-columns-element)
2. categorise this new accessible-autocomplete-multiselect repo as a utility

What do you think @DilwoarH?

ollietreend (commented):
Thanks for adding this. I just wonder about the 'type' key.

data/repos.yml
DilwoarH: I think to keep things simple; I'll just change it to utilities

ollietreend (approved):

PR #3874, opened 2023-02-02
Add 2nd line instructions for adding an HMRC manual

KludgeKML:
https://trello.com/c/5pJBXAQP/1784-documentation-for-hmrc-manuals-api-slug-add

### DIFF
--- a/source/manual/add-an-hmrc-manual.html.md
+++ b/source/manual/add-an-hmrc-manual.html.md
@@ -1,0 +1,11 @@
+---
+owner_slack: "#govuk-2ndline-tech"
+title: Add an HMRC Manual
+parent: "/manual.html"
+layout: manual_layout
+section: 2nd line
+---
+
+HMRC publish their own manuals on GOV.UK using the HMRC Manuals API, but new manuals can only be added once we have manually allowed the slug for the manual. When a ticket comes in to Zendesk asking for an HMRC manual to be published, it just means that we need to [add a new slug], which they will provide.
+
+[add a new slug]: https://github.com/alphagov/hmrc-manuals-api#adding-a-new-slug


ChrisBAshton (commented):
Thanks for this.

Rather than adding a new doc, is there a way we can better surface the [existing documentation](https://docs.publishing.service.gov.uk/repos/hmrc-manuals-api.html#adding-a-new-slug)? E.g. was there something you searched for that didn't surface the application README in the search results?

(Spotted an issue #3875 as part of this review, by the way!)

KludgeKML:
I searched for HMRC Manuals, which does bring up the application readme and a readme about the API, I looked through the API documentation. I think I'm just not expecting application README.md's to have zendesk task documentation? (this might be the only one?) Maybe there's a way to make it more obvious, though? (it might also just be me? If you think this is unnecessary, feel free to close).

source/manual/add-an-hmrc-manual.html.md
```diff
@@ -0,0 +1,11 @@
...
+owner_slack: "#govuk-2ndline-tech"
+title: Add an HMRC Manual
+parent: "/manual.html"
+layout: manual_layout
+section: 2nd line
+---
+
+HMRC publish their own manuals on GOV.UK using the HMRC Manuals API, but new manuals can only be added once we have manually allowed the slug for the manual. When a ticket comes in to Zendesk asking for an HMRC manual to be published, it just means that we need to add a new slug (which they will provide).
+
+[adding-a-new-slug]: https://github.com/alphagov/hmrc-manuals-api#adding-a-new-slug
```
ChrisBAshton, line 11: You're not currently using this link anywhere 👀

ChrisBAshton (approved):
Given we only have 2 'proper' search results for HMRC Manuals, and that this is a once-a-year task that's often a bit vague/forgotten, I guess the more docs the better 👍 

Please add the link to your paragraph somewhere, then this is good to go.

PR #3879, opened 2023-02-13
Move deploy job to separate workflow

theseanything:
This moves the deploy job to a separate workflow that specific to deploying the site. It was unclear the purpose of the CI workflow as it did both testing and deployments. This makes the scope of each workflow well defined and more understandable.

This PR has no functional change.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,49 @@
+name: Deploy
+
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+    branches: [main]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    permissions:
+      contents: write 
+    steps:
+      - name: Set commit message up front
+        id: commit_message_writer
+        run: | # `github.event.number` will be blank if this is a cron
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,48 @@
+name: Deploy
+
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+    branches: [main]
+
+jobs:
+  deploy:
```
ChrisBAshton, line 10: You may need to add a [`on-success:` property](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow).

If someone manages to merge broken code into `main` (not that they should be able to), we could potentially generate a bad output and deploy something completely broken to the user-facing Developer Docs. So it's better to check that the tests pass 🤞

ChrisBAshton (commented):
Thanks, this feels much cleaner! One comment below:

.github/workflows/deploy.yml
ChrisBAshton: @theseanything this convo has been resolved but I don't see any changes - am I missing something?

ChrisBAshton: Thanks, makes sense

ChrisBAshton (approved):

PR #3884, opened 2023-02-16
Display alternative versions of page for EKS platform

theseanything:
This PR enables a new feature that allows us to write alternative versions of a page. This is useful for allowing us to maintain new documentation for the new EKS platform along side our existing docs. For example, we can have two versions of "how to deploy the emergency banner", one for EC2 and another for EKS. These alternative versions of page are hidden from site navigation features (such as search and manual lists), so only a single result is displayed per manual. Navigation to alternative pages are provided by a selector adjacent the breadcrumbs, and is only visible if an alternative version is available.

This is feature is provided through the use of a Middleman extension and additional partial - so should be trivial to remove if no longer required (i.e. post migration to the new platform.) 

For example of a page using this feature see PR #3885:

Search results showing a single result (instead of duplicates):

image

Manual lists showing a single link:

image

Page version selector:

image

Alternative page selected:

image

Page without multiple versions (No selector):

image

### DIFF
--- a/app/manual.rb
+++ b/app/manual.rb
@@ -34,6 +34,7 @@ class Manual
     return [] if current_page.data.subsection.nil?
 
     sitemap.resources
+      .reject { |page| page.data.index == false }
       .select { |page| page.data.section == ICINGA_ALERTS }
       .select { |page| page.data.subsection == current_page.data.subsection }
       .sort_by { |page| page.data.title.downcase } - [current_page]
@@ -47,6 +48,7 @@ private
 
   def manual_pages
     sitemap.resources
+      .reject { |page| page.data.index == false }
       .reject { |page| page.data.section == ICINGA_ALERTS }
       .select { |page| page.path.start_with?("manual/") && page.path.end_with?(".html") && page.data.title }
       .sort_by { |page| [page.data.type || "how to", page.data.title.downcase] }
--- a/app/page_versions.rb
+++ b/app/page_versions.rb
@@ -1,0 +1,64 @@
+# PageVersions extension allows you to write alternative versions of a page and
+# have them hidden from site navigation features (such as search and page
+# lists).
+#
+# Page versions use a attribute in the frontmatter to specify the version. This
+# attribute is set using the version_attribute option. The default_version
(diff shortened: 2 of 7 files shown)


ChrisBAshton (commented):
Hey! 👋 This is pretty cool - but my hunch is that it's a bit over-engineered and comes with some downsides that might not be worth the feature.

In maintaining two separate document versions, we're [duplicating a lot of content](https://github.com/alphagov/govuk-developer-docs/pull/3885/files#diff-2a951dfa80ccafd8cf4ceefc422aad96251e2dbfef9882a83b14fff2005bed98R2-R43), which we'll have to try to remember to keep in sync for as long as we have two versions available.

Then, presumably, the EKS version is the one that will win out, and the EC2 version will be deleted. But that means if we want to do a `git blame` on any particular line of documentation, e.g. to find out why or when a certain instruction was added, we have to then jump through the extra hoop of finding the deleted file and `git blame`-ing that.

There's a bit of an assumption that developers are going to see the `EC2 | EKS` links at the top right - I think it could be easily missed, and cause confusion. I'm also not convinced that hiding the EKS content from search is a good idea - I assume if there's a particular keyword in the EKS version that is missing from the EC2 version, then searching for that word won't surface either the EKS or EC2 doc?

A more pragmatic approach, IMO, would be to document the EC2/EKS variants inline, as I've attempted to do in #3887, and keep a list of all the PRs where you apply this treatment, so that it's easier to go in later and remove the EC2 references once Replatforming has happened.

Sorry not to be more enthused about this feature! The code for it looks good (albeit lacking some tests), it's mostly the risk of content maintenance that I'm concerned about.

theseanything:
> Sorry not to be more enthused about this feature! The code for it looks good (albeit lacking some tests), it's mostly the risk of content maintenance that I'm concerned about.

Not a problem! And thanks for having a look! Was just experimenting with a way to present the EKS documentation along side the existing doc. I agree the content maintenance is a worry, originally was thinking this feature would probably only be used for month or so whilst we migrate, then switched off. Also didn't think of the commit history, and defo would be a pain to loose the trails.

PR #3891, opened 2023-02-20
Update "Reviewing Dependabot PRs" guidance

ChrisBAshton:
This is an attempt to bring the guidance up to date, and to better reflect the actual working practices of GOV.UK developers. See [RFC-156](https://github.com/alphagov/govuk-rfcs/pull/156) for details of the informal survey we ran.

Our review process is (and should be) different when comparing major and patch versions of a dependency, and when reviewing a Rails core dependency bump vs some obscure library we may have included in Whitehall 8 years ago.

This commit documents the minimum review steps each PR requires, and also details how you can perform a more in depth review when it is needed.

It also reflects a change in policy around auto-merging, which we expect to enable on 'internal' dependency PRs in the coming weeks.

PS the documentation steps were based on me looking at two different Dependabot PRs ([1](https://github.com/alphagov/govuk-developer-docs/pull/3890), [2](https://github.com/alphagov/govuk-developer-docs/pull/3889)), which have different expandable sections and links. I think what I've written can be followed for both.

### DIFF
--- a/source/manual/manage-ruby-dependencies.html.md
+++ b/source/manual/manage-ruby-dependencies.html.md
@@ -20,17 +20,26 @@ Dependabot updates occur relatively soon after a new version is published, which
 there’s a risk of updating to a rogue version. Some updates also contain breaking
 changes, irrespective of when they are published.
 
-When reviewing a Dependabot PR you should do the following:
+When reviewing a Dependabot PR you should do the following as a minimum:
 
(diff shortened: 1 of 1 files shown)


sihugh (approved):
Looks good as an enabler of auto-merge for internal dependencies 👍 

Would be worth keeping an occasional eye on real-world behaviour if we can (your survey was excellent). Feels like there might be tools (if not now, then soon) to flag up things like `eval` to make it easier to review.

PR #3910, opened 2023-03-08
Added workaround for host key problems on Ventura

kornysietsma:
Also added a clarification around setting USER as this bites a lot of people at the same point.

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -243,6 +243,26 @@ Test your SSH access by running:
 gds govuk connect --environment integration ssh backend
 ```
 
+If you see an error `Permission denied`, check the message shown later, similar to: `The SSH username used was: jsmith` - if this is not the user you specified in the puppet config above, you need to specify a username:
+
+```bash
+USER=jaysmith gds govuk connect --environment integration ssh backend
+```
+
+(or you can `export USER=jaysmith` separately to set it for a shell session)
+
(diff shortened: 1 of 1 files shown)


MuriloDalRi (approved):

source/manual/get-started.html.md
```diff
@@ -243,6 +243,26 @@ Test your SSH access by running:
 gds govuk connect --environment integration ssh backend
 ```
 
+If you see an error `Permission denied`, check the message shown later, similar to: `The SSH username used was: jsmith` - if this is not the user you specified in the puppet config above, you need to specify a username:
+
+```bash
+USER=jaysmith gds govuk connect --environment integration ssh backend
+```
+
+(or you can `export USER=jaysmith` separately to set it for a shell session)
```
ChrisBAshton, line 252: The above does duplicate some docs we have [here]https://github.com/alphagov/govuk-docker/blob/main/docs/troubleshooting.md#binsetup-fails-to-succeed) - not the end of the world, but it's nice to not create multiple places where we have to maintain docs! I wonder if we can consolidate these docs so that they're in the most appropriate location and easy to search for?

richardTowers:
I don't think it's worth it because of the relatively imminent move to k8s, but we could hardcode this option inside govuk-connect and avoid having to document it.

In https://github.com/alphagov/govuk-connect/blob/main/lib/govuk_connect/cli.rb#L283-L284, something like

```diff
+      "-o",
+      "HostKeyAlgorithms=+ssh-rsa",
+      "-o",
+      "PubkeyAcceptedAlgorithms=+ssh-rsa",
```

PR #3925, opened 2023-03-23
Update locations of apps and repos

theseanything:
This add support to show which apps are now primarily hosted in EKS.

It also fixes links to deployments and Grafana dashboards from the application page.


app/repo.rb
```diff
@@ -122,7 +134,13 @@ def deploy_url
   def dashboard_url
     return if repo_data["dashboard_url"] == false
 
-    repo_data["dashboard_url"] || "https://grafana.production.govuk.digital/dashboard/file/#{repo_name}.json"
+    default_url = if production_hosted_on_eks?
+                    query_string = argo_cd_apps.map { |app| "var-app=#{app}" }.join("&")
+                    "https://grafana.eks.production.govuk.digital/d/000000111?#{query_string}"
```
sengi, line 139: Oof, that reminds me we should give these dashboards permanent IDs (but don't worry about this right now!)

sengi (approved):
Thanks!

app/repo.rb
theseanything: Ah good to know that might be possible - these urls seemed brittle when I saw them using a dashboard ID.

PR #3929, opened 2023-03-28
Change who can review changes in get started docs

brucebolt:
The Get Started documentation implies that only tech leads or 2nd Line can review certain changes. This is not the case as it can be done by anyone with production access.

Updating the documentation to reflect this, since developers in the new starter's team are better placed to do this than 2nd Line or #govuk-developers.


ChrisBAshton (approved):
Thanks!

PR #3930, opened 2023-03-28
Updating ruby pages to reflect replatforming

KludgeKML:
- Shrink scope of where we use rbenv
- Remove obsolete information about adding new ruby versions to Puppet
- Expand a little more on changes needed if using bulk-changer
- Add link to how to get logs in EKS
- Remove item about hard-restarting (note needed in EKS)

Rendered view at: https://github.com/alphagov/govuk-developer-docs/blob/updating-ruby/source/manual/ruby.html.md


sengi (approved):
Fab, thank you!

PR #3936, opened 2023-04-06
Add analytics documentation

gclssvglx:
This change converts the existing [GA4 Implementation record](https://alphagov.github.io/govuk-google-analytics/) to be a sub-section of the GOV.UK developer docs. 

The original repository was held in this [GitHub repo](https://github.com/alphagov/govuk-google-analytics), was written using [Jekyll](https://jekyllrb.com/), and was hosted via [GitHub Pages](https://pages.github.com). 

Why convert? There are 3 main reasons:

* The team and other interested parties have decided to  make the implementation record the public face of GOV.UK's GA4 documentation.

* We should not use the [GOV.UK](http://gov.uk/) crown, unless we are hosting on a .gov.uk domain.

* We want the deployment of this documentation to be as simple as possible, and as much content as possible to be contained in easily edited [Markdown](https://en.wikipedia.org/wiki/Markdown) files.

For these reasons, adding the implementation record to the existing GOV.UK developer docs, and making it a sub-section of that documentation is the preferred option.

As part of the conversion process, it was noted that the existing developer docs links append to each page by the [core layout](https://github.com/alphagov/tech-docs-gem/blob/main/lib/source/layouts/core.erb) for navigation and other purposes, would be likely to cause the user confusion. This is because they would be taken away from the analytics pages and back to the GOV.UK developer documentation. Having explored the options for overriding, or removing these links (including the creation of a [custom extension](https://github.com/grahamcharleslewis/analytics_config_switcher) for evaluation purposes) - it was decided that simplest and most pragmatic solution would be to simply supply a custom layout that does not extend from the [layouts in the tech-docs-gem](https://github.com/alphagov/tech-docs-gem/tree/main/lib/source/layouts). We have however, endeavoured to use as many elements from the original [core layout](https://github.com/alphagov/tech-docs-gem/blob/main/lib/source/layouts/core.erb) as possible (meta tags, styling etc). Inevitably, this will cause a slight maintenance overhead, but is seen as the least hackiest approach.

[Trello](https://trello.com/c/v7EXVpba/424-get-the-implementation-record-online)

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -57,3 +57,23 @@ unless ENV["SKIP_PROXY_PAGES"] == "true"
     proxy resource[:path], resource[:template], resource[:frontmatter]
   end
 end
+
+# Configuration for Analytics pages
+ignore "analytics/templates/*"
+
+page "analytics/*", layout: :analytics_layout
+
+data.analytics.events.each do |event|
+  event_name = event["name"].downcase.gsub(" ", "_")
+  proxy "analytics/event_#{event_name}.html", "analytics/templates/event.html", locals: { event: }
+end
+
+data.analytics.attributes.each do |attribute|
+  attribute_name = attribute["name"].downcase.gsub(" ", "_")
+  proxy "analytics/attribute_#{attribute_name}.html", "analytics/templates/attribute.html", locals: { attribute: }
+end
+
+data.analytics.trackers.each do |tracker|
+  tracker_name = tracker["name"].downcase.gsub(" ", "_")
+  proxy "analytics/tracker_#{tracker_name}.html", "analytics/templates/tracker.html", locals: { tracker: }
+end
--- a/helpers/analytics_helpers.rb
+++ b/helpers/analytics_helpers.rb
@@ -1,0 +1,38 @@
+module AnalyticsHelpers
+  def urlize(string)
+    string.downcase.gsub(" ", "_").gsub("-", "_")
+  end
+
+  def build_event(data, attributes = [], output = {})
+    data.sort_by { |k, _v| k["name"] }.each do |item|
+      name = item["name"]
+      value = item["value"]
+      if value
+        case value
+        when String
+          output[name] = value
+        when Array
+          output[name] = build_event(value, attributes)
+        end
+      else
+        attribute = attributes.find { |x| x["name"] == name }
+        output[name] = attribute ? attribute["type"] : value
+      end
+    end
+    output
+  end
+
+  def to_html(hash, html = "")
+    html += "<ul class='govuk-list indented-list'>"
+    hash.each do |key, value|
+      case value
+      when String
+        html += "<li><a href='/analytics/attribute_#{urlize(key)}.html'>#{key}</a>: #{value}</li>"
+      when Hash
+        html += "<li>#{key}: #{to_html(value)}</li>"
+      end
+    end
+    html += "</ul>"
+    html
+  end
+end
--- a/source/analytics/templates/event.html.erb
+++ b/source/analytics/templates/event.html.erb
@@ -1,0 +1,43 @@
+<h1 class="govuk-heading-l">
+  <span class="govuk-caption-l">Event</span>
+  <%= event.name %>
+</h1>
+
+<% if event.description %>
+  <p class="govuk-body"><%= event.description %></p>
+<% end %>
+
+<% event.events.each do |page_event| %>
+  <div class="event">
+    <h2 class="govuk-heading-m">
+      <%= page_event.name %>
+    </h2>
+    <% if page_event.description %>
+      <p class="govuk-body">
+        <%= page_event.description %>
+      </p>
+    <% end %>
+    <ul class="govuk-list govuk-list--bullet">
+      <% if page_event.example_url %>
+        <li>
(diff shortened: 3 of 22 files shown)


helpers/analytics_helpers.rb
```diff
@@ -0,0 +1,42 @@
...
+  def to_html(hash, html = "")
+    html += "<ul class='govuk-list indented-list'>"
+    hash.each do |key, value|
+      case value
+      when String
+        html += "<li><a href='/analytics/attribute_#{urlize(key)}.html'>#{key}</a>: #{value}</li>"
+      when Hash
+        html += "<li>#{key}: "
+        html += to_html(value)
+        html += "</li>"
```
andysellick, line 36: Non-blocking: you might be able to use `content_tag` to reduce some of the lines here, although from reading the docs I'm not certain if it returns text, you might have to change the whole approach, in which case don't worry.

```Ruby
html += content_tag(:li, to_html(value))
```

source/analytics/templates/event.html.erb
```diff
@@ -0,0 +1,43 @@
...
+  <%= event.name %>
+</h1>
+
+<% if event.description %>
+  <p class="govuk-body"><%= event.description %></p>
+<% end %>
+
+<% event.events.each do |page_event| %>
+  <div class="event">
+    <h3 class="govuk-heading-m">
```
andysellick, line 12: The heading before this is a H1, so this should be a H2 (and the next one a H3, not a H4). Sorry, should have noticed this when I was pushing changes.

andysellick (commented):
Looks great to me, although I don't claim to understand all of the backend complex stuff.

gclssvglx (commented):
Looks great - thanks @andysellick 👍

config.rb
```diff
@@ -57,3 +57,25 @@ def sanitize(contents)
     proxy resource[:path], resource[:template], resource[:frontmatter]
   end
 end
+
+# Configuration for Analytics pages
+ignore "analytics/templates/*"
+
+page "analytics/*", layout: :analytics_layout
+
+# rubocop:disable Style/HashSyntax
```
richardTowers, line 66: Why do we need this comment?

helpers/analytics_helpers.rb
```diff
@@ -0,0 +1,42 @@
+module AnalyticsHelpers
+  def urlize(string)
+    string.downcase.gsub(" ", "_").gsub("-", "_")
+  end
+
+  def build_event(data, attributes = [], output = {})
+    data.sort_by { |k, _v| k["name"] }.each do |item|
+      if item["value"]
```
richardTowers, line 8: Might be worth pulling out `item["name"]` and `item["value"]` into locals for this block, since they're both referenced a few times.

helpers/analytics_helpers.rb
```diff
@@ -0,0 +1,42 @@
...
+        case item["value"]
+        when String
+          output[item["name"]] = item["value"]
+        when Array
+          output[item["name"]] = build_event(item["value"], attributes)
+        end
+      else
+        attribute = attributes.select { |x| x["name"] == item["name"] }
```
richardTowers, line 16: Because you always call `.first` on the result, you could replace this `.select` with `.find` and save yourself the calls to `.first`

spec/helpers/analytics_helpers_spec.rb
```diff
@@ -0,0 +1,179 @@
...
+
+  describe "#urlize" do
+    it "returns a url safe version of the given string" do
+      expect(helper.urlize("An event-name")).to eq("an_event_name")
+    end
+  end
+
+  describe "#build_event" do
+    it "returns an ordered hash given an array of un-ordered hashes" do
+      actual = [
```
richardTowers, line 15: I would call this `input` instead of `actual`. I always think "actual" means "the actual result", which you'd get if you did

```
actual = helper.build_event([...
```

and then `expect(actual).to eq(expected)`

richardTowers (approved):
I like the overall approach - I think the custom layout is the least hacky of the ways we discussed, and the maintenance overhead of one custom layout is much much lower than a whole separate static site, so 👍🏻 

I couldn't resist a few nitpicks, which you can feel free to ignore. I think I've reviewed it closely enough to ✅ , but feel free to ask for another review if you think it would be good to have someone more experienced with dev docs look at it.

config.rb
gclssvglx: Ah, that sets the layout for all files under `/analytics` - I feel this is better, safer and easier than having to add the `layout` metadata to each page. 

I'll add a comment to the commit to make it clear why I'm adding this.

richardTowers: Sorry, I meant "why do we need rubocop:disable Style/HashSyntax?"

gclssvglx: Oh, sorry! I don't believe Middleman is Ruby 3.2 friendly, so unsure (if removed and Rubocop fixed up) this will work - I'll give it a try...

gclssvglx: Nope you're right we don't need this. Great spot - thanks @richardTowers 👍

helpers/analytics_helpers.rb
gclssvglx: Thanks @andysellick - I've cleaned this up a bit.

PR #3945, opened 2023-04-12
Remove references to run Rake tasks on Jenkins

theseanything:
We should be running rake tasks on EKS now, and these links to Jenkins may be confusing.

### DIFF
--- a/source/manual/alerts/email-alerts-travel-medical.html.md
+++ b/source/manual/alerts/email-alerts-travel-medical.html.md
@@ -69,11 +69,13 @@ investigate was the email sent. Some avenues to explore are:
 
 ## Resending travel advice emails
 
-If you need to force the sending of a travel advice email alert, there
-is a rake task in Travel Advice Publisher, which you can run using
-[this Jenkins job][resend travel advice job] where the edition ID of the
-travel advice content item can be found in the URL of the country's edit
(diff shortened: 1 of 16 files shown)


ChrisBAshton:
It might create some chore work here, but I think it would be useful to add a link to the [Run a rake task](https://docs.publishing.service.gov.uk/manual/running-rake-tasks.html) docs in every case that used to link directly to a Jenkins job. It's still an extra step or two compared to the old process, but at least devs wouldn't have to scramble to remember how to run rake tasks on the new platform? 🙏 

(Also worth moving the EKS instructions on that page above the Jenkins ones, and marking the Jenkins ones as deprecated - but probably still worth keeping the Jenkins instructions around for as long as it's technically still possible to run them?)

sengi (approved):
+1 to Chris's suggestions but I'd further suggest implementing them in a later PR and getting this in sooner, as-is. Perfect is the enemy of good.

We should just have a macro for the run-a-rake-task examples (in fact I think we already do; we just don't use it consistently — [example here](https://github.com/alphagov/govuk-developer-docs/blob/0f000cb/source/manual/reindex-elasticsearch.html.md.erb#L70)).

theseanything:
Will do a follow up PR to implement your suggestions @ChrisBAshton

PR #3954, opened 2023-04-14
Update database restore docs

chao-xian:
To check for a newly created database via RDS instead of Route 53 because it's easier and makes more sense in this context.


theseanything (approved):
Huh - weird that was our original suggestion.

PR #3961, opened 2023-04-18
Proposal on consistent approach for running GitHub actions

kevindew:
This has been produced following the discussiona in: https://github.com/alphagov/govuk-developer-docs/pull/3950 and https://github.com/alphagov/govuk-developer-docs/pull/3816

This is to try resolve a consistency issue we have in GOV.UK where we have an agreed approach to using GitHub Actions documented which doesn't reflect the configurations used across GOV.UK.

Since we agreed the approach of running CI GitHub Action on `[push, pull_request]` and adopted it has become clear that this is not unanimously popular and, due to the appearance of duplicate jobs, can imply misconfiguration.

The motivation to document a new approach is to try form consensus on an approach that can be good enough for most cases and cover common concerns. This can then reduce the time we spend debating how to configure individual projects.

Aspects I've wanted to cover in this are:

- Stay broadly consistent with approach replatforming popularised since that is what is predominantly in use
- Remove the appearance of duplicate runs by default in pull requests
- Allow CI for external contributors (our [current process for dealing with them][1] seems more of a legacy Jenkins concern)
- Offer an approach for running CI without opening a PR, as that seems to be a common point of contention (it's a key part of some workflows, yet not a concern for others).

A risk of this approach is that the use of workflow_dispatch is somewhat ignored and thus unreliable. This doesn't seem a huge concern to me as I imagine this would be fixed if it's used frequently or be a sign that it can be removed.

[1]: https://docs.publishing.service.gov.uk/manual/merge-pr.html#a-change-from-an-external-contributor

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,15 @@ name: CI
 
 on:
   push:
+    branches:
+      - main
   pull_request:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: The branch, tag or SHA to checkout
+        default: main
+        type: string
   schedule:
     # Deploy hourly between 9am and 7pm on weekdays
     - cron: "0 9-19 * * 1-5"
@@ -12,6 +20,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ inputs.ref || github.ref }}
       - name: Clone publishing api
         uses: actions/checkout@v3
         with:
--- a/source/manual/test-and-build-a-project-with-github-actions.html.md
+++ b/source/manual/test-and-build-a-project-with-github-actions.html.md
@@ -27,21 +27,64 @@ You should name the file used for configuring the CI workflow `ci.yml`.
 
 ### When the CI workflow should run
 
-The workflow should be configured to run on [`push`][push-event] and
-[`pull_request`][pull-request-event] events. The `push` event occurs for every
-push to the repository and this will generate a build against the branch
-that was pushed. The `pull_request` event occurs whenever a pull request
-is opened and it is triggered again any time the pull request changes (such
-as extra commits or rebasing), these differ to the `push` event by merging in
-the main branch before running operations. Pull requests from forks of a
-repo will only trigger the `pull_request` event.
+The workflow should be configured to run when there is a [`push`][push-event]
+to the default branch (typically "main") and when a
+[`pull_request`][pull-request-event] opened. This means CI runs when a branch
+is merged (the `push` to main) and it runs against any changes introduced
+in a pull request. We prefer to run against a `pull_request` event to all
+`push` events, as this allows external contributors to have pull requests
+tested and because `pull_request` runs against a version of the codebase
+[that is merged with the repository default branch][merge-comment].
 
-For example:
+Example configuration:
 
 ```yml
 # ./github/workflows/ci.yml
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
 ```
+
+To run CI on-demand, outside a pull request, for example before opening a PR,
+you may configure a repository to have a `workflow_dispatch` event so you can
+run it from the GitHub Actions user interface. If you do this you will need
+to configure the checkout action to reference the appropriate commit.
+
+Example configuration:
+
+```yml
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'The branch, tag or SHA to checkout'
+        default: main
+        type: string
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ inputs.ref || github.ref }}
+```
+
+We previously recommended CI jobs to run on `[push, pull_request]` as per
+[RFC-123][] however this caused duplicate runs of the CI jobs
+of a pull request (one for the `push` another for the `pull_request`)
+which were both superfluous for our testing, a point of confusion and were
+depleting our allowance quotas.
+
+[merge-comment]: https://github.com/alphagov/govuk-developer-docs/pull/3961#discussion_r1171071337
+[RFC-123]: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-123-github-actions-ci.md#findings-from-some-initial-explorations-into-using-github-actions
 
 ### Avoid superfluous naming of workflows, jobs and steps
 


sengi:
Just to add a bit of extra background: the practical motivation for eliminating the duplicate runs is that in at least one repo (govuk-developer-docs) they've been exacerbating flakiness because of expensive tests exhausting quotas and this has been a non-negligible time sink for Replatforming team because we're trying to be helpful and make lots of updates to existing documentation in that repo.

.github/workflows/ci.yml
```diff
@@ -2,7 +2,16 @@ name: CI
 
 on:
   push:
+    branches:
+      - main
+      - ci-proposal
```
sengi, line 7: I'm guessing the idea would be to remove this line before merge, once we're happy that it works?

source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -27,22 +27,64 @@ You should name the file used for configuring the CI workflow `ci.yml`.
...
-push to the repository and this will generate a build against the branch
-that was pushed. The `pull_request` event occurs whenever a pull request
-is opened and it is triggered again any time the pull request changes (such
-as extra commits or rebasing), these differ to the `push` event by merging in
-the main branch before running operations. Pull requests from forks of a
-repo will only trigger the `pull_request` event.
+The workflow should be configured to run when there is a [`push`][push-event]
+to the default branch (typically "main") and when a
+[`pull_request`][pull-request-event] opened. This will mean that there is a
+CI run when a branch is merged (the `push` to main) and it will run against
```
sengi, line 33:
```suggestion
[`pull_request`][pull-request-event] opened. This means CI runs when a branch is merged (the `push` to main) and it runs against
```

source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -27,22 +27,64 @@ You should name the file used for configuring the CI workflow `ci.yml`.
...
+  push:
+    branches:
+      - main
+  pull_request:
+
 ```
 
+To have CI runs be able to run outside of a pull request, for example before a
+pull request is opened, you may configure a repository to have a
+`workflow_dispatch` event so it can be ran from the GitHub Actions user
```
sengi, line 53:
```suggestion
To run CI on-demand or outside of a pull request, for example before a
pull request is opened, you may configure a repository to have a
`workflow_dispatch` event so you can run it from the GitHub Actions user
```

source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -27,22 +27,64 @@ You should name the file used for configuring the CI workflow `ci.yml`.
...
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ inputs.ref || github.ref }}
+```
+
+We previously recommended CI jobs to run on `[push, pull_request]` as per
+[RFC-123][] however this caused an appearance of duplicate runs of a CI job
```
sengi, line 82: It definitely runs real duplicate jobs, unfortunately.
```suggestion
[RFC-123][] however this caused duplicate runs of a CI job
```

source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -27,22 +27,64 @@ You should name the file used for configuring the CI workflow `ci.yml`.
...
+of a pull request (one for the `push` another for the `pull_request`)
+which were both superfluous for our testing and a point of confusion.
```
sengi, line 84:
```suggestion
which were both superfluous for our testing and a point of confusion, as well as wasting resources and in some cases exacerbating quota issues caused by expensive tests.
```

sengi (approved):
Thanks! This looks good to me 👍

Definitely worth getting @theseanything's input as well.

sengi:
Happy to help with rolling this out btw!

.github/workflows/ci.yml
```diff
@@ -2,7 +2,16 @@ name: CI
...
   pull_request:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'The branch, tag or SHA to checkout'
+        required: false
```
theseanything, line 13: Nit: [think require = false is the default](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputsinput_idrequired) anyway so could omit. 

might be worth be explicit about what the default is:
```suggestion
        default: main
```

theseanything (commented):
Looks good - thanks for updating this!

theseanything (approved):

source/manual/test-and-build-a-project-with-github-actions.html.md
kevindew: I've put a more succinct, and slightly softer, phrasing in.

kevindew:
Thanks @theseanything and @sengi, I think I've got all your suggestions in now, do shout if I've got any wrong.

I'll go ask in #govuk-developers

robinjam (approved):
Love this, especially the addition of the `workflow_dispatch` trigger to allow folks run CI on their branch without raising a PR (if they wish).

source/manual/test-and-build-a-project-with-github-actions.html.md
ChrisBAshton: Worth updating the PR description too, as it still references "appearance of", @kevindew

.github/workflows/ci.yml
```diff
@@ -2,7 +2,15 @@ name: CI
 
 on:
   push:
+    branches:
+      - main
   pull_request:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'The branch, tag or SHA to checkout'
```
ChrisBAshton, line 11: Nit: we can omit these quotes I think.

source/manual/test-and-build-a-project-with-github-actions.html.md
```diff
@@ -27,22 +27,64 @@ You should name the file used for configuring the CI workflow `ci.yml`.
...
-the main branch before running operations. Pull requests from forks of a
-repo will only trigger the `pull_request` event.
+The workflow should be configured to run when there is a [`push`][push-event]
+to the default branch (typically "main") and when a
+[`pull_request`][pull-request-event] opened. This means CI runs when a branch
+is merged (the `push` to main) and it runs against any changes introduced
+in a pull request. We prefer to run against a `pull_request` event to all
+`push` events, as this allows external contributors to have pull requests
+tested and because `pull_request` merges the changes with
+the repository default branch.
```
ChrisBAshton, line 37: I don't understand this last bit. Don't GitHub actions run against the branch as-is, whether or not the trigger was `push` or `pull_request`?

ChrisBAshton (approved):
LGTM ⭐ Thanks for raising this.

source/manual/test-and-build-a-project-with-github-actions.html.md
sengi: So I only found out about the details of this recently myself (thanks to @kevindew for putting me onto it) — pull_request uses the `refs/pull/<pr_number>/merge` reference that GitHub uses for setting up the merge. (I believe this is essentially the PR branch rebased onto the destination branch, but don't quote me on that.) Compare https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request with https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push

If the PR branch is up-to-date with main then there should be no practical difference, but if they've diverged significantly and "Require branches to be up-to-date before merging" isn't enabled then it could mean you're testing against an old version of the rest of the codebase.

ChrisBAshton: Thanks for the explanation! Perhaps we should amend that sentence to link back to your comment? 🙏

kevindew:
Thanks all!

PR #3967, opened 2023-04-20
Fix un-linkified URL in Publish Special Routes.

sengi:
No idea why devdocs doesn't linkify URLs, but hey 🤷

While we're at it, make that a page a plain Markdown document instead of an ERB template, since it contains no Ruby.

Tested: renders as expected when running `bundle exec middleman server` locally.


ChrisBAshton (approved):
You can also wrap the URL in `<>` and it will linkify.

sengi:
> You can also wrap the URL in `<>` and it will linkify.

Huh, TIL 😀 thanks!

PR #3970, opened 2023-04-21
Document how to archive a repo

ChrisBAshton:
There are a few more steps than simply clicking 'archive' in the repo settings. This commit documents those steps and removes the duplication from the "Retire an application" steps.


AgaDufrat (approved):
Nice one. 
For frontend apps we also need to remove the project from Heroku if it has PR previews set up.

PR #3979, opened 2023-04-26
Remove redundant references to Jenkins in 'run a rake task' 

GDSNewt:
The segment on EC2 is no longer relevant, to avoid confusion we should remove it.

### DIFF
--- a/source/manual/running-rake-tasks.html.md
+++ b/source/manual/running-rake-tasks.html.md
@@ -7,7 +7,7 @@ parent: "/manual.html"
 important: true
 ---
 
-Rake is a task runner for Ruby. There are several different methods of running a rake task:
+Rake is a task runner for Ruby. Here are the methods for running a rake task:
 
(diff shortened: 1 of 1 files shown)


source/manual/running-rake-tasks.html.md
```diff
@@ -21,24 +21,9 @@ kubectl -n apps exec deploy/publishing-api -- rake 'represent_downstream:publish
 
 The output of the command will be streamed to your terminal.
 
-## Run a rake task on EC2
+## Run rake tasks from the command line
```
richardTowers, line 24:
```suggestion
## Run rake tasks on EC2 from the command line
```

Just to clarify the difference with "Run a rake task on EKS" (which also uses the command line)

richardTowers: Ah, I was too slow. No worries, fine as it is :)

PR #3993, opened 2023-05-05
Update architecture diagram

richardTowers:
The architecture diagram was looking a little out of date.

In particular:
- Mapit instead of locations-api
- manuals-frontend still there even though it's been retired
- info-frontend making calling out to performance platform (which has been retired)

I fixed these things in the source diagram, but that seems to have broken the original embedding code (draw.io has moved to diagrams.net, so I'm not surprised the embed broke).

This commit updates the embed, and attempts to size the diagram so it takes up more of the screen.

My CSS is a bit hacky, but the idea is "the whole view width, minus the sidebar and margins, appart from on mobile where it's the whole view width minus the margins". The 700px height is arbitrary. Seems to just about work.

### DIFF
--- a/source/stylesheets/screen.css.scss
+++ b/source/stylesheets/screen.css.scss
@@ -93,9 +93,12 @@ ul.subdir__menu {
 }
 
 .architecture-diagram {
+  // https://github.com/alphagov/tech-docs-gem/blob/main/lib/assets/stylesheets/modules/_app-pane.scss#L2
+  $toc-width: 330px;
+
   height: 700px;
-  width: calc(100vw - 400px);
+  width: calc(100vw - #{$toc-width} - #{$govuk-gutter} * 2);
   @include govuk-media-query($until: tablet) {
-    width: calc(100vw - 60px);
+    width: calc(100vw - #{$govuk-gutter} * 2);
   }
 }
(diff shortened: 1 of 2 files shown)


source/stylesheets/screen.css.scss
```diff
@@ -91,3 +91,11 @@ ul.subdir__menu {
 .indented-list {
   margin: 0 0 0 20px;
 }
+
+.architecture-diagram {
+  height: 700px;
+  width: calc(100vw - 400px);
+  @include govuk-media-query($until: tablet) {
+    width: calc(100vw - 60px);
```
jkempster34, line 99: Hello 👋 I wonder if this would be clearer with some variables `calc(100vw - (#{$govuk-gutter} * 2))`

I haven't tested this ☝️, but it should work

source/stylesheets/screen.css.scss
```diff
@@ -91,3 +91,11 @@ ul.subdir__menu {
 .indented-list {
   margin: 0 0 0 20px;
 }
+
+.architecture-diagram {
+  height: 700px;
+  width: calc(100vw - 400px);
```
jkempster34, line 97: And here. `calc(100vw - ((#{$govuk-gutter} * 2) + 330px))` (I'm not sure where the 330px width of the sidebar is defined)

Also not tested!

source/stylesheets/screen.css.scss
richardTowers: Ah yeah, I was thinking of doing that but I was too lazy to look up the variable name 😅 

I'll give it a try, thanks!

PR #3997, opened 2023-05-11
Improve link and step by step documentation

AshGDS:
Fixes missing/incorrect data on the link and step by step documentation.

### DIFF
--- a/data/analytics/attributes.yml
+++ b/data/analytics/attributes.yml
@@ -135,7 +135,7 @@
   example: |
     {1=>"/conditions/vaccinations/flu-influenza-vaccine/", 2=>"undefined", 3=>"undefined", 4=>"undefined", 5=>"undefined"}
   required: yes
-  type: string
+  type: object
   redact: true
 
 - name: location
--- a/data/analytics/attributes.yml
+++ b/data/analytics/attributes.yml
@@ -76,10 +76,10 @@
   redact: false
 
 - name: index
-  description: Contains the index of a thing being clicked on, for example the nth accordion section, or the nth link in a list.
+  description: Contains an object with the indexes of a thing being interacted with on, for example the nth accordion section, or the nth link in a list.
   example:
   required: no
-  type: string (1-indexed number)
+  type: object
   redact: false
 
 - name: index_link
--- a/data/analytics/attributes.yml
+++ b/data/analytics/attributes.yml
@@ -56,7 +56,7 @@
   description: Used for link tracking. True means the link points to a URL not on the current site.
   example:
   required: yes
-  type: string (true or false)
+  type: string ("true" or "false")
   redact: false
 
 - name: first_published_at
@@ -76,7 +76,7 @@
   redact: false
 
 - name: index
-  description: Contains an object with the indexes of a thing being interacted with on, for example the nth accordion section, or the nth link in a list.
+  description: Contains an object with the indexes of a thing being interacted with, for example the nth accordion section, or the nth link in a list.
   example:
   required: no
   type: object
@@ -91,7 +91,7 @@
   redact: false
 
 - name: index_section
-  description: The index of the section that was interacted with, for example the index of the section the clicked link is in
+  description: The index of the section that was interacted with, for example the index of the section the clicked link is in. Sections are usually segmented by headings within a component.
   value:
   example: 1
   required: no
@@ -99,7 +99,7 @@
   redact: false
 
 - name: index_section_count
-  description: The total number of sections in the thing being tracked, such as the header or sidebar
+  description: The total number of sections in the thing being tracked, such as the header or sidebar. Sections are usually segmented by headings within a component.
   value:
   example: 1
   required: no
--- a/data/analytics/events.yml
+++ b/data/analytics/events.yml
@@ -384,7 +384,7 @@
         - name: event_name
           value: navigation
         - name: external
-          value: string (true)
+          value: '"true"'
         - name: link_domain
         - name: link_path_parts
         - name: method
@@ -416,7 +416,7 @@
       implemented: true
       priority: low
       description: 'When a link points to an email address using mailto:'
-      tracker: specialist_link_tracker.
+      tracker: specialist_link_tracker
       data:
       - name: event
         value: event_data
@@ -425,7 +425,7 @@
         - name: event_name
           value: navigation
         - name: external
-          value: 'string (true)'
+          value: '"true"'
         - name: link_domain
         - name: link_path_parts
         - name: method
@@ -437,7 +437,7 @@
       implemented: false
       priority: medium
       description: 'Links on our guidance pages that contain file attachments, such as https://www.gov.uk/government/publications/equality-act-guidance'
-      tracker: specialist_link_tracker.
+      tracker: specialist_link_tracker
       data:
       - name: event
         value: event_data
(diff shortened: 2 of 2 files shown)


data/analytics/attributes.yml
```diff
@@ -76,10 +76,10 @@
   redact: false
 
 - name: index
-  description: Contains the index of a thing being clicked on, for example the nth accordion section, or the nth link in a list.
+  description: Contains an object with the indexes of a thing being interacted with on, for example the nth accordion section, or the nth link in a list.
```
andysellick, line 79: I think you meant to remove the word `on`?

data/analytics/attributes.yml
```diff
@@ -76,10 +76,10 @@
   redact: false
 
 - name: index
```
andysellick, line 78: Do we need data for the `index` attribute now? It's only a container for sub attributes. We don't have one for `event_data`.

data/analytics/events.yml
```diff
@@ -384,7 +384,7 @@
         - name: event_name
           value: navigation
         - name: external
-          value: true
+          value: string (true)
```
andysellick, line 387: Although this isn't a new entry for some reason the `external` attribute isn't showing on the current deployment, we might need to check.

andysellick, line 387: I think this is a bit confusing, although I get that you're trying to clarify that the `true` value is a string. Maybe instead if we had it display as

```
  external: "true"
```

...although that might require some code changes to make the quotes render.

There might be a separate conversation to be had about how much information we want to display in the event about the attribute (for example we could put all the value types after the values e.g. `external: "true" (string)` but I'd be concerned about overloading this view and reducing the value of the attribute page itself.

data/analytics/events.yml
```diff
@@ -413,17 +413,46 @@
...
+        - name: method
+        - name: text
+        - name: type
+          value: email
+        - name: url
     - name: attachment links
       implemented: false
       priority: medium
+      description: 'Links on our guidance pages that contain file attachments, such as https://www.gov.uk/government/publications/equality-act-guidance'
+      tracker: specialist_link_tracker.
```
andysellick, line 440: The `.` at the end needs to be removed - the code doesn't recognise this tracker and doesn't render the link (I think you've copied and pasted this in a couple of other places).

data/analytics/events.yml
```diff
@@ -413,17 +413,46 @@
...
+          value: 'string (true)'
+        - name: link_domain
+        - name: link_path_parts
+        - name: method
+        - name: text
+        - name: type
+          value: email
+        - name: url
     - name: attachment links
       implemented: false
```
andysellick, line 437: Is this not implemented yet? I thought it was?

data/analytics/events.yml
```diff
@@ -784,12 +813,18 @@
...
+        - name: event_name
+          value: select_content
         - name: action
           value: opened/closed
-        - name: event_name
         - name: index
-        - name: index_total
+          value:
+          - name: index_section
+            value: The section that this step is in. The section in this case refers to the step number.
```
andysellick, line 823: This is a good addition. However I think it should go into the data entry for the `index_section`. Unless it's changed in the move to dev docs, if no value is supplied in an event for an attribute, it'll pull through the value from the attribute data. That means for something generic like this we could have this description in one place and it'll propagate to all events that include it.

andysellick: (same for `index_section_count`)

data/analytics/events.yml
```diff
@@ -784,12 +813,18 @@
...
+          - name: index_section_count
+            value: The total amount of steps within this step by step component.
         - name: text
+          value: (the opened/closed step's title)
```
andysellick, line 827: Suggestion: `title of the opened/closed step`

data/analytics/events.yml
```diff
@@ -832,9 +870,8 @@
             value: The section that this link is in. The section in this case refers to the step number.
           - name: index_link
             value: The position of the link within this section.
-          - name: index_section_count
-            value: The total amount of links within this section.
         - name: index_total
+          value: The total amount of links within this section.
```
andysellick, line 874: Same as earlier - good addition, but if we can let's put it in the attribute data so it pulls through.

andysellick (commented):
Great stuff, thanks for adding and checking. Just a few comments.

data/analytics/events.yml
AshGDS: Yep, it doesn't display because just putting `true` makes it a `boolean` instead of a `string` value.

data/analytics/events.yml
AshGDS: It hasn't been deployed to production yet, it requires republishing all the guidance documents, and the PAs are changing it from just `html attachment` to `attachment`, so I'm not sure how long it'll be before it's fully implemented.

data/analytics/attributes.yml
AshGDS: Sure, I can remove it, but if I do it will break the `print this page` and `share this page` pages, as they are using the defaults for `index` (i.e. they aren't setting the value themselves), so without it, the whole `index` value disappears on them. Is this OK?

data/analytics/events.yml
AshGDS: No worries, I can just do `external: '"true"'` instead :+1:

data/analytics/events.yml
AshGDS: Good spot, thanks!

data/analytics/events.yml
AshGDS: @andysellick When I remove the `value`, instead it pulls in the `type`. So it just becomes `string (number)`. Is this OK?

data/analytics/events.yml
AshGDS: When I remove the value, instead it pulls in the `type`. So it just becomes 
```yml
index_section: number
index_section_count: number
```

Is this OK?

AshGDS: I've added some extra context to `index_section` and `index_section_count` in `attributes.yml` but if you think it isn't needed I can remove it :+1:

data/analytics/attributes.yml
andysellick: Oh of course, thanks. Okay, we'll update those two and as part of that remove the `index` attribute as part of a separate PR, thanks.

data/analytics/events.yml
andysellick: Fair enough, thanks.

data/analytics/events.yml
andysellick: Ah yes, starting to remember how this works! Yes I think it's fine that it just says `number` and that users can then click on the attribute to find out more information about what that number represents. Also good other additions 👍

data/analytics/events.yml
andysellick: Yes, I think that's okay. Although I'm now questioning which is better: `string (number)` or `number (string)`. Any thoughts?

andysellick (approved):
Looking good, a few outstanding comments but I think they're to pick up separately and shouldn't be blocking 👍

data/analytics/events.yml
AshGDS: @andysellick I think both make sense, so I don't mind!

data/analytics/attributes.yml
andysellick: Starting to pull that PR together here: https://github.com/alphagov/govuk-developer-docs/pull/4002

PR #4001, opened 2023-05-15
Mark assets-{businesslink,directgov} repos retired.

sengi:
Dunno whether it's actually of any value keeping these in here, but 🤷


richardTowers (approved):

sengi:
Found another bit. I think that's the last of it though https://sourcegraph.com/search?q=assets-directgov&groupBy=repo (not worth cleaning up govuk-app-deployment as we go along, I think)

PR #4003, opened 2023-05-16
Add analytics to the analytics pages

andysellick:
## What
Adds GA4 analytics tracking to the implementation record pages, using Google Tag Manager. This includes:

- cookie banner
- default opt out
- Javascript to handle cookie consent/deny
- a cookies page where users can change their preferences and learn about cookies

Note that this only adds GA4 to the implementation record part of the site, not to any of the rest of the dev docs (the JS to do this is only included in the implementation record page layout).

Also updates all of the pages of the implementation record to have page titles, so that they can be more easily distinguished in GA4.

### Things still to do
This PR is a slightly rush job so here's a quick list of things to fix soon:

- the JS needs refactoring and improving, ideally as a proper GOV.UK module (if possible)
- cookie handling in particular in the JS is a bit rough around the edges, if you accept cookies and then go to the cookies page and unconsent, it doesn't actually delete the GTM cookies (even though consent is successfully denied)
- need some tests for the JS and the new title generator in the helper
- changing preferences on the cookies page isn't as clear as it could be, particularly if you try to do it twice

## Why
We're about to share the implementation record and it would be helpful to know what kind of traffic we get and how people use the site.

## Visual changes

Screenshot 2023-05-16 at 11 49 50

Trello card: https://trello.com/c/JbLX3n2C/388-implementation-record-additions-and-improvements

### DIFF
--- a/source/javascripts/cookie-consent.js
+++ b/source/javascripts/cookie-consent.js
@@ -1,0 +1,78 @@
+window.GOVUK = window.GOVUK || {}
+
+window.GOVUK.gtm = {
+  init: function () {
+    window.dataLayer = window.dataLayer || [];
+    this.gtag();
+    this.checkConsent();
+  },
+
+  gtag: function () {
+    dataLayer.push(arguments);
+  },
+
+  setupGtm: function () {
+    (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+    'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+    })(window,document,'script','dataLayer','GTM-TNKCK97');      
+  },
+
+  checkConsent: function () {
+    var cookies = this.readCookies();
+    cookies = cookies.split(';');
+
+    for (var i = 0; i < cookies.length; i++) {
+      var thisCookie = cookies[i].split('=');
+      if (thisCookie[0] == 'analytics_consent') {
+        // consent cookie exists and is yes, init gtm
+        // if it exists and is no, we just hide the banner
+        if (thisCookie[1] == 'true') {
+          this.setupGtm();
+        }
+        this.hideCookieBanner();
+        break;
+      }
+    }
+  },
+
+  denyCookies: function () {
+    this.gtag('consent', 'default', {
+      'ad_storage': 'denied'
+    });
+    this.cookieConsent("false");
+    this.hideCookieBanner();
+  },
+
+  acceptCookies: function () {
+    this.gtag('consent', 'update', {
+      'ad_storage': 'granted'
+    });
+    this.setupGtm();
+    this.cookieConsent("true");
+    this.hideCookieBanner();
+  },
+
+  cookieConsent: function (trueOrFalse) {
+    var consentCookie = "analytics_consent=" + trueOrFalse + "; Secure" + this.readCookies();
+    this.writeCookies(consentCookie);
+  },
+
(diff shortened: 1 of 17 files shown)


gclssvglx (approved):
Given caveats you list @andysellick , and the fact that this change only affects the analytics part of GOV.UK dev docs - happy to approve.

As a side note: I do wonder if we should enable analytics across the entire docs 🤔 This might surface some interesting things around dev docs usage.

JamesCGDS (approved):
Code looks good to me - although I'm having trouble running developer docs locally so I'm not able to click around and test it

source/javascripts/cookie-consent.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+  },
+
+  denyCookies: function () {
+    this.consent = false;
+    this.gtag('consent', 'default', {
+      'ad_storage': 'denied'
+    });
+    this.cookieConsent();
+    this.hideCookieBanner();
```
AshGDS, line 52:
```suggestion
    this.hideCookieBanner();
    
    window.google_tag_manager = null
    window.google_tag_data = null
    document.cookie = '_ga=; Max-Age=0; path=/; domain=' + location.hostname;
    document.cookie = '_ga_EYTJMCJ2LP=; Max-Age=0; path=/; domain=' + location.hostname;
    var gtmScript1 = document.querySelector('script[src="https://www.googletagmanager.com/gtm.js?id=GTM-TNKCK97"]')
    var gtmScript2 = document.querySelector('script[src="https://www.googletagmanager.com/gtag/js?id=G-EYTJMCJ2LP&l=dataLayer&cx=c"]')
    if(gtmScript1) {
      gtmScript1.remove()
    }
    if(gtmScript2) {
      gtmScript2.remove()
    }
```

This should remove the cookies successfully. It clears the Google script from the window, sets the cookies to remove, and deletes the script from the head of the page. Up to you if you want to implement it, but thought it's best to make sure the rejection works properly.

AshGDS:
I think the PA’s might need to enable the domain in their dashboard for the analytics to be collected - (see [this comment](https://trello.com/c/uySAO8PX/418-add-assetspublishingservicegovuk-to-load-analyticsjs-domains#comment-63dd4fd6303db46d97e0b8c0), where a PA mentions configuring GA4 to accept traffic from a domain)

source/javascripts/cookie-consent.js
andysellick: That's very thorough. Maybe a bit too thorough as if you then navigate to another page those scripts will be removed anyway? Also I wasn't certain whether the script would always have that exact name, although maybe it's based on the property ID.

andysellick:
> I think the PA’s might need to enable the domain in their dashboard for the analytics to be collected - (see [this comment](https://trello.com/c/uySAO8PX/418-add-assetspublishingservicegovuk-to-load-analyticsjs-domains#comment-63dd4fd6303db46d97e0b8c0), where a PA mentions configuring GA4 to accept traffic from a domain)

It's currently working from my dev machine, I've got access to the property and can see my hits in realtime. Haven't tested it on the live site yet though.

source/javascripts/cookie-consent.js
AshGDS: @andysellick Yeah I guess it was a temporary solution, but it looks like you can remove the last part and it will still work. I think the ID would stay the same (I worked on National Archives' cookie mechanism, and the ID never changed.) This may be cleane:r:

```JavaScript
    window.google_tag_manager = null
    window.google_tag_data = null
    document.cookie = '_ga=; Max-Age=0; path=/; domain=' + location.hostname;
    document.cookie = '_ga_EYTJMCJ2LP=; Max-Age=0; path=/; domain=' + location.hostname;
```

The issue seems to be that if you just delete the cookie, GTM is still running on the actual cookie page, so it sets a cookie again after deleting the cookie. But `window.google_tag_manager = null  window.google_tag_data = null` seems to solve that.

PR #4012, opened 2023-05-24
Move Signon to Access and Permissions team

chrisroos:
The Access and Permissions team are now responsible for Signon.

I believe this change is enough to update the docs, dependapanda and seal.

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -1162,7 +1162,7 @@
 
 - repo_name: signon
   type: Supporting apps
-  team: "#govuk-publishing-platform"
+  team: "#govuk-publishing-access-and-permissions-team"
   production_hosted_on: eks
 
 - repo_name: slimmer


floehopper (approved):
LGTM 👍

floehopper:
@chrisroos Do you think we need to add an entry for the A&P team to [this file](https://github.com/alphagov/govuk-rota-announcer/blob/main/config/weekday.yml) as well...?

chrisroos:
> @chrisroos Do you think we need to add an entry for the A&P team to [this file](https://github.com/alphagov/govuk-rota-announcer/blob/main/config/weekday.yml) as well...?

Ah, good spot. It looks like it to me. I've just found https://github.com/alphagov/govuk-rota-announcer/pull/55 which added the Navigation and Presentation bot monitor entry, in case that helps.

PR #4017, opened 2023-05-25
CodeCommit IAM authentication setup and update previous github unavailable process

martynipratt:
Moving away from the shared credentials method of checking out code from CodeCommit as it is insecure and using IAM authentication instead.

### DIFF
--- a/source/manual/howto-checkout-and-commit-to-codecommit.html.md
+++ b/source/manual/howto-checkout-and-commit-to-codecommit.html.md
@@ -1,0 +1,84 @@
+---
+owner_slack: "#govuk-2ndline-tech"
+title: How to check and commit code to AWS CodeCommit
+section: AWS
+layout: manual_layout
+parent: "/manual.html"
+---
+
+CodeCommit is used as a backup/mirror for our repositories in GitHub. It is regularly populated by a [GitHub Action in govuk-infrastructure](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/mirror-repos.yml).
+
+Should we need to commit code manually to CodeCommit (if GitHub is down, for example), we can leverage the access granted to us via [gds-cli](https://github.com/alphagov/gds-cli) in order to authenticate our local user and commit code.
+
+## Quick reference guide
+
+- The following example shows how to clone the whitehall-prototype-2023 repository, create a local branch, make a local change then push the change to a remote branch in aws codecommit
+
+```
+gds aws govuk-tools-poweruser git clone codecommit::eu-west-2://whitehall-prototype-2023
+cd whitehall-prototype-2023
+git checkout -b mybranch
+touch mychange
+git add mychange
+git commit -m "my change"
+gds aws govuk-tools-poweruser git push origin mychange
+```
+
+## Install dependencies and set up local environment
+
+- For this to work you need to have [set up the GDS command line tools](https://docs.publishing.service.gov.uk/manual/get-started.html#3-install-gds-command-line-tools)
+
+- You will also need to install git-remote-codecommit
+
+```
+brew install git-remote-codecommit
+```
+
+- You will also need to add the below lines to your git config as per [AWS documentation](https://docs.aws.amazon.com/codecommit/latest/userguide/setting-up-https-unixes.html#setting-up-https-unixes-credential-helper)
+
+```
+git config --global credential.helper '!aws codecommit credential-helper $@'
+git config --global credential.UseHttpPath true
+```
+
+- You should see the following entry in the gitconfig file (`cat ~/.gitconfig`)
+
+```
+[credential]
+    helper = !aws codecommit credential-helper $@
+    UseHttpPath = true
+```
+
+## Cloning repositories
+
+- login to the AWS console to see the available repositories using [the GDS command line tool](https://docs.publishing.service.gov.uk/manual/get-started.html#3-install-gds-command-line-tools)
+
+```
+gds aws govuk-tools-poweruser --login
+```
+
+- search "codecommit" in the aws search bar and ensure that "London" is selected as the region
+
+- find the repository you are wanting to clone using the repository search bar then select the "HTTPS (GRC)" __repository url__
+
+- open a terminal and navigate to the directory you want to clone the repository to then clone the repository using
+
+```
+gds aws govuk-tools-poweruser git clone <repository url>
+```
+
+- an example of cloning the whitehall prototype 2023 repository looks like this
+
+```
+gds aws govuk-tools-poweruser git clone codecommit::eu-west-2://whitehall-prototype-2023
+```
+
+## Pushing to a branch
+
+- After cloning the repository you can create local branches and make local changes as you would with a normal git repository
+
+- When it comes to pushing to a remote branch prefix the git command with the GDS command line tool so the correct credentials are used to authenticate with AWS
+
+```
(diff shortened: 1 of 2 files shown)


source/manual/howto-checkout-and-commit-to-codecommit.html.md
```diff
@@ -0,0 +1,90 @@
+---
+owner_slack: "#govuk-2ndline-tech"
+title: How to check and commit code to AWS codecommit
+section: AWS
+layout: manual_layout
+parent: "/manual.html"
+---
+Codecommit is used as a backup/mirror to our repositories in github as such we don't need to push code there all the time.
```
ChrisBAshton, line 8:
```suggestion

CodeCommit is used as a backup/mirror for our repositories in GitHub. It is regularly populated by a [GitHub Action in govuk-infrastructure](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/mirror-repos.yml).
```

source/manual/howto-checkout-and-commit-to-codecommit.html.md
```diff
@@ -0,0 +1,90 @@
...
+We can leverage the access granted to us via the GDS cli in order to authenticate our local user and commit code.
```
ChrisBAshton, line 9:
```suggestion

Should we need to commit code manually to CodeCommit (if GitHub is down, for example), we can leverage the access granted to us via [gds-cli](https://github.com/alphagov/gds-cli) in order to authenticate our local user and commit code.
```

Though given we can't deploy directly from CodeCommit, I am struggling to think of a situation where we'd actually need to commit code to CodeCommit... perhaps you can think of a better example above? 🤔

source/manual/howto-checkout-and-commit-to-codecommit.html.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Installing dependancies and setting up local environment
```
ChrisBAshton, line 11:
```suggestion
## Install dependencies and set up local environment
```

source/manual/howto-checkout-and-commit-to-codecommit.html.md
```diff
@@ -0,0 +1,90 @@
...
+git config --global credential.UseHttpPath true
+```
+
+- check the git config file has the following entry in it by printing the file to the screen or checking with an editor
+
+```
+cat ~/.gitconfig
+```
+
+- You should see the following entry in the gitconfig file
```
ChrisBAshton, line 34:
```suggestion
- You should see the following entry in the gitconfig file (`cat ~/.gitconfig`)
```

source/manual/howto-checkout-and-commit-to-codecommit.html.md
```diff
@@ -0,0 +1,90 @@
...
+
+- search "codecommit" in the aws search bar and ensure that "London" is selected as the region
+
+- find the repository you are wanting to clone using the repository search bar then select the "HTTPS (GRC)" __repository url__
+
+- open a terminal and navigate to the directory you want to clone the repository to then clone the repository using
+
+```
+gds aws govuk-tools-poweruser git clone <repository url>
+
```
ChrisBAshton, line 58:
```suggestion
```

source/manual/howto-checkout-and-commit-to-codecommit.html.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+- an example of cloning the whitehall prototype 2023 repository looks like this
+
+```
+gds aws govuk-tools-poweruser git clone codecommit::eu-west-2://whitehall-prototype-2023
+
```
ChrisBAshton, line 65:
```suggestion
```

source/manual/howto-checkout-and-commit-to-codecommit.html.md
```diff
@@ -0,0 +1,90 @@
...
+```
+gds aws govuk-tools-poweruser git clone codecommit::eu-west-2://whitehall-prototype-2023
+cd whitehall-prototype-2023
+git checkout -b mybranch
+touch mychange
+git add mychange
+git commit -m "my change"
+gds aws govuk-tools-poweruser git push origin mychange
+```
```
ChrisBAshton, line 90: I'd move this to the top of the file for quick reference, i.e. just above the `## Install dependencies and set up local environment` heading. (And delete the `Example workflow` heading).

ChrisBAshton (commented):
Thanks for updating these - the old instructions have [already been causing confusion on 2nd line](https://trello.com/c/aqn3tSLZ/1943-2nd-line-drill-special-deployment-conditions) so it will be great to see this updated and merged soon ⭐

source/manual/howto-checkout-and-commit-to-codecommit.html.md
martynipratt: I've added all of your suggestions and squashed the commits 😊

source/manual/howto-checkout-and-commit-to-codecommit.html.md
ChrisBAshton: I still wonder if a better example exists, @martynipratt? In other words, what's the main driver for us having the ability to push code to CodeCommit? It had its uses when we had a Jenkins job that communicated with it and allowed us to deploy from CodeCommit if GitHub was down, but now we've moved to Argo (which also relies on GitHub) our fallback deployment mechanism is different. I can see the benefit of CodeCommit as a mirror, but still not sure of a use case for pushing to it when GH is down.

source/manual/howto-checkout-and-commit-to-codecommit.html.md
```diff
@@ -0,0 +1,84 @@
...
+parent: "/manual.html"
+---
+
+CodeCommit is used as a backup/mirror for our repositories in GitHub. It is regularly populated by a [GitHub Action in govuk-infrastructure](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/mirror-repos.yml).
+
+Should we need to commit code manually to CodeCommit (if GitHub is down, for example), we can leverage the access granted to us via [gds-cli](https://github.com/alphagov/gds-cli) in order to authenticate our local user and commit code.
+
+## Quick reference guide
+
+- The following example shows how to clone the whitehall-prototype-2023 repository, create a local branch, make a local change then push the change to a remote branch in aws codecommit
```
ChrisBAshton, line 15:
```suggestion
The following example shows how to clone the whitehall-prototype-2023 repository, create a local branch, make a local change then push the change to a remote branch in aws codecommit. *You must* first [install dependencies and set up your local environment](#install-dependencies-and-set-up-local-environment).
```

ChrisBAshton (approved):
LGTM 👍

source/manual/howto-checkout-and-commit-to-codecommit.html.md
martynipratt: Now that we have Github for both source control and pipeline it seems wise to consider having a contingency for the pipeline part as well. 

I see the scope of the PR/ticket be to replace the previous process of using shared credentials. 
I've created another [ticket](https://trello.com/c/5eMaxzB5/2986-move-the-current-codecommit-backup-off-of-jenkins) to work on updating the codecommit backup process which depends on puppet and Jenkins anyway so will need to be updated. 
I and could add another ticket to look into adding contingency for github actions in the event that Github actions goes down. I see that being a significantly large piece of work that is outside the scope of this PR/ticket.

PR #4026, opened 2023-06-01
Explain Terraform Cloud first login behaviour

theseanything:
This makes it clearer that users will need to create a new Terraform Cloud account with it's own password, even when using SSO.


source/manual/dns.html.md
```diff
@@ -127,6 +127,8 @@ Some of these are not managed by Terraform. If you can't find a configuration fi
 
 If you're a member of 2nd-line, Platform Engineering or Platform Security and Reliability, you should automatically be able to [log into Terraform Cloud](https://accounts.google.com/o/saml2/initsso?idpid=C01ppujwc&spid=738388265440&forceauthn=false) using your digital.cabinet-office.gov.uk Google account.
 
+For your first login, you must link a new or an existing one Terraform Cloud account (which will have its own password) to your Google Account. After linking these accounts, you'll be able to access the organization by signing in through your Google Account.
```
ChrisBAshton, line 130: I think we can make this clearer still (but disclaimer, I can't verify the steps below as I went a slightly different route to creating my account±). I like your GitHub analogy in the Slack thread:

```suggestion
If you're logging in for the first time, you'll need to create a password for your new Terraform Cloud account, as it exists independently of your Google Account (much like your GitHub account exists independently of the `alphagov` org). Once your account is created, you'll be able to sign in through your Google Account (SSO) and access the `govuk` organisation on Terraform cloud.
```

± My experience was to visit https://app.terraform.io/session, click "Sign in with SSO", then provide "govuk" as the "Organization name". I then had to create a password, and click on an email confirmation link. I don't think I had to provide an email address, I think this may have been pre-populated...? Later I clicked the SSO link referenced on line 128, which seemed to automatically link up my 'manual' account with my Google one, no password needed anymore.

sengi: If new users are having to set passwords then something is quite wrong with the way we've set this up. If we're documenting a workaround then let's make it clear that it's a workaround and not the intended situation.

dj-maisy: I do think we need to experiment with this a little bit before this becomes the official advice. Having to create a TF Cloud account before using SSO seems redundant and defeating part of the point of having SSO in the first place, so I don't think it is meant to work like this.

This could just be that users who needed to be in the relevant Google Group were not in it prior to logging in.

theseanything: I think there's some confusion over what SSO provides here. In Terraform Cloud, [user accounts belong to individual people, not organisations](https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/users) (similar to GitHub's user model). Users can be invited to the `govuk` Terraform Cloud org to give them access to GOV.UK workspaces.

SSO provides a way to manage membership to the `govuk` organisation, rather than manage individual user accounts. The benefit of SSO here is centralised way to grant and redact permissions, rather automatic user account provisioning.

sengi: The Hashi folks confirmed it works this way :(

sengi (approved):

sengi:
Sorry for holding this up :(

PR #4029, opened 2023-06-02
Update documentation on running A/B tests

GDSNewt:
Updates the docs on running A/B tests. The current docs are ambiguous when it comes to generating tests so this change should make it clearer. It also removes old links connecting to archived repos and provides more up-to-date examples to look at.

### DIFF
--- a/source/manual/run-ab-test.html.md
+++ b/source/manual/run-ab-test.html.md
@@ -34,8 +34,8 @@ Make sure you monitor your test after deployment.
 
 1. Write the A/B test. Use the information in the [govuk_ab_testing gem][govuk_ab_testing] to understand how to serve different versions to your users.
 1. Add your test to the [A/B test register][register].
-1. If you want to use Google Analytics to monitor the A/B test, talk to a performance analyst and pick a [GA dimension][analytics-dimensions] to use for your test.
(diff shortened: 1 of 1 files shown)


source/manual/run-ab-test.html.md
```diff
@@ -34,8 +34,8 @@ Make sure you monitor your test after deployment.
 
 1. Write the A/B test. Use the information in the [govuk_ab_testing gem][govuk_ab_testing] to understand how to serve different versions to your users.
 1. Add your test to the [A/B test register][register].
-1. If you want to use Google Analytics to monitor the A/B test, talk to a performance analyst and pick a [GA dimension][analytics-dimensions] to use for your test.
-1. Create dictionary and A/B test files in the [govuk-cdn-config repo][govuk-cdn-config]. See an [example for the dictionaries][dictionary-config-example] and an [example for the A/B configuration][cdn-config-example]. For more information, see the [dictionaries README][dictionaries-readme].
+1. If you want to use Google Analytics to monitor the A/B test, talk to a performance analyst and pick a [GA dimension][analytics-dimensions] to use for your test. If you wish to run an A/B test in integration or staging you may need to ensure that the relevant analytics also exist in those environments.
```
Rosa-Fox, line 37: A bit picky but I am thinking maybe: ~~the relevant analytics also exist in those environments~~ -> analytics tracking exists in those environments. Just to make it super clear that it may not already exist. As far as I know, there are conversations happening about whether this will be added or not? So I suppose the result of that will have an impact, but its good for the current situation.

Rosa-Fox (approved):

PR #4030, opened 2023-06-02
Update "Purge a page from cache" playbook.

sengi:
- Our Varnish servers no longer exist 🎉
- We no longer have or need a Jenkins job for issuing cache purge requests, since everyone on tech support is given a Fastly account (and the Fastly UI is much easier to use than the Jenkins job) 🎉
- Simplify and update the playbook instructions.
- Remove the complicated procedure for sending a `PURGE` HTTP request from the command line. We'd never do that in a hurry and it's much easier to use the UI (and we don't want to encourage mucking about with Fastly API tokens if we can at all help it).
- Restructure the article so that we don't have to read through a bunch of background info before getting to the actual playbook steps.

Additional context: we almost never have to purge anything from Fastly these days since moving to `max-age: 300` for the vast majority of resources.

Tested rendering with `middleman server` (and a bit of hacking around devdocs bugs to get it to start up locally).

### DIFF
--- a/source/manual/purge-cache.html.md
+++ b/source/manual/purge-cache.html.md
@@ -7,116 +7,69 @@ parent: "/manual.html"
 important: true
 ---
 
-The `www.gov.uk` domain is served through Fastly, which honours the
-cache control headers sent by GOV.UK's running instance of [Varnish][].
-Most pages are set with cache headers of 5 minutes, so when new changes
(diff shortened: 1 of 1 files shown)


theseanything (approved):

sengi:
(removed the superfluous "Where:" heading)

PR #4032, opened 2023-06-03
Stop scraping docs from retired repos.

sengi:
Scraping docs from retired repos is more harmful than beneficial to user experience, because it causes unmaintained and often irrelevant pages to appear in search results. It also makes local development of devdocs painful because so many additional requests are made on startup.

Excluding these legacy pages will improve relevance of search results and make it easier for developers to find relevant, up-to-date information.

We already exclude retired repos from the lists of repos by category on `/repos`.

This doesn't stop us from scraping the repo metadata for retired repos/apps. Retired repos will still appear in search results when searching for all or part of their name, and we'll still generate pages for them under `/repos/${repo-name}.html`.

### DIFF
--- a/app/proxy_pages.rb
+++ b/app/proxy_pages.rb
@@ -9,7 +9,7 @@ class ProxyPages
   end
 
   def self.repo_docs
-    docs = Repos.public.map do |repo|
+    docs = Repos.active_public.map do |repo|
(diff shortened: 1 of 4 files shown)


ChrisBAshton (approved):
LGTM - thanks for doing this ⭐

sengi:
No probs! Thanks for the super-quick review! :D

PR #4035, opened 2023-06-05
Clarify how engineers can access Terraform Cloud

ollietreend:
I've refined the existing documentation to cover some specific quirks and provide some more detailed steps.

1. Explain that membership of specific Google Groups is required - and provide links to them.

2. Explain the relationship between Google and Terraform Cloud accounts. This tripped me up initially because I'd assumed it was a SSO flow, so I didn't understand why I needed to set an account password. This hopefully explains that a bit.

3. Advise users who to speak to if they have problems following the documented self-serve sign-up steps.

Screenshot of updated content

### DIFF
--- a/source/manual/dns.html.md
+++ b/source/manual/dns.html.md
@@ -125,8 +125,14 @@ Some of these are not managed by Terraform. If you can't find a configuration fi
 
 ## Getting Terraform Cloud access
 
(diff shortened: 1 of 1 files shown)


dj-maisy (approved):
👍

PR #4037, opened 2023-06-05
Update the playbook for deploying Static.

sengi:
The aim here isn't so much to change the procedure, rather to document what's (more or less) already happening. The existing doc was clearly out of date and out of sync with the note in the Release app.

- Extract the playbook steps and the background info into separate sections.
- Update the playbook steps to reflect the current state of GOV.UK.
- Clarify why the clunky manual process is the way it is for now, so that it's easier to figure out when it's ok to change or eliminate it in future.

Please review for accuracy in particular :) It's entirely possible that I've made mistakes based on misinterpreting what I've read in the code or other docs while putting this together.

I'd also be open to deleting this entirely, if people feel it's not useful/discoverable enough to be worth keeping.

### DIFF
--- a/source/manual/deploy-static.html.md
+++ b/source/manual/deploy-static.html.md
@@ -6,10 +6,35 @@ layout: manual_layout
 parent: "/manual.html"
 ---
 
-[Static](https://github.com/alphagov/static) requires extra care when deploying.
-You must wait for at least 30 minutes between environment deploys. This is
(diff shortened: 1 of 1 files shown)


chao-xian (approved):
Nice - thanks!

sengi:
Cool, thanks for the quick review @chao-xian!

I'll merge now, but if anyone spots anything that I've got wrong then please still comment and I'll be happy to fix it in a followup.

PR #4038, opened 2023-06-05
Remove temporary and supervised caveat for granted prod admin access

kevindew:
In practice we don't supervise these users or do it for temporary periods of time. We typically use this approach for people who've joined and won't go through 2nd line (senior role or part of a managed service). So this updates this doc to reflect actual process.


ChrisBAshton (approved):
Agreed

PR #4047, opened 2023-06-13
Update and specify hosts in the repos.yml

AgaDufrat:
Bouncer is on EKS: https://github.com/alphagov/govuk-helm-charts/blob/0043196/charts/app-config/values-production.yaml#L230

This also adds information on other apps. I'm hoping that by updating this .yml we can remove duplicated information in https://github.com/alphagov/release/blob/main/data/ec2_deployed_apps.yml and the risk of those two files getting out of sync.

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -35,7 +35,7 @@
 - repo_name: bouncer
   type: Transition apps
   team: "#govuk-navigation-tech"
-  production_hosted_on: aws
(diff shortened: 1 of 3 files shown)


ChrisBAshton (approved):
LGTM 👍 nice idea about getting the Release app to perhaps look at the repos.json for its source of truth.

PR #4050, opened 2023-06-13
Add missing 'govuk-replatform-test-app' repo

ChrisBAshton:
This is used by GOV.UK engineers for getting to grips with the new platform.

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -612,6 +612,12 @@
   sentry_url: false
   dashboard_url: false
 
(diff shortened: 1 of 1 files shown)


sengi (approved):
Thanks!

We should totally just pull these from a label so we don't have to update this any more!

ChrisBAshton:
Long term aspirations 😁 working through the repos to make sure the right things are labelled first!

PR #3913, opened 2023-03-10
Work through list of missing repos tagged as `govuk`

ChrisBAshton:
These missing repos were identified by a new script developed by Platform Reliability, here:
https://github.com/alphagov/govuk-saas-config/pull/142

Some of these repos just need archiving, but others really ought to be represented in the Developer Docs, for visibility of docs and to designate ownership.

I don't intend to merge this as-is, but am using GitHub's PR feature to surface these repos in a forum around which we can decide their fate!

EDIT: also want to add other repos e.g. govuk-app-deployment-private

Trello: https://trello.com/c/tsBL9lPH/2901-work-through-list-of-missing-repos

PR #4051, opened 2023-06-13
Add undocumented Data Products repos where GOV.UK needs visibility

ChrisBAshton:
govuk-s3-mirror is "important for GOV.UK to know about": https://github.com/alphagov/govuk-developer-docs/pull/3913/files#r1132282271

govuk-knowledge-graph-search "should be [referenced in the Developer Docs]. Not only it's going to be on signon alongside the publishing apps, but the dependency with gov.uk content is very strong so it's important that it's mentioned somewhere here, if only to link to its own documentation on the data community docs."

See
https://github.com/alphagov/govuk-developer-docs/pull/3913/files#r1132281588

There's a wider conversation to be had about how Data Products (and indeed other directorates like Pay and Notify) fit into GOV.UK. But in the meantime, our rough litmus test for whether or not an external repo should have the `govuk` topic (anf therefore be included in the GOV.UK Developer Docs) is this:

> Some repos are owned by the Data team but GOV.UK (and
> particularly GOV.UK 2nd line) need to know about it.
> Repos that aren't particularly tied to something that is maintained
> in GOV.UK don't need to be tagged with the `govuk` topic.

See https://github.com/alphagov/govuk-developer-docs/pull/3913/files#r1149043894

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -543,6 +543,10 @@
   sentry_url: false
   dashboard_url: false
 
+- repo_name: govuk-knowledge-graph-search
+  team: "#data-products"
+  type: Data science
+
 - repo_name: govuk-kubernetes-cluster-user-docs
   type: Utilities
   team: "#govuk-platform-engineering"
@@ -642,6 +646,12 @@
   description: Base container images for GOV.UK Ruby apps
   team: "#govuk-platform-engineering"
   type: Utilities
+  sentry_url: false
+  dashboard_url: false
+
+- repo_name: govuk-s3-mirror
+  team: "#data-products"
+  type: Data science
   sentry_url: false
   dashboard_url: false
 


ChrisBAshton:
NB @nacnudus I did have a question mark over whether or not we should include `govuk-knowledge-graph-gcp` here too - see https://github.com/alphagov/govuk-developer-docs/pull/3913#discussion_r1227971688

nacnudus:
This question is timely, because we're trying to find out what standards and protocols have to be gone through to deploy an app behind GOV.UK Signon in production.  The app is `govuk-knowledge-graph-search`, and `govuk-knowledge-graph-gcp` is the data pipeline and hosting infrastructure.

GOV.UK 2nd line sort of knows about the app, because it's in Signon Integration and recently there were some error logs in Sentry.  But it doesn't know about the data pipeline yet.

Perhaps only the app needs to be in the developer docs.

ChrisBAshton:
Thanks - agreed. In which case, please remove the `govuk` topic from https://github.com/alphagov/govuk-knowledge-graph-gcp @nacnudus 🙏 

Are you happy to approve the two repos being added? If so, please approve and I'll go ahead and merge.

PS: also think a wider conversation about this sort of cross-GDS ownership would be a good idea, but right now just trying to unblock some of our repo validation automation work :)

nacnudus (approved):

PR #4053, opened 2023-06-13
Remove untagged repos

ChrisBAshton:
These were all added in #3547, but without much context. We don't want to import _all_ Data Products' repos into the GOV.UK Developer Docs, just the ones that are pertinent to us. See justification in #4051.

By omitting the `govuk` topic from these repos, the Data Products teams have implied that these repos are not pertinent to GOV.UK and should therefore be removed from our repos.json list.

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -400,11 +400,6 @@
   team: "#govuk-publishing-platform"
   type: Utilities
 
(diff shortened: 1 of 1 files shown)


nicholsj (approved):
Makes sense, thanks!

PR #4056, opened 2023-06-13
Update redirect whitehall docs to have the new k8s commands

beccapearce:
For ease of use these commands have been updated with the new eks cluster in mind. This should make the docs easier to follow.


AshGDS (approved):
LGTM :+1: Thanks for adding it!

PR #4058, opened 2023-06-14
Support multiple apps per repo in our data models

ChrisBAshton:
## What

Defines new optional `app_name` property for our repos.yml config, to allow specifying multiple applications per repo.
This fixes the GitHub repo link (on various pages / JSON endpoints) for the two Licensify projects that live in the [licensify](https://github.com/alphagov/licensify) repo but do not share its name (i.e. `licensify-feed` and `licensify-admin`).

The Release app [will need updating)(https://github.com/alphagov/release/blob/8a2d2e1daefb41be3db6c0ce4bbbcc85f23fee72/app/models/repo.rb#L6) to use the new endpoint, as it cares about 'apps' and deployment status, as opposed to abstract repos. The `repos.json` endpoint is also used by govuk-dependencies, seal and govuk-saas-config but in these cases we _do_ care about the repo rather than the app, so they don't need updating.

## Why

We have a [script](https://github.com/alphagov/govuk-saas-config/pull/142) that compares the list of repos in the Developer Docs with all of the alphagov GitHub repos tagged with the 'govuk' topic. We would like it to report no inconsistencies!

It is currently highlighting this inconsistency:

```
Untagged govuk repos:
licensify-admin
licensify-feed
```

In other words, it is expecting a `licensify-admin` and `licensify-feed` repo under alphagov, and for those repos to be tagged with the `govuk` topic. But those repos don't exist: they're a result of a limitation in the Developer Docs' current data modelling, which equates `repo_name` with the application name, making a false assumption that there can only be 0-1 apps per repository. The [licensify](https://github.com/alphagov/licensify) repo actually has 3 apps (`licensify`, `licensify-admin` and `licensify-feed`).

## Screenshots

Repos page (now shows just `licensify`, not its 3 apps):

| Before | After |
|--------|------|
|before|after|

Repos JSON (now shows just `licensify`, not its 3 apps):

| Before | After |
|--------|------|
|before|after|

Apps page (unchanged):

| Before | After |
|--------|------|
|before|after|

Apps JSON (repo URL fixed, otherwise unchanged):

| Before | After |
|--------|------|
|before|after|

Repo page (repo URL fixed, otherwise unchanged):

| Before | After |
|--------|------|
|before|after|

### DIFF
--- a/app/proxy_pages.rb
+++ b/app/proxy_pages.rb
@@ -59,13 +59,13 @@ class ProxyPages
   def self.repo_overviews
     Repos.all.map do |repo|
       {
-        path: "/repos/#{repo.repo_name}.html",
+        path: "/repos/#{repo.app_name}.html",
         template: "templates/repo_template.html",
         frontmatter: {
           title: repo.page_title,
           locals: {
             title: repo.page_title,
-            description: "Everything about #{repo.repo_name} (#{repo.description})",
+            description: "Everything about #{repo.app_name} (#{repo.description})",
             repo:,
           },
         },
--- a/app/repo.rb
+++ b/app/repo.rb
@@ -7,13 +7,13 @@ class Repo
 
   def api_payload
     {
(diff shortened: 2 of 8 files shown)


robinjam (approved):
Nice!

PR #4059, opened 2023-06-14
Update redirect routes documentation

MartinJJones:
Update redirect routes documentation, replacing the GDS CLI and GOV.UK tools and code examples with EKS.


beccapearce (approved):
Looks good!

PR #4061, opened 2023-06-15
Rename puppet_name to shortname

ChrisBAshton:
`shortname` is the terminology used in the Release app, where exactly the same values have been provided there as they have here, e.g. `licence-finder` => `licencefinder`:
https://release.publishing.service.gov.uk/applications/licencefinder/edit

An app's 'shortname' is used for several things, all of which will probably go away at some point:

- name of the app in Puppet, e.g. https://github.com/alphagov/govuk-puppet/blob/603dc08f1a0e9148c844590d006ad22d0bc21737/modules/govuk/manifests/apps/licencefinder.pp
- Sentry DSN configuration, e.g. https://github.com/alphagov/govuk-saas-config/blob/ba6a612caa61920a57d388013b0a7968b51c60fd/sentry/Rakefile#L32
- For use in Graphite metrics
- Colloquially, e.g. "Contacts" shortname as opposed to "Contacts Admin" application name. (But not always, e.g. "Mainstream Publisher" term used by GDS people instead of "Publisher", but not represented as such in any of our config).

Renaming `puppet_name` to `shortname` reflects the fact that we're retiring govuk-puppet soon, but that we may still need to keep these values for the other reasons above. It is also more descriptive.

### DIFF
--- a/app/repo.rb
+++ b/app/repo.rb
@@ -10,7 +10,7 @@ class Repo
       app_name: repo_name, # beware renaming the key - it's used here: https://github.com/alphagov/govuk-dependencies/blob/b3a2a29eb80aefa08098a08633b4a08b05bcc527/lib/gateways/team.rb#L15
       team:,
       dependencies_team:,
-      puppet_name:,
+      shortname:,
       production_hosted_on:,
       links: {
         self: "https://docs.publishing.service.gov.uk/repos/#{repo_name}.json",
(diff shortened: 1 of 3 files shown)


AgaDufrat (approved):
Makes sense to use consolidate the terminology and use `shortname` given we are aiming to get rid of puppet.

ChrisBAshton:
Gives rise to https://trello.com/c/PUsKqIyq/2996-remove-shortname-field-from-release-app.

PR #4066, opened 2023-06-15
Add info on applying terraform changes

MartinJJones:
## What

Add info on applying terraform changes and link to the existing documentation on this.

### DIFF
--- a/source/manual/deploying-terraform.html.md
+++ b/source/manual/deploying-terraform.html.md
@@ -55,6 +55,10 @@ All other fields can be left as they are.
 
 Click "build" and terraform should deploy as expected. Remember to deploy to staging and/or production as required.
 
+### Applying Terraform
+
+Once you have deployed Terraform, you will need [apply the changes in Terraform](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/docs/applying-terraform.md)
+
 ### Other ways of invoking Jenkins
 
 As of version `v2.15.0` of `gds-cli`, you can use it to deploy terraform via Jenkins.


source/manual/deploying-terraform.html.md
```diff
@@ -55,6 +55,10 @@ All other fields can be left as they are.
 
 Click "build" and terraform should deploy as expected. Remember to deploy to staging and/or production as required.
 
+### Applying Terraform
+
+Once you have deployed Terraform, you will need [apply the changes in Terraform](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/docs/applying-terraform.md)
```
ChrisBAshton, line 60: Hi 👋 there's already a plan and apply set of docs up above, concerning govuk-aws.
Your addition here talks about govuk-infrastructure.

I think you only need to run the `apply` in one place, depending on whether you've changed Terraform in govuk-aws or govuk-infrastructure?

It would probably be more helpful to have a section at the beginning of this page, explaining the two different places we store Terraform and the two ways we apply them. (Worth adding a link to our third place too(!): https://docs.publishing.service.gov.uk/manual/terraform-cloud.html)

MartinJJones: Thanks Chris, I will update this section with the recommend changes

MuriloDalRi: @MartinJJones is this PR still being worked on or should we close it?

MartinJJones: Sorry for the late response, thanks for closing, don't think this change is needed anymore, will reopen in the future if required, thanks

PR #4068, opened 2023-06-16
Update production hosts

AgaDufrat:
`production_hosted_on` needs to be specified for a repo to be classified as and app.
It's necessary because we switched to use apps.json in the Release app in https://github.com/alphagov/release/pull/1210

manuals-publisher is hosted on EKS: https://github.com/alphagov/govuk-helm-charts/blob/bcfbda7b2dbc2a9cc21f6f6464c349d494ed45e5/charts/app-config/values-production.yaml#L1473

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -293,6 +293,7 @@
 - repo_name: email-alert-monitoring
   type: Utilities
   team: "#tech-content-interactions-on-platform-govuk"
+  production_hosted_on: eks
   sentry_url: false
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
Thanks, a good thing for the setting of this property to be made consistent / compulsory 👍

AgaDufrat:
Good idea. I will merge it for now. I keep on changing my mind about the best way to make it consistent / compulsory.

PR #4070, opened 2023-06-16
Update playbook for Elasticsearch reindexing.

sengi:
- Fix example commands to reflect the move to Kubernetes. (Also fixes other pages that use the `RunRakeTask` template helper.)
- Remove procedures that no longer work.
- Improve the structure for clarity.
- Minor fixes for house style.

Tested: `SKIP_PROXY_PAGES=true bundle exec rake` builds the manual successfully; also checked formatting/rendering.


app/run_rake_task.rb
```diff
@@ -2,49 +2,16 @@
...
-    Repos.all.select { |app| app.repo_name == name }.first
+    app_name = if application.respond_to?(:repo_name)
+                 application.repo_name
+               else
+                 application
+               end
+    rake_task_name = rake_task.presence || "<rake task>".freeze
+    <<~END_OF_MARKDOWN
+      ```sh
+      k exec deploy/#{app_name} -- rake #{rake_task_name}
```
gclssvglx, line 13: Should we not use the full `kubectl` here, just in case the alias is not set 🤔

source/manual/reindex-elasticsearch.html.md.erb
```diff
@@ -7,139 +7,110 @@ parent: "/manual.html"
...
+document
+types](https://github.com/alphagov/search-api/blob/main/docs/adding-new-fields.md),
+you need to reindex the affected index before the new fields and types can be
+used.
+
+> ⚠️ **Avoid reindexing in production during working hours, except to fix an
+> outage.**
+>
+> Reindexing locks the index for writes, so content is not updated in the
+> search index. Reindexing takes around 2 hours to complete.
```
sengi, line 22:
```suggestion
> search index. Reindexing takes around 2 hours to complete.
```

If someone can confirm that writes are queued and retried (and therefore indexing should eventually catch up on its own after the reindex finishes) then I'd love to clarify that here.

I'd expect that's the case but I didn't want to assume (and I don't have time to verify it experimentally).

source/manual/reindex-elasticsearch.html.md.erb
```diff
@@ -7,139 +7,110 @@ parent: "/manual.html"
...
-age:
-
-```
-rake search:timed_clean MAX_INDEX_AGE=number_of_days SEARCH_INDEX=alias_of_index_to_clean_up
-```
-
-This is running in a Jenkins job that clears any index over 7 days old, and will always leave at least one inactive index available (typically the
-  newest one created) alongside the active index for backup purposes.
+Avoid leaving old indices around, otherwise we may run into storage limitations
+and be unable to index new documents.
```
gclssvglx, line 73: Just an FYI - we've a card here to automate this process: https://trello.com/c/DgcJRu1j/551-automate-cleaning-of-stale-indexes-on-search-api

app/run_rake_task.rb
sengi: I considered this but I think the short version is friendlier and makes it easier to read the relevant part of the examples. It's trivial to set that alias if you don't already have it, and it's in the [install instructions](https://kubernetes.io/docs/tasks/tools/install-kubectl-macos/#enable-kubectl-autocompletion).

PR #4083, opened 2023-06-27
Reflect CSP enforcement status in documentation

kevindew:
As we've started to push the CSP into enforcement mode on a few applications, this updates the documentation to reflect that.

It also updates the guide on changing the CSP to put a greater emphasis on in app modifications as this seems to be an area that people don't know so well.


ChrisBAshton (approved):
LGTM 👍

PR #4088, opened 2023-06-29
You need poweruser or greater access to query CSP

kevindew:
As queries to Athena involve writing to an S3 bucket they can't be run as a readonly user. This seems surprising as you'd expect that you could run a query as readonly.

Hopefully this clarification stops people getting stuck.


ChrisBAshton (approved):
Makes sense - thanks!

PR #4095, opened 2023-07-06
Remove Cache clearing service

AgaDufrat:
RabbitMQ docs is left unchanged (cache-clearing-service is used to show an example of binding). Only the govuk_crawler is still on self hosted RabbitMQ and there are plans remove it in the next couple of weeks so that doc can be removed then.

[GIFT week summer 2023 card](https://trello.com/c/ENPwB2vT/41-retire-cache-clearing-service)
[Tech debt card](https://trello.com/c/TX9wLihh/101-cache-clearing-app-exists-and-we-dont-think-its-necessary-replacing-cache-clearing-app-only-clears-path-not-query-strings)


PeterHattyar (approved):
Looks good to go, well done!

PR #4097, opened 2023-07-07
Keep non-production logs for one week instead of two.

sengi:
We currently have a 14-day retention window in Logit both for production and non-production.

Logit quota is quite expensive and I doubt we're getting value from storing integration/staging logs for more than a week.

I propose reducing the non-prod retention window from 14 to 7 days, saving $1092 / month ($832 / month once we drop the stacks for the old EC2/Puppet environments).

Raising this PR here for discussion/awareness since the actual setting is only in the Logit web UI.

Context: part of a broader push on [making logging more usable and cost-effective](https://docs.google.com/document/d/e/2PACX-1vS31XeKiFRcLmqcmjp2V6Zob-mxLV2oIALb_ZHj7MCK34-8Dko1PQCW0zjDtGG5cxLjChxKIN2g7f3x/pub).

### DIFF
--- a/source/manual/logit.html.md
+++ b/source/manual/logit.html.md
@@ -161,7 +161,12 @@ should be consistent across stacks.
 
 ## Retention period
 
-[14 days](https://dashboard.logit.io/a/1c6b2316-16e2-4ca5-a3df-ff18631b0e74/s/2dd89c13-a0ed-4743-9440-825e2e52329e).
+- Production: 14 days
+- Staging: 7 days
+- Integration: 7 days
+
+The retention window for an environment is configurable under the Settings page
+for the corresponding Logit stack, for users with the Stack Editor permission.
 
 ## Further reading
 


dgm34:
I can't foresee this being a problem and presumably we could change it back if needed in future.

sengi:
Good point — yes, it's very easy to increase it again if need be 👍

andrew-m-gds:
Would it be sensible to schedule an increase at (for example) Christmas, when there's a higher likelihood of investigating problems in the new year that occurred more than 1 week earlier? Or would that penalise the billing for (potentially) 2 calendar months (Dec and Jan)?

sengi:
> Would it be sensible to schedule an increase at (for example) Christmas, when there's a higher likelihood of investigating problems in the new year that occurred more than 1 week earlier? Or would that penalise the billing for (potentially) 2 calendar months (Dec and Jan)?

Production will be staying the same at 14 days' retention.

theseanything (approved):

sengi:
Turns out we already claim to retain non-prod logs for no more than 7 days in the GDS Way, though that page is years out of date and also says that Logit is deprecated (which is isn't, for GOV.UK), so I wouldn't set too much store by it.

Let's go ahead with this and see how it goes.

PR #4103, opened 2023-07-17
Fix rake task snippets for Whitehall

ChrisBAshton:
These were outputting commands like:

```
k exec deploy/whitehall -- rake publishing:overdue:list
```

...which would not work, as the pod name differs. It should actually be:

```
k exec deploy/whitehall-admin -- rake publishing:overdue:list
```


ollietreend (approved):

KludgeKML (approved):
LGTM

PR #4106, opened 2023-07-19
Update RabbitMQ connection docs

chao-xian:
We're on AmazonMQ so connecting has changed.


KludgeKML:
Looks good, but I think maybe also add that you need to be on the VPN / in the office? (I certainly couldn't get it to work without the VPN)

AgaDufrat:
There are separate docs for AmazonMQ: https://docs.publishing.service.gov.uk/manual/amazonmq.html
Those are for the self hosted RabbitMQ that govuk_crawler is still using.

PR #4111, opened 2023-07-20
Remove unnecessary instructions for aws integration access

AgaDufrat:
This step is not necessary as the relevant permissions should be granted in step 6. 'Get added to the GOV.UK user monitor system'.  
`admin` access is granted as it inherits from the `integration_admin_access` access level group. 
See: https://github.com/alphagov/govuk-user-reviewer/blob/48b85cd/config/govuk_tech.yml#L7 

The details are included in the govuk-user-reviewer README so there is no need to duplicate them here.

Also updated an anchor link in step 6.

I'm going to update the the user reviewer.

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -158,7 +158,7 @@ Follow the [VPN guide for Bring Your Own Devices (BYOD)](https://docs.google.com
 
 We store who has access to GOV.UK tooling in a git repo, [GOV.UK User Reviewer][govuk-user-reviewer] and use automation to alert when people should not have access. It's a private repo so you won't be able to access it until step 3 is completed.
 
-Get your tech lead to create a pull request to add you to the [tech users](https://github.com/alphagov/govuk-user-reviewer/blob/main/config/govuk_tech.yml). The repo readme specifies the [format new starters should have, with integration admin access](https://github.com/alphagov/govuk-user-reviewer/tree/main#properties-in-the-config-file). The `ssh_username` is in the format of `firstnamelastname` and is utilised in [8. Get SSH access to integration](#7-get-ssh-access-to-integration)).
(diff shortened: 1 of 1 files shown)


source/manual/get-started.html.md
```diff
@@ -158,7 +158,7 @@ Follow the [VPN guide for Bring Your Own Devices (BYOD)](https://docs.google.com
...
 
-Get your tech lead to create a pull request to add you to the [tech users](https://github.com/alphagov/govuk-user-reviewer/blob/main/config/govuk_tech.yml). The repo readme specifies the [format new starters should have, with integration admin access](https://github.com/alphagov/govuk-user-reviewer/tree/main#properties-in-the-config-file). The `ssh_username` is in the format of `firstnamelastname` and is utilised in [8. Get SSH access to integration](#7-get-ssh-access-to-integration)).
+Get your tech lead to create a pull request to add you to the [tech users](https://github.com/alphagov/govuk-user-reviewer/blob/main/config/govuk_tech.yml). The repo readme specifies the [format new starters should have, with integration admin access](https://github.com/alphagov/govuk-user-reviewer/tree/main#properties-in-the-config-file). The `ssh_username` is in the format of `firstnamelastname` and is utilised in [9. Get AWS access](#9-get-aws-access)).
```
kevindew, line 161: I think you've got a mistake here as the ssh_username is not used in AWS (at least as far as I'm aware) and is just a part of our pre-replatformed environment.

kevindew (commented):
Oops I did the same PR as you earlier today: https://github.com/alphagov/govuk-developer-docs/pull/4112 I'll close mine and review yours

source/manual/get-started.html.md
AgaDufrat: Ah yes, thanks! Now correctly updated the link.

ChrisBAshton (approved):

PR #4118, opened 2023-07-24
Add information on fixing an incident

anatron:

### DIFF
--- a/source/manual/incident-fixing.html.md
+++ b/source/manual/incident-fixing.html.md
@@ -1,0 +1,21 @@
+---
+owner_slack: "@ana.botto"
+title: Fixing an Incident
+parent: "/manual.html"
+layout: manual_layout
+type: learn
+section: Incidents
+---
+
+When there is an incident in an application you are supporting, it is important to make a quick fix so that users can get back to use the application as usual.
+
+You should look into the logs on Sentry and Logit to see where the error is coming from.
+
+You should have a look at the latest commits to the application on Github. When seeing the latest commits you will be able to see if there are any commits that are related to the error.
+
+When discovering what commit could be the one causing the issue, and it is not an issue that can be solved straight away, you should revert that commit.
+This revert commit can be made on Github from the pull request where the issue could have come from. This should ensure that your revert commit is fix forwards too.
+
+You should test this in integration to see if it has fixed the issue.
+
+Sometimes it could be necessary to freeze automatic deployment so we can see if we have identified the correct fix.


sivsubra-tw:
@anatron - Can you add the description on this PR? It'll be super helpful if the description also contains a screenshot of what the text actually looks like on the UI (particularly if I can see the hierarchy of pages etc)

source/manual/incident-fixing.html.md
```diff
@@ -0,0 +1,21 @@
...
+title: Fix an Incident
+parent: "/manual.html"
+layout: manual_layout
+type: learn
+section: Incidents
+---
+
+When there is an incident in an application you are supporting, it is important to make a quick fix so that users can get back to use the application as usual.
+
+You should look into the logs on Sentry and Logit to see where the error is coming from.
```
FelixMarcusMillne, line 12: I wouldn't say 'should' in this manner, as it isn't a mandatory requirement - we're offering a list of suggestions of where someone _might_ look

FelixMarcusMillne: Worth trying to respect the sort of semantic requirements that are used for RFCs, as they're clear and well documented:

https://datatracker.ietf.org/doc/html/rfc2119

source/manual/incident-fixing.html.md
```diff
@@ -0,0 +1,21 @@
...
+
```
FelixMarcusMillne, line 13: I'd combine a few of these paragraphs into a bulleted list, Sentry and Logit should be separate lines, checking the past history of the commit messages another

source/manual/incident-fixing.html.md
```diff
@@ -0,0 +1,21 @@
...
+You should have a look at the latest commits to the application on Github. When seeing the latest commits you will be able to see if there are any commits that are related to the error.
+
+When discovering what commit could be the one causing the issue, and it is not an issue that can be solved straight away, you should revert that commit.
```
FelixMarcusMillne, line 16: We haven't outlined here what we should do if we _can't_ simply revert a commit, or provided examples of why that might be the case...

source/manual/incident-fixing.html.md
```diff
@@ -0,0 +1,21 @@
...
+This revert commit can be made on Github from the pull request where the issue could have come from. This should ensure that your revert commit is fix forwards too.
```
FelixMarcusMillne, line 17: This is probably explaining how to use Github again - worth putting in a link to an external source rather than reproducing it here

source/manual/incident-fixing.html.md
```diff
@@ -0,0 +1,21 @@
...
+
+You should test this in integration to see if it has fixed the issue.
```
FelixMarcusMillne, line 19: This is probably worth being a harder requirement - I would expect that _any_ fixes _must_ be tested in integration

source/manual/incident-fixing.html.md
```diff
@@ -0,0 +1,21 @@
...
+
+Sometimes it could be necessary to freeze automatic deployment so we can see if we have identified the correct fix.
```
FelixMarcusMillne, line 21: We don't have anything here about freezing deployments or _how_ we would test a fix in integration - I think this is documented elsewhere in the docs, so we should put in a link to that content

source/manual/incident-fixing.html.md
```diff
@@ -0,0 +1,21 @@
+---
+owner_slack: "@ana.botto"
+title: Fix an Incident
+parent: "/manual.html"
+layout: manual_layout
+type: learn
+section: Incidents
+---
+
+When there is an incident in an application you are supporting, it is important to make a quick fix so that users can get back to use the application as usual.
```
FelixMarcusMillne, line 10: There is existing documentation on responding to an incident - the two should be linked to each other for easy access and discovery in a crisis

FelixMarcusMillne (changes requested):

PR #4119, opened 2023-07-24
Add information about reverting commits

anatron:


ChrisBAshton (commented):
Hi 👋 apologies for the drive-by comment - what is the motivation for documenting this?
As a general rule, we want to avoid documenting things that have no specific GOV.UK context and which are fairly easy to Google.

PR #4127, opened 2023-07-26
Additional instructions on how to clear production and staging K8s when retiring an app

KludgeKML:

### DIFF
--- a/source/manual/retiring-an-application.html.md
+++ b/source/manual/retiring-an-application.html.md
@@ -122,6 +122,18 @@ Once the PR is merged ([Example PR][]), the app pods will automatically be remov
 [govuk-helm-charts] (https://github.com/alphagov/govuk-helm-charts/)
 [Example PR] (https://github.com/alphagov/govuk-helm-charts/pull/1236)
 
-## 16. Archive the repo
+## 17. Delete the app's Kubernetes resources (if applicable)
+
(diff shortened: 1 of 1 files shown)


source/manual/retiring-an-application.html.md
```diff
@@ -122,6 +122,36 @@ Once the PR is merged ([Example PR][]), the app pods will automatically be remov
...
+
+You don't need to worry about deleting pods and replicasets. Once you've done the deletion,
+confirm that everything has been removed with:
+
+```
+kubectl get all | grep <app-name>
+```
+
+(Pods might remain, but as long as they are in the Terminating state, you can just wait and
+they'll remove themselves)
```
sengi, line 153:
```suggestion
## 17. Delete the app's Kubernetes resources

In the staging and production environments, deleting an Argo CD app will not automatically delete the Kubernetes resources that it manages, such as `deployments`, `services`, `jobs` and `cronjobs`.

You will need to list and delete these resources manually, for example:

    kubectl get all | grep <app-name>

    kubectl delete deploy/<app-name> svc/<app-name> job/<app-name>-upload-assets
```

sengi (approved):
Thanks for improving the docs. I've suggested a much simpler / pared down version that should suffice while being clearer and quicker to follow.

PR #4131, opened 2023-07-28
Resolve unnecessary absolute links

kevindew:
This is a find and replace which updates internal links in the developer docs to be host and scheme independent.

The motivation for this was that I got frustrated at clicking links when developing locally and ending up out of my local version of dev docs and into the production one.


ChrisBAshton (approved):
Thanks! We should consider a lint rule to prevent these creeping in again too 🤔

kevindew:
Yup, would be good to automate it. There's a few legitimate cases (examples of using APIs, accessibility statement) but I think generally they're not in markdown so a markdown one could do it.

PR #4141, opened 2023-08-03
Transfer ownership of some repos to Platform Engineering

ChrisBAshton:
These repos were owned by Platform Security and Reliability (PSR), however it makes more sense for them to be owned by Platform Engineering (PE). PE are retiring Puppet and Jenkins this quarter, which will render all of these repositories obsolete. In the meantime, therefore, there is no point in PSR attempting to make any improvements to these repos.

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -270,7 +270,7 @@
   type: Data science
 
 - repo_name: docker-collectd-plugin
-  team: "#govuk-platform-security-reliability-team"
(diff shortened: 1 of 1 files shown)


ChrisBAshton:
On second look, there are slightly more repos than the 3 we discussed yesterday @sengi, but hopefully they all make sense to transfer!

sengi (approved):
Ooh thanks, looking forward to archiving some of these! 😄

PR #4143, opened 2023-08-03
Delete obsolete doc about Unicorn workers.

sengi:
Redirect to the "Scale your app" page in GOV.UK k8s cluster docs.

Also update the reference from the 2nd-line "drills" page.

Tested: redirect works when running locally with `SKIP_PROXY_PAGES=true NO_CONTRACTS=true bundle exec middleman server`


robinjam (approved):
Nice!

PR #4145, opened 2023-08-04
Update CodeCommit / GitHub down playbooks.

sengi:
@rtrinque and I tried this out this afternoon and found a few bits missing so thought I'd update it.

I reckon we could simplify the [deploy-with-argo-cd](https://github.com/alphagov/govuk-developer-docs/blob/sengi/dril/source/manual/github-unavailable.html.md#deploy-the-image-to-kubernetes-using-argo-cd) procedure by a few steps if we use `kubectl patch` instead of the Argo CD UI, but I ran out of time to figure out the details for that so I'll leave that for a possible future improvement.

### DIFF
--- a/source/manual/github-unavailable.html.md
+++ b/source/manual/github-unavailable.html.md
@@ -6,59 +6,62 @@ layout: manual_layout
 parent: "/manual.html"
 ---
 
-## Backup to AWS CodeCommit
+## AWS CodeCommit
 
-We mirror all non-archived GitHub repositories tagged with `govuk` to AWS CodeCommit. This is achieved using a ["Mirror repositories" GitHub workflow](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/mirror-repos.yml), which uses a [`github_action_mirror_repos_role` IAM role configured in Terraform](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/github/mirror.tf).
+We mirror all non-archived GitHub repositories tagged with `govuk` to AWS CodeCommit via the ["Mirror repositories" GitHub Actions workflow](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/mirror-repos.yml). The workflow uses the [`github_action_mirror_repos_role` IAM role](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/github/mirror.tf).
 
-### Checkout code in AWS CodeCommit
+### Build and deploy an app when GitHub is unavailable
 
-Since our deployment images are named via git tags, we'll need to checkout our code from AWS CodeCommit in order to make sure we're deploying the correct image and have an up-to-date copy of the code.
+If GitHub.com is down, you can still clone a GOV.UK repo from CodeCommit and
+build and deploy a container image from your workstation.
 
-Setup your local environment for [checking out code from CodeCommit](/manual/howto-checkout-and-commit-to-codecommit.html)
+Run the following commands from the root directory of the repository.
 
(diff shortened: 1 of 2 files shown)


source/manual/github-unavailable.html.md
```diff
@@ -6,59 +6,67 @@ layout: manual_layout
...
-1. Go to "DISABLE AUTO-SYNC" then select "ok" when prompted "Disable Auto-Sync?"
-1. Close the "APP DETAILS" menu, then select the three dots next to the Deploy object for the part of the application you'd like to redeploy. For example, `account-api-worker` would be for deploying the `account-api-workers`.
-1. Go to the "LIVE MANIFEST" and select "EDIT". Find the "image", then update it to the new image tag you'd like to deploy. (You can double-check you have the correct image name by doing a `git log` after [checking out the code from CodeCommit](/manual/howto-checkout-and-commit-to-codecommit.html)).
-1. Click "Save" and Argo will start the deployment, which should complete in under ten minutes.
+### Deploy the image to Kubernetes using Argo CD
+
+> ⚠️ This procedure will disable all automatic deployments not only in the
+> environment you are working on but also in higher environments. For example,
+> performing this procedure on integration will prevent all automatic
+> deployments in integration, staging and production.
```
ChrisBAshton, line 49: Nitpick: this can all be on one line (i.e. no hard line-breaks), for consistency with the rest of the document.

source/manual/github-unavailable.html.md
```diff
@@ -6,59 +6,67 @@ layout: manual_layout
...
+    ```
+
+1. Log into Argo CD (for example in [integration](https://argo.eks.integration.govuk.digital/)).
+1. Open the `app-config` application.
+1. Select the `App Details` button near the top of the page.
+1. Scroll down to the bottom of the page and press `Disable auto-sync`. Argo will prompt you before actually disabling auto-sync.
+1. Go back to the main `Applications` page (the first page you see when you log into Argo CD).
+1. Go into the details page for the application you wish to deploy, either by selecting its card from the main page or using the `Search applications...` box and selecting it from the drop-down list.
+1. Select the `App details` button near the top of the page.
+1. Scroll down to the bottom of the page and press `Disable auto-sync`. Argo will prompt you before actually disabling auto-sync.
```
ChrisBAshton, line 64: This took me a re-read or two. I think it can be shortened and some of the steps re-used:

```suggestion
1. Disable the auto-sync for the `app-config` application by:
  1. Clicking the `Applications` link on the sidebar. then opening the `app-config` application (you can use the search bar at the top of the page if you can't see it)
  1. Select the `App Details` button near the top of the page.
  1. Scroll down to the bottom of the page and press `Disable auto-sync`. Argo will prompt you before actually disabling auto-sync.
1. Repeat the steps above for the application you wish to deploy (e.g. swap `app-config` for `collections-publisher` in the steps above).
```

ChrisBAshton (approved):
Thanks @sengi ! Two nitpicks below, but good to merge 👍

source/manual/github-unavailable.html.md
sengi: ok, the rest of the doc already isn't consistent though.

sengi: Actually on second thoughts, nobody should have to spend time on this. I'll reformat the repo and add an mdlint check so that nobody else has to waste time on this in future 👍

source/manual/github-unavailable.html.md
sengi: Good idea — I usually try to avoid loops in playbooks but I think this is a good exception. Makes the intent a bit clearer in this case I think. Thanks 👍

PR #4158, opened 2023-08-11
Update ownership of bouncer

JonathanHallam:
It's been decided that this sits better with the publishing platform team because they own transition and transition config.

Doc on change - https://github.com/alphagov/govuk-rfcs/blob/main/rfc-161-technical-direction-in-publishing-2023.md

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -40,7 +40,7 @@
 
 - repo_name: bouncer
   type: Transition apps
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
Perhaps worth linking to https://github.com/alphagov/govuk-rfcs/blob/main/rfc-161-technical-direction-in-publishing-2023.md too 👍 
LGTM

PR #4161, opened 2023-08-14
Switch to multi-page navigation

ChrisBAshton:
WIP.

This is to enable the migration of https://govuk-kubernetes-cluster-user-docs.publishing.service.gov.uk/ into the Developer Docs.

1. Breadcrumb seems to link to topmost ancestor, e.g. 'Manual'
   instead of 'Alerts' on http://127.0.0.1:4567/manual/alerts/redis.html
2. Duplicate "Icinga Alerts" nav option alongside "Manual -> Icinga
   Alerts" option, on http://127.0.0.1:4567/manual/alerts/.
3. Pages under "Icinga Alerts" aren't in alphabetical order.
4. "Page not found" is a link in the navigation.
5. Top level menu items in sidebar are in random order and sidebar
   behaviour differs depending on the link clicked. Probably something
   to do with manual_layout.html.erb.


ChrisBAshton:
I fear I've hit a big blocker in moving to multi-page navigation: it's totally reliant that every doc we import has a h1 tag. We'd probably need to make a new version of [tech-docs-gem](https://github.com/alphagov/tech-docs-gem) to cope with it better (e.g. adding an option to have a h1 generated from the page title instead, which is how the Developer Docs have worked to date - see d34c91b6d785d1822fa85d9ef7cb76475d9db7d9 for the annoying workaround!

Screenshot 2023-08-25 at 11 51 02

ChrisBAshton:
Superseded by #4186

PR #4181, opened 2023-08-22
Remove workaround for broken Jenkins GitHub token.

sengi:
h/t @dj-maisy for tracking down the (ridiculously well-hidden copies of the) expired token.

Also fix headings to match house style.


dj-maisy (approved):
🚀

PR #4183, opened 2023-08-23
Update the backup and restore databases doc

deborahchua:
We just carried out a drill following these docs.


source/manual/howto-backup-and-restore-in-aws-rds.html.md
```diff
@@ -224,6 +247,12 @@ sudo psql -U aws_db_admin -h local-links-manager-postgres -d local-links-manager
 
 This requires updating the CNAME for `local-links-manager-postgres`.
 
+You will need to switch your AWS role to `poweruser` apply the changes.
```
sengi, line 250: This seems very strange — `admin` is supposed to be a superset of `poweruser` so this step shouldn't be necessary, at least in theory. Did it fail when you used your admin role and if so, do you still have the error?

sengi (approved):

source/manual/howto-backup-and-restore-in-aws-rds.html.md
sengi: Asking cos if this really was necessary to complete the procedure then we probably have a misconfiguration somewhere (or I'm missing something 😅)

deborahchua: Ah, this is me not remembering the available privileges - we originally tried `readonly` and when that didn't work tried `poweruser` 😅 
I just logged in and tested it with `admin` and that worked 👍 I've updated the doc now.

sengi: Ahah cool no worries! Thanks for checking that. I think in that case you're probably best off removing that line altogether then, because someone following the playbook will already have admin creds at that point in the instructions (so if anything it's a little confusing to randomly call it out on that particular step).

sengi:
Thanks for checking and updating this btw! Super helpful to know that this still works (albeit with this rather complicated procedure!)

PR #4185, opened 2023-08-25
Tweaked Kubernetes cheat sheet

ChrisBAshton:
We were lacking information on switching AWS environment. I've also frontloaded the 'common tasks' section and split out the 'prerequisites' section, as the former will be referred to time and again, and the latter should only be needed once.

Also nudges devs to use the [Rails console sandbox mode](https://andycroll.com/ruby/play-in-a-sandbox-in-production/) where possible.


gclssvglx (approved):
Looks great @ChrisBAshton thanks 👍

PR #4186, opened 2023-08-25
Import Kubernetes docs

ChrisBAshton:
This PR pulls in the existing documentation from [govuk-kubernetes-cluster-user-docs](https://github.com/alphagov/govuk-kubernetes-cluster-user-docs).

It also applies the minimum changes to retain the hierarchical/nested structure of the Kubernetes docs (which originally made use of the ["multipage documentation" feature of the Tech Docs Template](https://tdt-documentation.london.cloudapps.digital/configure_project/structure_docs/#build-a-multipage-documentation-site)), without breaking all of our other docs. This is achieved by adding a new layout that closely mimics the multipage documentation feature, and only opting the `/kubernetes` folder into it.

We're unable to use the native feature provided by the template (as attempted in #4161), for several reasons:

1. The existing documentation is massive in scale, and flat in structure. To avoid making the sidebar navigation too messy, a significant piece of work would need to be done to rearchitect the current docs, choose appropriate page weights, group related docs into nested folders, and so on.
2. A lot of documentation is pulled in externally at compile time - a great proportion of documentation doesn't actually live in the Developer Docs repo. The mutlipage doc feature has a requirement that every single page has an explicit h1 tag, which we could add relatively easily to the Developer Docs' "local" documentation, but would be a significant undertaking to roll out across all of the external repositories whose 'docs' folders we import. NB, currently we derive the h1 from each document's 'title' frontmatter, which has been working just fine and avoids duplication.
3. There is an [/analytics microsite](https://docs.publishing.service.gov.uk/analytics/) hosted on the Developer Docs, which has its own structure and layout, and which does not play nicely when enabling the multipage feature.

## Screenshots

The design isn't _identical_ to the previous Kubernetes Cluster User Docs site - see screenshots below. Whilst they're extremely similar, the Developer Docs version does lack the collapsible menu in the sidebar. This is an optimisation we can look at adding later.

On the separate Kubernetes Cluster User Docs site (old):

> Screenshot of the old site

On the Developer Docs site (new):

> Screenshot of the new multipage_layout layout

---

Trello: https://trello.com/c/KRDdRuEw/186-merge-eks-docs-into-dev-docs


sihugh (approved):
🚀

sengi (approved):
I'm good with losing the collapsible menu tbh! 👍 

Thanks so much for doing this!

PR #4193, opened 2023-08-31
Update CDN documentation to reference new govuk-fastly workflow

samsimpson1:

### DIFF
--- a/source/manual/cdn.html.md
+++ b/source/manual/cdn.html.md
@@ -20,40 +20,18 @@ The CDN is responsible for retrying requests against the [static mirror](/manual
 
 Most of the CDN config is versioned and scripted:
 
-- [govuk-cdn-config](https://github.com/alphagov/govuk-cdn-config/)
-- [govuk-cdn-config-secrets](https://github.com/alphagov/govuk-cdn-config-secrets)
+- [govuk-fastly]
+- [govuk-fastly-secrets]
 
-Some configuration isn't scripted, such as logging. The www, bouncer and assets services send logs to S3 which can be [queried](/manual/query-cdn-logs.html). These logging endpoints are configured directly in the Fastly UI.
+The www, bouncer and assets services send logs to S3 which can be [queried](/manual/query-cdn-logs.html).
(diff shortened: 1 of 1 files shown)


source/manual/cdn.html.md
```diff
@@ -155,7 +133,7 @@ Note that banning JA3s is potentially risky. If we get it wrong, we could ban a
 
 As well as blocking based on source IP address or JA3 fingerprint, we can also block abusive traffic based on headers, URL paths or any arbitrary criteria about the request that we can specify using VCL. This requires care and testing, but can be nonetheless a valueable incident response tool for mitigating DoS and spam attacks.
 
-We have a mechanism for including VCL code from the private `govuk-cdn-config-secrets` repo into the Fastly config, so that mitigations we make during an attack are not published to the public repo for the attacker to see and work around. An example of this is [alphagov/govuk-cdn-secrets#133](https://github.com/alphagov/govuk-cdn-config-secrets/pull/133/files).
+We have a mechanism for including VCL code from the private `govuk-cdn-config-secrets` repo into the Fastly config, so that mitigations we make during an attack are not published to the public repo for the attacker to see and work around. An example of this is [can be found here](https://github.com/alphagov/govuk-fastly-secrets/blob/45381b4e0af022dfb4ba51c485551e023c2e2e12/secrets.yaml#L27-L35).
```
sengi, line 136:
```suggestion
We have a mechanism for including VCL code from the private `govuk-cdn-config-secrets` repo into the Fastly config, so that mitigations we make during an attack are not published to the public repo for the attacker to see and work around. An example of this [can be found here](https://github.com/alphagov/govuk-fastly-secrets/blob/45381b4e0af022dfb4ba51c485551e023c2e2e12/secrets.yaml#L27-L35).
```

source/manual/cdn.html.md
```diff
@@ -20,40 +20,18 @@ The CDN is responsible for retrying requests against the [static mirror](/manual
 
 Most of the CDN config is versioned and scripted:
 
-- [govuk-cdn-config](https://github.com/alphagov/govuk-cdn-config/)
-- [govuk-cdn-config-secrets](https://github.com/alphagov/govuk-cdn-config-secrets)
+- [govuk-fastly]()
+- [govuk-fastly-secrets]()
```
sengi, line 24: You can omit the `()` here if you like.
```suggestion
- [govuk-fastly]
- [govuk-fastly-secrets]
```

sengi (approved):

PR #4195, opened 2023-09-04
Update CDN failover documentation

robinjam:
This documentation is out of date for two reasons:

- Our DNS config has moved from alphagov/govuk-dns-config to alphagov/govuk-dns-tf, and is now deployed using Terraform Cloud
- We have a new CloudFront failover distribution, which supports more functionality than the previous one

---

[Trello card](https://trello.com/c/51gZjhxH/3292-update-tech-docs-with-step-by-step-procedure-to-switch-from-fastly-to-cloudfront-and-back-1)
This PR is based on the documented [Cloudfront failover test procedure](https://docs.google.com/document/d/1RkYwVJAqEwQkWyAuOwAcsFu0Sg6qEspNhtYf8HFZIiY/edit)


source/manual/fall-back-to-aws-cloudfront.html.md
```diff
@@ -9,43 +9,29 @@ parent: "/manual.html"
...
-    minutes (from the start of the incident) and then start the fail over
-  - For a minor Fastly outage (less than 30% of requests failing) wait at least
-    15 minutes (from the start of the incident) before failing over - discuss
-    the decision with the GOV.UK SMT Escalations person
-- While you wait, confirm the new CNAMES for www-cdn.production.govuk.service.gov.uk and assets.publishing.service.gov.uk
-  - This [Draft PR to Failover to AWS CloudFront](https://github.com/alphagov/govuk-dns-config/pull/714) shows the CNAMES you need to change, and how to test that they are correct
-  - You can also get the CNAMEs to use for the secondary CDN from the AWS CLI: <div class="highlight"><pre class="highlight plaintext"><code># www.gov.uk
-  gds aws govuk-production-readonly aws cloudfront list-distributions --query "DistributionList.Items[?Comment=='WWW'].DomainName | [0]"
-  # assets.publishing.service.gov.uk
-  gds aws govuk-production-readonly aws cloudfront list-distributions --query "DistributionList.Items[?Comment=='Assets'].DomainName | [0]"</code></pre></div>
```
ChrisBAshton, line 40: These snippets look useful to copy into the PR description for govuk-dns-tf#69.

ChrisBAshton (commented):
Thanks James.

1) Agreed, let's provisionally say you can remove the section about escalations / grace periods. I'll check with Senior Tech and our SPM, but I agree it's no longer appropriate. Whilst it feels like a big change in policy, I think the previous policy was written fairly ad-hoc, so we shouldn't worry about making changes.
2) Agreed we can also remove the manual steps for changing DNS, relying on the automated approach instead. Or at the very least, move the manual steps to their own section, which you can point to as a "if you're unable to apply the Terraform for whatever reason" fallback.

(Aside: you'll probably use the manual steps for the drill instructions you'll write as part of this card, for describing how to fail over in Staging).

3) One last thing: the [test steps](https://docs.google.com/document/d/1RkYwVJAqEwQkWyAuOwAcsFu0Sg6qEspNhtYf8HFZIiY/edit) had a note about reducing the DNS TTL to 30 seconds, to make it quicker to switch back to Fastly. I'll defer to an SRE or Tech Architect on this one, but that sounds like a sensible addition to make to the draft govuk-dns-tf PR while we're here 👍

richardTowers:
> I’m not sure what the rationale is for manually performing the failover from the AWS and GCP consoles, and then merging a PR in govuk-dns-tf (formerly govuk-dns-config) to do the same thing in an automated way(?) - should I remove the manual failover section?

It's a funny story actually. The only time we've had to do this failover in anger, we tried to do it with Terraform. Can you guess which CDN Terraform uses to host the AWS / GCP terraform providers?

tl;dr - we found that terraform doesn't work reliably if Fastly is having a major incident. This _might_ be fixed now that we're using Terraform Cloud (e.g. if it caches terraform providers more reliably than Jenkins did), but it's very hard to say for sure.

source/manual/fall-back-to-aws-cloudfront.html.md
```diff
@@ -9,43 +9,29 @@ parent: "/manual.html"
...
-  - The records should look like `d0000000000000.cloudfront.net.` (with 0s replaced with letters and numbers)
 - When you are ready to fail over:
-  - Manually update the CNAME record for www-cdn.production.govuk.service.gov.uk in GCP and AWS
-  - Manually update the CNAME record for assets.publishing.service.gov.uk in GCP and AWS
+  - Manually update the CNAME record for `www-cdn.production.govuk.service.gov.uk` in both GCP and AWS to point to `d2uru0rbwu9sz0.cloudfront.net`
+  - Manually update the CNAME record for `assets.publishing.service.gov.uk` in both GCP and AWS to point to `d3d1psggl6xqso.cloudfront.net`
```
richardTowers, line 30: Previously we were a bit coy about publishing the CloudFront URLs (even though they were effectively public, see `dig +short gov.uk TXT`), hence the complicated stuff with the AWS CLI. I think the worry was that they'd get used for web scraping or something, and we'd run up a big bill. Now that they're in front of origin instead of S3, they're potentially a new vector for DoS attacks, but making them secret doesn't exactly secure that threat.

I'm happy for them to be published if we think that we're managing the risk of them being misused in other ways (or that we've at least considered those threats).

robinjam:
Thanks @ChrisBAshton!

> 3. One last thing: the [test steps](https://docs.google.com/document/d/1RkYwVJAqEwQkWyAuOwAcsFu0Sg6qEspNhtYf8HFZIiY/edit) had a note about reducing the DNS TTL to 30 seconds, to make it quicker to switch back to Fastly. I'll defer to an SRE or Tech Architect on this one, but that sounds like a sensible addition to make to the draft govuk-dns-tf PR while we're here 👍

Good shout! I've gone with 5 minutes though, last time Fastly went down it took several hours to recover and we don't want people hammering our DNS too hard in the meantime.

source/manual/fall-back-to-aws-cloudfront.html.md
robinjam: This is a good shout, I'll update this document to say "get the URLs from the linked PR" instead.

robinjam:
> It's a funny story actually. The only time we've had to do this failover in anger, we tried to do it with Terraform. Can you guess which CDN Terraform uses to host the AWS / GCP terraform providers?
> 
> tl;dr - we found that terraform doesn't work reliably if Fastly is having a major incident. This _might_ be fixed now that we're using Terraform Cloud (e.g. if it caches terraform providers more reliably than Jenkins did), but it's very hard to say for sure.

Oh no 🙈 

Thanks, this is useful context that I should add to this document. I think it probably makes sense to leave the manual failover steps there rather than splitting them out to a separate section too.

source/manual/fall-back-to-aws-cloudfront.html.md
ChrisBAshton: Ah yes. They'll be [removed soon](https://trello.com/c/vhKRQfLu/3296-delete-the-old-static-cloudfront-distributions), but we do want a working snippet in the meantime (or no snippet at all).

robinjam:
@ChrisBAshton Requesting a re-review as I think I've addressed all the feedback (n.b. I will raise a separate PR for the drill)

ChrisBAshton (approved):
LGTM 👍 thanks!

PR #4204, opened 2023-09-11
Document that GitHub access relies on invitation

ChrisBAshton:
It was unclear whether Terraform handled this or not. There is indeed still a manual step whereby the user being added must accept the email invite before they're added to the alphagov org.


KludgeKML (approved):
LGTM

PR #4208, opened 2023-09-12
Add command to start the app on the README

mirlabragagds:
Add command `./startup.sh` to the readme to start the application

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,12 @@ Disabling proxy pages means you'll get a "Not Found" error if you try to access 
 bundle exec rake
 ```
 
+### Running the app
+
+```
+./startup.sh
+```
+
 ### Updating the template
 
 You can pull down the latest version of the Tech Docs template by running:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,6 +39,12 @@ Disabling proxy pages means you'll get a "Not Found" error if you try to access
 bundle exec rake
 ```
 
+### Running the app
+
+```
+./startup.sh
```
ChrisBAshton, line 45: This repeats the information in `### Before running the app`, but I agree that a `Running the app` section would be good to have!

I'd be tempted to just rename `Before running the app` to `Running the app`, and move this line of code to the top of that section. Then the rest of that section should explain how you'll hit GitHub API limits by default, so you'll either need a GITHUB_TOKEN env var or will need to use the SKIP_PROXY_PAGES option.

ChrisBAshton (approved):
Approved - feel free to press Merge!

PR #4215, opened 2023-09-19
[WIP] Pages to be deleted

mirlabragagds:
List pages need to be deleted and references to it:

[Redis Alerts](https://docs.publishing.service.gov.uk/manual/alerts/redis.html) 
[Rollback Puppet if Jenkins is broken](https://docs.publishing.service.gov.uk/manual/rollback-puppet-if-jenkins-is-broken.html) 

SSH to Machines page been deleted and the reference to that page are also been updated, still few references need to be updated which are going to address in another PR as those references need more context to update:
[SSH into machines](https://docs.publishing.service.gov.uk/manual/howto-ssh-to-machines.html)


source/manual/access-to-licensing-for-third-parties.html.md
```diff
@@ -77,7 +77,7 @@ environments is as follows:
 
 Accessing machines using SSH in production and staging requires Production Admin Access. Third parties are usually only given Production Deploy Access, which only allows SSH in the integration environment.
 
-Follow the [instructions for connecting to a machine via SSH](/manual/howto-ssh-to-machines.html#connecting-with-plain-ssh). The machine classes you will need are `licensing_frontend` and `licensing_backend`. You will need to be on the VPN.
+The machine classes you will need are `licensing_frontend` and `licensing_backend`. You will need to be on the VPN.
```
ChrisBAshton, line 80: If we're removing the SSH doc, I'd be tempted to add an inline command here for the Licensify devs:

```suggestion
The machine classes you will need are `licensing_frontend` and `licensing_backend`. You will need to be on the VPN.

Connect to the instance by running, for example, `gds govuk connect -e production ssh licensing_backend`.
```

source/manual/rollback-puppet-if-jenkins-is-broken.html.md
ChrisBAshton, line 1: Might be worth holding onto this until the Jenkins switch-off (in the coming weeks). I'd suggest we keep this PR open as a draft until that's ready. We can still pre-emptively delete most Puppet/Jenkins related docs, in this draft PR, so that we're ready to merge when the switch-off has happened.

ChrisBAshton (commented):
Thank you, this looks good to me, just some comments below (the main one being that whilst I think we should get this PR ready, we should hold off on actually merging it until nearer the switch-off).

Also just a general point to please [explain why you're making the change](https://gds-way.cloudapps.digital/standards/source-code/working-with-git.html#commit-messages) in each Git commit message, e.g. why is it safe to delete a particular doc?

Thank you! ⭐

source/manual/rollback-puppet-if-jenkins-is-broken.html.md
ChrisBAshton: Yeah I think that'll be fine. Usually it would be a bit risky because of docs falling out of sync with `main` and then having lots of conflicts to resolve, but the docs you'll be editing/deleting are most likely docs that haven't been touched in ages and won't be touched while this PR is open, so we should have no conflicts 👍 I'm also not too worried about the size of the PR to review.

mirlabragagds: That's good! Thanks you 👍

Anusha-GDS: Awesome, Thank you @ChrisBAshton

Anusha-GDS:
FYI @ChrisBAshton, we have created two PRs One could address the [pages that can be deleted now](https://github.com/alphagov/govuk-developer-docs/pull/4232) and the other one can address the [pages to be deleted later](https://github.com/alphagov/govuk-developer-docs/pull/4231) after the completion of replatformation. we are thinking to delete this PR as we no longer going to work on this PR. Could you please let us know any comments or suggestions on this approach, Thank you !!

PR #4222, opened 2023-09-28
Clarify how users are removed from GitHub

ChrisBAshton:
Repeats info written in https://github.com/alphagov/govuk-user-reviewer#addingremoving-users, but worth having here for completion.


KludgeKML (approved):
LGTM!

PR #4228, opened 2023-10-03
Add a step to 'Transition a site to GOV.UK' docs

AgaDufrat:
The bouncer deploy script was migrated to terraform and now doesn't run automatically when a subdomain is added to transition-config. There is also an additional domain verification step.

Given the plans to consolidate transition-config into the transition app, updating a doc seems sufficient at this stage.  The product team can later decide on automation of this process.

### DIFF
--- a/source/manual/transition-a-site.html.md
+++ b/source/manual/transition-a-site.html.md
@@ -155,7 +155,25 @@ advance, and to be lowered to 300 seconds (5 minutes). It can be raised
 again once everyone is happy there is no need to switch back - normally
 the day after.
 
-### 8) Point the domain at us
+### 8) Add the domain to Fastly
+
+#### Verify the subdomain (for a *.gov.uk subdomain only)
+
+For security reasons, *.gov.uk subdomains are delegated upon request. When the steps below are not completed, there will be a `Domain 'gov.uk' is owned by another customer` error when the terraform apply command runs.
+
+We need to:
+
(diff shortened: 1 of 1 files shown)


source/manual/transition-a-site.html.md
```diff
@@ -155,7 +155,23 @@ advance, and to be lowered to 300 seconds (5 minutes). It can be raised
...
+
+We need to:
+
+1. Obtain the TXT record by submitting Fastly support request. You will need to provide 'Production bouncer' service ID and the subdomain you want to add. See an [example of support request](https://support.fastly.com/hc/en-us/requests/700875).
+
+2. Add the DNS record to [govuk-dns-tf](https://github.com/alphagov/govuk-dns-tf) and apply terraform configuration.
+
+  It's not possible to add additional records on a subdomain if a CNAME already exists. In such case Fastly accepts setting the record on a subdomain prefixed with `_fastly` as a proof of ownership. See an [example code change][code change].
+
+3. Manually trigger govuk-fastly 'Plan and apply' run in [Terraform Cloud UI](https://app.terraform.io/app/govuk/workspaces/govuk-fastly/runs). Review the plan with changes to `module.bouncer-production.fastly_service_vcl.service` and apply the configuration.
```
brucebolt, line 170: My (very poor) understanding is that steps 1 and 2 only need to be done for `*.gov.uk` domains, whereas step 3 needs to be done for all domains.

Is that right?  If so, could it be worth separating this into two parts?

AgaDufrat: yes, good point. I will split it up.

brucebolt (approved):
Looks good to me 👍.  Thanks for making the update 🎖️.

PR #4240, opened 2023-10-06
Document asset size limit

ChrisBAshton:
This question comes up quite often on 2nd line.

The limit is 500MB, imposed by our nginx config.
In practice, publishers would do well to keep far below this limit, for a variety of reasons.

That said, the biggest asset we have is 425 MB in size. I'll avoid linking to it directly here in case that info gets abused, but it's referenced in this Trello comment: https://trello.com/c/snmx8e2L/368-test-uploading-of-massive-pdf-to-whitehall-3#comment-5fe0840c15463b4930dd4c8e

### DIFF
--- a/source/manual/assets.html.md
+++ b/source/manual/assets.html.md
@@ -53,6 +53,14 @@ following features:
   the replacement. Currently only Whitehall and Specialist Publisher support
   this.
 
+### File size limit
+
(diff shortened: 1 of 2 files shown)


ollietreend (approved):
Thanks for documenting this @ChrisBAshton 🎉

PR #4242, opened 2023-10-10
Add note about encryption at rest

ChrisBAshton:
This question comes up every so often, so let's document it.

Note from @martynipratt: when you request encryption on S3 buckets, it returns that the use of a KMS key is "False" but that doesn’t mean it’s not encrypted - just that it doesn’t use KMS as the encryption method.


martynipratt (approved):
Yes

PR #4249, opened 2023-10-17
Update GA4 pageview schema and attributes

AshGDS:
Adds all of the new pageview attributes to our docs.

### DIFF
--- a/data/analytics/attributes.yml
+++ b/data/analytics/attributes.yml
@@ -1,3 +1,11 @@
+- name: ab_test
+  description: Used in the page_view object to record whether an ab test is running on the page.
+  value: content attribute of meta tag govuk:ab-test
+  example: BankHolidaysTest:Z
+  required: no
+  type: string
+  redact: false
+
 - name: action
   description: Shows the kind of thing the user has done.
   value:
@@ -7,8 +15,8 @@
   redact: false
 
 - name: browse_topic
-  description: Used in the page_view object to record the content section
-  value: content attribute of meta tag govuk:section
+  description: Used in the page_view object to record the parent browse topic for the page
+  value: content attribute of meta tag govuk:ga4-browse-topic
   example: working at sea
   required: no
   type: string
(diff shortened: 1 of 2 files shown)


data/analytics/attributes.yml
```diff
@@ -1,3 +1,11 @@
+- name: ab_test
+  description: Used in the page_view object to record wether an ab test is running on the page.
```
andysellick, line 2: Typo: `wether`

data/analytics/attributes.yml
```diff
@@ -22,6 +30,22 @@
...
+  value: true/undefined depending on the presence of data-ga4-cookie-banner on the page
+  example: "true"
+  required: no
+  type: string (boolean)
+  redact: false
+
+- name: devolved_nations_banner
+  description: Used in the page_view object to record the countries listed in a devolved nations banner on the page
+  value: The value of data-ga4-devolved-nations-banner on the page
+  example: "England, Scotland, Wales"
```
andysellick, line 44: Just to check, does the actual data for this example contain spaces between the countries?

data/analytics/attributes.yml
```diff
@@ -120,6 +152,14 @@
   type: string (number)
   redact: false
 
+- name: intervention
+  description: Used in the page_view object to record the presence of an intervention banner on the page
+  value: true/undefined depending on the presence of data-ga4-intervnetion-banner on the page
```
andysellick, line 157: Typo: `intervnetion`

data/analytics/attributes.yml
```diff
@@ -185,11 +225,35 @@
   type: string
   redact: false
 
+- name: navigation_list_type
+  description: Used in the page_view object to record the navigation list type
```
andysellick, line 229: Can we explain any more? This is in the browse, right? Could we something like `Used in the page view object to record the type of page being shown in browse, either 'curated' or 'alphabetical'`? (I think the alternative is alphabetical, but check).

data/analytics/attributes.yml
```diff
@@ -185,11 +225,35 @@
...
+  value: content attribute of meta tag govuk:navigation_list_type
+  example: curated
+  required: no
+  type: string
+  redact: false
+
+- name: navigation_page_type
+  description: Used in the page_view object to record the navigation page type
```
andysellick, line 237: Can we say any more about this? I'm not certain what the navigation page type is.

andysellick (commented):
Nice updating 👍 

Just a few typos and minor comments.

data/analytics/attributes.yml
AshGDS: @andysellick Yep 👍

data/analytics/attributes.yml
AshGDS: @andysellick Will update it but I'll have to take your word for it, I can't find anywhere where it gives an example of the values, I can only find the pages from the trello card which have `curated` as the value!

AshGDS:
Thanks @andysellick - should be ready for review again :+1:

data/analytics/attributes.yml
andysellick: Maybe ask a PA?

AshGDS: @andysellick Have done this now and updated the description as a result, so should be ready to review again :+1:

andysellick: Nice work 👍

andysellick (approved):
Great stuff, good to see the record getting some improvements 👍

PR #4250, opened 2023-10-17
Update Seal hosting

robinjam:
Seal is no longer hosted on Heroku 🎉

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -1235,7 +1235,6 @@
 - repo_name: seal
   team: "#govuk-platform-security-reliability-team"
   management_url: https://dashboard.heroku.com/apps/govuk-seal
-  production_hosted_on: heroku
   type: Utilities
   sentry_url: false
   dashboard_url: false


ChrisBAshton (approved):

data/repos.yml
```diff
@@ -1235,7 +1235,6 @@
 - repo_name: seal
   team: "#govuk-platform-security-reliability-team"
   management_url: https://dashboard.heroku.com/apps/govuk-seal
-  production_hosted_on: heroku
```
ChrisBAshton, line 1238: Is there a github actions equivalent in Dev Docs config?

ChrisBAshton: (If not, should there be?)

robinjam: Note that we determine whether or not a repo is an "app" by [looking at `production_hosted_on`](https://github.com/alphagov/govuk-developer-docs/blob/cc8d6ef03b100f7a6e3187b8c71f3672dd7a9d04/app/repo.rb#L49-L51), so assuming we agree that it no longer counts as an "app", we probably want it to be `nil`.

ChrisBAshton: Fair 👍

ChrisBAshton (approved):

PR #4254, opened 2023-10-19
Remove router-postgres repo

catalinailie:
router-postgres repo was archived


ChrisBAshton (approved):
LGTM - you could also mark it as `retired: true` with an explanation as to what purpose the repo served, but given this repo was so short-lived, seems ok to remove all references to it.

PR #4255, opened 2023-10-20
Update ckan support page

nicholsj:
This updates the documentation around CKAN api calls:

- adds a note about the default number of rows returned and how to change it
- adds an example for returning all datasets from a specific publisher (a common use case)


source/manual/data-gov-uk-2nd-line.html.md
```diff
@@ -281,6 +287,9 @@ There's some [API documentation][dgu-api-docs] aimed at general users of data.go
 
 # View a publisher record
  https://data.gov.uk/api/3/action/organization_show?id=government-digital-service
+
+#List datasets from a specific publisher
```
ChrisBAshton, line 291: Nitpick:

```suggestion
# List datasets from a specific publisher
```

ChrisBAshton (approved):

PR #4261, opened 2023-10-20
Update the checklist for archiving a GitHub repo.

sengi:
Try to reduce the burden involved in archiving a repo (since we want people to archive disused stuff and not be put off by a big toily process) by clarifying that there are situations where e.g. it may be overkill to add a signpost to the README (in addition to the banner that GitHub adds anyway).

Also simplify some of the wording and make it a bit clearer as to why the remaining toil is needed.

### DIFF
--- a/source/manual/retiring-a-repo.html.md
+++ b/source/manual/retiring-a-repo.html.md
@@ -6,21 +6,31 @@ layout: manual_layout
 parent: "/manual.html"
 ---
 
-## 1. Update README
+## 1. Update the README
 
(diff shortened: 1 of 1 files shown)


sengi:
It's nuts that GitHub doesn't at least give the option to close out all the remaining PRs/issues when archiving a repo, especially given the bizarre way that open PRs/issues in archived repos still show up everywhere but can't be closed or deleted.

ChrisBAshton (approved):
Looks sensible to me - thanks Chris 👍

PR #4262, opened 2023-10-20
Added vulnerability Scanner docs

martynipratt:
Adding URL to internal Vulnerability Scanner policy

### DIFF
--- a/source/manual/logging.html.md
+++ b/source/manual/logging.html.md
@@ -107,3 +107,11 @@ viewing logs. It is responsible for:
 - parsing user queries written in Lucene/KQL into Elastic
 - querying the Elasticsearch indices
 - displaying the results
+
+### Vulnerability Scanners
+
+Occasionally you may see some logs that look suspicious. For example, see the `controller` part of the following made up example:
+
+`http_request_duration_seconds_count { action="1", container="app", controller="../../../../../etc/passwd", endpoint="metrics", job="govuk", namespace="apps", pod="static-abc123-de7f" }`
+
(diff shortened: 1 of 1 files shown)


source/manual/logging.html.md
```diff
@@ -107,3 +107,6 @@ viewing logs. It is responsible for:
 - parsing user queries written in Lucene/KQL into Elastic
 - querying the Elasticsearch indices
 - displaying the results
+
+### Vulnerability Scanners
+Please check this internal documentation what to do with suspected [Vulnerability Scanners](https://docs.google.com/document/d/1BZ_SBPuZmO8pseniqj1tlq7pss9iU3JslqQHp_JS9Wg/edit#heading=h.17fwpvs3mt0x)
```
ChrisBAshton, line 112: I think we can be a little more prescriptive in when the documentation is useful:

```suggestion

Occasionally you may see some logs that look suspicious. For example, see the `controller` part of the following made up example:

`http_request_duration_seconds_count { action="1", container="app", controller="../../../../../etc/passwd", endpoint="metrics", job="govuk", namespace="apps", pod="static-abc123-de7f" }`

This sort of thing often originates from vulnerability scanners and isn't necessarily something to worry about. However, you should consider _how_ such activity ended up leaking into your logs, and whether or not there's anything you could/should do to patch it up. Read our [internal guidance on the issue](https://docs.google.com/document/d/1BZ_SBPuZmO8pseniqj1tlq7pss9iU3JslqQHp_JS9Wg/edit#heading=h.17fwpvs3mt0x).
```

ChrisBAshton (approved):

PR #4263, opened 2023-10-20
Update AB test docs to work with govuk-fastly

richardTowers:
We've retired govuk-cdn-config in favour of govuk-fastly, which means these docs need updating now.

The new process is significantly simpler than the old one (no Jenkins, no generating VCL configuration manually and committing it), but there are still some complexities around diffing the VCL which I've tried to call out.

I haven't mentioned the other govuk-fastly caveat, which is that it hits all three environments at the same time. That's surprising, but already documented in the govuk-fastly repo. For AB tests, it's probably not a massive problem having the environments locked together.

https://trello.com/c/jKWA7Nu3/543-ab-test-content-for-find-your-utr-number


cynthia-anya (approved):
LGTM 👍

PR #4269, opened 2023-10-25
Tweak mirror docs

ChrisBAshton:
See commits for details.

### DIFF
--- a/source/manual/fall-back-to-mirror.html.md
+++ b/source/manual/fall-back-to-mirror.html.md
@@ -28,7 +28,7 @@ If [Fastly, our primary CDN](/manual/cdn.html), cannot fetch a page from our bac
 
 ## How are the mirrors populated
 
-Every day the [govuk-mirror-sync] cronjob crawls the www and assets domains, saves pages and assets to disk and then uploads the files to the primary S3 bucket. The [govuk-mirror] repository contains the code responsible for crawling and saving pages to disk.
+Every day the [govuk-mirror-sync cronjob (configured in govuk-helm-charts)][govuk-mirror-sync configuration] crawls the www and assets domains, saves pages and assets to disk and then uploads the files to the primary S3 bucket. The [govuk-mirror] repository contains the code responsible for crawling and saving pages to disk.
 
 S3 Replication automatically copies any changes from the primary S3 bucket to the secondary S3 bucket. This is configured in [govuk-aws].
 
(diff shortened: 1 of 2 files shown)


source/manual/fall-back-to-mirror.html.md
```diff
@@ -26,15 +26,15 @@ We use multiple mirrors across various AWS regions and GCP to ensure redundancy
 
 If [Fastly, our primary CDN](/manual/cdn.html), cannot fetch a page from our backend servers (becuase of a timeout or a 5xx error), then Fastly will attempt to serve a page from a mirror in order of priority.
 
-## How are the mirrors populated
+## How are the mirrors populated?
 
-Every day the [govuk-mirror-sync] cronjob crawls the www and assets domains, saves pages and assets to disk and then uploads the files to the primary S3 bucket. The [govuk-mirror] repository contains the code responsible for crawling and saving pages to disk.
+Every day the [govuk-mirror-sync cronjob (configured in govuk-helm-charts)][govuk-mirror-sync configuration] crawls the www and assets domains, saves pages and assets to disk and then uploads the files to the primary S3 bucket. The [govuk-mirror] repository contains the code responsible for crawling and saving pages to disk.
```
richardTowers, line 31: Suggest you don't need the parenthetical here - we didn't used to say "configured in govuk-puppet" for everything configured in puppet, and lots of things are configured there so it would be noisy to say this every time. The link already makes it clear where it's configured.

```suggestion
Every day the [govuk-mirror-sync cronjob][govuk-mirror-sync configuration] crawls the www and assets domains, saves pages and assets to disk and then uploads the files to the primary S3 bucket. The [govuk-mirror] repository contains the code responsible for crawling and saving pages to disk.
```

richardTowers (approved):
One nitpick - take it or leave it :)

source/manual/fall-back-to-mirror.html.md
ChrisBAshton: Thanks - I agree. My aim was to differentiate between the job configuration, and the job in Argo (they could both be called "govuk-mirror-sync cronjob", with the destination unclear until you click on the link.

I'll give it another tweak before merging.

source/manual/fall-back-to-mirror.html.md
```diff
@@ -43,10 +43,13 @@ Certain page types aren't included in the mirrors:
...
 ## Troubleshooting
 
-Check the logs of the govuk-mirror-sync job to see there are any errors during crawling, saving pages or uploading to S3.
+Check the [logs of the govuk-mirror-sync job in Argo][govuk-mirror-sync job] to see there are any errors during crawling, saving pages or uploading to S3.
 
 Check buckets in AWS S3 or GCP to see if they are populated.
 
-[govuk-mirror]: https://github.com/alphagov/govuk-mirror
-[govuk-mirror-sync]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
+You can fetch pages directly from the mirrors by specifying the `Backend-Override` header, e.g. `curl -H 'Backend-Override: mirrorS3' https://www.gov.uk`. The [allowed values](https://github.com/alphagov/govuk-fastly/blob/68427d372df05fd23c6851cfbea610845c6c3997/modules/www/www.vcl.tftpl#L258-L289) are `mirrorS3`, `mirrorS3Replica` and `mirrorGCS`.
```
sengi, line 50:
```suggestion
You can fetch pages directly from the mirrors by specifying the `Backend-Override` header, e.g. `curl -H 'Backend-Override: mirrorS3' https://www.gov.uk`. The [allowed values](https://github.com/alphagov/govuk-fastly/blob/68427d3/modules/www/www.vcl.tftpl#L258-L289) are `mirrorS3`, `mirrorS3Replica` and `mirrorGCS`.
```

sengi (approved):

PR #4275, opened 2023-10-26
Documentation for dependency review action

wjmurphy:
https://trello.com/c/6KeYUJ98/3332-document-dependency-review


source/manual/dependency-review.html.md
```diff
@@ -0,0 +1,37 @@
...
+[reusable-workflow]: https://github.com/alphagov/govuk-infrastructure/pull/966
+
+## Add Dependency Review to a project
+
+To use the Dependency Review reusable workflow, add the following job to the `jobs` section of your CI workflow:
+
+```yaml
+dependency-review:
+    name: Dependency Review scan
+    uses: alphagov/govuk-infrastructure/.github/workflows/dependency-review.yml@main
```
ChrisBAshton, line 22: I believe we normally use two spaces for YAML indentation:

```suggestion
dependency-review:
  name: Dependency Review scan
  uses: alphagov/govuk-infrastructure/.github/workflows/dependency-review.yml@main
```

source/manual/dependency-review.html.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+## Where to find Security Alerts
+
+Alerts can always be found in the job logs. Additionally, there is a job summary dislpayed beneath the Github action run. Here changes are summarised along with any vulnerabilities which may have been found.
```
ChrisBAshton, line 27:
```suggestion
Alerts can always be found in the job logs. Additionally, there is a job summary displayed beneath the GitHub Action run. Here, changes are summarised along with any vulnerabilities found. 
```

source/manual/dependency-review.html.md
```diff
@@ -0,0 +1,37 @@
...
+    uses: alphagov/govuk-infrastructure/.github/workflows/dependency-review.yml@main
+```
+
+## Where to find Security Alerts
+
+Alerts can always be found in the job logs. Additionally, there is a job summary dislpayed beneath the Github action run. Here changes are summarised along with any vulnerabilities which may have been found. 
+
+## Dealing with Security Alerts
+
+Currently Dependency Review is configured to [find critical issues][dependency-review-config]. This is becuase it cannot be configured to pass the test in an event where new vulnerabilities are found. Therefore, we would only want to fail the test in the event of a critical vulnerability to ensure developer time is not wasted on potential false positves. Over time, we intent to tune the dial so this becomes more strict.
```
ChrisBAshton, line 31:
```suggestion
Currently, Dependency Review is [configured to find "critical" issues][dependency-review-config]. If issues are found, the check automatically 'fails' and therefore blocks merging of the PR. We therefore want to avoid false positives and wasting developer time, so have set the bar high and only alert on critical issues. Over time, we intent to tune the dial so this becomes more strict. 
```

source/manual/dependency-review.html.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Where to find Security Alerts
+
+Alerts can always be found in the job logs. Additionally, there is a job summary dislpayed beneath the Github action run. Here changes are summarised along with any vulnerabilities which may have been found. 
+
+## Dealing with Security Alerts
+
+Currently Dependency Review is configured to [find critical issues][dependency-review-config]. This is becuase it cannot be configured to pass the test in an event where new vulnerabilities are found. Therefore, we would only want to fail the test in the event of a critical vulnerability to ensure developer time is not wasted on potential false positves. Over time, we intent to tune the dial so this becomes more strict. 
+
+Alerts can be resolved only by fixing the issue and running the tests again. Vulnerability alerts provide a link to the [Github advisory][gh-advisory] database where advice on resolving the issue can be found. Dismissing an alert requires [adding a GHSA ID to the inline configuration][skip-alert]. As the workflow is a communnal resource, this should be done sparingly and include a well detailed comment about why it is being skipped.
```
ChrisBAshton, line 33:
```suggestion
Alerts can be resolved only by fixing the issue and running the tests again. Vulnerability alerts provide a link to the [GitHub advisory][gh-advisory] database where advice on resolving the issue can be found. Dismissing an alert requires adding the GitHub Advisory Database ID (GHSA) to the inline configuration [via the `allow-ghsas` property][skip-alert]. As the workflow is a communal resource, this should be done sparingly and include a well detailed comment about why it is being skipped. 
```

Hmmm is there no option for a more local override? It might be that the advisory is valid for one of our repos, but not another. I imagine we could extend the `dependency-review` config to pass an optional argument of ghsa IDs to the reusable workflow, overriding what's there centrally.

Out of scope for this documentation PR, but we should consider writing up a card for it and coming back to update this later.

ChrisBAshton (commented):
Thanks @wjmurphy , looks good to me, some small suggestions below.

ChrisBAshton (approved):

PR #4287, opened 2023-11-02
Update analytics section

andysellick:
## What
Updates the [GA4 implementation record](https://docs.publishing.service.gov.uk/analytics/) part of the dev docs to match actual live events.

## Why
Part of the GA4 migration.

## Visual changes
Includes a new jump menu for the events pages, to allow easier navigation to different events within the same category.

Screenshot 2023-11-02 at 16 53 44

Trello card: https://trello.com/c/JQPHy26m/661-check-implementation-record-is-up-to-date

### DIFF
--- a/data/analytics/events.yml
+++ b/data/analytics/events.yml
@@ -903,14 +903,14 @@
         - name: link_domain
         - name: link_path_parts
 
-- name: print page
+- name: printing
   events:
-    - name: print page
+    - name: print page button
       implemented: true
       priority: low
       description: When a 'print this page' button is used
       examples:
-        - text: Page with print button
+        - text: Register a trust as a trustee
           link: https://www.gov.uk/guidance/register-a-trust-as-a-trustee
       tracker: event_tracker
       data:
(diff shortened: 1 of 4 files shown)


data/analytics/events.yml
```diff
@@ -552,6 +555,9 @@
       implemented: true
```
AshGDS, line 672: This event is meant for links to gov.uk files which aren't in an attachment component, for example on this page: https://www.gov.uk/government/news/the-2023-government-chemist-conference-presentations the "Welcome address- Julian Braybrook" link.

For attachment components, there are two different `event_name`s that an attachment link can have. If it's a HTML attachment, it has `event_name: navigation`, and if it's a file, it has `event_name: file_download`.
These seem to be covered already by http://127.0.0.1:4567/analytics/event_html_attachments.html and http://127.0.0.1:4567/analytics/event_links.html#heading-3 (File attachment links)

Would you be able to make this event match the "Welcome address- Julian Braybrook" link? Thanks

data/analytics/events.yml
```diff
@@ -1058,11 +1058,24 @@
 - name: see all link
   events:
     - name: see all link click
```
AshGDS, line 1144: There's also a 'link' version of this event, when 'See all updates' goes to another page: https://www.gov.uk/hmrc-internal-manuals/air-passenger-duty

So that will need to be added. Feel free to do it in a separate PR (or get me to do it)

AshGDS (changes requested):
Thanks for doing all of this work 🎉 Looks good, have requested a couple small changes :+1:

data/analytics/events.yml
andysellick: Thanks @AshGDS, have updated the `Update links GA4 event data` commit with these changes, let me know if it's alright.

data/analytics/events.yml
andysellick: Thanks @AshGDS, great detail. Have updated the `Update see all updates GA4 event data` commit with this addition.

AshGDS (approved):

PR #4289, opened 2023-11-03
Update pagerduty drill docs

kentsanggds:
Update docs for pagerduty drills as they have been moved to alertmanager

## Reference

https://trello.com/c/atcbGHh1/2156-move-pagerduty-drill-out-of-puppet

### DIFF
--- a/source/manual/pagerduty.html.md
+++ b/source/manual/pagerduty.html.md
@@ -29,13 +29,12 @@ See the [Pagerduty documentation on how to schedule an override](https://support
 ## PagerDuty drill
 
 Every week we test PagerDuty to make sure it can phone to alert us to
-any issues. This happens every Wednesday morning at 10am UTC.
+any issues. This happens every Wednesday morning at 10am BST.
 
-Cron creates a file on disk. Icinga raises a critical alert when that
-file exists with an Icinga contact group set so that PagerDuty is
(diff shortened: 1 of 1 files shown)


source/manual/pagerduty.html.md
```diff
@@ -47,22 +46,22 @@ anyone else being phoned.
 
 To trigger the drill manually, follow these steps:
 
-1. SSH onto the monitoring box:
+1. Exec onto the alertmanager box after logging onto the production cluster:
```
AgaDufrat, line 49: Can we elaborate on logging into production cluster? I get an error when I try to run it:
```
Error from server (Forbidden): pods "alertmanager-kube-prometheus-stack-alertmanager-0" is forbidden: User "aga.dufrat-poweruser" cannot get resource "pods" in API group "" in the namespace "monitoring"
```

I've done:
- `eval $(gds aws govuk-production-poweruser -e --art 8h)`
- `k config use-context production`

kentsanggds: hmmm, I wonder if special permissions are needed to access it?

kentsanggds: I've updated the notes to say that for manual triggering of pagerduty we should ask on the #platform-engineering channel.

AgaDufrat (approved):

PR #4302, opened 2023-11-15
Add Kibana urls to app pages

KludgeKML:
Adds logit links - these links for apps go to a now-3h view in Kibana/Logit.io filtered to the appropriate app, showing time/request/status and message, giving people an easy way into viewing appropriate logs.

https://trello.com/c/n8ZCNgzv/189-add-logit-links-into-apps-pages-links-in-dev-docs

### DIFF
--- a/app/repo.rb
+++ b/app/repo.rb
@@ -147,6 +147,18 @@ class Repo
     repo_data["dashboard_url"] || default_url
   end
 
+  def kibana_url
+    return if repo_data["kibana_url"] == false
+
+    kibana_url_for(app: app_name, hours: 3)
+  end
+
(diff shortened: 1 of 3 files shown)


1pretz1 (approved):
Good idea! Happy with changes and conversation offline below for further context:

Do the links solely go to the production instances?

> Yeah, solely to prod - we could add other ones, but this is the sort of foot-in-the-door to debugging things on logit.

May also be useful to add a fuzzy search field so other docs could narrow down the query but that's just an idea

> Good idea about options for particular docs, I will think about a way to access that (but if this is okay, then would be useful to merge now)

PR #4307, opened 2023-11-17
Fix "Access an EKS cluster" first-time setup guide.

sengi:
- Use an example kubectl command that works for all of the roles, not just admin.
- Simplify the wording, add some important background info, omit unnecessary/confusing details and generally streamline the process.


ChrisBAshton (approved):
LGTM 👍

sengi:
Thanks!

PR #4322, opened 2023-11-27
Remove stylelint-config-gds repo

MuriloDalRi:
The GOV.UK Design System owns this repo and it shouldn't be listed here, just as their other repos aren't.


robinjam (approved):
Good catch!

PR #4324, opened 2023-11-29
Update dgu docs k8s

kentsanggds:
Additional sections of dev docs located with mention of PaaS or PaaS commands updated, also updated the architecture diagram for DGU.

## Reference 

https://trello.com/c/2qXYzZYb/1239-switchover-paas-apps-to-k8s-cluster


data/repos.yml
```diff
@@ -222,7 +222,7 @@
 
 - repo_name: datagovuk-tech-docs
   production_url: https://guidance.data.gov.uk/
-  production_hosted_on: paas
+  production_hosted_on: gh-pages
```
samsimpson1, line 225: Looks good but this might not be necessary, this repo doesn't have that value set

https://github.com/alphagov/govuk-developer-docs/blob/6ae7c5cbb0ecef4fa609c7c085a1579d9af1a5fc/data/repos.yml#L552-L556

samsimpson1 (approved):

PR #4325, opened 2023-11-29
Consistent naming of Tech 2nd Line, and more

nicholsj:
We should use the term "Technical 2nd Line" consistently throughout the docs, so I fixed some here. 
Plus some other tidying up, and removed the potentially confusing word "downtime" :)


KludgeKML (approved):
Thanks, looks good!

PR #4330, opened 2023-12-05
Analytics browse update

andysellick:
## What
Update the implementation record (GA4 events data).

- add action link events in browse pages
- remove `link_path_parts` references as no longer in use
- update `taxonomy_all` and `taxonomy_all_ids` as now strings, no longer objects

## Why
Part of the GA4 migration.

Trello card: https://trello.com/c/lstNprcf/717-add-tracking-to-action-link

### DIFF
--- a/data/analytics/attributes.yml
+++ b/data/analytics/attributes.yml
@@ -369,19 +369,17 @@
   redact: false
 
 - name: taxonomy_all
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
Thanks for doing this :+1: Added a non-blocking comment.

PR #4336, opened 2023-12-08
Update Ruby doc

deborahchua:
The use of a service account/machine user was a workaround to replicate someone creating a trivial PR, but is not recommended as it is a security loophole.


AgaDufrat (approved):

sengi:
Thanks!

PR #4340, opened 2023-12-11
Update guidance: `test` "job" is now `CI` "workflow"

ChrisBAshton:
As described in the documentation change, we no longer define
`test` jobs, and now define `CI` workflows, which work slightly
differently insofar as they're made up of several parallel jobs.

This commit makes a few other updates, including:

- Remove reference to small steps within a CI workflow:
  We've actually been heading in the opposite direction: lots of
  small steps defined within the CI workflow. This is partly because
  we're now making use of a lot of shared workflows, so what we lose
  in ease of local development, we gain in cross-GOV.UK consistency.
- Remove old example code
  Looking at the Content Publisher GitHub Actions, we no longer have
  the complexity referred to in these docs. I've deleted the docs
  that are no longer relevant.
  I've kept the instructions for testing Ruby gems, as that is still
  being applied.
- Move "previous recommendation" section further up:
  Moving it closer to where we're justifying our choices for "when
  the CI workflow should run".
  Also creates a new subsection for running CI on demand.
- Softening the language around omitting names for jobs and steps.
  In practice, a great many of our repos needlessly define job and
  step names that override the job/step key, but there is no harm
  in doing so.

### DIFF
--- a/source/manual/test-and-build-a-project-with-github-actions.html.md
+++ b/source/manual/test-and-build-a-project-with-github-actions.html.md
@@ -7,11 +7,9 @@ parent: "/manual.html"
 ---
 
 [GitHub Actions](https://github.com/features/actions) is an automated workflow
-system that GOV.UK uses for [Continuous Integration (CI)][ci]. We also have
-[Jenkins](/manual/testing-projects.html) (and previously had Concourse) which
-provides similar functionality. In [GOV.UK RFC 123][] we decided that
-GitHub Actions is the preferred platform for GOV.UK CI usage where the wider
-platform integration of Jenkins is not required.
+system that GOV.UK uses for [Continuous Integration (CI)][ci]. In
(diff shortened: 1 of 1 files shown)


AgaDufrat (approved):
What do you think about bumping the “Reuse workflows where possible’ section up to the top of this doc? It has fairly general content, relevant for the later, more specific parts of this manual. We also mention it in the “Name of CI workflow and jobs”. 
When thinking about dev process for these, we would need to define the reusable workflow first to then us in the CI, so seems like a logical order. But I don't have strong feelings about it, so good to merge as it is ;)

ChrisBAshton:
Thanks @AgaDufrat - I like the suggestion, but also like the top-down approach of "this is the file", "this is how it's structured", and then eventually "this is where we define our reusable workflows"...

But I think a future restructure would be good, making it easier to dive in and find what you need (and can probably be done in a far less wordy way than the current iteration). 👍

PR #4342, opened 2023-12-12
Add basic documentation for new app on k8s

csutter:
This adds a documentation page on how to get a brand new app up and running on the Kubernetes platform.

### DIFF
--- a/source/kubernetes/create-app/index.html.md
+++ b/source/kubernetes/create-app/index.html.md
@@ -1,0 +1,113 @@
+---
+title: Create a new application on the GOV.UK Kubernetes Platform
+weight: 20
+layout: multipage_layout
+---
+
+# Create a new application
+
+## Assumptions
+
+This guide assumes the following:
+
+- You have completed all the steps in the ["get started" guide][get-started] of the developer docs
+- You are deploying an application named `example` (make sure to replace this with your
+  application name in all steps as needed)
+- You have pushed the application source code to a Github repository under the `alphagov`
+  organisation with the `govuk` and `container` labels applied (this triggers setup of a number of
+  Github Actions secrets from the organisation and permissions required for workflows)
+- You have a Sentry project configured for your application (through `govuk-infrastructure`)
+
+## Allow your application to be built as a Docker image
+
+Add a `Dockerfile` to your application. If your application is written in Ruby, this should normally
+be based on the [GOV.UK Ruby Images][images]. This should do everything necessary to create a
+standalone image for your application.
+
+> **Note**
+> The best way to establish what you need to do is to find an existing GOV.UK application similar to
+> yours, and adapt its `Dockerfile`.
+
+## Add Github Actions workflows
+
+You will need to add (at least) the following workflows:
+
+- `CI` (to test your application)
+- `release` (to create new releases and version tags)
+- `deploy` (to build/push an image and trigger deployment by updating image tags)
+
+Your Continuous Integration (CI) workflow will depend on the linting and testing needs of your
+application, but will act as a dependency for further workflows to block deployment in case of
+failure.
+
+For the `release` and `deploy` workflows, you can copy the workflow definitions from an existing
+GOV.UK application (in `.github/workflows`).
+
+## Add image tag files to `govuk-helm-charts`
+
+[GOV.UK Helm Charts](https://github.com/alphagov/govuk-helm-charts) uses image tag files to keep
+track of the image tag deployed into each environment. Add an initial image tag file named after
+your application into the `image-tags` folder for each of the four environments, i.e.
+`charts/app-config/image-tags/{test,integration,staging/production}/example`:
+
+```
+image_tag: latest
+automatic_deploys_enabled: true
+promote_deployment: false
+```
+
+## Add initial secrets to AWS Secrets Manager
+
+### Sentry
+
+In order to set up Sentry for your app, you will need a _DSN_ from the Sentry UI. This must be
+manually added as a key/value pair to the `govuk/common/sentry` secret in all three environments.
+
+## Add app to `app-config` chart values
+
+### Start with integration
+
+Start off by adding your app to the integration environment by adding it to the array of application
+Helm chart values in `charts/app-config/values-integration.yaml`. This can be as simple as:
+
+```yml
+- name: example
+```
+
+This will use the [generic-govuk-app][generic-app] Helm chart to set up your application. You can
+find the default values in `values.yaml` in the chart folder, and amend your configuration's
+`helmValues` value to override them.
+
+For example, if your app needs a `RETICULATE_SPLINES` environment variable set to "1" in a given
+environment, your integration configuration could look like this:
+
+```yml
+- name: example
+  helmValues:
+    extraEnv:
+      - name: RETICULATE_SPLINES
+        value: 1
+```
+
+There are configuration options available for a wide range of different needs your application may
+have, from mounting volumes to disabling the default asset upload.
+
+### Validate your deployment
+
+You can now trigger your Deploy Github action to push an image to AWS ECS, and visit [Integration
+Argo CD][argo]. After a short wait, your new application will be available to find on the
+dashboard.
+
+### Continue with the other environments
+
+Repeat the steps above for the other environments' values files (`values-staging.yaml` and
+`values-production.yaml`).
+
+## Congratulations!
+
+You've deployed an application to the Kubernetes platform!
+
+[argo]: https://argo.eks.integration.govuk.digital/applications
+[generic-app]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/generic-govuk-app/
+[get-started]: /manual/get-started.html
+[images]: https://github.com/alphagov/govuk-ruby-images


sengi (approved):
Thanks, this looks great!

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
+---
+title: Create a new application on the GOV.UK Kubernetes Platform
+weight: 20
+layout: multipage_layout
+---
+
+# Create a new application on the GOV.UK Kubernetes Platform
```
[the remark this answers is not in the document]
csutter: Good question 💡 

The manual does explicitly call it `GOV.UK Kubernetes cluster documentation for developers` - so it _is_ there, but also kinda redundant!

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+
+This guide assumes the following:
+
+- You have completed all the steps in the ["get started" guide](/manual/get-started.html) of the
+  developer docs
+- You are deploying an application named `example` (make sure to replace this with your
+  application name in all steps as needed)
+- You have generated an initial application and pushed it to a Github repository under the
+  `alphagov` organisation with the `govuk` and `container` labels applied (this triggers setup of a
+  number of Github Actions secrets from the organisation and permissions required for workflows)
```
theseanything, line 19: Nit: simplification
```suggestion
- You have pushed the application source code to a Github repository under the
  `alphagov` organisation with the `govuk` and `container` labels applied (this triggers setup of a
  number of Github Actions secrets from the organisation and permissions required for workflows)
```

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+- You have set up the application in [GOV.UK Docker](https://github.com/alphagov/govuk-docker) (not
+  covered by this guide) so that others can run it locally (this will also help you understand what
+  environment configuration your application needs, especially if it is an existing application
+  rather than a brand new one)
```
theseanything, line 23: Don't think this assumption necessary to make to get an application deployed to our k8s cluster. Maybe remove it here? or make it a suggestion later on?

theseanything: Maybe what's more pertinent is that the application is containerised? I.e. has a Dockerfile

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+- You have a Sentry project configured for your application (through `govuk-infrastructure`)
+
+## Add Github Actions workflows
+
+### Add CI workflow
+
+What your Continuous Integration (CI) workflow looks like will depend on the linting and testing
+needs of your application, but will act as a dependency for further workflows to block deployment in
+case of failure. A Rails application would usually run RSpec unit and integration/system tests and
+Rubocop.
```
theseanything, line 33: Nit: Avoid going into specific examples here - (makes this document less brittle)

```suggestion
What your Continuous Integration (CI) workflow looks like will depend on the linting and testing
needs of your application, but will act as a dependency for further workflows to block deployment in
case of failure.
```

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+    name: Release
+    uses: alphagov/govuk-infrastructure/.github/workflows/release.yml@main
+    secrets:
+      GH_TOKEN: ${{ secrets.GOVUK_CI_GITHUB_API_TOKEN }}
+```
+
+> **Note**
+> You may need to manually create a `v0` tag (e.g. against the most recent commit) in your
+> repository for the `release` workflow to pick up on and start automatically creating releases with
+> incremented versions.
```
theseanything, line 66: We need to be specific that they need to create a GitHub release and not just a git tag.
```suggestion
> You may need to manually create a `v0` GitHub release (e.g. against the most recent commit) in your
> repository for the `release` workflow to pick up on and start automatically creating releases with
> incremented versions.
```

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+    branches: [main]
+
+jobs:
+  release:
+    if: github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success'
+    name: Release
+    uses: alphagov/govuk-infrastructure/.github/workflows/release.yml@main
+    secrets:
+      GH_TOKEN: ${{ secrets.GOVUK_CI_GITHUB_API_TOKEN }}
+```
```
theseanything, line 61: Because these workflow definition get updated not infrequently - we might wanna point to an existing workflow definition (e.g. in Whitehall) to copy, rather than have another bit of code in docs to maintain.

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+    needs: build-and-publish-image
+    uses: alphagov/govuk-infrastructure/.github/workflows/deploy.yml@main
+    with:
+      imageTag: ${{ needs.build-and-publish-image.outputs.imageTag }}
+      environment: ${{ inputs.environment || 'integration' }}
+    secrets:
+      WEBHOOK_TOKEN: ${{ secrets.GOVUK_ARGO_EVENTS_WEBHOOK_TOKEN }}
+      WEBHOOK_URL: ${{ secrets.GOVUK_ARGO_EVENTS_WEBHOOK_URL }}
+      GH_TOKEN: ${{ secrets.GOVUK_CI_GITHUB_API_TOKEN }}
+```
```
theseanything, line 119: Same as above - this is already out of date

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Add image tag files to `govuk-helm-charts`
+
+[GOV.UK Helm Charts](https://github.com/alphagov/govuk-helm-charts) uses image tag files to keep
+track of the image used for the active deployment in each environment. Add an initial tag file named
+after your application into the `image-tags` folder for each of the four environments, i.e.
+`charts/app-config/image-tags/{test,integration,staging/production}/example`:
+
+```
+image_tag: v1
```
theseanything, line 129: If we use latest, the image tag is valid right off the bat.
```suggestion
image_tag: latest
```

theseanything: It'll obviously get updated to v1 as soon as the deploy happens

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+      WEBHOOK_TOKEN: ${{ secrets.GOVUK_ARGO_EVENTS_WEBHOOK_TOKEN }}
+      WEBHOOK_URL: ${{ secrets.GOVUK_ARGO_EVENTS_WEBHOOK_URL }}
+      GH_TOKEN: ${{ secrets.GOVUK_CI_GITHUB_API_TOKEN }}
+```
+
+## Add image tag files to `govuk-helm-charts`
+
+[GOV.UK Helm Charts](https://github.com/alphagov/govuk-helm-charts) uses image tag files to keep
+track of the image used for the active deployment in each environment. Add an initial tag file named
+after your application into the `image-tags` folder for each of the four environments, i.e.
```
theseanything, line 125: Nit: simplification
```suggestion
[GOV.UK Helm Charts](https://github.com/alphagov/govuk-helm-charts) uses image tag files to keep
track of the image tag deployed into each environment. Add an initial image tag file named
after your application into the `image-tags` folder for each of the four environments, i.e.
```

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+`charts/app-config/image-tags/{test,integration,staging/production}/example`:
+
+```
+image_tag: v1
+automatic_deploys_enabled: true
+promote_deployment: true
```
theseanything, line 131: Nit: we don't promote this initial deployment
```suggestion
promote_deployment: false
```

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Add initial secrets to AWS Secrets Manager
+
+### Sentry
+
+In order to set up Sentry for your app, you will need a _DSN_ from the Sentry UI. This must be
+manually added as a key/value pair to the `govuk/common/sentry` secret in all three environments.
+
+You can use `gds-cli aws govuk-{integration/staging/production}-poweruser -l` to log in to the AWS
+Console for each environment and make the change.
```
theseanything, line 142: I'd remove this, we don't need to tell people how to log into AWS in this document.

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+There are configuration options available for a wide range of different needs your application may
+have, from mounting volumes to disabling the default asset upload.
+
+### Validate your deployment
+
+You can now visit [Integration Argo CD][argo] and after a short wait, your new application will be
+available to find on the dashboard.
+
+When your application has successfully deployed, you can try to access it through `kubectl`'s
+[port forwarding feature](/kubernetes/fix-app/#port-forwarding-to-a-pod).
```
theseanything, line 179: Ah yeah, it's difficult to know where to stop here. I'd say this might be confusing, as most apps should have an ingress associated which would just make them accessible.

source/kubernetes/create-app/index.html.md
```diff
@@ -0,0 +1,191 @@
...
+        value: 1
+```
+
+There are configuration options available for a wide range of different needs your application may
+have, from mounting volumes to disabling the default asset upload.
+
+### Validate your deployment
+
+You can now visit [Integration Argo CD][argo] and after a short wait, your new application will be
+available to find on the dashboard.
```
theseanything, line 176: They'll need to have make sure their container image has been built and successfully pushed to ECR beforehand. Essentially they just need to run the deploy GitHub Action.

theseanything (approved):
Thanks @csutter

source/kubernetes/create-app/index.html.md
csutter: Interesting - I've definitely only created a `v0` tag rather than a release, and then relied on the workflow to create the `v1` release and tag based on that. The only reason this is really necessary (I think) is because the workflow fails to run if there isn't an initial tag around to increment 😕

PR #4348, opened 2023-12-14
Implementation record document copy event

andysellick:
## What
Update the GA4 implementation record. See individual commits for details.

## Why
Part of the GA4 migration.

Trello card: https://trello.com/c/s6uLhovR/751-add-tracking-browser-copy-events


AshGDS (approved):
👍

PR #4351, opened 2023-12-15
Make the note about AWS MFA device name bold

peteglondon:
2 new starters on the team have missed this.

When I asked in slack others had already asked - in fact it's in the `reliability-eng` channel header.

***Making it bold and obvious***


gclssvglx (approved):
Nice one!

PR #4367, opened 2023-12-29
Add info about Snyk and slack bot

MuriloDalRi:
https://trello.com/c/R0ioTXZS/3380-document-ci-pipeline-auditing-and-alerting


MuriloDalRi:
> Nicely written! 💅 I don't think we are auditing the Snyk scans just yet I think it's ok to publish it like this (we will be adding that check shortly)

Ahh yes, I forgot I have a draft PR for that https://github.com/alphagov/seal/pull/485

PR #4376, opened 2024-01-03
Extend GA4 to the whole of the dev docs

andysellick:
## What

**NOTE this PR depends upon the changes in https://github.com/alphagov/govuk_publishing_components/pull/3802 being included here in a new version of the components gem.** (now done)

Extend GA4 analytics to the whole of the dev docs, instead of just to the implementation record pages in `/analytics`.

This uses the cookie banner component and code from `govuk_publishing_components`. 

We can't modify the general layout template to include the cookie banner, as it comes directly from the tech docs gem, and we don't want to include a copy of the layout as that would introduce future maintenance. Instead, we render the component as a JavaScript string in the `:head` block and use it as the markup for an element that we then prepend to the `body`. This is either quite clever or quite hacky depending on your point of view. 

## Why
Part of the GA4 migration.

## Visual changes
Currently the banner is limited to a max width inline with GOV.UK, which looks a bit strange in the full width layout here.

Screenshot 2024-01-03 at 17 13 36

Trello card: https://trello.com/c/kA66mxak/755-extend-analytics-to-all-of-dev-docs


source/partials/_cookie_wrapper.html.erb
```diff
@@ -0,0 +1,11 @@
+<script>
+  document.addEventListener('DOMContentLoaded', function () {
+    <% markup = capture do %>
+      <%= partial 'partials/cookie_banner' %>
+    <% end %>
+    var markup = `<%= markup.squish %>`;
```
AshGDS, line 6: Note that the backtick doesn't work in IE11, if we're still supporting this browser.

AshGDS (approved):
Nice work - I like the approach to import the banner 👍 

I've added a comment, it's either blocking or non-blocking depending on what we do with IE11 these days!

source/partials/_cookie_wrapper.html.erb
AshGDS: @andysellick No worries. I think you may be able to JSON stringify the `markup`, and then parse it back to a regular string when it's inserted into the DOM, but I appreciate IE11 isn't a high priority these days.

PR #4385, opened 2024-01-09
Update documentation for deploying the emergency banner

brucebolt:
In https://github.com/alphagov/whitehall/pull/8709, a user interface has been created for deploying the emergency banner.

Therefore removing instructions to run a rake task and instead replacing it with links to the UI.


ChrisBAshton (approved):
⭐

PR #4386, opened 2024-01-09
Add timestamp value to GA4 docs

AshGDS:
Adds the `timestamp` value to the GA4 docs. I placed it in `schemas.yml` but noticed it should probably link to something, so I kept the initial `attributes.yml` entry so that it could something with a description of the value. However I'm happy to remove the reference to it in `attributes.yml` if preferred.

### DIFF
--- a/data/analytics/attributes.yml
+++ b/data/analytics/attributes.yml
@@ -438,6 +438,13 @@
   type: string
   redact: true
 
+- name: timestamp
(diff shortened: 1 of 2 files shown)


andysellick (approved):
We used to have a thing where if the value of something wasn't set in the data file it would pull it automatically from the attributes.yml, but that doesn't seem to be working, might be worth some investigation - but this PR looks good 👍

PR #4387, opened 2024-01-10
Fix JS modules initialisation

andysellick:
## What/why
Fix the search. The search should work internally, but [recent changes](https://github.com/alphagov/govuk-developer-docs/pull/4376) broke the JS for this, so instead search was going to an external Google search result page.

Specifically:

- extra JavaScript from the components gem (for GA4 tracking) has conflicted with existing JS in the tech docs gem, specifically the modules start code
- the JS modules start code in the gem only works for modules that have an `init` function, but the tech docs gem code is based on older JS (using jQuery) where the module is initialised with a `start` function
- to fix this we include a reworked version of https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/modules.js here, with manual changes to call the `start` function, with a jQuery element
- this fixes the problem with the search not starting correctly, as well as other existing JS modules from the tech docs gem

### DIFF
--- a/source/javascripts/application.js
+++ b/source/javascripts/application.js
@@ -5,3 +5,72 @@
 //= require govuk_publishing_components/lib/cookie-settings
 //= require govuk_publishing_components/components/cookie-banner
 //= require govuk_publishing_components/load-analytics
+
+// reworked version of modules.js from govuk_publishing_components
+// needed to manually start tech docs gem modules now that we're including
+// the modules code from govuk_publishing_components
+var devdocsModulesFind = function () {
+  container = document
+
+  var modules
+  var moduleSelector = '[data-module]'
+
+  modules = container.querySelectorAll(moduleSelector)
+  var modulesArray = []
+  // convert nodelist of modules to array
+  for (var i = 0; i < modules.length; i++) {
+    modulesArray.push(modules[i])
+  }
+
+  // Container could be a module too
+  if (container !== document && container.getAttribute('data-module')) {
+    modulesArray.push(container)
+  }
+  return modulesArray
+}
+
+var devdocsModulesStart = function () {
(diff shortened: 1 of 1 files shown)


AshGDS (commented):
@andysellick Thanks for fixing this 👍  Happy to approve, but I had some questions:

1. Why do we need to use jQuery instead of just vanilla JS? 

2. Was there a reason this was picked over making `govuk_publishing_components`'s `modules.js` support calling `.start` as a fallback? i.e. doing

```
if init function exists:
    call init function
else if start function exists:
   call start function
```

as that would prevent some code duplication / looping through DOM elements twice.

Thanks 👍

andysellick:
@AshGDS excellent questions. We need to pass jQuery to the tech docs gem modules because they expect a jQuery object - one of the first things the search module does is try to use jQuery's `.find` on whatever gets passed (which errors if it's not a jQuery object).

To your second question. The tech docs gem modules code looks a bit like the code in the gem used to - with a `start` function to get the module going. When we started removing jQuery from the gem we had a problem - we had some modules using jQuery and some without, and we needed to know whether we should pass a jQuery object or a regular JS DOM node. So we made non-jQuery modules have an `init` function, so the modules code could check for the presence of either `start` or `init` and pass the right thing accordingly.

Over time we removed jQuery and one of the last things to go was the bit in the modules code where we did this check and called one of the two functions. I could put this back in, but it feels like a step backwards - particularly since it would add code to the live site specifically only for the dev docs site. It is a little inefficient, but as the dev docs isn't doing very much JS outside of this, I don't think it's going to make the site noticeably slower.

AshGDS (approved):
Makes sense - thanks for the explanation! Happy to approve

PR #4389, opened 2024-01-10
Spike putting OAuth proxy in front of Developer Docs

ChrisBAshton:
This heavily borrows from a PR doing the same thing for the Pay team manual back in September 2021:
https://github.com/alphagov/pay-team-manual/pull/1121

It uses [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy) to require signing in via GitHub to view any pages in the `source/private` folder.

Notes:
- This is a proof of concept for local development only. The Pay PR had an additional oauth2-proxy.cfg file for production, which we would need too.
- We'd move hosting to something like Heroku, so that we can have this proxy in front of the docs (not supported by GitHub Pages, which allows for static sites only).
- The GitHub provider settings currently allow anyone with a GitHub account to view private pages. In production we'd need to set the `github_org` setting to "alphagov" to lock it down to GDS folks.
- The `startup_local_auth.sh` script has logic in it that is supposed to remove the need to run the middleman server in another shell, but this wasn't working so I commented it out.

Decisions to make from here:
1. *Whether to be public or private by default*. The config is currently public by default, in that it allows unrestricted access to all routes except those under `/private`. We could instead deny access to all routes by default, and remove authentication from certain routes where desired. Ideally, the following routes would be public:
  - Homepage (to describe what this site is and how to access it)
  - The `/analytics` prefix, which is the public documentation for our GA4 implementation record.
  - Our API endpoints (e.g. repos.json).
2. *How to store private documentation*. Having a `private` folder in our public repo would not make much sense! We could make the repo private, or make use of git submodules to pull in a private repo dedicated to private docs. Implementation largely depends on the decision made in 1.
3. *Which OAuth provider to use*. Probably Google, as it means access is automatically lost when someone leaves the org (as opposed to GitHub Access, which has a manual/Terraform removal step). Also makes it more likely that OAuth won't impede the initial onboarding. Looks marginally more difficult to set up than GitHub, as it doesn't allow setting up against localhost URLs and custom ports.

Other functional considerations:
- Keeping search public would require that we omit the indexing of private documentation - potentially impacting its usefulness. If everything is private, we don't have to worry about this.
- Having some things public and some private means the private docs need to be declared private up front and stored in the 'private' folder. Alternatively, we could have a full allowlist of docs that should be private, but we then need to ensure that a rename of a private file won't accidentally make it publicly visible.

Trello: https://trello.com/c/BpaOgcsY/43-explore-putting-developer-docs-behind-sso

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,11 @@
 # Ignore the build directory
 /build
 
+# Ignore OAuth2-related files created by `startup_local_auth.sh`
+/auth/oauth2-proxy
+/auth/*.tar.gz
+/auth/*-sha256sum.txt
+
 # Ignore cache
 /.sass-cache
 /.cache
--- a/README.md
+++ b/README.md
@@ -37,6 +37,16 @@ env SKIP_PROXY_PAGES=true ./startup.sh
 
 Disabling proxy pages means you'll get a "Not Found" error if you try to access them locally.
 
+You can test the GitHub OAuth aspects of the Developer Docs by running the startup script above in one shell, and then running the following in another shell:
+
+```
+OAUTH2_PROXY_CLIENT_ID=abc123456 OAUTH2_PROXY_CLIENT_SECRET=def789101112 ./startup_local_auth.sh
+```
+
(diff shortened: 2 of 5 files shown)


ChrisBAshton:
Superseded by new repo: https://github.com/alphagov/govuk-developer-docs-proxy

ChrisBAshton:
Alternative Ruby-based proxy now available: https://itsi.fyi/

PR #4396, opened 2024-01-16
Some tweaks to the CKAN support docs

nicholsj:
A few small tweaks to the documentation for supporting CKAN which have come up in recent tickets - should be self-explanatory.


AshGDS (approved):
LGTM 👍

PR #4402, opened 2024-01-18
Update examples for connecting to CKAN DB with psql.

sengi:
Also add an example to the _Databases on GOV.UK_ page (so people don't have to backtrack if they come across that page first).


KludgeKML (approved):
LGTM, thanks!

PR #4404, opened 2024-01-19
Document change note update process for publishing apps

ryanb-gds:
Unfortunately at present only Whitehall and Content Publisher have an automated method for updating change notes. Hopefully this note will save time spent on 2nd line looking for something that isn't there.

The Publishing Experience team will look into providing a better way of doing this going forward, as it's obviously something that comes up on 2nd line fairly often.


JonathanHallam (approved):
nice one, thanks!

PR #4419, opened 2024-01-29
Fe updates

andysellick:
## What

- update the analytics documentation with details of the GA4 data for events on the back link component
- update the local frontend development page with a workaround to a newly discovered problem

## Why
Needed for the GA4 work.

Trello card: https://trello.com/c/Z5FYBU5o/782-back-links

### DIFF
--- a/data/analytics/events.yml
+++ b/data/analytics/events.yml
@@ -76,6 +76,28 @@
         - name: external
         - name: method
         - name: link_domain
+
+- name: back link
+  description:
+  events:
+    - name: link click
+      description: When the back link is used, on pages like the email subscription journeys
+      implemented: true
+      priority: low
+      data:
+      - name: event
+        value: event_data
+      - name: event_data
+        value:
+        - name: event_name
+          value: navigation
+        - name: type
+          value: back
+        - name: url
+        - name: text
+        - name: external
+        - name: link_domain
+        - name: method
 - name: banners
   description: Note that the presence of banners is tracked via our Pageview tracker.
   events:
--- a/source/manual/local-frontend-development.html.md
+++ b/source/manual/local-frontend-development.html.md
@@ -128,6 +128,22 @@ Some components, such as the cookie banner, are pulled in by Static. To test cha
 
 ## Troubleshooting
 
+### JavaScript changes not appearing/JavaScript not working properly
+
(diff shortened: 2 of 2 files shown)


source/manual/local-frontend-development.html.md
```diff
@@ -128,6 +128,22 @@ Some components, such as the cookie banner, are pulled in by Static. To test cha
 
 ## Troubleshooting
 
+### JavaScript changes not appearing/JavaScript not working properly
+
+Sometimes the main JavaScript files cannot be served properly when running locally. This happens when Rails automatically inserts fingerprinting into the script URLs when it is not needed e.g. `application-834y2r2hr292y9r.js` is being requested, but only `application.js` exists.
+
+This problem does not currently have a solution, but a workaround is to modify `_layout_for_public.html.erb` in your local `govuk_publishing_components` to hardcode the URLs to the broken JavaScript files (`application.js`, `load-analytics` and any others you need). For example, change:
```
AshGDS, line 135: Sorry to be nitpicky here, but regarding `This problem does not currently have a solution` -  have we asked the GOVUK devs if they've also experienced this but had a solution for it? Just a bit mindful of documenting  that there's no solution if we haven't had this confirmed.

data/analytics/events.yml
```diff
@@ -76,6 +76,28 @@
         - name: external
         - name: method
         - name: link_domain
+
+- name: back link
+  description:
+  events:
+    - name: link click
+      description: When the back link is used, on pages like the email subscription journeys
```
AshGDS, line 84: Is it worth using the `examples` value here to provide a link to an example?

AshGDS (commented):
Thanks for doing this - had a couple small comments

source/manual/local-frontend-development.html.md
andysellick: Reasonable point. I've reworded for now.

data/analytics/events.yml
andysellick: Yeah I wanted to but linking directly into the subscriptions would only work if someone was signed in. There's got to be other examples around the site but I can't find any right now. I'm hoping once the tracking is enabled we'll be able to put an actual example in here.

AshGDS: No worries 👍

source/manual/local-frontend-development.html.md
AshGDS: Thanks 👍

AshGDS (approved):
🚀

PR #4420, opened 2024-01-29
Add preflight checks to OpenSearch reindexing

sihugh:
This is the process that is run to adopt OpenSearch schema changes within search-api.

People are meant to clean up old indices once they're content that their change worked, but this has a habit of slipping through the net.

This process is eminently automatable, but writing it down helps, as does the instructions for finding out the state of the search cluster.


baisa (approved):
LGTM!

sihugh:
Thanks!

PR #4430, opened 2024-01-31
Re-add rack-logstasher

ChrisBAshton:
This was removed in #4424, but is still used by search-api: https://github.com/alphagov/search-api/blob/faf0e74546aa3a491c19c648e61ada17c3f58329/Gemfile#L26


KludgeKML (approved):
LGTM

PR #4431, opened 2024-01-31
Remove two more repos from repos.yml list

ChrisBAshton:
Following convention in #4424, we'll remove these rather than mark them as retired, as they have little value in being surfaced in the developer docs from now on.


KludgeKML (approved):
LGTM

PR #4433, opened 2024-02-01
Remove reference to content-store Mongo cluster

aldavidson:
The backups page still had a reference to content-store running a Mongo cluster - this PR removes it


nicholsj (approved):
Thanks!

PR #4438, opened 2024-02-02
Update data.gov.uk architecture docs

nicholsj:
This removes some outdated parts and updates the data.gov.uk architecture diagram


ChrisBAshton (approved):
LGTM 👍 if I had to nitpick, I'd give a bit more info in the architecture diagram commit to explain what's changed. (Looks like it just adds a bit more info to what the Find and CKAN services are, and that Fastly is a CDN)

nicholsj:
> LGTM 👍 if I had to nitpick, I'd give a bit more info in the architecture diagram commit to explain what's changed. (Looks like it just adds a bit more info to what the Find and CKAN services are, and that Fastly is a CDN)

Good point, thanks @ChrisBAshton . Done.

PR #4453, opened 2024-02-07
Fix cookie banner showing when it shouldn't

andysellick:
- cookie banner was recently updated to be hidden when JS is disabled
- but we had styles in here that also handled this behaviour, which were aligned with the component until the change
- removing unnecessary styles fixes the problem
- fortunately cookie consent was still being done correctly, even though the banner wasn't disappearing


richardTowers (approved):
Thank you! 🙇🏻

PR #4454, opened 2024-02-08
Mark `govuk-user-journey-analysis-tools` as retired

brucebolt:
This repo has been archived,so can be marked as retired in the list of repos.


nicholsj (approved):
Thanks

PR #4457, opened 2024-02-08
Clean up docs around retiring and publishing gems

theseanything:
This does the following:

- Groups the docs together in the same section
- Removes documentation referencing Jenkins and geocache
- Make docs for publishing a gem more concise.


sengi (approved):
🙌

PR #4458, opened 2024-02-08
Remove docs referring to old infra

theseanything:
This removes the following:
- How to configure a new app on the old infra
- A few references to Jenkins
- How to restart an application on the infra
- Information about our old tooling (new tooling is referenced elsewhere in docs)


sengi (approved):
Thanks!

PR #4463, opened 2024-02-12
Remove references to "leavers tickets" on 2nd line

ChrisBAshton:
We no longer process leavers on 2nd line. It is currently handled by members of Senior Tech, but is also ultimately a Platform responsibility and should be handled internally there, rather than delegated to 2nd line.

This commit therefore removes documentation that instructed devs to raise leavers tickets.

It also softens the documentation around revoking access if you will be away for 6 months, since we don't consistently do this in practice.


AgaDufrat (approved):
🙌

PR #4482, opened 2024-02-17
Replace legacy GPG usage with Secrets Manager.

sengi:
Get rid of our sketchy use of GPG.

- Add a playbook entry for retrieving a shared credential from Secrets Manager (with a subtle warning about the evils of allowing such things to exist in the first place).
- Replace links to the govuk-secrets pass store with references to Secrets Manager.
- Secure and simplify a clunky procedure for rotating Fastly API tokens which the Emergency Alerts feature apparently uses to purge objects from the CDN cache.
- Remove all references to GPG, including a page about using it to transfer files (yowch!)

This unblocks deleting the security dumpster-inferno that is the govuk-secrets repo.

#### Background

It's 2024. Even 5-10 years ago, infosec professionals were [blogging] about [how problematic GPG is] (though some folks [still like using it for email] 🤷). Our usage of it has always been pretty iffy, so let's make the most of the opportunity to get rid of it while we can.

[blogging]: https://blog.cryptographyengineering.com/2014/08/13/whats-matter-with-pgp/
[how problematic GPG is]: https://www.latacora.com/blog/2019/07/16/the-pgp-problem/
[still like using it for email]: https://articles.59.ca/doku.php?id=pgpfan:tpp


KludgeKML (approved):
LGTM

sengi:
Rebase to fix merge conflict.

PR #4500, opened 2024-02-22
Update the RDS restore docs post-replatforming

robinjam:
[Trello card](https://trello.com/c/uBUSc9zC/3436-fix-the-database-restore-steps-in-the-dev-docs)

The previous instructions involved changing a CNAME record to point to the newly restored database instance. This approach had a few problems:

- It involved a lot of complicated steps, and would have been difficult to perform during an incident
- It assumed that all of our apps were configured to refer to their database indirectly via a CNAME record, which is not always true
- It required extra steps post-incident to update our Terraform configuration (else the changes would be overwritten on the next Terraform apply)

Instead we should recommend updating the secret that stores the app's database connection details.

## Testing

I've tested these steps in staging, on the `local-links-manager` app.


robinjam:
(Split into two commits to make the diffs easier to follow)

source/manual/howto-backup-and-restore-in-aws-rds.html.md
```diff
@@ -105,172 +107,45 @@ gds-cli aws govuk-<environment>-admin aws rds wait db-instance-available --db-in
...
+Through the app's user interface, or via the app console, make a change that you
+can use as a sense check to verify that the database switch has been successful.
 
 For example, you might create a draft edition of something, or modify or delete
 a record. After switching to the restored database, your changes should be undone.
 
-In this example we want to make a change to the database `local-links-manager_production`
-such as delete an old record:
+In this example we open a Rails console on local-links-manager, from which we can
+make a change such as deleting an old record:
```
sengi, line 127: This step seems a bit vague. Not sure if I've correctly understood the intent but it sounds like it's about checking that the client app can read/write the restored database? If so, can we replace it with a more specific and straightforward check?

Could maybe delete this paragraph and the Rails console snippet and just leave the bit about about checking that the app can still write stuff by e.g. creating a draft edition.

source/manual/howto-backup-and-restore-in-aws-rds.html.md
```diff
@@ -105,172 +107,45 @@ gds-cli aws govuk-<environment>-admin aws rds wait db-instance-available --db-in
...
-### 1. Make a change to the database contents
-
-Through the app's user interface, or via the app console or database console, make
-a change that you can use as a sense check to verify that the database switch has
-been successful.
+Through the app's user interface, or via the app console, make a change that you
+can use as a sense check to verify that the database switch has been successful.
 
 For example, you might create a draft edition of something, or modify or delete
 a record. After switching to the restored database, your changes should be undone.
```
sengi, line 124: Best to avoid choices in playbooks where possible.

```suggestion
For example, you might create a draft edition and use the preview feature to check that the app saved your changes.
```

source/manual/howto-backup-and-restore-in-aws-rds.html.md
```diff
@@ -105,172 +107,45 @@ gds-cli aws govuk-<environment>-admin aws rds wait db-instance-available --db-in
...
-
-Once you have restored a database in AWS RDS, you now need to point the corresponding app towards it.
-
-### 1. Make a change to the database contents
-
-Through the app's user interface, or via the app console or database console, make
-a change that you can use as a sense check to verify that the database switch has
-been successful.
+Through the app's user interface, or via the app console, make a change that you
+can use as a sense check to verify that the database switch has been successful.
```
sengi, line 121:
```suggestion
Check that the app can still read and write to its database.
```

sengi (approved):
Cool, thanks for cleaning this up — it's a massive improvement!

Made some minor editorial suggestions but feel free to amend or pick and choose them.

source/manual/howto-backup-and-restore-in-aws-rds.html.md
robinjam: I think the intent was to make sure the restore was successful by making a change before the pre-restore, and then checking if the change was reverted post-restore.

It's probably easier/less confusing if I remove that whole section and suggest checking which DB ActiveRecord is connected to via the Rails console instead.

robinjam:
Thanks @sengi! Super useful suggestions 🙏

PR #4502, opened 2024-02-23
Updates to GA4 documentation

AshGDS:
Adds updates to:
- Callout tracking
- Contact tracking
- Webchat tracking


data/analytics/events.yml
```diff
@@ -303,14 +303,103 @@
...
-- name: call out box
+- name: call out boxes
   events:
     - name: call out box link click
-      implemented: false
-      priority: medium
+      description: Link tracking for call to action, warning text, advisory callout, information callout, and content banner components.
+      tracker: link_tracker
+      implemented: true
+      priority: low
```
andysellick, line 312: Are there any examples you could add for call out boxes?

data/analytics/events.yml
```diff
@@ -303,14 +303,103 @@
...
+          - name: event_name
+            value: form_response
+          - name: type
+            value: contact
+          - name: action
+            value: Send message
+          - name: section
+            value: "What's it to do with?"
+          - name: text
+            value: "The chosen 'What's it to do with?' radio button text, and the URL of the page theyre contacting us about, if relevant."
```
andysellick, line 352: Nitpick: `they're`

data/analytics/events.yml
```diff
@@ -303,14 +303,103 @@
...
+      - name: event
+        value: event_data
+      - name: event_data
+        value:
+          - name: event_name
+            value: form_error
+          - name: type
+            value: contact
+          - name: section
+            value: "(Comma separated list of form inputs that errored)"
```
andysellick, line 373: Is using brackets like this a convention we use? Also wondering if it would be worth writing an example, so this would become something like:

`Comma separated list of form inputs that errored, e.g. "name,message"`

data/analytics/events.yml
AshGDS: Thanks - my yml file was crashing so I eradicated apostrophes at one point as I thought they were causing it!

data/analytics/events.yml
AshGDS: Thanks, have added a banner example - we're not sure the warning text component ever contains links, so I don't have an example at the moment.

andysellick (approved):

PR #4503, opened 2024-02-23
Update docs on Emerg. Alerts Fastly automation token.

sengi:
- We only provide them one token now, because it's not a very privileged token. Having separate tokens per environment was no better for security and definitely worse for maintainability.
- Remove extraneous detail. The details are already in the linked playbook on changing the token.
- In the playbook, mention an acceptable way to transfer the token after generating it.
- Use the same Markdown heading syntax as the rest of devdocs.


theseanything (approved):

sengi:
(Just simplified the wording slightly on the new playbook step.)

PR #4508, opened 2024-02-26
Remove .erb extension from docs that contain no ERB.

sengi:
It's not ideal that we have ERB in our docs, but where we do, it's better to be explicit about it.


ChrisBAshton (approved):
⭐

PR #4514, opened 2024-02-27
Update transition docs: move domain verification to be first transition step

brucebolt:
Failing to verify a *.gov.uk domain blocks the transition of any other sites subsequently added.

Therefore moving this to be the first instruction so it can be carried out before we add the domain into the transition app.


jkempster34 (approved):
I've never actually done a transition, so idk if you want a second (third?) opinion. But, this all makes sense to me

ChrisBAshton (approved):
LGTM - thanks Bruce

PR #4527, opened 2024-03-01
Update playbook for rotating a Signon API token.

sengi:
- Put the sections in order so that e.g. the information about special cases is presented before the reader starts down the non-special-case procedure.
- Avoid ambiguity and unnecessary choices in the procedure.
- Call out that this is toil and should be eliminated; we just haven't got around to it yet.
- Fix syntax highlighting in source examples and make them easier to copy-paste by using code blocks.
- Use ordered lists for procedures where the order of steps is important.
- Split up a bunch of compound steps into simpler, individual steps that are hopefully a bit easier to follow.
- Remove the step where we trigger a rolling update of the application; [reloader](https://github.com/stakater/Reloader) takes care of this automatically.
- Give a practical example of testing an API call from a client app.

[Preview](https://github.com/alphagov/govuk-developer-docs/blob/sengi/signon-token-toil/source/manual/alerts/signon-api-user-token-expires-soon.html.md)

### DIFF
--- a/source/manual/alerts/signon-api-user-token-expires-soon.html.md
+++ b/source/manual/alerts/signon-api-user-token-expires-soon.html.md
@@ -1,63 +1,86 @@
 ---
-owner_slack: "#govuk-2ndline-tech"
-title: Signon API User Token Expires Soon
+owner_slack: "#govuk-platform-engineering"
+title: Signon API user token expires soon
 parent: "/manual.html"
 layout: manual_layout
 section: Alertmanager alerts
 ---
 
-[signon]: https://signon.publishing.service.gov.uk/api_users
-[gds-api-adapters]: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api.rb
-
 One or more tokens for API Users are about to expire. You should rotate
 expiring tokens to ensure the associated application keeps working.
 
-- First login to [Signon], go to API Users and click on the appropriate API User.
+## Rotate a Signon API token
 
-> Check the *Last synced at* time to see if the API User is still using the
-> application. If you are confident the token is unused, the you can just
-> click the *Revoke* button to remove it and there's no need to continue.
+> 🪨 This procedure is [toil](https://sre.google/workbook/eliminating-toil/) and should be eliminated or automated away. For now, it's unfortunately still a manual process.
 
-- Click *Add application token*, select the application and click *Create
-  access token*.
+### Special cases
 
-- At this point the new token is created, and will be picked up automatically by a job that runs once a day (around 1am) to sync tokens to the applications in Kubernetes. If you want to check that the token has been updated, you can wait until the next day and query the app: `kubectl -n apps exec -it deploy/<your-app-name> -- printenv | grep BEARER_TOKEN`, then find the relevant token in the output and compare the first and last 8 characters with the hidden token shown in Signon.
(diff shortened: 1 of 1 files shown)


source/manual/alerts/signon-api-user-token-expires-soon.html.md
```diff
@@ -1,63 +1,86 @@
...
+
+### 2. Check that the client application works with the new token
+
+1. Trigger the token sync cronjob to run now (or wait until the next day).
 
-> Check the *Last synced at* time to see if the API User is still using the
-> application. If you are confident the token is unused, the you can just
-> click the *Revoke* button to remove it and there's no need to continue.
+    ```sh
+    k create job --from cronjob/signon-sync-token-secrets-to-k8s signon-token-sync-$USER
```
kentsanggds, line 47: I went through the commands and they worked after adding `apps` namespace to the command, could this be added to the command or add a command to set the namespace in the docs.

sengi: Thanks for testing the steps!

We get users to set their context namespace to `apps` in the [getting started guide](https://docs.publishing.service.gov.uk/kubernetes/get-started/access-eks-cluster/#access-a-cluster-for-the-first-time) and the [quick-reference guide](https://docs.publishing.service.gov.uk/kubernetes/cheatsheet.html#prerequisites) so that we don't have to repeat `-n apps` in hundreds of examples. Definitely let me know if there's some other place where it's missing from setup instructions tho!

kentsanggds: ok, thats fine then, I guess I'm a bit different as I mainly use the `datagovuk` space for things

source/manual/alerts/signon-api-user-token-expires-soon.html.md
```diff
@@ -1,63 +1,86 @@
...
-At this point you can check that the app has the new token as above.
+    ```ruby
+    ENV["PUBLISHING_API_BEARER_TOKEN"]
+    ```
 
-## Revoke the old token
+    Check that the result matches the new token in Signon and not the old one.
 
-Once the new token is in place, you should check that it can be used:
+1. You can also check that the token works by making an API call yourself from the client app's Rails console. For example, to check that Short URL Manager can talk to Publishing API:
```
dnkrj, line 70:
```suggestion
1. You can also check that the token works by making an API call yourself from the client app's Rails console. For example, to check that Short URL Manager can talk to Publishing API ([as per the API's documentation](https://github.com/alphagov/publishing-api/blob/main/docs/api.md)):
```
The other thing that we thought might be useful to include was a link to the API documentation, so it's clearer what to call for other APIs.

dnkrj (approved):
Thanks for this Chris, just one comment from following the runbook

source/manual/alerts/signon-api-user-token-expires-soon.html.md
sengi: Ah yeah that's a fair point, it's a little inconvenient if your day-to-day is mostly in a different namespace, sorry :( You could have another context or use the [ns](https://github.com/ahmetb/kubectx) kubectl plugin (I recommend installing via [krew](https://krew.sigs.k8s.io/docs/user-guide/quickstart/)) to help with this a bit.

source/manual/alerts/signon-api-user-token-expires-soon.html.md
sengi: Linked to gds-api-adaptors since the Rubydoc in there is pretty good and corresponds directly to the method calls that you'd need to make.

PR #4530, opened 2024-03-04
Update the licensing team slack channel

sihugh:
This is more likely to reach relevant engineers.

As well as the docs, this config powers the Release app's [deploy badger](https://docs.publishing.service.gov.uk/manual/slack-integrations.html#release-app-badger). I think it will get more response going to the new channel.


nicholsj (approved):
LGTM thanks

PR #4533, opened 2024-03-05
Fix implementation record filtering

andysellick:
## What / why
Fix a small bug in the filtering functionality recently added to some pages in the implementation record.

The JS was using `innerText` but we include hidden elements to match the title of the sections on events by type. `innerText` is less reliable than `textContent` as it doesn't include hidden text, and during testing it produced some slightly odd results. This change fixes them.

## Visual changes
None.

Trello card: https://trello.com/c/m6vL7lJy/186-add-filters-to-events-and-attributes-pages-on-implementation-record


andysellick:
@AshGDS I guess I just found the value of having tests for our code... 🤦

AshGDS (approved):
😅

PR #4537, opened 2024-03-05
Update retire app docs

1pretz1:
The mark an app as retired step is already covered in the last step [1] in the retiring an app doc and doing this prematurely of archiving the repo will cause consistency checks to fail in govuk-sass-config [2].

Also updates the kube command for retrieving resources for an app as the old command doesn't return anything.

Trello:
https://trello.com/c/Mv4qiBEI/2386-check-and-remove-kubernetes-resources

[1]: https://docs.publishing.service.gov.uk/manual/retiring-a-repo.html
[2]: https://github.com/alphagov/govuk-saas-config


KludgeKML (approved):
LGTM

PR #4543, opened 2024-03-06
Add components to implementation record

andysellick:
## What

In the GA4 implementation record, where an event occurs on a component, document and link to the component guide page for that component. This is done per-event, as some event groupings apply to more than one component.

## Why
Part of improving the implementation record for users.

## Visual changes

Screenshot 2024-03-06 at 09 31 08

### DIFF
--- a/data/analytics/events.yml
+++ b/data/analytics/events.yml
@@ -10,6 +10,7 @@
         - text: Manual page
           link: https://www.gov.uk/guidance/immigration-rules/immigration-rules-part-11b
       tracker: event_tracker
+      component: accordion
       data:
       - name: event
         value: event_data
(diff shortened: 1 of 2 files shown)


AshGDS (approved):
Nice 👍 How does this work when the event applies to more than one component? E.g. the `callout` tracking would have worked on both the warning text component, and the `govspeak` component before I removed it from the `govspeak` component due to it affecting external links.

andysellick:
@AshGDS yeah I thought about that. I think if we encounter that in future we can either extend this functionality (so the `component` is a comma separated list or something) or separate the event out into two events. 

There are some other slightly complicated things that I just decided were too nuanced for the purposes of this documentation. For example when we have attachments rendered inside govspeak - explaining how that all fits together for a non-technical audience seemed too much, so I just pointed it at the attachment component page instead.

PR #4546, opened 2024-03-08
Implementation record events listing updates

andysellick:
## What
Changes to the events pages on the GA4 implementation record, specifically:

- rename tabs on the events pages to better reflect their content
- rename top level events to more closely match their events types (where possible)
- minor event data additions
- fix a missing bit of functionality on the events pages where form submission wasn't being prevented, which could frustrate users (as it's in-page filtering)

## Why
Part of improving the implementation record for users.

Trello card: https://trello.com/c/7pwKVz4k/188-update-names-of-tabs-on-implementation-record

### DIFF
--- a/data/analytics/events.yml
+++ b/data/analytics/events.yml
@@ -80,8 +80,8 @@
         - name: method
         - name: link_domain
 
-- name: back link
-  description:
+- name: back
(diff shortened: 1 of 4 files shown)


AshGDS (approved):
Nice 👍

PR #4548, opened 2024-03-08
Add specific detail to Ruby gem versions

kevindew:
The most recent change to this documentation had lost the specifics which has led to a bulk change where GOV.UK gems now mostly specify an unnecessarily strict version constraint, which means some apps aren't keeping up to date with gems anymore.

I've added more detailed to try rectify that in future.

### DIFF
--- a/source/manual/publishing-a-ruby-gem.html.md
+++ b/source/manual/publishing-a-ruby-gem.html.md
@@ -36,8 +36,15 @@ Contact a [GOV.UK GitHub Owner][] to grant your repository
 ## Ruby version compatibility
 
 Ensure gems are compatible with all [supported minor Ruby
-versions][supported-rubies], specifying the minimum Ruby version in the [gemspec
-file][gemspec-ruby-version] and matching it in .ruby-version.
+versions][supported-rubies], specifying the minimum minor Ruby version in the gemspec
+file (for example: `spec.required_ruby_version = ">= 3.1"`) - note, this should not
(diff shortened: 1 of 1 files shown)


source/manual/publishing-a-ruby-gem.html.md
```diff
@@ -36,8 +36,15 @@ Contact a [GOV.UK GitHub Owner][] to grant your repository
 ## Ruby version compatibility
 
 Ensure gems are compatible with all [supported minor Ruby
-versions][supported-rubies], specifying the minimum Ruby version in the [gemspec
-file][gemspec-ruby-version] and matching it in .ruby-version.
+versions][supported-rubies], specifying the minimum minor Ruby version in the gemspec
+file (for example: `spec.required_ruby_version = ">= 3.1"`) - note, this should not
+be more specific than a minor Ruby version unless there is a particular issue with
+a Ruby patch release.
```
sengi, line 42: minor edit for clarity / tech writing style: (avoid "note", prefer callout, avoid bracketed clauses other than for acronym expansions etc.)

```suggestion
file. For example: `spec.required_ruby_version = ">= 3.1"`.

> Do not specify a patch (or "tiny") Ruby version unless there is a particular issue with a Ruby patch release.
```

sengi (approved):
Thanks!

deborahchua (approved):

sengi:
I don't suppose there's a suitable linter for gemspecs that would help with this?

PR #4549, opened 2024-03-08
Updates following the codecommit drill

kentsanggds:
During the drill we discovered a number of things which can improve the docs:

- add missing tags, the `docker push` would fail otherwise
- re-enable auto-sync after confirming that the deployment of the image from code commit has worked
- use `govuk-replatform-test-app` rather than `whitehall` as the example as we spent quite a long time downloading and waiting for `whitehall` to build, and we should be able to make visible changes to the test app without affecting a live service.


sengi (approved):
Nice!

PR #4560, opened 2024-03-15
Add section on how to get an asset's ID

deborahchua:
Some of the rake tasks require the ID, but we are sometimes only provided with the URL when the request to remove it comes through Zendesk.

### DIFF
--- a/source/manual/manage-assets.html.md.erb
+++ b/source/manual/manage-assets.html.md.erb
@@ -53,6 +53,9 @@ available for this:
 
 <%= RunRakeTask.links("asset-manager", "assets:redirect[ASSET_ID,REDIRECT_URL]") %>
 
+Also see [getting an asset's ID](#getting-an-asset39s-id) if you have only been
+provided the URL.
+
 ## Uploading an asset
 
 Some publishing apps such as [Mainstream Publisher](/repos/publisher.html) do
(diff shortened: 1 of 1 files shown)


source/manual/manage-assets.html.md.erb
```diff
@@ -128,3 +128,28 @@ changing the URL, follow these steps:
    attachment_data = asset.assetable
    attachment_data.update(file_size: 123, number_of_pages: 28)
    ```
+
+## Getting an asset's ID
```
beccapearce, line 132: If I'm being fussy could I ask you to link to the `## Getting an asset's` ID section in `## Redirecting an asset`, not a necessity but that's then I need it (unless it's not covered by the page fold, then it's all good :) )

source/manual/manage-assets.html.md.erb
```diff
@@ -128,3 +128,28 @@ changing the URL, follow these steps:
...
+
+You can get an asset's ID by using it's URL:
+
+1. Open a Rails console:
+
+  ```
+  k exec -it deploy/asset-manager -- rails c
+  ```
+
+1. Find the asset:
```
JamesCGDS, line 142: Very minor typo - should this be `2. Find the asset:`?

JamesCGDS (approved):
LGTM! 👍

PR #4565, opened 2024-03-18
Make improvements to the GA4 implementation record

andysellick:
## What

Adds the following to the GA4 implementation record:

- fix link to data docs
- add links from events page to GA4/Looker studio reports for traffic for those events
- minor code cleaning

## Why
Part of improving the GA4 implementation record for users.


AshGDS (approved):
LGTM 👍

PR #4572, opened 2024-03-22
Make StringToId more robust to arbitrary input.

sengi:
- Add some tests.
- Don't "crudely remove HTML tags"; use a library that does it properly. Not a huge deal cos this is only gonna see semi-trusted input, but it's not a good look to be rolling our own with that.
- Remove comments that just give false reassurance.
- Use the simplest string functions for the job rather than hitting everything with the regex hammer.

Fixes https://github.com/alphagov/govuk-developer-docs/security/code-scanning/2.

Gemfile.lock changes were generated by `bundle install`.

Alternatives considered:

Since we're already pulling in `activesupport`, we could just use `ActiveSupport::Inflector::parameterize` and do away with this little utility module altogether. I assumed that would probably break a bunch of existing links to heading anchors, though. (If not then let's just ditch this thing!)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gem "ffi"
 gem "rake"
 gem "rspec"
 gem "rubocop-govuk", require: false
+gem "sanitize"
 gem "simplecov"
 gem "webmock"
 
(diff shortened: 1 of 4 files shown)


ChrisBAshton (approved):
LGTM 👍 looks like the [original code was added in 2017](https://github.com/alphagov/govuk-developer-docs/commit/4c412444391cccf64e1328cff5488f95984f52c8) 😅

PR #4574, opened 2024-03-22
Add GA4 'filter' events to the analytics docs

AshGDS:
## What
- Adds the filter events to the docs
- Adds the focus loss tracker to the docs
- Updates the `.gitignore` as it tries to add `node_modules` to your commits and `.DS_Store` if you've ever worked with images in this repo

### DIFF
--- a/data/analytics/events.yml
+++ b/data/analytics/events.yml
@@ -673,7 +673,60 @@
           value: Get emails from GOV.UK
         - name: type
           value: email subscription
-
+- name: filter
+  description: Interactions that can be conducted after filtering the content using CSS instead of HTTP or API requests.
+  events:
+  - name: filter search
+    implemented: true
+    priority: low
+    description: When a search box is used to filter a long list of things on a page, by hiding irrelevant matches, rather than fetching a new page of results from the server.
+    tracker: focus_loss_tracker
+    examples:
+      - text: Government organisations page
+        link: https://www.gov.uk/government/organisations
+      - text: Help and services around the world page
+        link: https://www.gov.uk/world
+    data:
+      - name: event
+        value: event_data
+      - name: event_data
+        value:
+        - name: event_name
+          value: filter
+        - name: type
+          value: filter
+        - name: action
+          value: filter
+        - name: section
+        - name: text
+          value: the user's search term
+  - name: filter link clicks
+    implemented: true
+    priority: low
+    description: Link clicks in a filter search list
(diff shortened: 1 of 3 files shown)


data/analytics/events.yml
```diff
@@ -673,7 +673,43 @@
...
+      - name: event_data
+        value:
+        - name: event_name
+          value: filter
+        - name: type
+          value: filter
+        - name: action
+          value: filter
+        - name: section
+  - name: filter link clicks
```
andysellick, line 696: Does the filter link click not also include all the regular link data, like href/url, text, domain, method etc.?

data/analytics/events.yml
```diff
@@ -673,7 +673,43 @@
...
         - name: type
           value: email subscription
-
+- name: filter
+  description: Interactions that can be conducted after filtering the content using CSS instead of HTTP or API requests.
+  events:
+  - name: filter search
+    implemented: true
+    priority: low
+    description: When a user searches for a link on a page, and the search box is one that filters all the results using CSS instead of HTTP or API requests.
```
andysellick, line 682: Suggestion: `When a search box is used to filter a long list of things on a page, by showing relevant matches using JavaScript, rather than fetching a new page of results from the server`

andysellick, line 682: Worth including some `examples:` here?

data/analytics/events.yml
AshGDS: Good point, thanks - will update

data/analytics/events.yml
AshGDS: Good point, will add them in, thanks

AshGDS: @andysellick I was hesitant to just say JavaScript, since our finder searches use client side JavaScript to update search results. Given this, should I still use your suggestion?

AshGDS:
Thanks @andysellick - should be ready for another review 👍

data/analytics/events.yml
andysellick: How about `When a search box is used to filter a long list of things on a page, by hiding irrelevant matches, rather than fetching a new page of results from the server` ?

AshGDS: @andysellick  Sounds good 👍 Have updated both descriptions, should be ready for another review

AshGDS: Updated now!

andysellick (approved):
👏

PR #4608, opened 2024-04-24
Update PagerDuty drill time

JamesCGDS:
## What
This PR updates the documented time for the PagerDuty drill to address the difference in UTC and BST times.

## Why
The drill is documented as being triggered at 10am UTC and 10am BST (and should be 10am UTC and 11am BST).

### DIFF
--- a/source/manual/2nd-line.html.md
+++ b/source/manual/2nd-line.html.md
@@ -86,7 +86,7 @@ We use Grafana dashboards to monitor the health of our applications and service 
 
 ### PagerDuty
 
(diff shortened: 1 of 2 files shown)


KludgeKML:
Could we update the docs so that we don't have to update them again in autumn (ie mention that it's 10 UTC, which is 11 BST)?

JamesCGDS:
Thanks @KludgeKML, I've made that change

KludgeKML (approved):
Thanks! LGTM

PR #4612, opened 2024-04-29
Update Status Page instructions to not use internal terms

brucebolt:
This adds guidance to suggest we shouldn't be using internal acronyms (e.g. DGU for data.gov.uk) on Status Page updates.

Also clarifies that names of internal applications shouldn't be named in the incident title.


sengi (approved):
💯

PR #4619, opened 2024-04-30
Update dependencies + work around govuk_tech_docs.

sengi:
Update all the Rubygems dependencies as far as possible.

Had to specify a couple of indirect dependencies missing from govuk_tech_docs and avoid html-pipeline >= 3, which appears to be incompatible with govuk_tech_docs for now.

Group the dev/test-only dependencies in the Gemfile so we can skip them in the build later.


ChrisBAshton (approved):
Thanks ⭐

sengi:
No probs! I'll have a look at what's up with tech_docs gem if I have time.

Looks like I need to rebase this.

PR #4627, opened 2024-05-01
Update instructions for "Fall back to AWS CloudFront"

matthillco:
This has been updated based on a 2ndline drill. The distinction between actions to take on **production** vs **staging** were unclear and have been improved. I've also added a couple of new details regarding how to do the manual failover, and how to check if the fallback domain is working. Bullet lists have been split up and nesting improved/removed for clarity.

### DIFF
--- a/source/manual/fall-back-to-aws-cloudfront.html.md
+++ b/source/manual/fall-back-to-aws-cloudfront.html.md
@@ -23,52 +23,100 @@ This backup CDN is currently provided by AWS CloudFront.
 
 > Note: These steps will have you make changes to our production environment. This is because our DNS records and domains for our integration and staging environments are configured in our AWS and GCP production accounts.
 
-- Confirm that Fastly is the cause of the incident (check [https://status.fastly.com/](https://status.fastly.com/)
+### Initial steps
+
+1. Confirm that Fastly is the cause of the incident (check [https://status.fastly.com/](https://status.fastly.com/)
   and keep an eye on twitter - if there's a major Fastly outage there will be a lot of noise)
-- Escalate to GOV.UK SMT as soon as you begin to consider failing over
-- Sign in to the AWS console as an admin (`gds aws govuk-production-admin -l`, or however you prefer to sign in to AWS)
-- Sign in to [the `govuk-production` project on GCP console](https://console.cloud.google.com/home/dashboard?project=govuk-production)
-- For **production DNS records**, open the following four pages as separate tabs:
-  - [GCP Cloud DNS www-cdn.production.govuk.service.gov.uk](https://console.cloud.google.com/net-services/dns/zones/govuk-service-gov-uk/rrsets/www-cdn.production.govuk.service.gov.uk./CNAME/edit?project=govuk-production)
-  - [AWS Route 53 govuk.service.gov.uk](https://console.aws.amazon.com/route53/v2/hostedzones#ListRecordSets/Z22RPYZA77J620)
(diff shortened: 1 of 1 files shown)


source/manual/fall-back-to-aws-cloudfront.html.md
```diff
@@ -23,49 +23,78 @@ This backup CDN is currently provided by AWS CloudFront.
...
+
+### Next steps
+
+In either case, the records should look like `d0000000000000.cloudfront.net.` (with 0s replaced with letters and numbers). Now you can manually update the `CNAME` records for both domains in both GCP and AWS, via the tabs you opened in your web browser earlier:
+
+- Change the canonical name from `www-gov-uk.map.fastly.net.` to the CloudFront domain name you found before, including the trailing period (e.g. `d0000000000000.cloudfront.net.`)
+- Test if the new Cloudfront domain is serving assets correctly, for example:
+
+```bash
+curl -vs https://assets.staging.publishing.service.gov.uk/media/662a74aa45f183ec818a72c2/dvsa-earned-recognition-vehicle-operators-accredited-list.csv/preview | grep cloudfront
```
kentsanggds, line 94: I think we need to tell the dev what to look for from the curl response that will inform them that the requests are being served by the updated CDN - 

missing key - `fastly-backend-name`

or the presence of `xxxxx.cloudfront.net` value in the `via` key.

source/manual/fall-back-to-aws-cloudfront.html.md
```diff
@@ -23,49 +23,78 @@ This backup CDN is currently provided by AWS CloudFront.
...
+```
+
+- Perform a manual failover, for example by temporarily changing the new canonical domain name with a random prefix.
```
kentsanggds, line 97: As we are now checking where the requests are being served from in the curl response we probably don't need this step to break it.

source/manual/fall-back-to-aws-cloudfront.html.md
```diff
@@ -23,49 +23,78 @@ This backup CDN is currently provided by AWS CloudFront.
...
 - **On production only**, after performing the manual failover, you should also update our infrastructure-as-code to match the changes you just made:
```
kentsanggds, line 98: Should this section be in the `Production` section?

kentsanggds (changes requested):
Just a few changes needed before it's good to go.

source/manual/fall-back-to-aws-cloudfront.html.md
matthillco: Do you mean move it to that section, or duplicate it? I'm unclear. @kentsanggds

matthillco: Actually, I'm not sure we need to implement your comment: this point is in the "Next Steps" section which applies to both Staging and Production. @kentsanggds

source/manual/fall-back-to-aws-cloudfront.html.md
matthillco: Updated with your suggestions

source/manual/fall-back-to-aws-cloudfront.html.md
matthillco: Updated with your suggestion.

source/manual/fall-back-to-aws-cloudfront.html.md
kentsanggds: I think that section needs to be moved as it only applies to production. 
I don't think for Staging its necessary to update the terraform as part of the drill, unless we want to test the changes on Staging first before applying to Production if there is a real need to switch CDN?

kentsanggds (approved):

PR #4629, opened 2024-05-02
Rewrite Zendesk doc

ChrisBAshton:
- Document the views, macros and triggers of relevant product areas (as these can be hard to dig out later)
- Document our implementation of one-time Slack alerts per ticket, and our mechanism for keeping tag configuration in one place
- Remove outdated or unuseful info that is no longer worth keeping as our support model evolves
- Make the document more about generic Zendesk usage, but encapsulating specific Technical 2nd Line knowledge in its own section

### DIFF
--- a/source/manual/zendesk.html.md
+++ b/source/manual/zendesk.html.md
@@ -7,126 +7,78 @@ section: 2nd line
 type: learn
 ---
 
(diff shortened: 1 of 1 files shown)


KludgeKML (approved):
Nice!

PR #4638, opened 2024-05-08
Add GA4 parameter names into GA4 implementation record

andysellick:
## What / why
Add data and functionality into the Implementation Record to display the GTM parameter names and GA4 display names of attributes in events. This allows analysts (familiar with reading the display names in GA4) to find the attributes, which are sometimes called something different in the dataLayer.

Complicating this is that dataLayer attributes and GA4 display names don't have a 1:1 relationship. Sometimes one attribute will have different GA4 display names. 

For example, `text` by default has a GA4 display name of `UI text`, but when included from a search event (where the `event_name` is `search`), this translates through to a GA4 parameter name of `Search term`. The implementation record didn't allow for this, so the code has been extended to allow for 'variants' of attributes. If an attribute has a variant containing the `event_name` from an existing event, that event will link to the variant page for that attribute. The variant page is the same page as the attribute page, but clearly marked as a variant and linking to the other variants and 'parent' page for that attribute. 'Parent' pages also show and link to their variants.

I've also put the GA4 display name in the attributes list, so analysts can find attributes more easily.

See this [list of attribute data](https://docs.google.com/spreadsheets/d/1QCZLhWW6GrnHY3spzzLAjMAWln0YxOW4loW3mJe1bAo/edit#gid=0) for the data on which this PR is based. Note that this data has a few gaps, but is mostly complete.

## Visual changes
Example of a variant attribute page:

Screenshot 2024-05-08 at 17 20 34

Example of the 'parent' attribute page:

Screenshot 2024-05-08 at 17 20 45

Searching for attributes:

Screenshot 2024-05-08 at 17 22 12

The events page remains unchanged and simply links to the correct 'parent' attribute or variant page.

Trello card: https://trello.com/c/needmOSF/788-add-ga4-field-names-to-implementation-record

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -70,7 +70,14 @@ end
 
 data.analytics.attributes.each do |attribute|
   attribute_name = attribute["name"].downcase.gsub(" ", "_")
-  proxy "analytics/attribute_#{attribute_name}.html", "analytics/templates/attribute.html", locals: { attribute: }
+  proxy "analytics/attribute_#{attribute_name}.html", "analytics/templates/attribute.html", locals: { attribute:, variant: nil }
+
+  next unless attribute.variants
+
+  attribute.variants.each do |variant|
+    variant_name = variant["event_name"].downcase.gsub(" ", "_")
+    proxy "analytics/attribute_#{attribute_name}/variant_#{variant_name}.html", "analytics/templates/attribute.html", locals: { attribute:, variant: }
+  end
 end
 
 data.analytics.trackers.each do |tracker|
--- a/data/analytics/attributes.yml
+++ b/data/analytics/attributes.yml
@@ -5,6 +5,8 @@
   required: no
   type: string
   redact: false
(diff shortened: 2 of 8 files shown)


AshGDS (approved):
Nice work 👍 I guess our data is small enough that having deeply nested loops isn't going to cause the server performance to degrade exponentially.

Just curious, is there a reason some links include the bracketed part, and some links have the bracketed part as regular text?  
image

andysellick:
@AshGDS thanks for the review. Yes, hopefully the overhead for the changes is minimal.

Good point on the attribute list, that was a bit unclear. I've tidied it up a little. Attributes with variants should now be named `Attribute / variant` and the bit in brackets is the GA4 parameter name (thought that analysts are likely to try to filter by that, so should make finding attributes easier for them).

AshGDS (approved):
Nice, thanks 👍

PR #4643, opened 2024-05-13
Add note about pact tests

KludgeKML:


source/manual/pact-testing.html.md
```diff
@@ -109,3 +109,25 @@ different applications.
 
 When an application calls an API directly, not via GDS API Adapters, this application will be the consumer. For example, Publishing API
 is the consumer in the pact tests between it and Content Store, since it makes [direct calls to Content store](https://github.com/alphagov/publishing-api/blob/main/app/adapters/content_store.rb).
+
+## How GDS API Adapters retrieves tests from the Providers
+
+Each provider defines some details of its tests in the `.github/workflows/pact-verify.yml` file. This workflow is used by both
+the provider and the consumer (gds-api-adapters), and behaves differently based on the value of the pact_artifact input, which can
+be confusing if you're unused to it.
```
kevindew, line 117: I started reviewing this and struggled to understand it so had a go at rewriting it with some more detail:

> ## How pacts operates as part of continuous integration (CI)
> 
> Consumers (for example GDS API Adapters) define the tests, and thus the contract, between the consumer and the provider (for example Places Manager API). These tests produce JSON files, known as pact artifcats, and describe how an API is expected to behave. These pact artifcats are then run against the provider codebase to confirm a pact.
> 
> As part of a CI process we need to test the pact whenever a consumer or provider changes. This involves executing [a GitHub action](https://github.com/alphagov/places-manager/blob/main/.github/workflows/pact-verify.yml) defined on the Provider. This action recieves different inputs depending on whether it is executed [by the consumer](https://github.com/alphagov/gds-api-adapters/blob/9f01b67805e915145a3cd7cbc80b6d4a7c9524a8/.github/workflows/ci.yml#L101-L105) or by [the provider](https://github.com/alphagov/places-manager/blob/03b06aa8e853020a1e1734786bab9e2ce2e68839/.github/workflows/ci.yml#L60-L64).  Where the consumer provides pact-artificats files generated as a result of changes to the consumer repository and where the provider downloads previously generated pact-artifacts based on the latest state of the consumer from a service, pact broker.

What do you think? 

I'd also suggest we move this section to be above the guide: "Running Pact tests locally" since it seems key information in proceeding with the further sections

source/manual/pact-testing.html.md
```diff
@@ -109,3 +109,25 @@ different applications.
...
+
+When the provider runs its CI pipeline, it does not set the `pact_artifact` input when it runs the `pact_verify` job, which causes it
+to retrieve the pact artifacts from Github and verify against them.
```
kevindew, line 120: The pacts are downloaded from the pact broker rather than from GitHub. You probably want to include the detail that the pacts that are downloaded default to those produced by the main branch of the consumer.

May also want to include the detail that this branch can be configured with pact_consumer_version, but may be able to get away with just a very short note that links to a an existing section on making breaking changes.

source/manual/pact-testing.html.md
```diff
@@ -109,3 +109,25 @@ different applications.
...
+
+When the consumer runs its CI pipeline, it:
+
+- generates the pact artifacts in the `generate_pacts` job. This generates the artifacts, but also uploads them to Github
+- runs the `pact_verify` job from each provider in turn, setting the `pact_artifact` input, which causes it to download the
+  pact artifacts from the pact broker (ie it's not using the artifacts it just created, but the previous versions). At this
```
kevindew, line 126: This is incorrect, the pacts that are used here are not from the Pact Broker - they are the ones uploaded to GitHub.

The GitHub upload/download is a bit of an annoying complex implementation detail - we just have to do that as a way of sharing files between actions and they're only uploaded temporarily for the duration of the CI run - I do wonder if going into much detail on that risks extra confusion.

source/manual/pact-testing.html.md
```diff
@@ -109,3 +109,25 @@ different applications.
...
+  point it uses the `pact_artifact_file_to_verify` value from the provider's `pact-verify.yml` file, which must be explicitly
+  set to match the implicit artifact names used when getting the pacts from Github (see below)
+- if all the verify jobs complete, it triggers the `publish_pacts` job, which publishes the generated pacts to the broker.
```
kevindew, line 129: I imagine we might want to make this section a little bit less GDS API Adapters specific and more generic to any app using pact tests - it also seems to be quite specific on specific implementation naming which seems risky for this staying in date.

Might want to go along the lines of saying:

The consumer CI job is responsible for generating new pact artifacts, verifying them against the provider and then, if successful, uploading them to the pact broker. The means it uses to achieve this is:
- running a task to generate the pacts (link to task) and uploading them to GitHub
- running the provider CI job to execute pacts with an argument that tells the provider to download these new pact artifacts (link to call) from GitHub to use as part of verifying
- running a task to upload the pact artifacts to the pact broker (link to task)

In some ways I do think perhaps this suggestion and are text are still a bit TMI for the dev docs but I'd understand in this pact case given how hard it is to understand.

source/manual/pact-testing.html.md
```diff
@@ -109,3 +109,25 @@ different applications.
...
+
+You should be careful that the names in the pact_helper.rb files in both the provider and consumer match, and that the value of
+`pact_artifact_file_to_verify` is "gds_api_adapters-<provider_id>.json", where the provider_id is the name from the pact_helper
+file in snake case.
```
kevindew, line 133: I'm not sure this is the right place for this gotcha, is this better included in the "Adding tests for a new app" ?

kevindew (commented):
Thanks so much for this @KludgeKML, I'm sorry I've suggested quite such extensive changes. I've just found myself spending the last hour and a half in the pact worm hole re-understanding and that involved a lot of rephrasing.

PR #4647, opened 2024-05-15
Rename imminence to places-manager

ChrisBAshton:
This should fix the broken deployment script.


KludgeKML:
Thanks both! I didn't realise this would have to be done this quick to allow dev docs to continue working.

PR #4657, opened 2024-05-17
Add note about cronjob

davidtrussler:
This change adds a note to the "Signon API user token expires soon" section to the effect that any alerts relating to Staging (currently) and Integration (in the future) do not need any action. This is because the overnight cronjob updates the token in this environment to match any changes made in Production.  The change is shown in the screenshot below. 

|Current|Updated|
|-|-|
|Screenshot 2024-05-17 at 12 30 47|Screenshot 2024-05-17 at 12 30 18|

### DIFF
--- a/source/manual/alerts/signon-api-user-token-expires-soon.html.md
+++ b/source/manual/alerts/signon-api-user-token-expires-soon.html.md
@@ -13,6 +13,10 @@ expiring tokens to ensure the associated application keeps working.
 
 > 🪨 This procedure is [toil](https://sre.google/workbook/eliminating-toil/) and should be eliminated or automated away. For now, it's unfortunately still a manual process.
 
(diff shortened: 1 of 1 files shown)


gclssvglx (approved):
Nice job @davidtrussler 🥇

ChrisBAshton:
Nice one! We can probably also stop the alerts from firing in Integration and Staging altogether 🤔

ChrisBAshton:
Now dropped from Staging, so we should consider revising this @davidtrussler 🙏 
https://github.com/alphagov/govuk-helm-charts/pull/2059

davidtrussler:
> Now dropped from Staging, so we should consider revising this @davidtrussler 🙏 [alphagov/govuk-helm-charts#2059](https://github.com/alphagov/govuk-helm-charts/pull/2059)

Cheers Chris, yes, I've passed that onto the next shift with an explanation of events.

PR #4666, opened 2024-05-22
Fix site search

MartinJJones:
## What

- Copy over `modules.js` and `dependencies.js` from the publishing components gem.
- Update developer docs version of `modules.js` to add support for initialising modules from the tech-docs gem that rely on jQuery

Although it would be possible to update `modules.js` in the publishing components gem to avoid copying over from the gem, we would likely need to include some checks for jQuery which doesn't feel like quite the right approach as it could send the wrong message for jQuery support in the govuk-publishing-components gem.

## Why

Site search in the developer docs was defaulting to use Google search following a recent update in `modules.js` in the publishing-components-gem - https://github.com/alphagov/govuk-developer-docs/pull/4664

It appears this was conflicting with the approach used in the developer docs where jQuery is still used, and modules.js would initialise all modules on the page, even where jQuery is used.

[Trello card](https://trello.com/c/TXRwnlvj/2621-fix-site-search-in-developer-docs-s)

## Further info

There appears to be a console error when testing locally, this error was present before the update to 38.4.0 and may need to be investigated further.

```
Error starting Search component JS: window.stripPIIFromString is not a function
```

### DIFF
--- a/source/javascripts/dependencies.js
+++ b/source/javascripts/dependencies.js
@@ -1,0 +1,24 @@
+/* eslint-env */
+
+// Copied without change from the govuk-publishing-components gem
+// to allow use of a custom modules.js with jQuery support
+
+// This adds in javascript that initialises components and dependencies
+// that are provided by Slimmer in public frontend applications.
+// = require ./modules.js
+
+$(document).ready(function() {
+  window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
+  window.GOVUK.analyticsVars = window.GOVUK.analyticsVars || {};
+
+  // if statements ensure these functions don't execute during testing
+  if (typeof window.GOVUK.loadAnalytics !== "undefined") {
+    window.GOVUK.loadAnalytics.loadExtraDomains();
+    if (typeof window.GOVUK.analyticsGa4.vars === "undefined") {
+      window.GOVUK.loadAnalytics.loadGa4();
+    }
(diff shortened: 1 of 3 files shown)


source/javascripts/dependencies.js
```diff
@@ -0,0 +1,26 @@
...
+    window.GOVUK.loadAnalytics.loadExtraDomains();
+    if (typeof window.GOVUK.analyticsGa4.vars === "undefined") {
+      window.GOVUK.loadAnalytics.loadGa4();
+    }
+    if (typeof window.GOVUK.analyticsVars.gaProperty === "undefined") {
+      window.GOVUK.loadAnalytics.loadUa();
+    }
+  }
+
+  window.GOVUK.modules.start();
```
andysellick, line 25: I don't think you need this line in here - the tech docs gem has a `document.ready` that calls this anyway. From running your code locally it looks like `start` gets called three times, removing this takes it to 2 (which is still too many, but better).

andysellick (approved):
Looks good, just one small change.

source/javascripts/dependencies.js
andysellick: https://github.com/alphagov/tech-docs-gem/blob/main/lib/assets/javascripts/_start-modules.js#L10C1-L12C3

MartinJJones: Good spot!, thanks. I've removed this line as recommended. Doing this has also fixed the console error message above 😄

PR #4670, opened 2024-05-24
Fix broken search JavaScript

andysellick:
## What

- update the local dev docs copy of the modules initialisation code from the gem to include handling of modules that have a start function and expect a jQuery element to be passed to them (like the Search JavaScript - the tech docs modules are old and still rely on jQuery)
- update this whole thing to actively fully replace not only the modules.start code from the gem but also from the tech docs - this is still included earlier in the JS stack (slightly inefficient, but I think we're stuck with it until a change is made upstream) but then we overwrite it. It's later called by the tech docs JS in a document.ready
- include a copy of the dependencies.js code from the gem, to initialise the GA4 code

## Why
Recent changes to how modules are initialised in the gem meant that the tech docs modules weren't being initialised anymore, and therefore the dev docs search stopped working properly.

## Visual changes
None.

### DIFF
--- a/source/javascripts/application.js
+++ b/source/javascripts/application.js
@@ -1,5 +1,5 @@
 //= require govuk_tech_docs
-//= require govuk_publishing_components/dependencies
+//= require dependencies
 //= require govuk_publishing_components/lib/trigger-event
 //= require govuk_publishing_components/lib/cookie-functions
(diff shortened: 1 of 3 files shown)


andysellick:
@MartinJJones just raised this to check I understood your solution - we've basically arrived at the same conclusion, so I'm going to close this one 😁

PR #4690, opened 2024-06-11
Update policy on staffing 2nd line during away days

ChrisBAshton:
2nd line is no longer the full time job it used to be: the platform is much more stable, and most application support work is now being handled directly by the product teams. 2nd line is therefore largely an incident response team now.

It therefore seems difficult to justify someone's exclusion from away days, if most of the time there is very little for 2nd line to do. So the updated guidance allows it, on the proviso that the 2nd line engineers regularly check Zendesk and continue to be interruptible by PagerDuty.


KludgeKML (approved):
LGTM!

source/manual/2nd-line.html.md
```diff
@@ -70,6 +68,14 @@ If your working patterns are not compatible with a 9.30am-5.30pm shift, let the
 
 If you do not work a 5-day week, please talk to your delivery manager to arrange cover with another developer on your team.
 
+### Away days and all-staff events
+
+Before attending an all-staff event, team away-day or any other event that could keep you away from your laptop for long periods at a time: try to swap your shift with someone who is _not_ attending ([see above](#shift-swaps-working-patterns-and-sickness)).
```
leenagupte, line 73: Could we add that, it's not possible to swap we recommend that people keep their laptops on them and that tether from their (work) phone for internet access if necessary?

leenagupte (commented):
This looks good to me. It's great that in-hours 2ndline is changing to be more on-call in nature.

I've added one question in the away days section.

source/manual/2nd-line.html.md
leenagupte: Ah, it wasn't clear that they'd be expected to work on the incident whilst at an away day. So perhaps expand "and are prepared to drop what you're doing if anything urgent comes in" to append something like " to work on the incident.

So yes, you can go to the away day, but you must be able to sign in and work even if it's from a park bench.

ChrisBAshton: Added a tweak 👍

PR #4700, opened 2024-06-21
Mark CKAN as retired

AgaDufrat:
The repo has been archived


kentsanggds (approved):
Just a quick note that the CKAN fork is being retired and not CKAN in DGU

PR #4701, opened 2024-06-21
Update repos.yml

MuriloDalRi:
Remove retired repos. They will keep showing up in the search if we don't do this.


sengi (approved):
Thanks!

PR #4706, opened 2024-06-25
Update some links to Kubernetes pages

jackbot:
It looks like the `/kubernetes` prefix was added to these pages at some
point, but a couple of links got missed.


kevindew (approved):
Nice work fixing these. I've checked and they all seem good.

PR #4709, opened 2024-06-26
Ignore content_block_email_address schema

pezholio:
This schema uses a `type` of `email_address`, something which is not supported in older versions of `govuk_schemas`. We can’t bring in the new version yet, because of dependency issues with Middleman.

We have an [upstream PR](https://github.com/middleman/middleman/pull/2709) to fix this, but in the interests of unblocking things, we can safely ignore this new schema for now.

### DIFF
--- a/app/schema_names.rb
+++ b/app/schema_names.rb
@@ -1,0 +1,9 @@
+class SchemaNames
+  # TODO: We can't use `content_block_email_address`, as is relies on functionality
+  # in the later versions of `GovukSchemas`, which is blocked by dependency issues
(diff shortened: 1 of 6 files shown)


app/schema_names.rb
```diff
@@ -0,0 +1,7 @@
+class SchemaNames
+  # Return the names of all schemas, except `content_block_email_address` which is
+  # still in development
```
ChrisBAshton, line 3: I suggest adding a link to the upstream PR, alongside a "TODO" comment along the lines of "remove this when the PR is merged".

ChrisBAshton (approved):
LGTM, one suggestion below. Thanks! ⭐

PR #4716, opened 2024-07-02
Add timed updates info in the global banner documentation

KludgeKML:
https://trello.com/c/upXxaeBR

(Don't merge until https://github.com/alphagov/static/pull/3369 is merged)


source/manual/global-banner.html.md
```diff
@@ -26,6 +26,18 @@ In [`app/views/components/_global_bar.html.erb`]:
    not disappear after being seen 3 times.
 1. [Deploy Static]. See the [guidance on deploying Static](/manual/deploy-static.html).
 
+### Timed Updates
+
+If you now about banner changes in advance and need the updates to show at a
```
leenagupte, line 31: Typo: should be "know" rather than "now"

leenagupte (approved):
Thanks for adding this.

There's one typo, but that can be fixed without a re-review

source/manual/global-banner.html.md
KludgeKML: Thanks! Fixed.

PR #4718, opened 2024-07-03
Update RabbitMQ web control panel docs

davidgisbey:
## Description 

While doing some bulk requeuing of messages to populate an index on Opensearch, we followed this guide. It was good, but there was one thing that could potentially catch people out, which is you need to admin credentials to access the dashboard.

When you run

```bash
k relay host/publishingmq.<environment>.govuk-internal.digital 4430:443
```

it doesn't give you any useful feedback if you're using poweruser creds so i think it's worth documenting.


kevindew (approved):
Nice one.

PR #4719, opened 2024-07-03
Update govuk-helm-charts URLs

jackbot:
It looks like the links to these workflow config files has changed at
some point, so this updates them to point to the right files in the
repo.

Also adds a link to the Argo docs; see that commit's message for details.


kevindew (approved):
Great - thanks for fixing.

PR #4729, opened 2024-07-10
Remove retired repos

MuriloDalRi:

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -557,10 +557,6 @@
   type: Utilities
   team: "#govuk-whitehall-experience-tech"
 
-- repo_name: nested_form
-  type: Utilities
-  team: "#govuk-publishing-mainstream-experience-tech"
-
 - repo_name: optic14n
   team: "#govuk-publishing-platform"
   type: Gems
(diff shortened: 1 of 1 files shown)


sengi (approved):

brucebolt:
`widdershins` is [still being used in govuk-content-api-docs](https://github.com/alphagov/govuk-content-api-docs/blob/b83875275a6762b878f6cc8c027ab1f4700efb82/package.json#L3).  Could you add it back into the list of repos?

sengi:
I don't think we should do that — it's still an unmaintained fork so it ought to stay archived unless you're willing to become its maintainer and keep it up to date.

There's no immediate problem with depending on an archived repo — it just makes it clearer that you have a (pre-existing) problem to address sometime in the not too distant future.

sengi:
Otherwise we're just sweeping serious security risks under the rug.

brucebolt:
> Otherwise we're just sweeping potentially serious security risks under the rug.

Surely we're doing that by archiving the repo and therefore hiding the security alerts associated with it?

sengi:
Not really, we're making it clear that people are depending on unmaintained and insecure software, which is a problem that you already have right now.

You're free to become a maintainer of the fork if you wish.

PR #4734, opened 2024-07-11
Manual for disabling test coverage checking

unoduetre:
## What

Documentation for disabling test coverage by 2nd line.

## Why

As explained in the comment in this [Trello ticket](https://trello.com/c/6cX3Xipx/2664-investigate-adding-javascript-test-coverage-and-implement-it-if-possible-l)

### DIFF
--- a/source/manual/disable-test-coverage-checking.html.md
+++ b/source/manual/disable-test-coverage-checking.html.md
@@ -1,0 +1,75 @@
+---
+owner_slack: "#govuk-2ndline-tech"
+title: Disable test coverage checking
+section: 2nd line
+layout: manual_layout
+parent: "/manual.html"
+---
+
+> ⚠️ This information should only be used by 2nd line when it is necessary to
+> apply a hotfix e.g. because of an ongoing incident, and they need to act
+> quickly. Test coverage checking should not be disabled during normal
+> development.
+
+Some applications check for test coverage in continuous integration. If test
+coverage requirement (the standard value is 95%) is not met, the testing step in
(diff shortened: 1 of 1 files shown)


source/manual/disable-test-coverage-checking.html.md
ChrisBAshton, line 1: Thanks for documenting this. I'm not against merging it, but have a few reasons for reconsidering it:

1. Applying a hotfix during an incident is more easily and reliably achieved by [manually deploying](https://docs.publishing.service.gov.uk/manual/deployments.html#manual-deployments) the branch or release.
2. A [relatively small proportion of GOV.UK apps](https://github.com/search?q=org%3Aalphagov%20minimum_coverage&type=code) have `minimum_coverage` set, so these instructions will often not be applicable.
3. I'd wager that someone who _does_ encounter the minimum coverage issue blocking their deployment, would either force it through by following 1, or would figure it out or ask for help on Slack, rather than searching for and finding this document.

So on balance I'd suggest we don't merge this (perhaps adding a line or two of example contexts to https://docs.publishing.service.gov.uk/manual/deployments.html#manual-deployments instead). What do you think?

unoduetre: I'm fine with that. If you think it's not required and we already have a workaround (1), then I think we don't need it.

unoduetre:
I'm closing this pull request based on the comment above. We already have a workaround for issues like that: https://docs.publishing.service.gov.uk/manual/deployments.html#manual-deployments

PR #4735, opened 2024-07-11
Update "Change an organisation's slug" instructions

ChrisBAshton:
Updating the slug in Manuals Publisher is much easier as of https://github.com/alphagov/manuals-publisher/pull/2427.

And updating 'best bets' has changed.

Trello: https://trello.com/c/G6OS639d/2731-reslug-ministry-of-housing-communities-local-government


ryanb-gds (approved):
Bonus drive-by search update 👀

PR #4740, opened 2024-07-15
Include worker process in secret env var steps

kevindew:
Worker processes also need a rolling restart to set the env var.


theseanything:
We've fixed the automatic restarts for worker deployments in https://github.com/alphagov/govuk-helm-charts/pull/2179 - so shouldn't need this extra step.

kevindew:
Great - even better. Thanks

PR #4741, opened 2024-07-15
Update govsearch apps in repos.yml

richardTowers:
Adding `production_hosted_on` should cause these to show up in the list of "apps", rather than just repos.

I think #data-insights-team is a more accurate team name than #data-products

This should make it a little more difficult to forget that govsearch exists, which we should be careful not to do because the app meets an important need for publishing users that is otherwise not coverd by GOV.UK.

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -108,7 +108,7 @@
 
 - repo_name: data-community-tech-docs
   type: Utilities
-  team: "#data-products"
+  team: "#data-insights-team"
 
 - repo_name: datagovuk-tech-docs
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
LGTM 👍 (Should all the other #data-products repos in this file go into #data-insights-team instead? That's in line with the "[GOV.UK Data Support](https://docs.publishing.service.gov.uk/manual/zendesk.html#summary-of-gov-uk-zendesk-groups) area we've set up in Zendesk)

ChrisBAshton (approved):

PR #4750, opened 2024-07-24
Fix Kibana links for Whitehall

ChrisBAshton:
Fixes the Kibana links on https://docs.publishing.service.gov.uk/repos/whitehall.html. See commits for details.

> screenshot of whitehall doc page


ryanb-gds (approved):
🖊️

PR #4781, opened 2024-08-22
Add guidance for the finder schema

GDSNewt:
To increase knowledge of the Specialist Publisher application, we thought we could do some digging into the finder schema and provide some additional documentation on what it controls on GOV.UK. 

The constraints of the developer docs mean that it is currently just taking the form of a simple table, I hope this proves readable enough, but suggestions on how this could be improved would be welcome. 

Whilst researching the finder schema I came across a few properties that seem deprecated and barely used. I believe there is a discussion to be had on whether we should remove some of these. I have [raised an issue](https://github.com/alphagov/publishing-api/issues/2851) in the publishing api repo. 

 For now though I have simply documented what the intention behind each property is/was. 

[Trello](https://trello.com/c/FrdGMVRI/2745-document-specialist-publisher-system-behaviours)


ryanb-gds (approved):
Thanks @GDSNewt, this is exceedingly useful 🙌

PR #4799, opened 2024-09-13
Update 'Drill updating homepage popular links' docs

jon-kirwan:
## What

Update 'Drill making changes to the homepage' docs

## Why

It is no longer necessary to drill into updating popular links on the homepage, as this can now be managed through Signon / Homepage. Therefore, I have removed that section.


nicholsj (approved):
Thanks!

PR #4817, opened 2024-09-27
Runbook for mobile backend remote config

jflm:
First pass at a runbook for making updates to remote config for 2nd line support - focusing mainly on the 'emergency' changes that might be needed, e.g. switching off a particular platform


gclssvglx (approved):
Cool - seems straight-forward to me 👍

PR #4818, opened 2024-09-30
Remove mention of auto-rotating tokens.

callumknights:
This behaviour was removed with [this PR](https://github.com/alphagov/signon/pull/2294)

Trello: https://trello.com/c/l6JLGp25/1381-spike-how-and-when-do-auto-rotating-api-tokens-rotate


Gweaton (approved):

brucebolt (commented):
Are you sure this functionality was removed?

[These lines of code](https://github.com/alphagov/signon/blob/7906b1205ab389e77195e6e80462a179b2859a3d/lib/sso_push_credential.rb#L13-L17) look like they're issuing a new authorisation (i.e. token) that lasts for 10 years if the current token expires in the next 4 weeks.

callumknights:
> Are you sure this functionality was removed?
> 
> [These lines of code](https://github.com/alphagov/signon/blob/7906b1205ab389e77195e6e80462a179b2859a3d/lib/sso_push_credential.rb#L13-L17) look like they're issuing a new authorisation (i.e. token) that lasts for 10 years if the current token expires in the next 4 weeks.

Damn you might be right...

I'm going to close this PR for now now then and leave this in place, this is very confusing...

PR #4846, opened 2024-10-22
Update on-call docs

ChrisBAshton:
- Move on-call charter to top of doc - this is the bit of documentation we'll be pointing on-call engineers towards on a regular basis, so shouldn't be tucked away at the bottom
- Removes duplication of on-call checklist, which is covered by the Trello card template

PS: I hope to soon consolidate https://docs.publishing.service.gov.uk/manual/2nd-line.html into https://docs.publishing.service.gov.uk/manual/on-call.html, but we can do that in https://trello.com/c/d2z5WSze/254-rebrand-2nd-line-to-simply-on-call

Trello: https://trello.com/c/WAVXDbAr/263-review-and-consolidate-checklist-template-cards

### DIFF
--- a/source/manual/on-call.html.md
+++ b/source/manual/on-call.html.md
@@ -1,50 +1,36 @@
 ---
 owner_slack: "#govuk-2ndline-tech"
-title: Out of hours support (on-call)
+title: On-call support
 section: 2nd line
 layout: manual_layout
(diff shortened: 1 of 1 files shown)


KludgeKML (approved):
LGTM!

PR #4847, opened 2024-10-22
Remove references to shadow and standby roles

ChrisBAshton:
Standby roles were reduced from 2 to 1 in Q2, and from 1 to 0 in Q3.

Shadow roles are of very little value now, as activity on Tech 2nd Line is now negligible. We'll instead explore a system for shadowing incidents.

References to these two roles have now been removed, and I've fixed up some other obviously broken / out of date info while I'm here. A more full review of the doc will happen when we consolidate the Tech 2nd Line and On Call docs.

Trello: https://trello.com/c/BDcPvHJM/262-consider-archiving-the-shadowing-2nd-line-on-govuk-template-card


KludgeKML (approved):
LGTM

PR #4862, opened 2024-10-29
Add a lead time to transition requests

JonathanHallam:
People keep requesting transition sites be added immediately which isn't really practicable.

Trello - https://trello.com/c/SOPCsAsD/1421-update-the-public-guidance-about-transition-to-include-a-lead-time


brucebolt (commented):
Should this be in the [public guidance for departments](https://www.gov.uk/guidance/govuk-transition-guidance-for-agencies) instead of the internal (albeit pubic) developer docs?

JonathanHallam:
> Should this be in the [public guidance for departments](https://www.gov.uk/guidance/govuk-transition-guidance-for-agencies) instead of the internal (albeit pubic) developer docs?

I had no idea this existed. Yes, yes it should!

PR #4866, opened 2024-10-30
Update guidance on assets

lauraghiorghisor-tw:
[Trello](https://trello.com/c/YIs8YZmF/3146-fix-live-assets-on-superseded-editions)

### DIFF
--- a/source/manual/manage-assets.html.md.erb
+++ b/source/manual/manage-assets.html.md.erb
@@ -205,3 +205,42 @@ steps to remove the asset in Asset Manager:
    attachment_data = Asset.find_by(asset_manager_id: asset_id).assetable # e.g. 57a9c52b40f0b608a700000a
    attachment_data.update(file_size: 123, number_of_pages: 28) # file_size in bytes
    ```
+
+## Whitehall known asset bugs
+
+There are a few known bugs in Whitehall, that cause assets to be live when they should not be.
+These are currently (end of 2024) being worked on, and a data clean up has been run.
+In case you see anything that matches the patterns below, here's how you can fix it.
+You can still just delete the asset, but this approach is less destructive and more aligned with the intended behaviour.
+
+1. On a new draft of a published edition, if the user replaces a previously published attachment, then deletes it, and publishes the draft. The original asset remains live. However, the attachment is deleted, so the user does not know anything has gone wrong.
+
+The problem seems to be that the deleted replacement's draft state is not set to false.
+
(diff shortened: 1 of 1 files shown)


source/manual/manage-assets.html.md.erb
```diff
@@ -19,6 +19,8 @@ For example, for `https://assets.publishing.service.gov.uk/government/uploads/up
 
 ## Remove an asset
 
+Some applications, such as Whitehall, might require setting a redirect, rather than deleting the asset. Look out for replacements, or unpublished state of the parent edition, in the publishing application, and set the `replacement_id` or `redirect_url` in asset manager accordingly.
```
ChrisBAshton, line 22: Could we expand this out into its own "Redirect an asset" section instead? And maybe add some blurb at the top of the doc, where we explain when to redirect and when to simply remove.

I'd love to see this section expanded out with code examples (e.g. how to set the `redirect_url`? Whether to set `replacement_id` or `redirect_url` - which is preferred? etc)

lauraghiorghisor-tw: There already is a redirect a section but I wanted to put it in the delete section as a reminder for people like me who totally forgot there are other states and just fixed all zendesk ticket assets with deletion 🙈

I can do an intro section with an explanation of the various states of assets, at least what I understand of it so far. And move this bit in that section, at the top of the file.

The [redirect section](https://github.com/alphagov/govuk-developer-docs/blob/f635b86f7833b73fa437ab6e55bfcaf2e138c2c6/source/manual/manage-assets.html.md.erb#L48) uses a bit of an odd syntax for the rake task. [Rake task ](https://github.com/alphagov/asset-manager/blob/bf1189b102a2c6d104fe5c9df601c3fa801d895d/lib/tasks/assets.rake#L15)seems alright. Never used it though.

ChrisBAshton: Sounds good - let me know if you want a hand with syntax etc.

ryanb-gds (commented):
I don't know if the developer docs are quite the right place for some of this stuff - would probably be good to move the heavier stuff into the Asset Manager or Whitehall docs and then link to them from here. I feel the developer manuals are more aimed at "I want to quickly fix a problem without having to understand the whole of asset publishing, what do I do?"

lauraghiorghisor-tw:
On the back of comments: 
- Extracted the fields summary out into this PR - https://github.com/alphagov/asset-manager/pull/1568/files

minhngocd (approved):
Looks great - thank you 🙏

Issue #4865, opened 2024-10-30
Deployment broken


ChrisBAshton:
Last successful run: https://github.com/alphagov/govuk-developer-docs/actions/runs/11499000463
October 24th, 1:11pm

First bad run: https://github.com/alphagov/govuk-developer-docs/actions/runs/11500022582/job/32009292992
October 24th, 2:17pm

```
I, [2024-10-24T13:17:02.286597 #2071]  INFO -- response: Status 200
/home/runner/work/govuk-developer-docs/govuk-developer-docs/app/github_repo_fetcher.rb:22:in `repo': alphagov/govuk-mobile-remote-config not found (RuntimeError)
	from /home/runner/work/govuk-developer-docs/govuk-developer-docs/app/github_repo_fetcher.rb:40:in `docs'
	from /home/runner/work/govuk-developer-docs/govuk-developer-docs/app/proxy_pages.rb:13:in `block in repo_docs'
	from /home/runner/work/govuk-developer-docs/govuk-developer-docs/app/proxy_pages.rb:12:in `map'
	from /home/runner/work/govuk-developer-docs/govuk-developer-docs/app/proxy_pages.rb:12:in `repo_docs'
	from /home/runner/work/govuk-developer-docs/govuk-developer-docs/app/proxy_pages.rb:3:in `resources'
	from /home/runner/work/govuk-developer-docs/govuk-developer-docs/config.rb:56:in `evaluate_configuration!'
	from /home/runner/work/govuk-developer-docs/govuk-developer-docs/vendor/bundle/ruby/3.3.0/gems/middleman-core-4.5.0/lib/middleman-core/application.rb:329:in `instance_eval'
	from /home/runner/work/govuk-developer-docs/govuk-developer-docs/vendor/bundle/ruby/3.3.0/gems/middleman-core-4.5.0/lib/middleman-core/application.rb:329:in `evaluate_configuration!'
	from /home/runner/work/govuk-developer-docs/govuk-developer-docs/vendor/bundle/ruby/3.3.0/gems/middleman-core-4.5.0/lib/middleman-core/application.rb:286:in `initialize'
	from /home/runner/work/govuk-developer-docs/govuk-developer-docs/vendor/bundle/ruby/3.3.0/gems/middleman-cli-4.5.0/bin/middleman:49:in `new'
[...]

PR #4885, opened 2024-11-15
Add Slack notification on deployment failure

ChrisBAshton:
See #4865 - our docs were failing to deploy for around a week and nobody noticed. This is an attempt to remedy that.

The 'notify failure' action is copied from govuk-rota-generator and the secret used is `govuk/slack-webhook-url` from AWS Secrets Manager.


AgaDufrat (approved):
🌟

PR #4898, opened 2024-11-20
Add ga4-dataform repo

AgaDufrat:
New GOV.UK repo

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -170,6 +170,10 @@
     - frontend
     - draft-frontend
 
+- repo_name: ga4-dataform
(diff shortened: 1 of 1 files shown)


Nyzl:
thanks @AgaDufrat i'm a bit slow getting the repo's in order. Would it be alright to have a new type because this isn't really data-science, it's more data-engineering?

AgaDufrat:
Can't see a reason why not. In this case it's only used to group repos in the docs: https://docs.publishing.service.gov.uk/repos.html#repos-by-type

MuriloDalRi (approved):

PR #4899, opened 2024-11-20
Add instructions for running a load test with k6

brucebolt:
This gives instructions for running a load test on AWS using k6.

The need originated from [this Trello card](https://trello.com/c/P7cdLbnK).

### DIFF
--- a/source/manual/load-test.html.md
+++ b/source/manual/load-test.html.md
@@ -1,0 +1,124 @@
+---
+owner_slack: "#govuk-developers"
+title: Run a load test with k6
+section: Testing
+layout: manual_layout
+parent: "/manual.html"
+---
+
+These instructions are based on [this blog post](https://grafana.com/blog/2022/06/23/running-distributed-load-tests-on-kubernetes/).
+
+1. Determine the name of the Kubernetes context to use:
+
+    ```sh
+    kubectl config get-contexts
+    ```
+
+    The output will list the environments and their names (e.g. `integration`, `staging` or `production`, or however they are named in your `~/.kube/config` file).
+
+1. Log into AWS and set the context to the environment you wish to test (this is the context name from the previous step):
+
+    ```sh
+    ENVIRONMENT=staging
+    ROLE=admin
+
+    eval $(gds aws govuk-${ENVIRONMENT}-${ROLE} -e --art 8h)
+    kubectl config use-context $ENVIRONMENT
+    ```
+
+    > You will need to assume the `admin` role in order to deploy k6 to the cluster.
+
+1. Clone the `k6-operator` repo and make this:
+
+    ```sh
+    git clone git@github.com:grafana/k6-operator.git && cd k6-operator
+    make deploy
+    ```
+
+    > If you get an error similar to `invalid go version '1.22.0': must match format 1.23` then you need to upgrade `go` to the correct version.
+    > This can be done by using Homebrew, e.g. `brew install go@1.23`
+
+1. Write your test script, for example to test 3000 virtual users against the Collections app's `/world` page over 30 seconds (bypassing router):
+
+    ```javascript
+    import http from 'k6/http';
+    import { check } from 'k6';
+
+    export const options = {
+      vus: 3000,
+      duration: '30s',
+    }
+
+    export default function() {
+      const res = http.get('http://collections.apps.svc.cluster.local/world');
+
+      check(res, {
+        'is status 200': (r) => r.status === 200,
+        'is status 2xx, not 200': (r) => r.status >= 201 && r.status <= 299,
+        'is status 3xx': (r) => r.status >= 300 && r.status <= 399,
+        'is status 4xx': (r) => r.status >= 400 && r.status <= 499,
+        'is status 5xx': (r) => r.status >= 500 && r.status <= 599,
+      });
+    }
+    ```
+
+1. Run your code locally to make sure it works:
+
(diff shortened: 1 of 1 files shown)


source/manual/load-test.html.md
```diff
@@ -0,0 +1,116 @@
...
+
+1. Log into AWS and set the context to the environment you wish to test:
+
+    ```sh
+    ENVIRONMENT=staging
+    ROLE=admin
+    AWS_REGION=eu-west-1
+
+    eval $(gds aws govuk-${ENVIRONMENT}-${ROLE} -e --art 8h)
+    kubectl config use-context $ENVIRONMENT
```
yndajas, line 19: **suggestion:** clarification on values

Looks like this assumes that you've named your context something like `staging`/`integation`/`production`. [This guidance](https://docs.publishing.service.gov.uk/kubernetes/get-started/access-eks-cluster/#access-a-cluster-for-the-first-time) suggests a format more like `govuk-staging`, although some people might have missed that advice. Might be worth linking to that guidance or something else in case anyone is confused about where these values come from and what they should be if this doesn't work for them

source/manual/load-test.html.md
```diff
@@ -0,0 +1,116 @@
...
+---
+
+These instructions are based on [this blog post](https://grafana.com/blog/2022/06/23/running-distributed-load-tests-on-kubernetes/).
+
+1. Log into AWS and set the context to the environment you wish to test:
+
+    ```sh
+    ENVIRONMENT=staging
+    ROLE=admin
+    AWS_REGION=eu-west-1
```
yndajas, line 16: **question:** variable usage

Is the `AWS_REGION` environment variable used under the hood somewhere? I can't see where it's used here

source/manual/load-test.html.md
```diff
@@ -0,0 +1,116 @@
...
+    eval $(gds aws govuk-${ENVIRONMENT}-${ROLE} -e --art 8h)
+    kubectl config use-context $ENVIRONMENT
+    ```
+
+    > You will need to assume the `admin` role in order to deploy k6 to the cluster.
+
+1. Clone the `k6-operator` repo and make this:
+
+    ```sh
+    git clone https://github.com/grafana/k6-operator && cd k6-operator
```
yndajas, line 27: **question:** HTTPS vs SSH

Any reason for the example using HTTPS rather than SSH? I thought SSH was generally preferred, but from a quick search it sounds as though GitHub flip flop on what they recommend

source/manual/load-test.html.md
```diff
@@ -0,0 +1,116 @@
...
+    make deploy
+    ```
+
+    > If you get an error similar to `invalid go version '1.22.0': must match format 1.23` then you need to upgrade `go` to the correct version.
+    > This can be done by using Homebrew, e.g. `brew install go@1.23`
+
+1. Write your test script, for example to test 3000 virtual users against the Collections app over 30 seconds (bypassing router):
```
yndajas, line 34: **suggestion:** testing scope

`against the Collections app's "/world" page`?

source/manual/load-test.html.md
```diff
@@ -0,0 +1,116 @@
...
+    ```sh
+    k6 run test.js
+    ```
+
+    > You may see errors related to connections as you may not be able to establish the correct number of connections from your machine.
+
+1. Deploy the test script:
+
+    ```sh
+    kubectl create configmap my-test-name --from-file /path/to/our/test.js
```
yndajas, line 70: **suggestion:** naming guidance

What's the meaning of the test name here? Is it something that will appear in some interface? Is there anything to bear in mind when choosing a name?

source/manual/load-test.html.md
```diff
@@ -0,0 +1,116 @@
...
+
+      check(res, {
+        'is status 200': (r) => r.status === 200,
+        'is status 2xx, not 200': (r) => r.status >= 201 && r.status <= 299,
+        'is status 3xx': (r) => r.status >= 300 && r.status <= 399,
+        'is status 4xx': (r) => r.status >= 400 && r.status <= 499,
+        'is status 5xx': (r) => r.status >= 500 && r.status <= 599,
+      });
+    }
+
```
yndajas, line 56: **typo:** trailing line break

This trailing line break isn't present in the other code blocks

yndajas (approved):
Looks good. Just a few minor questions/suggestions

source/manual/load-test.html.md
brucebolt: Thanks.  Hadn't spotted that.

source/manual/load-test.html.md
brucebolt: It's just a name that is used throughout to refer to the configmap (and therefore the name of the pods).  I'll add a note about this.

source/manual/load-test.html.md
brucebolt: Good idea.  I'll add on that clarification about the scope.

source/manual/load-test.html.md
brucebolt: Just stole it from the linked blog.  I believe we normally suggest SSH elsewhere, so I'll use that instead.

source/manual/load-test.html.md
brucebolt: I don't believe it is.  I'll remove this as the config was taken from my notes for doing some other things in AWS.

PR #4900, opened 2024-11-20
Update guidance on Govspeak regex Rake task

yndajas:
[Trello](https://trello.com/c/DuYxt7Gn/1474-improve-govuk-data-replication-and-whitehall-regex-rake-task-documentation)

- Provide more advice on how to produce an appropriately-escaped regex string for passing to the Rake task
- Provide guidance on how to run this locally should a regex timeout error prevent running it on a pod
- Rename file to match title


brucebolt (approved):
Maybe it's worth adding [a redirect](https://github.com/alphagov/govuk-developer-docs/blob/b3357e7b2f5aec32a92f720e18a994b73ba0fba5/config/tech-docs.yml#L30), in case anyone has the old URL bookmarked?

PR #4911, opened 2024-11-26
Fix broken incident management link

MartinJJones:
## What

Updated the "Incident Management" linked on the Status Page.

## Why

The previous link would result in a 404 error.


nimalank7 (approved):
LGTM!

PR #4915, opened 2024-11-27
Define alerts channel for Platform Engineering

nicholsj:
The ability for the Seal to use a separate channel for alerts was added [here](https://github.com/alphagov/seal/pull/570). 

This just defines that team for Platform Engineering-owned repos.


AgaDufrat:
Thanks! We just talked about it in a retro. It also needs updating in Seal https://github.com/alphagov/seal/blob/main/config/alphagov.yml#L254

AgaDufrat (approved):

PR #4949, opened 2025-01-14
Document trigger for "Support app" routed Platform tickets

ChrisBAshton:
Somehow this trigger rule was never created, and we only noticed today! Trigger rule now created and documented.


KludgeKML (approved):
LGTM

PR #4951, opened 2025-01-15
Remove secrets from Brakeman workflow example

jackbot:
The reusable workflow for running Brakeman[1] doesn't rely on any
secrets, so there's no need to pass `secrets: inherit` when calling the
workflow.

[1] https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/brakeman.yml


kevindew (approved):
Nice catch

PR #4953, opened 2025-01-17
Update data.gov.uk docs following removal of publish and opensearch

kentsanggds:
Remove references to Publish and opensearch and add information on how to connect to Solr to make direct queries.


source/manual/data-gov-uk-2nd-line.html.md
```diff
@@ -272,43 +214,13 @@ This is a generic response for such cases:
...
-```
-
-A [reindex](#reindexing-find) is then needed to update the status with the Elastic instance that serves data.gov.uk.
+### Getting data directly from Solr
 
-### Datasets published in CKAN are not appearing on data.gov.uk
+There are times when it is useful to connect to the Solr search index server in order to get the raw data that generates the search results and dataset information. To do this you need this command - 
 
-Check the Sidekiq queue (see [monitoring section](/manual/data-gov-uk-monitoring.html#sidekiq-publish)) to ensure the queue length is not too long.  You should not see more jobs than the number of datasets in CKAN.
+`kubectl port-forard ckan-solr-0 8983:8983 -n datagovuk`
```
nicholsj, line 221:
```suggestion
`kubectl port-forward ckan-solr-0 8983:8983 -n datagovuk`
```

nicholsj (approved):
Thanks @kentsanggds - one suggestion to fix a typo but looks good to me.

AgaDufrat:
Looks good. I may later add some examples of Solr queries and may be good to mention that Solr tends to have default limits on results it returns. We still need to [fully retire](https://docs.publishing.service.gov.uk/manual/retiring-a-repo.html) Publish.

PR #4961, opened 2025-01-22
Add skip_docs option, and skip siteimprove docs

richardTowers:
The siteimprove client includes [extensive auto-generated API documentation](https://docs.publishing.service.gov.uk/repos/siteimprove_api_client.html), which doesn't need to be mirrored to the developer docs.

If someone wants to read this documentation while working with the siteimprove client, they can still read it in GitHub.

### DIFF
--- a/app/proxy_pages.rb
+++ b/app/proxy_pages.rb
@@ -9,7 +9,7 @@ class ProxyPages
   end
 
   def self.repo_docs
-    docs = Repos.active_public.map do |repo|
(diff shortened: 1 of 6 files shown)


KludgeKML (approved):
LGTM (also, sorry! Unintended consequence, I didn't consider that the autogenerated Siteimprove stuff would get dragged into dev docs!)

jasonBirchall (approved):

PR #4966, opened 2025-01-28
Document Whitehall retagging form

ChrisBAshton:
Also reorders the instructions for the other publishing apps, and brings the Publishing API rake task to the bottom of the list, since we should be making all changes in the publishing apps themselves where possible.

Trello: https://trello.com/c/o9iYyNxf/623-add-ui-for-re-tagging-organisations-to-content-in-bulk


ryanb-gds (approved):
Worth adding a line to say "prefer updating via Publishing apps where possible" to the Publishing API docs?

PR #4977, opened 2025-01-31
Improve Implementation Guide page titles and meta descriptions

andysellick:
## What
Improves page titles and adds metadata descriptions for the [analytics section of the dev docs](https://docs.publishing.service.gov.uk/analytics/).

## Why
Previously no meta descriptions, so sharing pages didn't provide enough information.

## Visual changes
None.

Trello: https://trello.com/c/PmI26YFF/839-update-ga4-implementation-record-page-titles

### DIFF
--- a/data/analytics/navigation.yml
+++ b/data/analytics/navigation.yml
@@ -1,10 +1,12 @@
 analytics:
   name: Google Analytics 4 implementation record
   desc: This site contains a record of the data implementation for Google Analytics 4 (GA4) on GOV.UK Publishing. It was built to aid developers on GOV.UK Publishing but is available to anyone who may find our approach useful.
+  metadesc: This site contains a record of the data implementation for Google Analytics 4 (GA4) on GOV.UK Publishing.
 
 data:
   name: Data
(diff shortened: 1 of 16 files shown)


AshGDS (commented):
Looks OK to me 👍 , but I've noticed two things:
1. We're using the old GOVUK Crown, since you're modifying the analytics layout already would you like to include that change in this repo?
2. The social media preview image no longer exists: https://docs.publishing.service.gov.uk/images/govuk-large.png

No worries if you think they should be done in separate PRs.

andysellick:
@AshGDS I think best done in a separate PR. Not sure who's responsible for maintenance of the docs but probably something in their remit.

AshGDS (approved):
@andysellick OK - it looks like the crown is only wrong on the analytics docs though, so it would probably come back to us. The repo is owned by #govuk-developers so I think  the responsibility is for everyone to manage it. Do you want me to look into fixing these?

andysellick:
@AshGDS sorry, didn't quite get that, yes I'll take a look, thanks.

PR #4979, opened 2025-02-03
Set correct GOV.UK crown on analytics pages

andysellick:
## What / why
Changes the crown on the /analytics pages to match the new crown (already changed on the rest of the dev docs) to keep up to date with GOV.UK.

## Visual changes
Before

Screenshot 2025-02-03 at 09 56 02

After

Screenshot 2025-02-03 at 09 56 10


andysellick:
@AshGDS have fixed the crown as noted in https://github.com/alphagov/govuk-developer-docs/pull/4977#pullrequestreview-2587208802

However the broken sharing image you referred to seems to be part of the main tech docs template and not controlled directly inside the dev docs, and is broken on every page, so seems a bit trickier and best addressed separately.

AshGDS (approved):
Nice, thanks 👍

PR #4984, opened 2025-02-06
Add Rails upgrade guidance

yndajas:


brucebolt (commented):
This looks good to me.  I'll hold off approving until the rest of the team have taken a look.

source/manual/how-to-upgrade-rails.html.md
```diff
@@ -0,0 +1,190 @@
...
+
+   ```sh
+   bundle exec rails app:update
+   ```
+
+1. Accept all the changes with `a`.
+
+   > You can alternatively review each individual change via the interactive
+   > update process if you're comfortable reviewing and making decisions on
+   > diffs in the command line. It's a similar interface to `git add --patch`.
```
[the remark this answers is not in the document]
yndajas: Yeah, I added the bit below to try to indicate that this is an example workflow, but the first step (the one before this) is the important bit

> The exact process for this step might vary by app, developer, and team, but an example process is provided below. The first step is the key entry point.

source/manual/how-to-upgrade-rails.html.md
```diff
@@ -0,0 +1,190 @@
...
+
+- depend on another package which is incompatible with the new version of Rails;
+  or
+- still use [previously-deprecated now-unsupported Rails
+  syntax][unresolved-deprecation].
+
+Review the results of the test suite run via continuous integration and/or
+locally, and if possible, address any failing tests. If a fix is not currently
+possible, make a note (via a Trello card, for example) to come back to the
+upgrade at a later date.
```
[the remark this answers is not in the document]
yndajas: Maybe, although the mention of incompatible dependencies was only really intended as an example of why tests might fail, and I wonder it's a bit too case-by-case to specify?

source/manual/guidance-for-upgrades-to-rails-8-0.html.md
```diff
@@ -0,0 +1,56 @@
...
+- Accept the change to enable query log tags. This seems useful and harmless.
+- If the disallowed depecation warnings array is empty, accept the change to
+  remove two related settings. This setting does not affect whether deprecation
+  warnings are logged.
+
+## `config/environment/production.rb`
+
+- In general, we don't use Rails' public file server, so this should be/remain
+  disabled and its headers setting commented out.
+- SSL settings should be commented out. These settings have caused issues in the
```
AgaDufrat, line 40: For more context, this is an issue because Kubernetes liveness probes are configured to make HTTP requests to the apps. If a Rails app is configured to force SSL (redirect HTTP traffic to HTTPS), the probe gets 301 response (redirect). Since the probe doesn’t follow redirects, it fails as it’s expecting 200 response. This then causes the container to restart.

For the frontend apps, SSL is enforced by the CDN in our [Fastly config](https://github.com/alphagov/govuk-fastly/blob/4231616bb074792e6d8fb3560d80ad837abed892/www/www.vcl.tftpl#L224-L227).

source/manual/how-to-upgrade-rails.html.md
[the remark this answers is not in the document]
callumknights: I do the same as Bruce as well, I've been burned a few times by trusting the command line tools a bit too much. 
But I agree that as an example workflow the above is pretty good.

callumknights (approved):
Happy to approve, obviously still waiting on other team members though.

source/manual/guidance-for-upgrades-to-rails-8-0.html.md
yndajas: We actually [turned these settings on in Publishing API](https://github.com/alphagov/publishing-api/pull/3071/commits/94925c781b5f4fec61076385e31b70e90c8744ec#diff-da60b4e96eff2b132991226d308949e23f4ef3aad45ad59edd09cbc32cc6251eR22-R25) (somewhat by mistake) and have been surprised to see that everything _seems_ to be working as before

AgaDufrat: It's probably because it also enables `assume_ssl`

```
# Assume all access to the app is happening through a SSL-terminating reverse proxy.
  config.assume_ssl = true
```

source/manual/how-to-upgrade-rails.html.md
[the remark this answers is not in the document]
brucebolt: Fair point.  A more general sentence (as you've already written) seems fine here.

JonathanHallam (approved):
This looks good to me, thanks for writing it up!

source/manual/how-to-upgrade-rails.html.md
```diff
@@ -0,0 +1,194 @@
...
+
+- [8.0][]
+
+[8.0]: /manual/guidance-for-upgrades-to-rails-8-0
+
+### Switch to the new version defaults
+
+1. Update the version number provided to the `config.load_defaults` setting in
+   `config/application.rb`.
+1. Delete the `config/new_framework_defaults_*.rb` file.
```
brucebolt, line 179: I think this file exists at `config/initializers/new_framework_defaults_*.rb`, rather than at the top level of the `config` directory.

yndajas: Oops, thanks for spotting!

pezholio (approved):
Happy with this - thanks for documenting!

PR #4995, opened 2025-02-17
Fix broken links in production rules document

brucebolt:
We are no longer using `govuk-aws-data`, with access to AWS granted through `govuk-user-reviewer`.

Updating the documentation to reflect this.


KludgeKML (approved):
LGTM. Thanks!

PR #5010, opened 2025-03-04
CI Improvements

pezholio:
The build is [currently failing](https://github.com/alphagov/govuk-developer-docs/actions/runs/13657932250/job/38181773085), but it's difficult to see why. This PR adds verbose logging as well as failing fast so the build bails at the first error, so we can see what's going on a bit more easily. I've also fixed the bad JSON that was preventing the Slack alerts from sending.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -62,7 +62,7 @@ jobs:
                   "type": "section",
                   "text": {
                     "type": "mrkdwn",
-                     "text": "The <https://github.com/alphagov/govuk-developer-docs/blob/main/.github/workflows/deploy.yml|Developer Docs deployment script> failed.",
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -62,15 +62,14 @@ jobs:
...
+                     "text": "The <https://github.com/alphagov/govuk-developer-docs/blob/main/.github/workflows/deploy.yml|Developer Docs deployment script> failed."
                   },
                   "accessory": {
                     "type": "button",
                     "text": {
                         "type": "plain_text",
                         "text": "Check the build logs for details"
-                    },
-                    "url": "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}",
+                    },                    "url": "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}",
```
ChrisBAshton, line 72: This looks unintentional?

ChrisBAshton (approved):

PR #5019, opened 2025-03-10
Add mermaid.js as a dependency

pezholio:
https://trello.com/c/JfbTFW0e/917-tech-documentation

A lot of docs are loaded in from Github. As Github-flavoured Markdown [supports Mermaid for diagrams](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams), Mermaid markup is used in a lot of our Readmes and docs ([For example](https://github.com/alphagov/signon/blob/main/docs/access_and_permissions.md#dependencies-by-route)). This renders great in Github, but when we load it into developer docs, we just see the source code, for example:

image

This loads in Mermaid.js as a dependency, so now docs render like so:

image

This is the first time we've introduced JS-level dependencies to this repo, so I'd appreciate a steer that I've done it correctly!


ChrisBAshton (approved):
⭐

PR #5022, opened 2025-03-11
Alert Richard about graphql repo because it's not really platforms

JonathanHallam:
I don't fully know if this works, I guess we'll see!

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -327,7 +327,7 @@
 
 - repo_name: govuk-graphql
   team: "#govuk-publishing-platform"
-  alerts_team: "#govuk-publishing-platform-system-alerts"
+  alerts_team: "UAEH5J62J"
   type: Publishing apps
   description: Experimental high-performance GraphQL API for GOV.UK
   sentry_url: false


data/repos.yml
```diff
@@ -327,7 +327,7 @@
 
 - repo_name: govuk-graphql
   team: "#govuk-publishing-platform"
-  alerts_team: "#govuk-publishing-platform-system-alerts"
+  alerts_team: "@richard.towers"
```
brucebolt, line 330: It appears the use of usernames in place of a channel name has been [deprecated from the Slack API](https://api.slack.com/methods/chat.postMessage#channels).  You'll have to use Richard's user ID instead.

richardTowers: My ID is `UAEH5J62J` (I think)

richardTowers:
```suggestion
  alerts_team: "UAEH5J62J"
```

richardTowers (approved):

data/repos.yml
JonathanHallam: done

PR #5032, opened 2025-03-19
Update documentation for retiring an application

brucebolt:
This makes various changes to bring the documentation up-to-date.  See the individual commit messages for details.

[Trello card](https://trello.com/c/YV6VNRFO)

### DIFF
--- a/source/manual/retiring-an-application.html.md
+++ b/source/manual/retiring-an-application.html.md
@@ -17,7 +17,7 @@ Remove any [end-to-end tests][end-to-end] specific to the application.
 Mark the application as archived in the Release app.
 
 Edit the application in the release app (you'll need the `deploy` permission to
-do this), and check the `archived` checkbox. This will hide it from the UI.
+do this), and press the 'Delete application' button.
 
 ## 3. Update Signon
 
--- a/source/manual/retiring-an-application.html.md
+++ b/source/manual/retiring-an-application.html.md
@@ -64,6 +64,9 @@ items need to be unpublished and do it via the Publishing API.
 
 Since the application has been retired, it shouldn't be tracked in Sentry.
 
+Do not do this through the Sentry UI, instead remove the application from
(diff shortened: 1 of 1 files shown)


source/manual/retiring-an-application.html.md
```diff
@@ -82,10 +82,7 @@ Remove the app's entry in [govuk-helm-charts] from:
 - (if present) /charts/external-secrets/templates/<app name>
 
 It's also wise to search that repo for other references to the app being retired.
-Once the PR is merged ([Example PR][]), the app pods will automatically be removed by Argo.
-
-[govuk-helm-charts] (https://github.com/alphagov/govuk-helm-charts/)
-[Example PR] (https://github.com/alphagov/govuk-helm-charts/pull/1236)
```
yndajas, line 85: **note:** I think these lines were just missing colons after the square brackets, and line 76 was missing the second empty pair of square brackets after the first. There seems to be a slight preference for that form of URL markup in the docs from what I've seen, but it's a mixed bag, so this works fine too!

source/manual/retiring-an-application.html.md
```diff
@@ -47,7 +47,7 @@ these.
 
 ## 6. Remove from the GOV.UK architecture diagram
 
-- Remove the application from the [GOV.UK architecture diagram](/manual/architecture.html)
+Remove the application from the [GOV.UK architecture diagram](/manual/architecture.html).
```
yndajas, line 50: **praise:** nice thematic use of a full stop in the associated commit message 😅

yndajas (approved):

source/manual/retiring-an-application.html.md
brucebolt: I tend to use the colon/alias syntax (don't know the real name for it) when a link is being reference more than once.  Otherwise I prefer the in-line link.

But, as you say, there's a good mixture of both throughout the docs.

source/manual/retiring-an-application.html.md
brucebolt: Not intentional at all 😆

PR #5044, opened 2025-03-26
Add section on checking deployment on Argo CD

kentsanggds:
The new section will make it clearer that before testing the application the developer should check that their release has been successfully deployed.

### DIFF
--- a/source/manual/data-gov-uk-deployment.html.md
+++ b/source/manual/data-gov-uk-deployment.html.md
@@ -75,6 +75,14 @@ Builds can take up to 10 minutes. You can view progress in GitHub Actions under 
 
 Deployments typically take 5 minutes. You can view progress in Argo CD.
 
+### How to check that your release has been successfully deployed
+
+In order to validate that your release has been deployed you will need to check the relevant pod in the relevant environment for Argo CD. 
+
+For changes to `CKAN` you will need to look under the `ckan` application and for the `Find` app you will need to look under the `datagovuk` application. 
+
+If the Argo app health status is not green then the deployment has not been successful and will need investigating before you continue testing the web application as you will not be testing your release but the previous release as that will still be running.
+
 ### Promote a release to staging or production
 
 Once you are happy that your change works well in the integration environment, you can promote it to staging and then to production.


source/manual/data-gov-uk-deployment.html.md
```diff
@@ -75,6 +75,14 @@ Builds can take up to 10 minutes. You can view progress in GitHub Actions under
 
 Deployments typically take 5 minutes. You can view progress in Argo CD.
 
+### How to check that your release has been successfully deployed
+
+In order to validate that your release has been deployed you will need to check the relevant pod in the relevant environment for Argo CD.
```
AgaDufrat, line 80: I'm not sure 

```suggestion
In order to validate that your release has been deployed, you will need to check the status of the relevant pod in the appropriate environment in Argo CD.
```

source/manual/data-gov-uk-deployment.html.md
```diff
@@ -75,6 +75,14 @@ Builds can take up to 10 minutes. You can view progress in GitHub Actions under
...
+
+For changes to `CKAN` you will need to look under the `ckan` application and for the `Find` app you will need to look under the `datagovuk` application.
```
AgaDufrat, line 82: I'd add a link to speed up the checks.

```suggestion
For changes to `CKAN` you will need to look under the `ckan` application and for the `Find` app you will need to look under the `datagovuk` application. For example [datagovuk application in integration in Argo CD](https://argo.eks.integration.govuk.digital/applications/cluster-services/datagovuk)
```

source/manual/data-gov-uk-deployment.html.md
```diff
@@ -75,6 +75,14 @@ Builds can take up to 10 minutes. You can view progress in GitHub Actions under
...
+
+If the Argo app health status is not green then the deployment has not been successful and will need investigating before you continue testing the web application as you will not be testing your release but the previous release as that will still be running.
```
AgaDufrat, line 84: I'd avoid using colours to be more inclusive. 

```suggestion
If the Argo app health status is not "Healthy" then the deployment has not been successful and will need to be investigated by checking the pod's logs. You should not continue testing the web application in this case, as the new release hasn't been successfully deployed, and you will be testing the previous release instead.
```

AgaDufrat, line 84: Is there's a more explicit way in Argo to check what version (commit SHA) is currently running in the container?

AgaDufrat (commented):
I made a couple of minor mostly stylistic comments.

source/manual/data-gov-uk-deployment.html.md
kentsanggds: what do you mean by you are not sure?

AgaDufrat: I meant to delete that started sentence. My first thought was that this sentence was very generic, didn't really say what needed checking (there's a lot of info in Argo).

source/manual/data-gov-uk-deployment.html.md
kentsanggds: There are probably a number of ways to investigate why a pod is failing so I'm going to expland on this

kentsanggds: what do you mean, by an inspection of the source page? I don't think thats necessary, I'll add another step to check that the cluster is synced with the HEAD of the main branch commit though as it might not have synced with the PR merge

AgaDufrat: What I'd want to know is how can I check *in Argo* what version is deployed (currently running).  If it's the SYNC STATUS then would be good mention it.

The way I'd check it outside of Argo (which maybe a bit cowboy), would be to ssh into the machine and `cat` a file I expect to be changed to see if the change is there.

AgaDufrat (approved):

PR #5049, opened 2025-03-28
Add combined repos by team and type breakdown

yndajas:
It's easy to find all of a team's apps or all of a team's repos using the by team sections of the apps and repos pages. In Publishing Platform we recently had a need to find all of our gems (perhaps when looking to remove support for Ruby 3.1, which is reaching end of life). This was less easy. We could compare our repos by team with the list of gems in repos by type, or try to use our memory with either list, but neither solution is ideal, and the latter wouldn't work very well for people outside of or new to our team

This adds a combined breakdown of repos by team and type to help in such situations. Given the use cases for this are probably quite niche, I've added it as a separate page linked from the main repos page, rather than clogging up an already-complex page


ChrisBAshton (approved):
Nice - I love it!

yndajas:
@ChrisBAshton I've made a small addition since your review, but nothing controversial: just an opening sentence with a link to the main repos page in lieu of working breadcrumbs

PR #5055, opened 2025-04-01
Fix A/B testing documentation

brucebolt:
Two issues are fixed here:
- a broken link
- an instruction to remove a dictionary when you actually need to add it

[Trello card](https://trello.com/c/stxnESVI)


JonathanHallam (approved):
👍

PR #5061, opened 2025-04-04
Consolidate documentation on app deployment

ChrisBAshton:
Whenever I look up anything about our deployment process, I found myself flicking between three very similar documents:

- https://docs.publishing.service.gov.uk/manual/deployments.html
- https://docs.publishing.service.gov.uk/manual/development-pipeline.html
- https://docs.publishing.service.gov.uk/kubernetes/manage-app/access-ci-cd/#how-apps-are-deployed

There was a fair bit of duplication and outdated info in some of these, so I’ve consolidated them into one go-to doc.


source/manual/deployments.html.md
```diff
@@ -6,78 +6,139 @@ layout: manual_layout
...
+1. Open the [Argo CD dashboard](https://argo.eks.integration.govuk.digital/applications) for your app/environment.
+1. Click “Details” to verify the `ANNOTATIONS` include your commit SHA.
+1. Use “Sync” to speed up propagation.
+1. Confirm the Deployment pod is up-to-date (green tick).
 
-Deployments are triggered by changes to [app-config/image-tags](https://github.com/alphagov/govuk-helm-charts/tree/main/charts/app-config/image-tags), which Argo polls periodically.
-This includes image tags for all of the app repositories (e.g. Whitehall).
+## Deployment rules
 
-If you don't want to wait for Argo to poll the `app-config` chart, you can click the Refresh button in the app-config application in Argo (e.g. [app-config on integration](https://argo.eks.integration.govuk.digital/applications/cluster-services/app-config)).
```
richardTowers, line 49: I think we've lost some information here...

Your version says "Use sync to speed up propagation", but often times the thing that delays deployments is that Argo only poll's GitHub every few mins to refresh the app-config chart.

It goes:

- A commit is made to update `app-config/image-tags`
- DELAY
- Argo poll's GitHub and notices that `app-config/image-tags` has been updated, so it syncs the app-config project
- This updates the charts for the actual apps
- A synchronise is automatically triggered in the app with the new image-tag, which does the actual deployment

What that means is that clicking "Sync" in the app you care about usually doesn't make things go faster - It will sync the latest version of the chart that Argo has, but that won't have been updated from GitHub until app-config has refreshed.

So yeah, if you want to rush out a deploy, you need to refresh the app-config chart, not sync the app chart.

richardTowers: Ah no, it's still there, just further down.

Maybe it would be good to link to it here? I'm not sure whether `Use “Sync” to speed up propagation` is accurate in some cases, but in my experience it's usually refreshing app-config that makes things go.

ChrisBAshton: Good spot. I've added a link 👍

dj-maisy (approved):
👍

PR #5069, opened 2025-04-14
Remove App team repos from GOV.UK dev docs

sihugh:
These are managed by the app team independently of GOV.UK now.


leenagupte (approved):

ChrisBAshton (commented):
Does this need removing too, @sihugh ? https://docs.publishing.service.gov.uk/mobile/

PR #5079, opened 2025-04-25
add instructions for adding new schema to Content Data API

Harriethw:
https://trello.com/c/Fa5E0qo6/1031-tech-task-create-basic-contact-content-block-schema

This adds step to add the new schema type to Content Data API, so that its integration tests don't stop Publishing API merging https://gds.slack.com/archives/CAB4Q3QBW/p1745595957204829


pezholio (approved):
Good catch!

PR #5081, opened 2025-04-28
Add content-modelling-e2e repo

pezholio:
This adds the new https://github.com/alphagov/content-modelling-e2e repo, which contains a more extensive suite of e2e tests for content modelling, which can be run on demand.


nimalank7 (approved):
LGTM!

PR #5099, opened 2025-05-12
BAU - Add search-with-select-component and govuk-design-guide repos

nimalank7:
Description:
- Add `search-with-select-component` repo
- Add `govuk-design-guide` repo

### DIFF
--- a/data/repos.yml
+++ b/data/repos.yml
@@ -288,6 +288,12 @@
   type: Utilities
   production_hosted_on: eks
 
+- repo_name: govuk-design-guide
+  team: "#govuk-publishing-design-guide"
+  type: Publishing apps
+  description: GOV.UK Publishing Design Guide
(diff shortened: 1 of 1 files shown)


MuriloDalRi (approved):

data/repos.yml
```diff
@@ -288,6 +288,12 @@
   type: Utilities
   production_hosted_on: eks
 
+- repo_name: govuk-design-guide
+  team: "#govuk-publishing-design-guide"
+  type: Publishing apps
```
AgaDufrat, line 293: I don't think this should have it's own type, something like 'Design System'. Also the "Design System" team doesn't sit with in Publishing.

nimalank7: This is design guidance for the publishing team so I think it's reasonable to stick it there. The team that contributes to that repository is in that Slack channel (I checked).

PR #5103, opened 2025-05-14
Update Roles on the EKS Clusters Access page

dj-maisy:
## What?
There are some examples of the old AWS roles still being displayed on the EKS Access page. This will use the correct role names and add some other bits of information to the documentation.


syed-ali-tw (approved):
LGTM

PR #5128, opened 2025-06-02
Document the Platform Engineer role and access

dj-maisy:
## What?
This documents the new `platformengineer` roles that have been implemented for use by GOV.UK Platform Engineers.

### DIFF
--- a/source/manual/rules-for-getting-production-access.html.md
+++ b/source/manual/rules-for-getting-production-access.html.md
@@ -20,6 +20,7 @@ We have two types of production access:
 
 1. [Production Deploy access](#production-deploy-access)
 2. [Production Admin access](#production-admin-access)
+3. [Platform Engineer access](#platform-engineer-access)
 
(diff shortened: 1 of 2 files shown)


source/manual/rules-for-getting-production-access.html.md
```diff
@@ -84,6 +85,14 @@ A new starter/engineer will be expected to work through the following checklist
...
 - Best practices around the principle of least privilege, how to safely debug production issues, and how to work with credentials and accounts. E.g. by pairing with another developer to practise a drill on your product team.
 
+### Platform Engineer access
+
+Platform Engineer access is a special set of access permissions that are very similar to Production Admin, except with specific additional access to aid Platform Engineers with their day-to-day operational needs. In addition to the access granted by Production Admin, it also gives:
+
+- A special set of `-platformengineer` IAM roles for each environment that provide an access level similar to the `-developer` roles except also grants "Cluster Admin" access to our EKS clusters to allow Platform Engineers to access and manage all namespaces and resources
+
+This is necessary because without this, the only way to obtain Cluster Admin access would be to assule the `fulladmin` role on a regular basis, which we are trying to discourage except for "break glass" type scenarios that may trigger alerting.
```
nicholsj:
```suggestion
This is necessary because without this, the only way to obtain Cluster Admin access would be to assume the `fulladmin` role on a regular basis, which we are trying to discourage except for "break glass" type scenarios that may trigger alerting.
```

PR #5154, opened 2025-07-03
Update example PR for Whitehall

ChrisBAshton:
It really is just three lines of config now, not >400 files of locale files as per the previous example.


beccapearce (approved):
NOICEEEEE ✨

PR #5160, opened 2025-07-10
Remove URLs from email exemptions doc

deborahchua:
We have recently reviewed the exemption list and these pages are no longer live.

Reference:
- https://github.com/alphagov/government-frontend/blob/deb2c61424e86cea14cc363c18f68b89dcc7fe85/app/presenters/content_item/single_page_notification_button.rb

### DIFF
--- a/source/manual/get-emails-about-this-page.html.md
+++ b/source/manual/get-emails-about-this-page.html.md
@@ -32,4 +32,4 @@ If departments ask for any more pages to be excluded, we should try to refuse th
 If you must add or remove an exemption, update the [exemption list][exemption-list] by adding or removing the `content_id` of the publication which needs to be added or removed from the list.
 
 [get-emails-about-this-page]: https://components.publishing.service.gov.uk/component-guide/single_page_notification_button
-[exemption-list]: https://github.com/alphagov/government-frontend/blob/6ecd50c198565dc9bafdb72cb42b15411e795ad2/app/presenters/content_item/single_page_notification_button.rb#L4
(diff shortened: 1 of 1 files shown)


source/manual/get-emails-about-this-page.html.md
```diff
@@ -15,9 +15,7 @@ Before rolling out the feature, the GOV.UK account team contacted 14 departments
 Departments asked for these 4 pages to be excluded from the roll-out:
```
sairamya93: Just a tiny one - should that be `2` pages instead of `4` pages as we removed 2 of them?

sairamya93 (commented):
Added a tiny observation in.
Apart from that looks great!

source/manual/get-emails-about-this-page.html.md
deborahchua: Great spot thank you! That's updated now 👍

source/manual/get-emails-about-this-page.html.md
```diff
@@ -12,12 +12,10 @@ The [‘get emails about this page’ button][get-emails-about-this-page] was ro
 
 Before rolling out the feature, the GOV.UK account team contacted 14 departments who had content that offered users a work-around for subscribing to page-level updates.
 
-Departments asked for these 4 pages to be excluded from the roll-out:
```
KludgeKML, line 15: non-blocking nitpick: I wonder if we should say "Departments asked for 4 pages to be exluded, of which 2 now no longer exist. The remaining 2 are:" (which is more historically accurate).

sairamya93 (approved):
LGTM too!

PR #5162, opened 2025-07-14
Add Fastly DDOS protection docs

sihugh:
This service has been enabled, so we should add some docs for it.

### DIFF
--- a/source/manual/cdn.html.md
+++ b/source/manual/cdn.html.md
@@ -96,6 +96,30 @@ Fastly publish their cache node [IP address ranges as JSON from their API][fastl
 
 [fastly_ips]: https://api.fastly.com/public-ip-list
 
+## DDOS protection
+
+Our services that sit behind Fastly have [DDOS protection enabled](https://manage.fastly.com/security/ddos/protection).
+
+If we are experiencing an attack, and this is providing insufficient mitigation [start an incident](/manual/incident-what-to-do.html).
+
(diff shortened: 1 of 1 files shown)


source/manual/cdn.html.md
```diff
@@ -96,6 +96,30 @@ Fastly publish their cache node [IP address ranges as JSON from their API][fastl
...
+
+### Unblocking legitimate traffic
+
+It's possible that the automatic protection may block legitimate traffic. In this case, you may wish to [start an incident](/manual/incident-what-to-do.html).
+
+Sign in to Fastly and check for [DDOS events](https://manage.fastly.com/security/ddos/protection/events). There should be more information about any detected events there.
+
+You can inspect all the [matched rules for each event](https://www.fastly.com/documentation/guides/security/ddos-protection/about-the-ddos-protection-controls/#about-the-events-page) to check if any of them look like they're incorrectly matching legitimate traffic.
+
+You can [override any of the generated DDOS protection rules](https://www.fastly.com/documentation/guides/security/ddos-protection/about-the-ddos-protection-controls/#modifying-rule-behavior) to block, log only, accept the default, or just turn them off.
```
leenagupte, line 119: This isn't you or your docs, but it's not clear on a page for a single event what the rules are. Perhaps because the ones I can see are inactive?

leenagupte (approved):
I could follow the docs and all of the links work.

What each rule means is still a mystery!

PR #5164, opened 2025-07-15
[PP-7225] Add instructions for adding new language to govspeak

brucebolt:
Govspeak also contains translations, so new languages need to be added to this gem for full support on GOV.UK.


source/manual/add-support-new-language.html.md
```diff
@@ -72,7 +72,13 @@ At the time of writing, Whitehall content is rendered by either Frontend or Gove
 
 Follow steps 2, 3 and 4 from the [previous section](#1-add-support-for-your-new-language-to-the-frontend-rendering-applications) from within the govuk-publishing-components repo.
 
-### 3. Update Whitehall
+### 3. Add support for your new language to govspeak
+
+Follow steps 2, 3 and 4 from the [first section](#1-add-support-for-your-new-language-to-the-frontend-rendering-applications) from within the govspeak repo.
+
+After making this change, [release a new version the gem](/manual/publishing-a-ruby-gem.html#releasing-gem-versions) then update Whitehall and Publishing API to use this new version of govspeak.
```
yndajas: **question:** is there a reason for actively updating Whitehall and Publishing API (and not other apps that use Govspeak)?

source/manual/add-support-new-language.html.md
```diff
@@ -72,7 +72,13 @@ At the time of writing, Whitehall content is rendered by either Frontend or Gove
 
 Follow steps 2, 3 and 4 from the [previous section](#1-add-support-for-your-new-language-to-the-frontend-rendering-applications) from within the govuk-publishing-components repo.
 
-### 3. Update Whitehall
+### 3. Add support for your new language to govspeak
+
+Follow steps 2, 3 and 4 from the [first section](#1-add-support-for-your-new-language-to-the-frontend-rendering-applications) from within the govspeak repo.
```
yndajas: **issue:** it looks like the Rake task from step 4 comes from [rails_translation_manager](https://github.com/alphagov/rails_translation_manager/blob/main/lib/tasks/translation.rake#L51), which we don't have as a dependency in govspeak (we do in govuk_publishing_components and the frontend apps)

We could add it or just use simpler instructions like for the apps below?

yndajas: **issue:** in step 4 it looks like the Rake task should be `translation:import:all` not `translation:import_all`. See:
- https://github.com/alphagov/rails_translation_manager/blob/main/lib/tasks/translation.rake#L51
- https://docs.publishing.service.gov.uk/repos/rails_translation_manager/rake-command-reference.html

source/manual/add-support-new-language.html.md
yndajas: **typo:** "new version **of** the gem"

brucebolt: Whitehall is the only publishing application that supports translations.

source/manual/add-support-new-language.html.md
brucebolt: I hadn't spotted this gem didn't have `rails_translation_manager`.  I'll update to suggest just editing the YAML files directly.

brucebolt: Good spot.  I've fixed this in an additional commit.

source/manual/add-support-new-language.html.md
yndajas: Isn't this more about which apps use Govspeak?

yndajas (approved):

source/manual/add-support-new-language.html.md
brucebolt: I think the only apps where you _must_ upgrade govspeak to support the new language will be Whitehall and Publishing API, as none of the others currently make use of locales other than English.  All the other apps will get the dependency bumped by dependabot.

PR #5165, opened 2025-07-16
Add docs for site search

leenagupte:
[Trello](https://trello.com/c/7sPykzFQ)

Adds docs that explain:
1. How the new site search works
2. How to debug errors
3. How to complete common tasks.

Co-authored-by: @hannako

### DIFF
--- a/source/manual/search-synchronisation-errors.html.md
+++ b/source/manual/search-synchronisation-errors.html.md
@@ -1,0 +1,64 @@
+---
+owner_slack: "#govuk-search"
+title: "Search synchronisation errors"
+parent: "/manual.html"
+layout: manual_layout
+section: Search
+---
+
+## A document doesn't appear in site search after it is published
+
+When a document is published, publishing API places a message on Rabbit MQ's  `search_api_published_documents` queue.
+
+Search API v2's `document_sync_worker` listens to this queue, and co-ordinates the creation of a PublishingAPIDocument that is then synchronised to the VAIS datastore.
+
+If a document has been successfully published, is live on gov.uk, but is not being returned in search results, the first step is to confirm if this is the expected behaviour.
+
+1. Check the locale of the missing content_item. The document_sync_worker only sends documents with an `en` locale to the datastore.
+2. Check the document_type of the missing ContentItem. If it's is on the [document_type_ignorelist] the content has been intentionally ignored by the worker.
+3. Check the state of the missing ContentItem. Withdrawn content is desynchronised i.e. removed from the datastore.
+
+If you are sure that the document should be visible in results, you can try the following debugging steps:
+
+1. Check #govuk-search-alerts for any helpful synchronisation errors coming from search-api-v2
+2. Inspect the application logs for the search-api-v2-worker in kibana. A quick hacky way to find
+  the logs is to set the appropriate time window, and just search for ' "DiscoveryEngine" AND "<content-id>" '. A more systematic approach would be to [filter by] kubernetes.labels.app_kubernetes_io/name: search-api-v2 worker and search for the [content_id].
+3. Look for a message that explains why the worker has failed to synchronise the document. One
+  possible cause is that the payload version of the message received from the queue is lower than the previous time the document was synced. If this is the cause, [clear the redis cache] and then [resynchronise] the document.
+
+## An unpublished/withdrawn document is still present in search results
+
+When a document is withdrawn or unpublished via the UI in Whitehall, a message is placed on Rabbit MQ's `search_api_published_documents` queue.
+Search API v2's document_sync_worker listens to this queue, and co-ordinates the desynchronisation of the document from the VAIS datastore if the document type is on the [UNPUBLISH_DOCUMENT_TYPES list].
+
+If an unpublished document is still visible in search results, you can take the following debugging steps
+
+1. Confirm if the document_sync_worker received a request to update the document. Do this by checking in Kibana by searching for the [content_id].
+2. Check for a message which explains why the desynchronisation failed.
+3. If there is no log for this request, it could be that the document was manually deleted from the   whitehall database and so no message was placed on the search_api_published_documents queue. If this is the case, you can manually [delete the document from the datastore].
+
+## Wider synchronisation issues
+
+When documents are published or unpublished via a publishing app a message is placed on Rabbit MQ's `search_api_published_documents` queue.
+
+Search API v2's document_sync_worker listens to this queue and should co-ordinate the creation of a PublishingAPIDocument that is then synchronised to or desynchronised from the VAIS datastore.
+
+If all documents that have been successfully published or unpublished on gov.uk, but are not being returned as expected in search results it could be a sign that synchronisation process is failing in general for all documents. You can take the following debugging steps.
+
+1. Check #govuk-search-alerts for any helpful synchronisation errors coming from search-api-v2
+1. Confirm if the document_sync_worker received a request to update the document. Do this by checking in Kibana by searching by [content_id] for a few example documents.
+2. Check for messages which explains why the desynchronisation failed.
+4. One cause is that the payload version of the message received from the queue is lower than the previous time the document was synced. If this is the cause, [clear the redis cache] and then [resynchronise] all documents.
+5. If you are see a lot of `Google::Cloud` errors in the logs the likely cause is that VAIS is having issues. [Check the error rates for site search] and follow the steps for what to do [if site search is unavailable].
+
+[document_type_ignorelist]: https://github.com/alphagov/search-api-v2/blob/1c3e8115b15703a44691311a2971ce2dbee10c59/config/document_type_ignorelist.yml
+[filter by]: https://kibana.logit.io/s/13d1a0b1-f54f-407b-a4e5-f53ba653fac3/app/data-explorer/discover#?_a=(discover:(columns:!(_source),isDirty:!f,sort:!()),metadata:(indexPattern:'filebeat-*',view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))&_q=(filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'filebeat-*',key:kubernetes.labels.app_kubernetes_io%2Fname,negate:!f,params:(query:search-api-v2-worker),type:phrase),query:(match_phrase:(kubernetes.labels.app_kubernetes_io%2Fname:search-api-v2-worker)))),query:(language:kuery,query:''))
+[content_id]: https://kibana.logit.io/s/13d1a0b1-f54f-407b-a4e5-f53ba653fac3/app/data-explorer/discover#?_a=(discover:(columns:!(_source),isDirty:!f,sort:!()),metadata:(indexPattern:'filebeat-*',view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-7d,to:now))&_q=(filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'filebeat-*',key:kubernetes.labels.app_kubernetes_io%2Fname,negate:!f,params:(query:search-api-v2-worker),type:phrase),query:(match_phrase:(kubernetes.labels.app_kubernetes_io%2Fname:search-api-v2-worker)))),query:(language:kuery,query:%22225d80c5-01bb-47d0-b57c-6862efbed7b3%22))
+[ignore list]: https://github.com/alphagov/search-api-v2/blob/main/config/document_type_ignorelist.yml
+[clear the redis cache]: /manual/how-to-clear-the-redis-cache
+[resynchronise]: /manual/how-to-resync-content-in-vertex
+[document_sync_worker]: https://github.com/alphagov/search-api-v2/blob/1c3e8115b15703a44691311a2971ce2dbee10c59/lib/tasks/document_sync_worker.rake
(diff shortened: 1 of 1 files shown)


source/manual/boost-a-document-in-search.html.md
```diff
@@ -0,0 +1,19 @@
+---
+owner_slack: "#govuk-search"
+title: "Boosting a document in site search"
```
hannako: Maybe it should be titled: "Boost or demote a document in site search"

source/manual/published-document-does-not-appear-search.html.md
```diff
@@ -0,0 +1,24 @@
+---
+owner_slack: "#govuk-search"
+title: "A document doesn't appear in site search after it is published"
+parent: "/manual.html"
+layout: manual_layout
+section: Search
+---
+
+If a document doesn't appear in GOV.UK site search results after it has been published, you can follow a series of troubleshooting steps:
+
```
hannako: Do you think we should split this to make three separate pages? 

```
"A document doesn't appear in site search after it is published"  - how to debug a single missing doc
"A document is present in site search which should not be" - how to debug a doc in search that should have been deleted
"Search synchronisation errors" - a wider outage and how to investigage/fix.
```
So we could have:
```
---
owner_slack: "#govuk-search"
title: "A document doesn't appear in site search after it is published"
parent: "/manual.html"
layout: manual_layout
section: Search
---

When a document is published, publishing API places a message on Rabbit MQ's `search_api_published_documents` queue. 
Search API v2's DocumentSyncWorker listens to this queue, and co-ordinates the creation
 of a PublishingAPIDocument that is then synchronised to the VAIS datastore. 

If a document has been successfully published, is live on gov.uk, but is not being returned in search results, the first step is 
to confirm if this is the expected behaviour.

1. Check the locale of the missing content_item. The DocumentSyncWorker only sends documents with an `en` locale to the datastore.
2. Check the document_type of the missing ContentItem. If it's is on the [document_type_ignorelist](https://github.com/alphagov/search-api-v2/blob/main/config/document_type_ignorelist.yml),
 the content has been intentionally ignored by the worker. 
3. Check the state of the missing ContentItem. Withdrawn content is desynchronised i.e. removed from the datastore.

If you are sure that the document should be visible in results, you can try the following debugging steps:

1. Check #govuk-search-alerts for any helpful synchronisation errors coming from search-api-v2
2. Inspect the application logs for the search-api-v2-worker in kibana. A quick hacky way to find 
the logs is to set the appropriate time window, and just search for ' "DiscoveryEngine" AND "<content-id>" '.
But a more systematic approach would be to filter by kubernetes.labels.app_kubernetes_io/name: search-api-v2 worker and search for the content_id. 
3. Look for a message that explains why the worker has failed to synchronise the document.
One possible cause is that the payload version of the message received from the queue is lower than the previous
time the document was synced. 
If this is the cause, clear the redis cache and then resynchronise the document.

### How to clear the redis cache and resynchronise a document

I have no idea

```
I reckon we give kibana steps rather than a links so it's not tied to an environment?

hannako:
```
---
owner_slack: "#govuk-search"
title: "An unpublished/withdrawn document is still present in search results"
parent: "/manual.html"
layout: manual_layout
section: Search
---

When a document is withdrawn or unpublished via the UI in Whitehall, 
a message is placed on Rabbit MQ's `search_api_published_documents` queue. 
Search API v2's DocumentSyncWorker listens to this queue, and co-ordinates the 
desynchronisation of the document from the VAIS datastore
 if the document type is on the UNPUBLISH_DOCUMENT_TYPES [list](https://github.com/alphagov/search-api-v2/blob/main/app/models/concerns/publishing_api/action.rb#L6). 

If an unpublished document is still visible in search results, you can take the following debugging steps

1. Confirm if the DocumentSyncWorker received a request to update the document. Do this by checking in Kibana blah
2. Check for a message which explains why the desynchronisation failed. 
3. If there is no log for this request, it could be that the document was manually deleted from the whitehall database and so no message was placed on the search_api_published_documents queue. 
If this is the case, you can manually delete the document from the datastore.

### How to manually delete a document from search

```
I just sent a delete request to the api last time, but do you think we should have a rake task?!

source/manual/how-to-clear-the-redis-cache.html.md
```diff
@@ -0,0 +1,64 @@
...
+
+Search API V2 uses [Redis] to track the latest synced payload version for a document. It compares the payload version of the document being synced with the `latest_synced_version` number for that version in the cache.
+If the new `payload_version` is greater than the `latest_synced_version` the document will fail to sync. The Kibana logs should show an error for the document:
+
+```
+[DiscoveryEngine::Sync::Put] Ignored as newer version already synced
+```
+
+If this error is occurring for a lot of documents then the best option may be to clear the Redis cache.
+The [comparison logic] only executes if publishing-api sends a payload ID and the content_id has been seen before. Therefore clearing the cache would make it seem as though the content_id hasn't been seen before, thus triggering a sync regardless of payload version.
```
hannako: Doesn't it always execute, but if the content id isn't present in the cache then it will return true? i.e. [Redis.get](https://github.com/alphagov/search-api-v2/blob/main/app/services/coordination/document_version_cache.rb#L44) returns nil, and nil.to_i. returns 0, and https://github.com/alphagov/search-api-v2/blob/main/app/services/coordination/document_version_cache.rb#L24 returns true.

Are there any side effects of doing this?

And how come the resync just happens? Would you not need to put it back on the document_sync_worker queue?

leenagupte: If the content item isn't present in the cache, then `latest_synced_version` is `nil` and `sync?` [returns true](https://github.com/alphagov/search-api-v2/blob/main/app/services/coordination/document_version_cache.rb#L22C22-L22C43), so it doesn't reach the comparison logic.

The side effect of doing this is that if you represent every document from publishing-api, every document will be resynced, whereas if it existed in the cache, the sync would have been skipped as it would be known that Vertex already has the latest version of the document.

And I need to reword the last bit. The reync doesn't just happen, you'd need to manually requeue the document.

hannako: Ah gotcha thank you

source/manual/prevent-a-page-being-returned-in-search-results.html.md
```diff
@@ -0,0 +1,15 @@
...
+parent: "/manual.html"
+layout: manual_layout
+section: Search
+---
+
+## How to filter a page from Site search results
+
+To prevent a page being returned in site search results, a large negative boost can be applied to the DiscoveryEngine serving configuration.
+
+Add the path the filtered_pages array defined in [here][link-1]
```
leenagupte: Regarding the link, I'd limit the lines to only the array, and use a unique sha rather than linking to main which will change. E.g. https://github.com/alphagov/govuk-infrastructure/blob/0fbbc7017fd0a6334b38df72731f664252f14dbd/terraform/deployments/search-api-v2/serving_config_global_default.tf#L116-L120

source/manual/add-a-best-bet.html.md
```diff
@@ -0,0 +1,20 @@
...
+Sometimes we need to lift specific content to the top of search results, for example when a crucial new piece of content is added that hasn’t had time to naturally rise to the top by means of its popularity, or a page isn’t ranked well enough naturally.
+
+### How to add a best bet
+
+Add the path of the page to the [best bets YAML configuration file][link-1] in Search API v2, which will apply a [heavy query time boost][link-2] to the document.
+
+Note that it is not currently possible to have a document as a best bet if it is completely unrelated to a query, i.e. if it would not naturally occur in results for the query. For example, you wouldn’t be able to add a document about strawberry yoghurt to the best bets for the query “pool noodle”.
+
+[link-1]: https://github.com/alphagov/search-api-v2/blob/main/config/best_bets.yml
+[link-2]: https://github.com/alphagov/search-api-v2/blob/main/app/services/discovery_engine/query/best_bets_boost.rb#L11
```
leenagupte: Minor: but I'd link to the unique commit SHA rather than main so that the contents of the link never changes even as main does.

source/images/search-debugging-steps.png
leenagupte, line 1: Love this!

source/manual/govuk-site-search.html.md
```diff
@@ -0,0 +1,111 @@
...
+
+- Search API v2 receives the query, applies some query-time boosts to it and translates it to a VAIS API call, then maps the API response back to the format expected by Finder Frontend
+
+- Finder Frontend renders the results, together with a VAIS attribution token that is tracked by the analytics code
+
+- Google Analytics (GA4) tracks how the user interacts with results
+
+- GA4 data is processed overnight by the existing GOV.UK analytics pipeline, and ingested into GCP BigQuery.
+
+- Three times a day, the GA4 data data is imported and pushed into VAIS as user events. VAIS adjusts the internal ranking model for the engine based on these user events (and some data from its own proprietary search and internet analytics data)
```
leenagupte: Typo: An extra "data" on this line.

source/manual/published-document-does-not-appear-search.html.md
leenagupte: I've added these suggestions to the search synchronisation errors page.

source/manual/boost-or-demote-a-document-in-search.html.md
```diff
@@ -0,0 +1,19 @@
...
+owner_slack: "#govuk-search"
+title: "Boost or demote a document in site search"
+parent: "/manual.html"
+layout: manual_layout
+section: Search
+---
+
+Some kinds of content are more useful to our users than others. To account for this, we configure the VAIS engine to always apply certain “boosts” to search results. These are generally applied by [document type] but some documents are boosted based on their [base path].
+
+Boosts and demotions are configured in the [serving config] in the [govuk-infrastructure] repo and applied through Terraform.
```
beccapearce: I think we're missing the link to the terraform here
```suggestion
Boosts and demotions are configured in the [serving config] in the [govuk-infrastructure] repo and applied through [Terraform].
```

source/manual/govuk-site-search.html.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Architecture summary
+
+![site-search-diagram][link-1]
+
+Broadly, the new search stack for site search introduces the following new components to the GOV.UK technical ecosystem:
+
+- A deployment of **Vertex AI Search** (VAIS) in Google Cloud Platform (GCP) managed by Terraform, comprising per-environment GCP projects with a datastore (and schema), an engine, and serving configuration and controls
+
+- **Search API v2**: A Ruby on Rails application deployed on the standard GOV.UK AWS Kubernetes infrastructure, providing an API for querying (search and autocomplete) the VAIS serving configuration, a synchronisation module for updating the state of the VAIS datastore, and a scheduled task performing ongoing automated relevance evaluations.
```
beccapearce, line 36: This makes me think it might be worth having an explainer somewhere of the difference between search v1 vs v2 (maybe it exists but Might be useful to have it somewhere in these docs?)

source/manual/govuk-site-search.html.md
```diff
@@ -0,0 +1,111 @@
...
+
+- GA4 data is processed overnight by the existing GOV.UK analytics pipeline, and ingested into GCP BigQuery.
+
+- Three times a day, the GA4 data is imported and pushed into VAIS as user events. VAIS adjusts the internal ranking model for the engine based on these user events (and some data from its own proprietary search and internet analytics data)
+
+## Infrastructure
+
+### Deployment & environments
+
+The new search stack follows the same environment conventions as other GOV.UK applications, having integration, staging, and production environments. Our GCP account has separate projects for each environment with fully isolated resources.
```
beccapearce: Could we add a link to the GCP dev docs page here?
https://docs.publishing.service.gov.uk/manual/google-cloud-platform-gcp.html

beccapearce (commented):
This is all so helpful! Just a couple of comments of things i noticed 😄

source/manual/govuk-site-search.html.md
leenagupte: Do you mean how they work? That might take a bit of time as it's been a while since anyone dug into the ins and outs of how search-api v1 works,
or just which bits are powered by search-api / Elasticsearch and which bits search-api-v2 / VAIS?

leenagupte: @beccapearce There's this existing page: https://docs.publishing.service.gov.uk/manual/govuk-search.html
All of these pages should appear alongside this existing page under a new "Search" section in the manual.

beccapearce: I was a high level kind of thing covering which bits are v1. Along the lines of:
Search API v1 is a ruby on rails application which is used to produce search results for our finders using elasticsearch.

Very high level just to explain the difference. My motivation for bringing it up is more thinking about new starters and how v1 and v2 don't really mean much and this use of v1 and v2 is very specific, since they both do different things.
Maybe it doesn't need to be part of this PR. Or maybe we could add a link to this [doc](https://docs.publishing.service.gov.uk/manual/govuk-search.html) that has a nice bit explaining what each is for

beccapearce: @leenagupte yep I think that page does it!!  I'll resolve this :)

leenagupte:
@beccapearce Thanks for your review. I've responded to your comments and fixed up the requested changes.

beccapearce (approved):
Fabulous!!!! ✨ ✨ ✨

PR #5181, opened 2025-07-31
Updated URL for Request an AWS User

davidr-gds:
As per the change announced here:  https://gds.slack.com/archives/C05DC0SJ6UW/p1753954435102749

### DIFF
--- a/source/manual/get-started.html.md
+++ b/source/manual/get-started.html.md
@@ -62,7 +62,7 @@ You should [use your Yubikey as your MFA device][yubikey-aws-mfa] if you have on
 
 [aws-account-info]: https://reliability-engineering.cloudapps.digital/iaas.html#amazon-web-services-aws
 [iam-role-creation]: #6-get-permissions-for-aws-github-and-other-third-party-services
-[request-aws-user]: https://request-an-aws-account.gds-reliability.engineering/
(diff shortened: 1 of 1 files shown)


source/manual/get-started.html.md
```diff
@@ -62,7 +62,7 @@ You should [use your Yubikey as your MFA device][yubikey-aws-mfa] if you have on
 
 [aws-account-info]: https://reliability-engineering.cloudapps.digital/iaas.html#amazon-web-services-aws
 [iam-role-creation]: #6-get-permissions-for-aws-github-and-other-third-party-services
-[request-aws-user]: https://request-an-aws-account.gds-reliability.engineering/
+[request-aws-user]: https://request-an-aws-user.digital.cabinet-office.gov.uk/user
```
ChrisBAshton, line 65: It seems to redirect to this one, so may as well fix it at source:

```suggestion
[request-aws-user]: https://request-an-aws-user.digital.cabinet-office.gov.uk
```

davidr-gds: Yeah, can be argued both ways, I reckon. /user is there because that's the page you want to get to. It's just that the auth fails to return the user to that page after login. But maybe that will be fixed in future

PR #5199, opened 2025-08-14
Add app healthchecks and k8s probes info

jfharden:
Add documentation describing how the kubernetes probes interact with the app health checks, and how developers can configure which endpoints to hit for each probe type.


source/kubernetes/manage-app/control-healthchecks/index.html.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+## Kubernernetes Probes
+
+When kubernetes executes a Pod, it will perform 3 types of [Kubernetes
+Probes](https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/#startup-probe).
+
+### Startup Probe
+
+As soon as the pod launches kubernetes will start querying the startup probe.
```
jaskaransarkaria, line 68: Unless there is an `initialDelaySeconds` defined later down the line in the helm chart.

jaskaransarkaria (commented):
I know there is probably a good reason for this and I have probably missed some naming context, but what's the reason behind the startup probe pointing towards `/ready` and not it's own `/startup`?

jaskaransarkaria (approved):

jfharden:
The mermaid diagram renders wayyyyy too small, I don't think this is mergable while the diagram looks like this :(

source/kubernetes/manage-app/control-healthchecks/index.html.md
jfharden: Yeah, I've purposefully chosen to omit this since we don't support it yet in our charts anyway

PR #5226, opened 2025-09-03
Add asset-related documentation

pezholio:
This caught me out when deploying a new app - as the upload-assets job requires this to be present when booting the app (see https://github.com/alphagov/govuk-helm-charts/blob/c9e5255fc485af4b7b1877a1f1ef6a30f44d18bd/charts/generic-govuk-app/templates/assets-upload-job.yaml#L3)

### DIFF
--- a/source/manual/setting-up-new-rails-app.html.md
+++ b/source/manual/setting-up-new-rails-app.html.md
@@ -140,6 +140,31 @@ You should:
     end
     ```
 
+1. If your app will include assets (such as images, CSS, Javascript etc), replace the content  of `config/initializers/assets.rb` with the following (where `YOUR_APP_NAME` is the repo name of your application):
+
+   ```rb
+   # Be sure to restart your server when you modify this file.
+
+   # Path within public/ where assets are compiled to
+   Rails.application.config.assets.prefix = "/assets/YOUR_APP_NAME"
+
+   # Version of your assets, change this if you want to expire all your assets.
+   Rails.application.config.assets.version = "1.0"
+   ```
+
+  This can also be done in `config/application.rb` if you prefer (this is done in some existing apps) eg:
+
+   ```
+   module YourApp
+     class Application < Rails::Application
+       ...
+       # Path within public/ where assets are compiled to
+       config.assets.prefix = "/assets/your-app"
+       ...
+     end
+   end
+   ```
+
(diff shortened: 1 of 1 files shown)


source/manual/setting-up-new-rails-app.html.md
```diff
@@ -140,6 +140,18 @@ You should:
     end
     ```
 
+1. If your app will include assets (such as images, CSS, Javascript etc), replace the content  of `config/initializers/assets.rb` with the following (where `YOUR_APP_NAME` is the repo name of your application):
+
+   ```rb
+   # Be sure to restart your server when you modify this file.
+
+   # Path within public/ where assets are compiled to
+   Rails.application.config.assets.prefix = "/assets/YOUR_APP_NAME"
```
ChrisBAshton, line 149: Is this definitely needed? Picking two existing apps at random, I don't see this line:

- https://github.com/alphagov/content-publisher/blob/e0ddfbc4bc8cb171f2603ddf4ce8a4a41e006068/config/initializers/assets.rb
- https://github.com/alphagov/whitehall/blob/0ac4a7864831bf972d9afcbd0c21b234a4acc74b/config/initializers/assets.rb

pezholio: Ah yeah, in those apps it's in application.rb:

- https://github.com/alphagov/content-publisher/blob/e0ddfbc4bc8cb171f2603ddf4ce8a4a41e006068/config/application.rb#L68
- https://github.com/alphagov/whitehall/blob/0ac4a7864831bf972d9afcbd0c21b234a4acc74b/config/application.rb#L117

I've seen it in assets.rb in other apps, eg:

- https://github.com/alphagov/feedback/blob/b561c5ad4fe396088e9d7b3f75d3523175f16814/config/initializers/assets.rb#L7

Which seemed a more logical place

ChrisBAshton: Cool, thanks for checking 👍 probably worth tweaking the doc to just say that that config property needs setting, and that it can be done in `config/initializers/assets.rb` but that some apps do this in `config/application.rb`?

pezholio: Done! 👍

ChrisBAshton (approved):
Thanks!

PR #5228, opened 2025-09-03
Switch Bouncer cert provider to GlobalSign

sihugh:
We have a number of GlobalSign certificates already paid for in our account, so this works out cheaper for us at present.


nicholsj (approved):
LGTM thanks

PR #5244, opened 2025-09-17
Fix list formatting in rename country guidance

richardTowers:
| Before | After |
|----|----|
| image | image |


mtaylorgds (approved):
LGTM

PR #5247, opened 2025-09-18
Add how-to disable an app instructions

jfharden:
Add a how-to which describes how to disable, or re-enable an app.

This guide stems from RDS maintenance work where we had to figure this out on the fly.


AgaDufrat (approved):
Nicely written 👌 
I wonder if `kubernetes/manage-app/` would be a better home for this manual.

jfharden:
> Nicely written 👌 I wonder if `kubernetes/manage-app/` would be a better home for this manual.

Oh definitely, I'll move it 👍

PR #5251, opened 2025-09-23
Add howto encrypt rds at rest

jfharden:
* Add a page explaining how to convert an RDS instance to be encrypted at rest
* Remove the info that says we only have automated backups in production

### DIFF
--- a/source/manual/howto-convert-an-unencrypted-rds-instance-to-be-encrypted.html.md
+++ b/source/manual/howto-convert-an-unencrypted-rds-instance-to-be-encrypted.html.md
@@ -8,15 +8,14 @@ parent: "/manual.html"
 
 This playbook describes how to convert an AWS RDS instance which does not have encryption at rest enabled, into an encrypted at rest instance.
 
-These instructions will incur around 1 to 2 hours of downtime. It is possible to do this with only a minute or so downtime, but this requires
-manually setting up logical replication to a secondary RDS instance which is beyond the scope of this guide.
+These instructions will incur around 1 to 2 hours of downtime.
 
 ## Overview
 
 The general process of converting to an encrypted database is
 
 1. [Take your app(s) offline and ensure there are no possible writes to the RDS instance](#take-your-app-offline)
-2. [Make an encrypted snapshot of your RDS instance](#make-an-encrypted-snapshot-of-your-rds-instance). This cannot be done in 1 step so you will need to:
+2. [Make an encrypted snapshot of your RDS instance](#make-an-encrypted-snapshot-of-your-rds-instance). This cannot be done in a single step so you will need to:
     1. Take an unencrypted snapshot of the RDS instance; and then
     2. Copy the unencrypted snapshot to a new snapshot while enabling encryption
 4. [Recreate the RDS instance from the encrypted snapshot](#recreate-the-rds-instance-from-the-encrypted-snapshot)


source/manual/howto-convert-an-unencrypted-rds-instance-to-be-encrypted.html.md
```diff
@@ -0,0 +1,103 @@
...
+title: Convert an unencrypted AWS RDS instance into an encrypted one
+section: Backups
+layout: manual_layout
+parent: "/manual.html"
+---
+
+This playbook describes how to convert an AWS RDS instance which does not have encryption at rest enabled, into an encrypted at rest instance.
+
+These instructions will incur around 1 to 2 hours of downtime. It is possible to do this with only a minute or so downtime, but this requires
+manually setting up logical replication to a secondary RDS instance which is beyond the scope of this guide.
```
jasonBirchall: I think you should mention only the incurred downtime and omit the logical replication part. Alternatively, we could say, "If you want to explore this, then you can contact PE..." etc.

We can discuss the next evolution of RDS encryption, which probably contains logical replication.

What do you think?

source/manual/howto-convert-an-unencrypted-rds-instance-to-be-encrypted.html.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Overview
+
+The general process of converting to an encrypted database is
+
+1. [Take your app(s) offline and ensure there are no possible writes to the RDS instance](#take-your-app-offline)
+2. [Make an encrypted snapshot of your RDS instance](#make-an-encrypted-snapshot-of-your-rds-instance). This cannot be done in 1 step so you will need to:
```
jasonBirchall: I'm not sure I understand. This is how I read this: 
> Make an encrypted snapshot of your RDS instance. This cannot be done in 1 step (Take your app(s) offline and ensure there are no possible writes to the RDS instance), so you will need to[...]

Why would this be done in step 1? What am I missing here?

source/manual/howto-convert-an-unencrypted-rds-instance-to-be-encrypted.html.md
```diff
@@ -0,0 +1,103 @@
...
+    1. Take an unencrypted snapshot of the RDS instance; and then
+    2. Copy the unencrypted snapshot to a new snapshot while enabling encryption
```
jasonBirchall, line 20: Use a different list structure for a secondary list. i.e. (i), (ii)

source/manual/howto-convert-an-unencrypted-rds-instance-to-be-encrypted.html.md
```diff
@@ -0,0 +1,103 @@
...
+5. [Bring your app(s) back online](#bring-your-app-back-online)
+
+## Take your app offline
+
+Follow the [instructions for disabling an app in EKS](https://docs.publishing.service.gov.uk/kubernetes/how-to-disable-or-re-enable-an-app/#disabling-an-app).
+
+If multiple apps can write to the RDS instance you will need to disable all of them.
+
+Once your app is offline you need to check that there are no database connections to your RDS instance. AWS CloudWatch has a DatabaseConnections metric
+which will show you how many connections are open, this should show 0 (or possibly, if you have read replicas 1 per read replica).
```
jasonBirchall, line 31: Is it worth showing an image here, or does that quickly go out of date?

source/manual/howto-convert-an-unencrypted-rds-instance-to-be-encrypted.html.md
```diff
@@ -0,0 +1,103 @@
...
+       account_api = {
+         engine         = "postgres"
+         ...SNIP...
+         encryption_at_rest           = false
++        create_encrypted_snapshot    = true
++        deletion_protection          = false
+       }
+```
+
+The terraform will handle the whole process of making an unencrypted snapshot, and then copying it to an encrypted snapshot encrypted with the `govuk/rds` KMS key.
```
jasonBirchall, line 54: ❤️

jasonBirchall (approved):

source/manual/howto-convert-an-unencrypted-rds-instance-to-be-encrypted.html.md
jfharden: No, 1 step, not step 1. 

1 step as in a single step. You cannot say "Make an encrypted snapshot of an unencrypted database"

I'll reword this to "This cannot be done in a single step"

source/manual/howto-convert-an-unencrypted-rds-instance-to-be-encrypted.html.md
jfharden: I'll just leave it out and saying it involves downtime.

Generally given the incredibly low risk and low effort of this method, I would much prefer people do it this way if possible anyway. For those DBs which really cannot take a 2 hour downtime we should probably think about architecture anyway :)

source/manual/howto-convert-an-unencrypted-rds-instance-to-be-encrypted.html.md
jfharden: Markdown ordered lists require numbers for each list item (see https://daringfireball.net/projects/markdown/syntax#list), you can kinda force rendering of different characters for the sublists, but doing so loses the semantic context of a sub list that's ordered. The actual usage of a number in the output is a styling decision made by the CSS of the page, so using numbers is a choice of the tech docs gem (or maybe a choice by omision), either way to preserve the semantic meaning I need to stick to numbers in the raw markdown.


Just as a test though in case our renderer supports non-numbers I tried them and they don't work either with i, ii:
Screenshot 2025-09-24 at 15 30 50
and also with a. b. etc:
Screenshot 2025-09-24 at 15 13 02

For completeness, when you use numbers:


Screenshot 2025-09-24 at 15 52 59

source/manual/howto-convert-an-unencrypted-rds-instance-to-be-encrypted.html.md
jfharden: AWS change this enough, and there's a bunch of ways to get there that I think it's probably better without an image (if I were going to put an image it would be better to provide a complete description of how to get there, but I think it's probably not worth it in this case

PR #5255, opened 2025-09-25
BAU - Add Prometheus Operator information

nimalank7:
Description:
- Add a brief line on the Prometheus Operator

### DIFF
--- a/source/manual/prometheus-grafana-and-alertmanager.html.md
+++ b/source/manual/prometheus-grafana-and-alertmanager.html.md
@@ -9,6 +9,8 @@ parent: "/manual.html"
 
 ## How we use Prometheus on GOV.UK
 
+The Prometheus Operator installs and configures Prometheus, AlertManager and Grafana through Custom Resource Definitions.
+
 [Prometheus][link-1] is a systems monitoring and alerting toolkit. Its features include:
 
 - A time-series database, specifically optimised for graphing data over time.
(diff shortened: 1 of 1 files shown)


source/manual/prometheus-grafana-and-alertmanager.html.md
```diff
@@ -38,17 +40,17 @@ Grafana provides a frontend for visualizing the Prometheus data. A common workfl
 
 ## Alertmanager
 
-Alertmanager reads data directly from Prometheus, and sends notifications when metrics cross predefined thresholds. These thresholds are defined as [rules][link-11] in the monitoring-config folder of [govuk-helm-charts][link-12].
+Alertmanager reads data directly from Prometheus, and sends notifications when metrics cross predefined thresholds. These thresholds are defined as [rules][link-11] in [govuk-helm-charts][link-12]. See [here][link-13] on how to write these rules.
```
jfharden: we shouldn't use 'here' as a link, it should say what it's linking to (So maybe:


```suggestion
Alertmanager reads data directly from Prometheus, and sends notifications when metrics cross predefined thresholds. These thresholds are defined as [rules][link-11] in [govuk-helm-charts][link-12]. See [the prometheus operator manual][link-13] on how to write these rules.
```

AP-Hunt: Agree. Link text must say what the link is to on its own.

PR #5261, opened 2025-09-26
Update CDN failover instructions

jfharden:
Prior to the CDN failover drill, update the instructions with things already known prior to the drill.

### DIFF
--- a/source/manual/fall-back-to-aws-cloudfront.html.md
+++ b/source/manual/fall-back-to-aws-cloudfront.html.md
@@ -58,10 +58,10 @@ You can also get the `CNAME`s to use for the secondary CDN from the AWS CLI:
 
 ```bash
   # www-cdn.production.govuk.service.gov.uk
-  gds aws govuk-production-developer aws cloudfront list-distributions \
+  gds aws govuk-production-developer aws cloudfront list-distributions --output text \
     --query "DistributionList.Items[?Aliases.Items[0]=='www.gov.uk'].DomainName | [0]"
   # assets.publishing.service.gov.uk
-  gds aws govuk-production-developer aws cloudfront list-distributions \
+  gds aws govuk-production-developer aws cloudfront list-distributions --output text \
     --query "DistributionList.Items[?Aliases.Items[0]=='assets.publishing.service.gov.uk'].DomainName | [0]"
 ```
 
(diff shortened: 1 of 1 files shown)


source/manual/fall-back-to-aws-cloudfront.html.md
```diff
@@ -64,33 +67,50 @@ You can also get the `CNAME`s to use for the secondary CDN from the AWS CLI:
 
 The records should look like `d0000000000000.cloudfront.net.` (with 0s replaced with letters and numbers). Now you can manually update the `CNAME` records for both domains in both GCP and AWS, via the tabs you opened in your web browser earlier:
 
+Prior to changing the CNAMEs, do a quick test to ensure CloudFront is serving requests:
+
+```bash
+curl --fail -vs --header "Host: www.gov.uk" \
```
samsimpson1: Does this work with SNI? Afaik setting the host header in curl doesn't change the hostname curl indicates its connecting to during the TLS handshake

samsimpson1: it is possible to get curl to do that, with something like `curl --connect-to www.gov.uk:443:d0000000000000.cloudfront.net:443 https://www.gov.uk`

jfharden: It doesn't, but it still works with the cloudfront distro since it has it's own certificate. You'll get the cert for *.cloudfront.net:

But this is less than ideal actually since it doesn't guarantee we have a good cert attached to the CF distribution. I'll update it to:

```
curl --fail -vs \
  --connect-to assets.staging.publishing.service.gov.uk:443:<cloudfront distro>:443 \
  --header "Host: assets.staging.publishing.service.gov.uk" \
  https://assets.staging.publishing.service.gov.uk/media/662a74aa45f183ec818a72c2/dvsa-earned-recognition-vehicle-operators-accredited-list.csv/preview
```

jfharden: Of course I don't need the host header then, so I'll leave that off

jfharden: @samsimpson1 Updated, I also added `--output text` to the aws cli commands because it was annoying having the CF domain URLs include surrounding double quotes :)

PR #5263, opened 2025-09-30
Point to updated manual to archive a repo

AgaDufrat:
The instructions were updated. There's no need to have the instructions duplicated in the Dev docs so just adding a link.

https://github.com/alphagov/govuk-infrastructure/issues/2776


nicholsj (approved):
LGTM thanks

PR #5280, opened 2025-10-08
[SCH-1491] Add search-api-v2-beta-features

leenagupte:


data/repos.yml
```diff
@@ -686,6 +686,11 @@
   team: "#govuk-search"
   production_hosted_on: eks
 
+- repo_name: search-api-v2-beta-features
+  type: APIs
```
hannako: Is it maybe better classed as `Supporting apps`, like search admin?

leenagupte: I've updated and fixed up.

PR #5513, opened 2026-04-23
Update documentation about GOV.UK Notify service account

KludgeKML:
Removes reference to Feedback and makes clear that nothing is using this service at the moment and it doesn't have keys.

TODO:
- [x] Confirm that's true of all environments

### DIFF
--- a/source/manual/govuk-notify.html.md
+++ b/source/manual/govuk-notify.html.md
@@ -43,11 +43,9 @@ for each environment):
 
 - **GOV.UK**
 
-  This service is for any public-facing application that does not use the Email Alert Api
-  to send emails.
-
-  Currently it is used by [Feedback](https://github.com/alphagov/feedback) to email a survey link
-  to users who click a survey banner or use the `Is this page useful?` feature.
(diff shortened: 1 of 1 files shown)


source/manual/govuk-notify.html.md
```diff
@@ -43,11 +43,9 @@ for each environment):
 
 - **GOV.UK**
 
-  This service is for any public-facing application that does not use the Email Alert Api
-  to send emails.
-
-  Currently it is used by [Feedback](https://github.com/alphagov/feedback) to email a survey link
-  to users who click a survey banner or use the `Is this page useful?` feature.
+  This service is reserved for public-facing applications that do not use the Email Alert Api
+  to send emails. It it currently not used, and has no API Keys (which avoids the overhead
```
sairamya93: Typo: It`is` instead of `it`

sairamya93 (approved):
Looks good 👍 
Just a typo in the doc.

source/manual/govuk-notify.html.md
KludgeKML: Thanks! fixed.

PR #5518, opened 2026-04-28
Include app workers in rolling restart instructions

nacnudus:
When restarting an app in kubernetes, only the app pods will be restarted. A second command is necessary to restart worker pods.

### DIFF
--- a/source/kubernetes/manage-app/set-env-var/index.html.md
+++ b/source/kubernetes/manage-app/set-env-var/index.html.md
@@ -62,10 +62,13 @@ To update an existing secret:
     k delete secret foo-app-api-key
     ```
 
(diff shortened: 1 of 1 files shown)


AgaDufrat (approved):

source/kubernetes/manage-app/set-env-var/index.html.md
```diff
@@ -62,11 +62,14 @@ To update an existing secret:
     k delete secret foo-app-api-key
     ```
 
-3. Do a rolling restart of the affected app:
+3. Do a rolling restart of the affected app and its workers:
```
AgaDufrat: Not all apps have workers but it's definately worth mentioning here.

PR #5528, opened 2026-05-06
[PP-7735] Add instructions for dealing with CAA errors in Transition

brucebolt:
This is a rare scenario, so not previously documented. But we have come across this on the latest transition request, so should note it down for the future.


yndajas (dismissed):

source/manual/request-fastly-tls-certificate.html.md
```diff
@@ -66,6 +66,8 @@ When transitioning HTTPS domains, these are the steps to request a TLS certifica
 
 1. At this point, a unique domain [ownership validation](https://docs.fastly.com/en/guides/serving-https-traffic-using-fastly-managed-certificates#verifying-domain-ownership) record (`_acme-challenge`) is generated by Fastly.
 
+   > If you see the error `Found conflicting CAA record(s). Expected globalsign.com` alongside the domain, this means the domain owner has restricted which providers can issue certificates for the domain. Ask the domain owner to add `globalsign.com` as a CAA record for the domain.
```
yndajas, line 68: **question:** did you run this locally to confirm this gets formatted correctly? I know the dev docs are very particular. It looks like on the Rails upgrade guidance when using 'block quotes' under ordered list items we use three space indent (like this) but don't leave a line break: https://github.com/alphagov/govuk-developer-docs/blob/20618476dcadbb6d9d74ae9322485c51ef8798aa/source/manual/how-to-upgrade-rails.html.md?plain=1#L85

brucebolt: Good spot.  I've removed the line break, as the list numbering restarted after the quote otherwise.  Seems we don't parse this markdown in the same way as GitHub.

yndajas (approved):

PR #5533, opened 2026-05-13
Sch 2050 update documentation for new vertex name

emmalowe:
Google have rebranded Vertex AI Search to Agent Search, so we need to update the marketing name in the docs.

We are also changing some of the internal documentation to use "Discovery Engine", which is the more stable API name, to avoid future re-namings and because it should be a more familiar name for developers.

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-2050


stringiest (approved):
This looks great!

PR #5540, opened 2026-05-15
[SCH-2046] Add instructions on how to query the autocomplete endpoint

stringiest:
In April we had an incident where autocomplete suggestions were not being shown to users because they were not being returned by Google Vertex.

In responding to this incident we identified some shortcomings in our documentation on autocomplete

Rename file to reflect its broader scope. Add instructions for querying the `completeQuery` endpoint & examples of what responses might look like.

See [Jira ticket SCH-2046](https://gov-uk.atlassian.net/browse/SCH-2046)


source/manual/interacting-with-discovery-engine.html.md
```diff
@@ -1,6 +1,6 @@
 ---
 owner_slack: "#govuk-search"
-title: "How to get or delete a document from Discovery Engine"
+title: "How to interact with the Discovery Engine REST API"
```
emmalowe: I'm wondering if we should split this manual into two or more separate ones. If I were looking at the list of [search "how to"s](https://docs.publishing.service.gov.uk/manual.html#undefined) I would want to see a list of tasks I might want to complete. This title doesn't tell me much about what I can do by interacting with the Discovery Engine API. Maybe we want to have a manual of "How to delete a document from Discovery Engine" and "How to debug issues in Discovery Engine" (with everything else) as separate manuals - what do you think?

source/manual/search-synchronisation-errors.html.md
```diff
@@ -59,6 +59,6 @@ If all documents that have been successfully published or unpublished on gov.uk,
 [resynchronise]: /manual/how-to-resync-content-in-discovery-engine
 [document_sync_worker]: https://github.com/alphagov/search-api-v2/blob/1c3e8115b15703a44691311a2971ce2dbee10c59/lib/tasks/document_sync_worker.rake
 [UNPUBLISH_DOCUMENT_TYPES list]: https://github.com/alphagov/search-api-v2/blob/1c3e8115b15703a44691311a2971ce2dbee10c59/app/models/concerns/publishing_api/action.rb#L6
-[delete the document from the datastore]: /manual/get-or-delete-a-document-from-discovery-engine#delete-a-document-from-the-datastore
+[delete the document from the datastore]: /manual/interacting-with-discovery-engine#delete-a-document-from-the-datastore
```
emmalowe: I think this link needs to be updated

source/manual/interacting-with-discovery-engine.html.md
stringiest: Sounds sensible! Now done :)

source/manual/debug-issues-with-discovery-engine.html.md
```diff
@@ -84,3 +77,50 @@ curl -XPOST \
...
+            "suggestion": "proof of immigration status"
+        },
+        {
+            "suggestion": "proof of right to work"
+        }
+    ]
+}
+```
+
+#### Problems with autocomplete
```
emmalowe, line 124: Maybe a sentence or two to explain this a bit more would help?

e.g. "If autocomplete is not working, it is likely to return an empty array, which looks like this:"

emmalowe: We might also want to point out that just because autocomplete has no suggestions for a query, that doesn't mean that autocomplete is not working. We could list some common queries that we do expect responses for e.g. "universal" (should lead to universal credit related things) , "tax", "income" etc.

stringiest: Both sound great - now amended

emmalowe (approved):

PR #5543, opened 2026-05-18
Improve search alert documentation

hannako:
Improve documentation around the cause and actions to take when search api v2 success rates drop

Jira: https://gov-uk.atlassian.net/jira/software/c/projects/SCH/boards/994

Rendered file 👉 https://github.com/alphagov/govuk-developer-docs/blob/e9df9747d1f5316d6c8d88caf0fab48421a364e3/source/manual/search-alerts-and-monitoring.html.md

### DIFF
--- a/source/manual/search-alerts-and-monitoring.html.md
+++ b/source/manual/search-alerts-and-monitoring.html.md
@@ -41,31 +41,68 @@ We monitor and alert on both Search API v2 success rates, and Google Discovery E
 
 We have an informal SLO to maintain a search and autocomplete success rate of about 99.99% over any 24 hour period. There are currently four Alertmanager rules configured in govuk-helm-charts to send notifications on Slack, if rates drop below this:
 
-- [SearchDegradedAcute][link-6] 5 minute rolling success rate for search requests has dropped below 99% for more than 10 minutes.
+##### Search alerts
 
-- [SearchDegradedMid][link-7] 1 hour rolling success rate for search requests has dropped below 99.9% for more than 2 hours.
+- [SearchDegradedAcute][link-6] (__Critical__) 5 minute rolling success rate for search requests has dropped below 99% for more than 10 minutes.
 
-- [SearchDegradedLong][link-8] 24 hour rolling success rate for search requests has dropped below 99.99% for more than 24 hours.
+- [SearchDegradedMid][link-7] (__Critical__) 1 hour rolling success rate for search requests has dropped below 99.9% for more than 2 hours.
 
-- [AutocompleteDegradedAcute][link-9] 5 minute rolling success rate for autocomplete requests has dropped below 90% for more than 10 minutes.
+- [SearchDegradedLong][link-8] (__Warning__) 24 hour rolling success rate for search
+
+##### Autocomplete alerts
+
+- [AutocompleteDegradedAcute][link-9] (__Critical__) 5 minute rolling success rate for autocomplete requests has dropped below 90% for more than 10 minutes.
+
+- Note on future work: The Search team is developing improved autocomplete monitoring. Currently our success rate will remain 100% if the Discovery Engine API responds with a 200 success code but provides no autocomplete suggestions for queries where suggestions are available.
 
 #### Google Cloud Discovery Engine request durations
 
 There is currently one Alertmanager rule configured in govuk-helm-charts, [HighVertexP90Latency](link-14), which sends notifications in Slack if requests to Google Cloud Discovery Engine search endpoint exceed acceptable duration thresholds.
 
-#### Causes and steps to take in the event of a degradation of service alert firing
+#### Causes of degradation of service alerts firing
 
-We are aware of the following occasional errors which should not be considered critical and do not need intervention unless they occur consistently for a large number of users and don’t go away by themselves within a few minutes.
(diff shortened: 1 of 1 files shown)


source/manual/search-alerts-and-monitoring.html.md
```diff
@@ -41,32 +41,66 @@ We monitor and alert on both Search API v2 success rates, and Google Discovery E
...
 
+#### Quantifying the user impact of low search success rates
+
+A degradation in Search API v2 success rates indicates that users will be experiencing a problem with site search on GOV.UK. However it is not a quantitative represenation of the number of users being impacted. A better measure would be to look at the error rates for Finder Frontend or Router for requests to search/all with a keyword.
+
+Suggested Kibana query:
+
+```
+logjson.request_uri: *search/all?keywords=*
+kubernetes.labels.app_kubernetes_io/name: finder-frontend
```
stringiest, line 87: Do we need to add `kubernetes.container.name: app` here to remove duplicates?

source/manual/search-alerts-and-monitoring.html.md
```diff
@@ -41,32 +41,66 @@ We monitor and alert on both Search API v2 success rates, and Google Discovery E
...
+NOT logjson.status: 200 to 310
```
stringiest: Did we say that the errors that users experience are only 503s?

source/manual/search-alerts-and-monitoring.html.md
```diff
@@ -41,32 +41,66 @@ We monitor and alert on both Search API v2 success rates, and Google Discovery E
...
+##### Search alerts
 
-- [SearchDegradedMid][link-7] 1 hour rolling success rate for search requests has dropped below 99.9% for more than 2 hours.
+- [SearchDegradedAcute][link-6] (__Critical__) 5 minute rolling success rate for search requests has dropped below 99% for more than 10 minutes.
 
-- [SearchDegradedLong][link-8] 24 hour rolling success rate for search requests has dropped below 99.99% for more than 24 hours.
+- [SearchDegradedMid][link-7] (__Critical__) 1 hour rolling success rate for search requests has dropped below 99.9% for more than 2 hours.
 
-- [AutocompleteDegradedAcute][link-9] 5 minute rolling success rate for autocomplete requests has dropped below 90% for more than 10 minutes.
+- [SearchDegradedLong][link-8] (__Critical__) 24 hour rolling success rate for search
```
stringiest: Is SearchDegradedLong now a `warning` rather than `critical`?

source/manual/search-alerts-and-monitoring.html.md
```diff
@@ -41,32 +41,66 @@ We monitor and alert on both Search API v2 success rates, and Google Discovery E
...
+
+##### Autocomplete alerts
+
+- [AutocompleteDegradedAcute][link-9] (__Critical__) 5 minute rolling success rate for autocomplete requests has dropped below 90% for more than 10 minutes.
+
+- Note on future work: The Search team is developing improved autocomplete monitoring. Currently our success rate will remain 100% if the Discovery Engine API responds with a 200 success code and no autocomplete suggestions.
```
stringiest: Could we add something like 'This gives the false impression that autocomplete is working correctly' to the end of this paragraph?

source/manual/search-alerts-and-monitoring.html.md
```diff
@@ -41,32 +41,66 @@ We monitor and alert on both Search API v2 success rates, and Google Discovery E
...
+A common cause of drops in search success rate, is high latency from the DiscoveryEngine API. This will result in the [Google::Cloud::DiscoveryEngine Ruby client][link-15] timing out and raising `Google::Cloud::DeadlineExceededError` errors, which in turn lead Search API v2 to respond with 499 errors for search result requests.
+
+We are also aware of the following occasional errors which should not be considered critical and do not need intervention unless they occur consistently for a large number of users and trigger critical alerts.
 
-- `Google::Cloud::DeadlineExceededError` A timeout occurred on the Google API
 - `Google::Cloud::InternalError` An internal error occurred on the Google API
 - `AMQ::Protocol::EmptyResponseError` RabbitMQ sent an unexpected response, possibly due to restarting (the listener will restart by itself in most cases)
 
-If these errors persist and trigger the degradation of service alerts, this indicates an issue with GCP or Discovery Engine.
+If these errors persist and trigger a critical degradation of service alerts, this indicates an issue with GCP or Discovery Engine that we should raise with Google.
```
stringiest: I find this section quite confusing. I think that, as a dev debugging a problem with search, I'd only have come to this section because a degradation of service alert was firing. Are we saying that for any errors that have triggered a critical degradation of service alert we should raise a ticket with Google? 

Currently the wording implies that there would be some circumstances in which we would not raise with Google, but I'm not convinced that those reasons are related to the types of error mentioned above (perhaps they're just if the alert was `warning` level, don't raise a ticket with Google). I wonder if the section about the types of error we're aware of should be removed? Or we could move it into an info box that isn't related to whether or not we raise a ticket with Google?

source/manual/search-alerts-and-monitoring.html.md
hannako: Yes

hannako: Oh I see I've got critical! Sorry cut and paste error when adding the bold markdown!

stringiest (approved):

PR #5552, opened 2026-05-20
[WHIT-3439] Replace removed rake tasks in Whitehall runbook

TonyGDS:
## Summary
The runbook references two rake tasks (`publishing:overdue:list` and `publishing:overdue:publish`) that were removed from Whitehall in alphagov/whitehall@41ae17e71d (Nov 2024). Per that commit message, neither task is needed: the PagerDuty alert is itself the verification of overdue editions, and `publishing:scheduled:requeue_all_jobs` already covers the bulk publish case.

Point both alert sections at `publishing:scheduled:requeue_all_jobs`, and trim the "After a database restore" subsection since it's no longer routing the reader to a different rake task.


kashifatcha (approved):
LGTM

PR #5561, opened 2026-05-26
Update guidance for adding Transition mappings

brucebolt:
This guidance has not been updated for many years and does not reflect the current process.

Between 2022 and 2026, we received no requests for transitions with mappings, so the Content APIs (formerly Publishing Platform) team left this guidance unchanged when they inherited the Transition and Bouncer apps.

However, in recent transition requests, we have put more of an emphasis on the department to determine and add their mappings, rather than relying on GDS to do this, so updating the instructions to reflect this.

### DIFF
--- a/source/manual/configure-transition-mappings.html.md
+++ b/source/manual/configure-transition-mappings.html.md
@@ -18,77 +18,37 @@ They will need one new CNAME DNS entry for each domain/subdomain they wish to pr
 - www.domain.gov.uk → aka.domain.gov.uk
 - sub.domain.gov.uk → aka-sub.domain.gov.uk
 
-There are lots of examples of these in [hosts currently configured in Transition][transition-hosts].
+There are lots of examples of these in [hosts currently configured in Transition](https://transition.publishing.service.gov.uk/hosts).
 
-These AKA domains should be CNAMEd to: `bouncer-cdn.production.govuk.service.gov.uk`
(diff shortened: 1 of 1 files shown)


source/manual/configure-transition-mappings.html.md
```diff
@@ -18,77 +18,37 @@ They will need one new CNAME DNS entry for each domain/subdomain they wish to pr
...
+## 2) Set up user accounts for the department
 
-## 2) Get a list of old URLs
+Ask the requester for names and email addresses of users who need access to the Transition app.
 
-In order for us to redirect anything but the homepage, we need a list of URLs for the old site so that they can be mapped. In the past, this is something GDS did for the organisation as it requires some technical skill and experience.
+Give the user(s) the "Site Manager" permission in Signon for the Transition app.
 
-Getting traffic logs from a transitioning organisation is the best option, as we can get a list of URLs we know are actively used. If we can't get this then there are other options.
+> If they do not have a Signon account in production already, create one for them at 'Normal' level with access to Transition only (i.e. no ability to publish or preview documents, which may be granted by default when creating an account).
```
AgaDufrat: I think access to Support app is also granted to all new Signon users, which maybe useful here.

AgaDufrat (approved):
Nicely written. I added a comment about Signon permissions.

source/manual/configure-transition-mappings.html.md
brucebolt: That's a very good point.  I've added "and Support" into this sentence.

PR #5562, opened 2026-05-28
Fix command for uploading an asset manually

brucebolt:
The existing command would give an output in the format `pod/asset-manager-xxxyyy` which causes a failure in the next step.

Updating to output the pod name as `asset-manager-xxxyyy` only.

### DIFF
--- a/source/manual/manage-assets.html.md.erb
+++ b/source/manual/manage-assets.html.md.erb
@@ -120,13 +120,13 @@ best to simply perform the upload directly in production.
 1. Upload the file to the pod's `/tmp` directory:
 
     ```sh
-    k cp my_file.jpg apps/$POD:/tmp
+    kubectl cp my_file.jpg apps/$POD:/tmp
     ```
 
 1. Run the create_asset command on the same pod:
 
     ```sh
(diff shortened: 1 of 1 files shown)


AgaDufrat (approved):

source/manual/manage-assets.html.md.erb
```diff
@@ -113,7 +113,7 @@ best to simply perform the upload directly in production.
 1. Get the name of an asset-manager app pod.
 
     ```sh
-    POD=$(kubectl get pods -l app=asset-manager -o name | head -1)
+    POD=$(basename $(kubectl get pods -l app=asset-manager -o name | head -1))
```
yndajas: **suggestion:** this could just be piped instead of subshelled. You need `xargs` so that `basename` receives the argument properly though

```sh
POD=$(kubectl get pods -l app=asset-manager -o name | head -1 | xargs basename)
```

yndajas (approved):

source/manual/manage-assets.html.md.erb
brucebolt: Nice suggestion.  I've updated both instances.

PR #5564, opened 2026-06-01
Add a local copy of cookie-functions.js

MartinJJones:
## What

- Add a copy of `cookie-functions.js` from the [govuk_publishing_components gem](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js)
- Update the COOKIE_CATEGORIES list to be specific to developer docs

## Why

This will make it is easier to update the cookie functions and the configuration included to be specific to the developer-docs where required

Jira card: https://gov-uk.atlassian.net/browse/NAV-18951

### DIFF
--- a/source/javascripts/application.js
+++ b/source/javascripts/application.js
@@ -2,9 +2,9 @@
 //= require govuk_frontend
 //= require dependencies
 //= require govuk_publishing_components/lib/trigger-event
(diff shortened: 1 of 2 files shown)


MartinJJones:
Closing in favour of https://github.com/alphagov/govuk-developer-docs/pull/5593

PR #5571, opened 2026-06-03
Sch 1993 update publishing docs after decommissioning government index

emmalowe:
Updates publishing related manuals after decommissioning the government index.

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1993


source/manual/fix-out-of-date-search-indices.html.md
```diff
@@ -21,23 +21,5 @@ Run the `represent_downstream:published_between['2018-12-17T01:02:30,%202018-12-
...
-
-**This will not be necessary after whitehall content has been moved to the
-`govuk` index.**
-
-These indexes are populated by whitehall calling an HTTP API in Search API.
-Missing documents can be recovered by resending the content to Search API directly.
-
-Run the `search:index:published_between['2018-12-17T01:02:30,%202018-12-18T10:20:30']` rake task in Whitehall, remembering to change the two timestamps.
-
-## `metasearch` index
```
hannako, line 34: Instead of deleting this, maybe we should amend?

This index contains best bets, which used to be published by Search Admin until that functionality was removed in [PR]. There is currently no support for reindexing this index. This is known tech debt.

source/manual/fix-issues-with-external-search.html.md
```diff
@@ -0,0 +1,9 @@
+---
+owner_slack: "#govuk-search"
+title: Fix issues with external search
```
stringiest: I wonder if the title should be 'Fix issues with external search engines'? I'm not sure that I would immediately parse external search as meaning Google etc. That might just be me, though!

source/manual/add-a-new-field-to-an-elasticsearch-document-type.html.md
```diff
@@ -0,0 +1,41 @@
...
+
+In order for the new field to work as expected, you will need to reindex elasticsearch. Follow the guidance on [how to reindex elasticsearch][reindex-guide].
+
+If you consider the change low risk and are only adding new fields for which content doesn't yet exist, you can run the `search:update_schema` [task][update-schema-task].
+This task will attempt to update the Elasticsearch index schema in place without requiring a re-index.
+If you have made any changes which affect existing fields, Elasticsearch will reject the change and a full re-index will be required.
+
+[schema-documentation]: https://docs.publishing.service.gov.uk/repos/search-api/schemas.html
+[analysed]: https://www.elastic.co/guide/en/elasticsearch/guide/current/mapping-analysis.html
+[filter]: https://www.elastic.co/guide/en/elasticsearch/reference/6.8/query-filter-context.html
```
stringiest: I think this should be 'filters' rather than 'filter'

source/manual/fix-out-of-date-search-indices.html.md
```diff
@@ -1,25 +1,33 @@
...
-section: Backups
+section: Search on GOV.UK
+related_repos: [search-api]
 ---
 
-If the data in the search index is out-of-sync with the Publishing API,
-(for example, after [restoring a backup][restore-backups]), then any `publish`
-and `unpublish` messages that have not been processed need to be resent.
+The Elasticsearch cluster utilised by [Search API](/repos/search-api.html) can
+get out of sync with publishing applications. This affects any part of the site
```
hannako: I think it's more accurate to say out of sync with the Publishing API rather than with publishing applications, because Whitehall no longer writes directly.

source/manual/reindex-elasticsearch.html.md.erb
```diff
@@ -18,10 +18,11 @@ used.
...
 >
 > Reindexing locks the index for writes, so content is not updated in the
-> search index. Reindexing takes around 2 hours to complete.
+> search index. Reindexing takes around 2 hours to complete. If you need
+> to reindex during working hours, you will need to update the index once
+> the reindexing is complete.
 >
-> See [Fix out-of-date search indices](fix-out-of-date-search-indices.html) if you
-> need to reindex during working hours.
+> See [Fix out-of-date search indices](fix-out-of-date-search-indices.html)
```
stringiest: Is this saying that if you're reindexing during working hours you will then need to update the index, the instructions for which are on the fix out-of-date search indices doc? 

If so, I wonder if it would flow better to just have 'Reindexing takes around 2 hours to complete.' here, and then have the instruction to update the index as step 3 of the reindexing suggestions (below the clean up section)?

source/manual/fix-out-of-date-search-indices.html.md
```diff
@@ -26,8 +26,11 @@ Be aware that these options will replay the entire Publisher API history for tha
 
 ## Environment syncs
 
-** TO DO **
+There is a nightly job to [synchronise elasticsearch][sync-job] in integration and staging with the production environment.
+If documents that are published or unpublished in production are not reflected in integration and staging after 1 day, that
+could be because of issues with the synchronisation job.
```
hannako: Do we think it's safe to resubmit the elasticsearch env sync in prod or staging if it fails? I'm pretty sure I did that to fix it last time. If so we should probably add information to https://docs.publishing.service.gov.uk/manual/govuk-env-sync.html#header and link to it from here.

For this PR leave it as it is, but let's come back to it when we've written that documentation.

stringiest (commented):
This looks great - thanks! A couple of suggestions for readability.

source/manual/reindex-elasticsearch.html.md.erb
emmalowe: Yes, that is what I'm trying to say here. I agree that flow works better - I've updated the docs.

source/manual/fix-out-of-date-search-indices.html.md
emmalowe: Good idea. I've added that back in, and also added something on the page-traffic index for completeness.

source/manual/fix-issues-with-external-search.html.md
emmalowe: Good idea - I've updated that now.

source/manual/add-a-new-field-to-an-elasticsearch-document-type.html.md
emmalowe: Good catch, thanks.

source/manual/fix-out-of-date-search-indices.html.md
emmalowe: Good point, thanks.

source/manual/fix-issues-with-external-search.html.md
```diff
@@ -0,0 +1,9 @@
+---
+owner_slack: "#govuk-search"
+title: Fix issues with external search engines
+parent: "/manual.html"
+layout: manual_layout
+section: Search on GOV.UK
+---
```
stringiest, line 7: Just thought of this, but I wonder if it's helpful to add `related_repos: [search-api, search-api-v2]` here for cross-referencing purposes?

source/manual/fix-out-of-date-search-indices.html.md
```diff
@@ -1,43 +1,49 @@
...
-
 ## `govuk` index
 
-Content in the `govuk` index is populated from the [Publishing API message queue][queue].
-Missing documents can be recovered by resending the content to the message queue, using
-the `represent_downstream:published_between` rake task.
+GOV.UK content in [Search API](/repos/search-api.html) is stored in the `govuk` index in Elasticsearch.
+This is populated from the [Publishing API message queue][queue]. This can get out of sync with publishing API,
+which affects any part of the site using it, including navigation pages and related links.
+This can happen after [restoring a backup][restore-backups] or [reindexing search][reindex-search]).
```
stringiest: Is there an extra `)` at the end of this line?

stringiest (approved):
🌟 
Have made a couple of small suggestions, but also approved as the suggestions are non-blocking.

source/manual/fix-out-of-date-search-indices.html.md
emmalowe: I think that would be safe. The [elasticsearch docs](https://www.elastic.co/docs/deploy-manage/tools/snapshot-and-restore) say that there's no downtime when taking a snapshot.

Were you imagining we'd add some info [here](https://docs.publishing.service.gov.uk/manual/govuk-env-sync.html#whether-to-launch-a-new-sync-operation)? That guidance seems pretty generic, so I wonder if we're better off putting the info in this manual?

source/manual/fix-out-of-date-search-indices.html.md
emmalowe: Thanks, good spot!

source/manual/fix-issues-with-external-search.html.md
emmalowe: I decided not to, because it's not really about either of them. It's under 'search manuals', so hopefully that's enough of a cross reference.

source/manual/fix-out-of-date-search-indices.html.md
hannako: I don't feel strongly - as long as there's a link between the two piece of documentation so whoever comes across it feels confident that they've found all relevant info.

emmalowe: Sounds good. I've added it here so we don't have to come back to it later.

hannako (approved):

PR #5587, opened 2026-06-17
Update 'Get Started' guide for DSIT devices

brucebolt:
Some of the steps are different now we are using DSIT issued devices.

Updating the steps that are relevant. This is by no means comprehensive and I expect we'll need to make further edits as more people set up their devices.


ChrisBAshton (approved):
LGTM. I've followed these steps today, didn't have to do anything particularly different outside of that.

PR #5591, opened 2026-06-19
Sch 2144 update alerting documentation for new autocomplete alert

emmalowe:
Updating documentation after adding new alert for autocomplete: https://github.com/alphagov/govuk-helm-charts/pull/4307

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-2144

### DIFF
--- a/source/manual/search-alerts-and-monitoring.html.md
+++ b/source/manual/search-alerts-and-monitoring.html.md
@@ -235,12 +235,12 @@ If an evaluation fails for a reason other than a sample query set not existing, 
 [link-3]: https://govuk.sentry.io/insights/projects/app-search-api-v2-beta-features/?project=4510158725513217
 [link-4]: ./kibana.html
 [link-5]: https://grafana.eks.production.govuk.digital/d/govuk-search/gov-uk-search?orgId=1&from=now-24h&to=now&timezone=browser
-[link-6]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L63
-[link-7]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L115
-[link-8]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L152
+[link-6]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L45
+[link-7]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L89
+[link-8]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L141
 [link-9]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L104
-[link-10]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L188
-[link-11]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L201
+[link-10]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L164
+[link-11]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L190
 [link-12]: https://console.cloud.google.com/welcome?inv=1&invt=Ab3mhA&project=search-api-v2-production
 [link-13]: https://console.cloud.google.com/support/cases?inv=1&invt=Ab3mhA&project=search-api-v2-production
 [link-14]: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/monitoring-config/rules/search_api_v2.yaml#L31
--- a/source/manual/search-alerts-and-monitoring.html.md
+++ b/source/manual/search-alerts-and-monitoring.html.md
@@ -40,7 +40,7 @@ We monitor and alert on both Search API v2 success rates and Google Discovery En
 
 #### Search API v2 success rates
 
-We have an informal SLO to maintain a search success rate of about 99.99% over any 24-hour period. There are currently three Alertmanager rules configured in govuk-helm-charts to send notifications on Slack, if rates drop below this:
+We aim to maintain a search success rate of about 99.99% over any 24-hour period. There are currently three Alertmanager rules configured in govuk-helm-charts to send notifications on Slack, if rates drop below this:
 
 - [SearchDegradedAcute][link-6] (__Critical__) 5 minute rolling success rate for search requests has dropped below 99% for more than 10 minutes.
 
--- a/source/manual/search-alerts-and-monitoring.html.md
+++ b/source/manual/search-alerts-and-monitoring.html.md
@@ -52,6 +52,8 @@ We aim to maintain a search success rate of about 99.99% over any 24-hour period
 
 There is currently one Alertmanager rule configured in govuk-helm-charts, [HighVertexP90Latency][link-14], which sends notifications in Slack if requests to Google Cloud Discovery Engine search endpoint exceed acceptable duration thresholds.
 
+Note that, since we have [introduced a 2-second timeout][link-17] on search requests to Discovery Engine, this alert is not expected to fire until the levels are adjusted.
+
 #### Causes of degradation of search service alerts firing
 
 A common cause of drops in search success rate, is high latency from the DiscoveryEngine API. This will result in the [Google::Cloud::DiscoveryEngine Ruby client][link-15] timing out and raising `Google::Cloud::DeadlineExceededError` errors, which in turn lead Search API v2 to respond with 500 errors for search result requests.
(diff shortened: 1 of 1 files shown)


source/manual/search-alerts-and-monitoring.html.md
```diff
@@ -34,33 +34,27 @@ All environments are configured for Slack notifications. Production alerts are r
 
 ## Alert types and how to handle them
```
stringiest, line 36: This page is pretty long now. I wonder if it would be helpful to have a contents list at the start of this section, with links to the h3 headings?

stringiest (commented):
This looks good! 

I'm probably overthinking this, but is this document about how monitoring & alerting works on search (a 'learn' doc), or about what to do if there's an alert on search (a 'how to' doc)? 

I wonder if the 'Monitoring & alerting tools' section should be its own 'learn' document, and then the rest of the doc is a 'how to' guide on what to do if there's an alert on search?

source/manual/search-alerts-and-monitoring.html.md
```diff
@@ -105,6 +99,65 @@ GROUP BY status
...
+it doesn't make sense to monitor the success rate for autocomplete requests to `search-api-v2`, as this should always be 100%. Instead, we
+monitor the number of autocomplete suggestions returned from each request to the Discovery Engine API, and raise an alert if we haven’t received any suggestions
+for over an hour (subject to receiving a minimum of 500 responses during that time):
+
+- [AutocompleteDegradedMid][link-9] (__Warning__): Maximum number of autocomplete suggestions returned by Google Cloud Discovery Engine over 1 hour
+  has dropped to zero for more than 10 minutes.
+
+#### Causes of degradation of autocomplete service alerts firing
+
+There are three issues that could cause all autocomplete responses from the Discovery Engine API to be empty (i.e. contain zero suggestions):
```
hannako: Are we sure there are exactly three reasons? Perhaps we should say something less brittle like "common" or "likely"

source/manual/search-alerts-and-monitoring.html.md
```diff
@@ -105,6 +99,65 @@ GROUP BY status
...
+
+- Requests to the Discovery Engine API returning `Google::Cloud::DeadlineExceededError`
+- Requests to the Discovery Engine API returning `Google::Cloud::InternalError`
+- Requests to the Discovery Engine API returning a successful but empty response.
+
+It's worth noting that, even in healthy environments, we expect around 50% of autocomplete requests to Discovery Engine API to return
```
hannako, line 121: ... because not all search queries are frequent enough for an autocomplete suggestion to be calculated. Or something like that!

source/manual/search-alerts-and-monitoring.html.md
emmalowe: There's a navigation panel on the left when you view the manual in dev docs already:
https://docs.publishing.service.gov.uk/manual/search-alerts-and-monitoring.html

emmalowe:
> I'm probably overthinking this, but is this document about how monitoring & alerting works on search (a 'learn' doc), or about what to do if there's an alert on search (a 'how to' doc)?
> 
> I wonder if the 'Monitoring & alerting tools' section should be its own 'learn' document, and then the rest of the doc is a 'how to' guide on what to do if there's an alert on search?

Valid question. I think our approach on this across GOV.UK isn't very consistent. If you search 'alerts' in dev doc manuals you can see some under 'learn', some under 'how to' and some under the 'alert' subheading. I think probably the most important thing is that we link to this page from our Alertmanager alerts, which we do.

Given that we'll probably add to our alerts when we tackle monitoring and alerting more broadly, and then will have to reshuffle this page then, I'm inclined to leave this reorganisation for now.

source/manual/search-alerts-and-monitoring.html.md
stringiest: Good point. I only viewed it in GH.

stringiest (approved):

source/manual/search-alerts-and-monitoring.html.md
```diff
@@ -105,6 +99,65 @@ GROUP BY status
...
+
+#### Steps to take in the event of an alert being triggered
+
+If the [AutocompleteDegradedMid][link-9] alert is firing, the first step is to figure out what is causing the Discovery Engine
+API to return zero suggestions. To do this, go to Kibana and search for logs over the last hour that contain the following text:
+
+```DiscoveryEngine::Autocomplete::Complete: Did not get autocomplete suggestion```
+
+The `debug_error_string` should be able to tell you if this was because of a `DeadlineExceededError` or a `InternalError`.
+In the absence of these error logs, it's likely that Discovery Engine is returning successful but empty responses.
```
hannako: Maybe this section needs a bit of set up ? If the event fires I'm guessing we will link to this section directly, so the reader might not have absorbed the information in the sections above.. 

How about:

If the [AutocompleteDegradedMid][link-9] alert is firing, the DiscoveryEngine API is either responding to autocomplete requests with an error code. Or it's responding with a success code, but no suggestions. 

It's important to understand which scenario has occurred so that we can provide a detailed description of the issue we are experiencing to Google.

Use Kibana to locate the cause:

1. Search for logs over the last hour that contain the following text:

```DiscoveryEngine::Autocomplete::Complete: Did not get autocomplete suggestion```

2. Check the `debug_error_string` field for `DeadlineExceededError` or `InternalError`.

The absence of these error messages suggests that Discovery Engine is returning successful but empty responses. 

You can also try a few common queries that should always return autocomplete suggestions, and track the logs for these in Kibana and Google Cloud Platform (after [switching on increased logging][gcp-logging]) to see what's happening. Such queries include: "tax", "credit" and "work".

File: source/manual/search-alerts-and-monitoring.html.md [hunk 5]
```diff
@@ -105,6 +99,65 @@ GROUP BY status
...
+
+You can also try a few common queries that we expect to return suggestions, and track the logs for these in Kibana and
+Google Cloud Platform (after [switching on increased logging][gcp-logging]) to see what's happening. Such queries include:
+"tax", "credit" and "work".
+
+Once you know what is causing the issue, [raise a support ticket with Google](#how-to-contact-google-if-there-is-a-critical-issue-with-gcp-or-discovery-engine),
+with as much information about the problem as possible including the time the issue started and what the cause is.
+
+While waiting for a response from Google, it could be worth [turning off the autocomplete feature][turn-off-autocomplete],
+to prevent unhelpful suggestions from appearing.
```
hannako: I think we should remove this section. We turned it off for a very specific reason last time and there is existing documentation here: https://docs.publishing.service.gov.uk/repos/search-api-v2/search_autocomplete.html#disable-search-autocomplete

... which now I read it makes me realise that if we disable autocomplete, we will trigger this alert! 😂 🤔

source/manual/search-alerts-and-monitoring.html.md
```diff
@@ -105,6 +99,65 @@ GROUP BY status
...
+
+#### Quantifying the user impact of degraded autocomplete service
+
+Since we handle autocomplete request errors internally, users will still be able to use site search when the autocomplete service is degraded.
+This minimises the user impact, as autocomplete is considered a non-critical service. However, it might still be useful to see
+how many users are not getting autocomplete suggestions when they should be.
+
+When the reason for an autocomplete requests containing zero suggestions is a `Google::Cloud::DeadlineExceededError` or a
```
hannako: I'm actually not sure if this is how we'd do it, as potentially you'd get these errors for requests that would return 0 suggestions in a healthy scenario - as you've described below.

hannako: Maybe just chop back so it begins with the para:


>To estimate this you can look at how many requests were made during the window, and divide this by two since we expect about 50% of responses to legitimately return zero suggestions. 


The next bit sounds like a good idea, but the instruction on what you'd do with the metric to turn that into "how many users are impacted" is not clear to me.  Are you able to explain a bit? 

>To get a more up-to-date and accurate percentage of responses that should return zero responses, you can look at the Prometheus histogram metric `search_api_v2_discovery_engine_autocomplete_suggestions_response_bucket`,
which shows how many autocomplete suggestions are returned for each request. Remember to only consider responses recorded during a healthy scenario.

hannako (commented):
Thanks so much for getting this down Emma! Such a lot of information really nicely explained 😅 

I've made some suggestions, sorry didn't get to it earlier this week. If you've had enough of writing documentation I'm happy to push the suggested changes (provided you agree with them!)

source/manual/search-alerts-and-monitoring.html.md
emmalowe: I think if we turned off autocomplete, we wouldn't be pushing the metric to Prometheus anyway (we'd stop [here](https://github.com/alphagov/search-api-v2/blob/main/app/controllers/autocompletes_controller.rb#L9)) so part of the alert that checks we're recording a minimum number of observations should stop us from getting alerted while it's turned off.

I'll take this out anyway if it's not needed.

source/manual/search-alerts-and-monitoring.html.md
emmalowe: That reads better to me - thanks!

emmalowe: At the moment, we link to this whole page from the alerts, but I agree it would be better to link to the specific relevant section, given how long this has now got.

source/manual/search-alerts-and-monitoring.html.md
emmalowe: Oooh that's a great point, I hadn't though of that!

I had kind of run out of steam by this point, so agree that this explanation could be better! I'll add to that.

source/manual/search-alerts-and-monitoring.html.md
hannako: Ah yep OK phew!

source/manual/search-alerts-and-monitoring.html.md
emmalowe: Rewritten. What do you think now @hannako ?

hannako (approved):

PR #5593, opened 2026-06-22
Remove Google Analytics

MartinJJones:
## What

Remove Google Analytics from the develop docs, by removing:

- Cookie banner component partials and references
- JavaScript used for cookies and google analytics
- Cookie settings page
- Cookie banner component styles

## Why

The developer docs repo currently uses Google Analytics, enabled by the govuk_publishing_components gem.

However, Google Analytics and the related cookie consent code in the govuk_publishing_components gem is written in a way that is specific to GOV.UK and the publishing applications.

After reviewing the different options available, there does not appear to be a need for Google Analytics in the developer docs and we can simply remove support for this to avoid any future maintenance or customisation.

Jira card: https://gov-uk.atlassian.net/browse/NAV-18980


source/javascripts/application.js
```diff
@@ -2,9 +2,5 @@
 //= require govuk_frontend
 //= require dependencies
 //= require govuk_publishing_components/lib/trigger-event
```
jon-kirwan: One small thing - I’m wondering whether we actually need `triggerEvent`, as I can’t see it used anywhere in the developer docs (aside from the documentation page itself: [https://docs.publishing.service.gov.uk/repos/govuk_publishing_components/lib/trigger_event.html](https://docs.publishing.service.gov.uk/repos/govuk_publishing_components/lib/trigger_event.html)).

jon-kirwan (commented):
Also, I can’t find a way to navigate to this page: [https://github.com/alphagov/govuk-developer-docs/blob/main/source/analytics/consent.html.md.erb](https://github.com/alphagov/govuk-developer-docs/blob/main/source/analytics/consent.html.md.erb). However, it looks like it may be specifically related to implementing GA4 for developer docs (I think). If that’s the case, we should be able to remove it.

jon-kirwan (commented):
All looks good to me, thanks @MartinJJones!

Just two small things I’ve commented on.

Otherwise, I’ve run the docs site locally and everything works as expected - no cookie banner, and no links to the cookie settings page anywhere. I can also see the page has been removed, so it can’t be accessed directly either.

MartinJJones:
> Also, I can’t find a way to navigate to this page: https://github.com/alphagov/govuk-developer-docs/blob/main/source/analytics/consent.html.md.erb. However, it looks like it may be specifically related to implementing GA4 for developer docs (I think). If that’s the case, we should be able to remove it.

Ahhh, good spot, I've had a further look and it looks like this page is included in the implementation record and should remain in place - https://docs.publishing.service.gov.uk/analytics/consent.html

We will likely need to update this documentation in the future, so I have added this site as an example to https://gov-uk.atlassian.net/browse/NAV-18880

source/javascripts/application.js
MartinJJones: Agreed, looks safe to remove, I've removed this in the latest commit, thanks

jon-kirwan:
> > Also, I can’t find a way to navigate to this page: https://github.com/alphagov/govuk-developer-docs/blob/main/source/analytics/consent.html.md.erb. However, it looks like it may be specifically related to implementing GA4 for developer docs (I think). If that’s the case, we should be able to remove it.
> 
> Ahhh, good spot, I've had a further look and it looks like this page is included in the implementation record and should remain in place - https://docs.publishing.service.gov.uk/analytics/consent.html
> 
> We will likely need to update this documentation in the future, so I have added this site as an example to https://gov-uk.atlassian.net/browse/NAV-18880

Ah, that's good to know. Thanks for checking and updating the docs card 👍

jon-kirwan (approved):
@MartinJJones Thanks for the changes. Everything looks good to me, and I've now approved.

PR #5618, opened 2026-07-20
Add small clarifications around the new fulladmin role permissions

hannako:
Add small clarifications around the new fulladmin role permissions


dj-maisy (approved):
Reads better than what I wrote. Ship it! :shipit:

PR #5622, opened 2026-07-24
Update the link to the Incident management doc

AgaDufrat:
We migrated from Google to Microsoft. The new process has been drilled and updated accordingly.


sihugh (approved):
New link works for me

PR #5628, opened 2026-07-30
Add more detail to timezone specification of temp admin credentials file

jfharden:
Ruby will happily parse all kinds of weirdly specified timezones, but IAM policies [Date conditions](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html#Conditions_Date) only support the W3C implementation of ISO 8601 time formats. So be explicit in the temp admin details about what that means


nicholsj (approved):
LGTM thanks

PR #5636, opened 2026-08-07
Add frontend developer baseline and probation objectives

luke-booker:
## What
This PR introduces a dedicated objectives page for mid-level and senior frontend developers (including probation) covering modern standards (WCAG/accessibility, progressive enhancement, web performance, and GOV.UK Design System implementation)

## Why
Currently, we have [baseline objectives for backend developers](https://docs.publishing.service.gov.uk/objectives/backend.html) to help guide line managers and developers during performance reviews and probation periods. However, there is no equivalent page for frontend developers to reference as a baseline or on probation. 

The update gives line managers a clear, repeatable framework for probation reviews and 1:1 discussions.

Screenshot 2026-08-21 at 15 13 02


data/dashboard.yml
```diff
@@ -33,8 +33,11 @@
     - name: Developer Objectives
       sites:
         - name: Backend developers
-          url: /objectives/backend.html
-          description: Objectives that describes what is meant by a "good backend developer" on GOV.UK.
+          url: /backend-developer/objectives.html
```
maxgds: While I can see the logic in adding "developer" I'm not clear on the advantage of branching here rather than "objectives" being the parent to both. Most of the docs are generic to all disciplines and I don't envision that changing and supporting discipline "areas" within the docs - unless you can make a case for what else might be specific to disciplines to support this? Also, while I think backend-developer is clearer, if you make this change I think you need to ensure a redirect at some level to support anything that might have bookmarked or documented that page under its current url.

source/backend-developer/objectives.html.md
leenagupte, line 1: I think this file name change should be undone, and the frontend file moved to `source/objectives/frontend.html.md` instead.

That way when I get around to creating objectives for junior technologists, they can also be in `source/objectives/`

That was my original plan for creating an objectives folder.

source/frontend-developer/objectives.html.md
```diff
@@ -0,0 +1,67 @@
...
+|Modern standards|I will uphold high standards of frontend engineering. (Progressive enhancement, accessibility, semantic HTML, web performance, CSS/JS code quality, testing)|I will guide other developers to follow best frontend engineering practices and use the GOV.UK Design System|_I will measure my progress against this objective using the following:_<p><p>Evidence of best frontend engineering practices (e.g. scalable CSS, progressive enhancement)<p><p>I can ensure code written by myself and others considers user needs, including accessibility standards (WCAG), cross-browser compatibility, and performance<p><p>Evidence of holding others accountable for doing high quality, accessible work|
+|Learning|I take ownership of my learning goals relating to my teams work|I will manage my time effectively to achieve learning goals|_I will measure my progress against this objective using the following:_<p><p>I can demonstrate progress against learning goals with my line manager<p><p>I can share my learning at community meetings|
```
maxgds: I think this (here and for mid) one wants a minor tweak, something like 

I take ownership of my learning goals, prioritising those relating to my team's work where necessary.

Just to make it clear learning can be a parallel stream unrelated to current team stuff. Learning is supposed to be an incentive for working here so people should feel empowered to learn something unrelated to their team's work if they don't have an immediate need there; for example upskilling in AI, learning another language like Python, taking a public speaking course to support giving presentations etc.

maxgds: Also - apostrophe missing in "team's"

source/frontend-developer/objectives.html.md
```diff
@@ -0,0 +1,67 @@
...
+
+|Type                                  |What           |How            |Outcomes     |
+|:---                                  |:---           |:---           |:---         |
+|Taking initiative / <p>Being proactive|I will be an active and engaged member of my team|I will be an active participant in team activities<p><p>I will proactively get involved in the work and pick up tasks<p><p>I will communicate progress and blockers with my team in a timely manner<p><p>I will communicate absences to my line manager in a timely manner<p><p>(Stretch) I will suggest improvements to team processes|_I will measure my progress against this objective using the following:_<p><p>Feedback from the team shows that I was an active participant in meetings, e.g. stand-up, retros<p><p>Feedback shows that I have been self-motivated in getting involved in the work and picking up tasks<p><p>Feedback shows that I communicated progress and blockers with the team<p><p>My line manager knows if I have been absent and why e.g. sick leave, annual leave<p><p>(Stretch) Feedback shows that I made suggestions in team retros, emails, or slack when I saw things that could be improved|
```
maxgds: "I will communicate absences to my line manager in a timely manner" - probably worth adding "and team" throughout on this one.

source/frontend-developer/objectives.html.md
```diff
@@ -0,0 +1,67 @@
...
+|Learning|I will learn about the frontend architecture of my assigned product or service|I am able to describe a relevant frontend component or feature<p><p>(relevant topic to be decided with Line manager)|_I will measure my progress against this objective using the following:_<p><p>I can explain how parts of the frontend architecture work to another person<p><p>I can debug issues in this area<p><p>(Stretch) I can deliver a talk at a team or frontend community meeting|
+|Learning|I will learn how to use the GOV.UK Design System<p><p>(if required) I will learn Ruby on Rails|I will apply components and patterns from the GOV.UK Design System in my development work<p><p>I will learn to build interfaces using the expected templating languages of my product<p><p>I will familiarise myself with WCAG accessibility guidelines and testing tools|_I will measure my progress against this objective using the following:_<p><p>I am able to complete frontend features on my own<p><p>I can write code that considers user needs, including accessibility standards (WCAG) and performance<p><p>I can describe a progressive enhancement or accessibility concept to another person|
```
maxgds: Let's change this to something more generic about "my team's stack" as the Design System and Brand teams don't use Ruby but this is still a valid goal for someone new to their ways of coding.

source/frontend-developer/objectives.html.md
```diff
@@ -0,0 +1,67 @@
+---
+layout: multipage_layout
+title: Baseline objectives for frontend developers at GDS
```
maxgds: Thanks for doing this. I've done a slightly hasty review which means it's probably not exhaustive. Broadly speaking it all looks good. I didn't compare to the backend objectives so apologies if I'm critiquing things you've inherited and not iterated yourself. The effort is much appreciated.

maxgds (changes requested):

source/frontend-developer/objectives.html.md
leenagupte: I think the Learning goal should be separated into two rows, one for the learning the Design System, and the optional row for learning Ruby. Otherwise all of the Ruby guidance and outcomes are lost.

source/frontend-developer/objectives.html.md
```diff
@@ -0,0 +1,67 @@
...
+|Learning|I will learn about the frontend architecture of my assigned product or service|I am able to describe a relevant frontend component or feature<p><p>(relevant topic to be decided with Line manager)|_I will measure my progress against this objective using the following:_<p><p>I can explain how parts of the frontend architecture work to another person<p><p>I can debug issues in this area<p><p>(Stretch) I can deliver a talk at a team or frontend community meeting|
+|Learning|I will learn how to use the GOV.UK Design System<p><p>(if required) I will learn Ruby on Rails|I will apply components and patterns from the GOV.UK Design System in my development work<p><p>I will learn to build interfaces using the expected templating languages of my product<p><p>I will familiarise myself with WCAG accessibility guidelines and testing tools|_I will measure my progress against this objective using the following:_<p><p>I am able to work on frontend features independently<p><p>I can write code that considers user needs, including accessibility standards (WCAG) and performance<p><p>I can describe a progressive enhancement or accessibility concept to another person|
```
leenagupte: Again, here. I think the Ruby training should have it's own row like it does in the backend developer objectives. That way it can also be measured as a separate learning goal.

leenagupte (commented):
I think on the whole the objectives look good. As I expected, most of them are behaviour based and therefore the same as the backend developer ones 💚.

The specific frontend developer ones sound reasonable to me, but you should get a review from @maxgds as well.

There is an inline comment about the directory structure of objectives. As mentioned I did set it up so that all of the objectives could eventually be housed in the same folder location in the developer docs to make them easier find which this change undoes.

luke-booker:
@leenagupte @maxgds @AgaDufrat 

Thank you all for the useful feedback. 

I've update the dashboard structure and changed some of the wording as suggested 👍  

Overall, I've removed any references to Ruby to stay agnostic for teams using other frameworks/technologies. Ruby learning and recommended reading can certainly still be applied here. The probation learning is now split into three sections - GOV.UK architecture, GOV.UK Design System and learning new tech.

Happy to make further changes if needed

maxgds (approved):
I've realised there's a glaring omission in that we haven't mentioned community work which is very important. I'm approving for now though, can add an iteration later.

PR #5643, opened 2026-08-12
Add new 'how to' document to help diagnose and fix chromedriver/chrom…

hezzimo:
Adding a how to document to help diagnose and fix issues with Chromium and Chromedriver which can cause all/most frontend tests to fail during test runs.


baisa (commented):
Looks good, thank you for documenting!

baisa (approved):

PR #5648, opened 2026-08-18
Clarify which email address should be used when requesting AWS accounts

brucebolt:
We are onboarding a new starter and this was not clear.


callumknights (approved):
Good spot. I wonder how long this will be the case

PR #5662, opened 2026-09-10
Fix implementation record header

andysellick:
## What/why
- was appearing incorrectly styled with light blue text on a dark blue background
- fixed by updating the markup
- ideally would use the layout component for this rather than hard coding the markup, but we can't render components directly in the tech docs so have to use `GovukPublishingComponents.render`, which doesn't seem to support yielding to blocks, which is what the layout component depends on
- as an improvement have also added the service navigation component
- unfortunately couldn't call the component here directly either because there's a mystery CSS rule which causes the content to be full width, which misaligns it with the rest of the page which acted on `govuk-width-container` being the immediate child of the component, fix was to insert an extra div to break this layout, so the normal styles can function correctly

## Visual changes
Before | After
----- | -----
Screenshot 2026-09-10 at 12 12 38 | Screenshot 2026-09-10 at 12 12 46


source/layouts/analytics_layout.html.erb
```diff
@@ -20,54 +20,77 @@
...
+    </section>
   </header>
 
   <div class="govuk-width-container ">
     <main class="govuk-main-wrapper">
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-one-third">
-          <ul class="govuk-list">
-            <li>
-              <%= link_to "Implementation record", "/analytics/", class: "govuk-link" %>
```
AshGDS, line 65: How does a user get back to the homepage with the service nav? I think you might need to make the service name a link

AshGDS (commented):
LGTM, just had a question before approving

source/layouts/analytics_layout.html.erb
andysellick: Whoops, sorry, overlooked that, fixed now.

AshGDS (approved):
LGTM 👍

PR #5668, opened 2026-09-18
Update visually-hidden text class

yndajas:
Currently we see the text "Warning This document has not been updated for a while now. It may be out of date.". The intention is for the "Warning" part to be read out by screen readers, but not visible

The old class was removed in govuk-frontend v5.0.0: https://frontend.design-system.service.gov.uk/v5/changes-to-govuk-frontend-v5/

This was because it duplicates the `govuk-visually-hidden`, to which the 5.0.0-beta notes advised migrating:
https://github.com/alphagov/govuk-frontend/discussions/4389#:~:text=Update%20the%20HTML%20for%20warning%20text

### Before

Screenshot 2026-09-18 at 13 49 32

### After

Screenshot 2026-09-18 at 13 48 48


MartinJJones (approved):
Good spot, thanks for fixing 👍

